PR #8, opened 2014-07-13
added support of frequency snapshots

pavlo2014:
This code is from Arthur Thijssen's repository:
https://github.com/Arthur-Thijssen/MEEP-actt/issues

The near to far field transform code (MEEP-actt) written by Arthur requires snapshots of this kind.


stevengj:
I'm going to close this PR, as I currently have a branch with what should be a vastly more efficient implementation of near-to-far-field transformation (see the near2far branch).

PR #11, opened 2015-02-06
Band source

FilipDominec:
I propose to include the band source into MEEP, as I originally advertised in April 2014. It gives more control to the source spectrum, e.g. to selectively excite a group of resonance modes, whereas other unwanted modes are efficiently suppressed.

More information can be found in the dedicated chapter on my website, f.dominec.eu/meep/index.html#bandsource


src/sources.cpp
```diff
@@ -23,6 +23,13 @@
 #include "meep.hpp"
 #include "meep_internals.hpp"
 
+// For sine-integral function used in the band-source
+#define HAVE_LIBGSL_EXPINT
```
stevengj: Why are you unconditionally defining this rather than modifying `configure.ac` to check for it?

stevengj:
(Would be good to have a Scheme/libctl interface for this too, no?)

Once, this PR is in good shape, it would be good to squash the commits into a single one for merging.

stevengj:

PR #21, opened 2015-08-31
Fixed a bug concerned with multiple frequency output in near-to-farfi…

mnishida:
…eld spectra.


stevengj:
Looks good to me, thanks!

PR #25, opened 2015-11-29
Feature/large grid

matthiaszilk:
This is a quite invasive commit to enable very large grids with more than 2^31-1 grid points. Basically the changes consist of setting the type of `meep::grid_volume::the_ntot` and all related variables to `ptrdiff_t` (via a typedef) and tracking down all errors, conversion warnings and loop variables (for details see the commit message of [521cb13](https://github.com/stevengj/meep/commit/521cb1331fad734d427d063a2d6331642fa2f1bb)). 
So far the changes have been tested only on x86-64 with gcc 4.8 and 4.9. I ran some simulations with very large grids (>2^31-1 grid points) for a dispersive gold grating (x and y periodic, z pmls) and a dipole emitter in a large box terminated with pmls. The obtained resuts were consistent with smaller grid sizes and unchanged versions of meep. I can provide both the results and the ctl files if you are interested. If there are any further tests you would like me to perform please let me know.

### DIFF
--- a/src/bands.cpp
+++ b/src/bands.cpp
@@ -72,7 +72,7 @@ void fields::prepare_for_bands(const vec &p, double endtime, double fmax,
   bands->tend = t + (int)(endtime/dt) - 1;
 
   {
-    int ind[8];
+    meep::integer ind[8];
     double w[8];
     int indind = 0;
     while (bands->index[indind] != -1 && indind < num_bandpts) indind++;
@@ -93,7 +93,7 @@ void fields::prepare_for_bands(const vec &p, double endtime, double fmax,
 
   double cutoff_freq = 0.0;
   if (gv.dim == Dcyl) {
-    cutoff_freq = 1.84*a*dt/(2*pi)/gv.nr()/sqrt(epsmax);
+    cutoff_freq = 1.84*a*dt/(2*pi)/static_cast<double>(gv.nr())/sqrt(epsmax);
     if (m == 0) cutoff_freq *= 0.5;
   }
   bands->fmin = sqrt(cutoff_freq*cutoff_freq + abs(k[Z])*abs(k[Z])*(a*dt)*(a*dt)/epsmax); // FIXME
--- a/src/bicgstab.cpp
+++ b/src/bicgstab.cpp
@@ -73,14 +73,14 @@ using namespace std;
 
 namespace meep {
 
-static double dot(int n, const realnum *x, const realnum *y)
+static double dot(meep::integer n, const realnum *x, const realnum *y)
 {
   double sum = 0;
-  for (int i = 0; i < n; ++i) sum += x[i] * y[i];
+  for (meep::integer i = 0; i < n; ++i) sum += x[i] * y[i];
   return sum_to_all(sum);
 }
 
-static double norm2(int n, const realnum *x) { 
+static double norm2(meep::integer n, const realnum *x) {
   // note: we don't just do sqrt(dot(n, x, x)) in order to avoid overflow
   int i;
   double xmax = 0, scale;
@@ -96,11 +96,11 @@ static double norm2(int n, const realnum *x) {
     double xs = scale * x[i];
     sum += xs * xs;
   }
-  return xmax * sqrt(sum_to_all(sum));
+  return static_cast<double>(xmax * sqrt(sum_to_all(sum)));
 }
 
-static void xpay(int n, realnum *x, double a, const realnum *y) {
-  for (int m = 0; m < n; ++m) x[m] += a * y[m];
+static void xpay(meep::integer n, realnum *x, double a, const realnum *y) {
+  for (meep::integer m = 0; m < n; ++m) x[m] += a * y[m];
 }
 
 #define MIN_OUTPUT_TIME 4.0 // output no more often than this many seconds
@@ -108,7 +108,7 @@ static void xpay(int n, realnum *x, double a, const realnum *y) {
 typedef realnum *prealnum; // grr, ISO C++ forbids new (double*)[...]
 
 /* BiCGSTAB(L) algorithm for the n-by-n problem Ax = b */
-int bicgstabL(const int L, const int n, realnum *x,
+meep::integer bicgstabL(const meep::integer L, const meep::integer n, realnum *x,
 	      bicgstab_op A, void *Adata, const realnum *b,
 	      const double tol,
 	      int *iters,
@@ -119,7 +119,7 @@ int bicgstabL(const int L, const int n, realnum *x,
 
   prealnum *r = new prealnum[L+1];
   prealnum *u = new prealnum[L+1];
-  for (int i = 0; i <= L; ++i) {
+  for (meep::integer i = 0; i <= L; ++i) {
     r[i] = work + i * n;
     u[i] = work + (L+1 + i) * n;
   }
@@ -145,11 +145,11 @@ int bicgstabL(const int L, const int n, realnum *x,
   // rtilde = r[0] = b - Ax
   realnum *rtilde = work + (2*L+2) * n;
   A(x, r[0], Adata);
-  for (int m = 0; m < n; ++m) rtilde[m] = r[0][m] = b[m] - r[0][m];
+  for (meep::integer m = 0; m < n; ++m) rtilde[m] = r[0][m] = b[m] - r[0][m];
 
   { /* Sleipjen normalizes rtilde in his code; it seems to help slightly */
     double s = 1.0 / norm2(n, rtilde);
-    for (int m = 0; m < n; ++m) rtilde[m] *= s;
+    for (meep::integer m = 0; m < n; ++m) rtilde[m] *= s;
   }
 
   memset(u[0], 0, sizeof(realnum) * n); // u[0] = 0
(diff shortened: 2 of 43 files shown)


src/bands.cpp
```diff
@@ -93,7 +93,7 @@ void fields::prepare_for_bands(const vec &p, double endtime, double fmax,
 
   double cutoff_freq = 0.0;
   if (gv.dim == Dcyl) {
-    cutoff_freq = 1.84*a*dt/(2*pi)/gv.nr()/sqrt(epsmax);
+    cutoff_freq = 1.84*a*dt/(2*pi)/static_cast<double>(gv.nr())/sqrt(epsmax);
```
stevengj: Why is this cast needed?

src/bicgstab.cpp
```diff
@@ -96,11 +96,11 @@ static double norm2(int n, const realnum *x) {
     double xs = scale * x[i];
     sum += xs * xs;
   }
-  return xmax * sqrt(sum_to_all(sum));
+  return static_cast<double>(xmax * sqrt(sum_to_all(sum)));
```
stevengj: ?

src/bicgstab.cpp
```diff
@@ -108,7 +108,7 @@ static void xpay(int n, realnum *x, double a, const realnum *y) {
 typedef realnum *prealnum; // grr, ISO C++ forbids new (double*)[...]
 
 /* BiCGSTAB(L) algorithm for the n-by-n problem Ax = b */
-int bicgstabL(const int L, const int n, realnum *x,
+meep::integer bicgstabL(const meep::integer L, const meep::integer n, realnum *x,
```
stevengj: The return value here is just an error code; why can't it be an `int`?  And `L` is the size of a subspace that will always be `< 1000` in practice; why did this need to change?

stevengj:
A lot of the type conversions here seem superfluous, e.g. the calls to `static_cast<double>`.

src/bands.cpp
matthiaszilk: Those casts were inserted to silence compiler warnings. Converting from  a 64 bit int to a double is a narrowing operation (the double mantissa has only 52 bits). To find all potential problems with the new 64 bit int type I had to use the gcc option `-Wconversion` that warns when an implicit conversion may alter the value. The meep::integer -> double conversions gave hundreds of false positives which made it hard to find the real problems. So I decided to insert static casts to make the conversion explicit. It turns of the compiler warning and documents that the conversion is actually intented. It looks a bit ugly but it does not cause any overhead (the conversion has to happen anyway, no matter if it is implicit or explicit). So I would prefer to leave the casts.

src/bicgstab.cpp
matthiaszilk: The type of sum is long double. Implicit conversion to double upon return will trigger a compiler warning without static cast.

src/bicgstab.cpp
matthiaszilk: The return type is not just an error code (see below) but could also be the required workspace size which is derived from `n`. The type of  `n` has to be meep::integer because at the call site it is derived from various chunk specific sizes which were all promoted to meep::integer. `L` could stay an int. I probably just missed that.

src/bicgstab.cpp
```diff
@@ -119,7 +119,7 @@ int bicgstabL(const int L, const int n, realnum *x,
```
matthiaszilk: The returned value is not just an error code but could also be the required workspace. As the type of `n` is meep::integer, the return type must also be meep::integer.

matthiaszilk:
The calls to `static_cast<double>` are required to suppress compiler warnings when `-Wconversion` is activated. They tell the compiler explicitly that the narrowing conversion from `meep::integer` (a 64-bit int on 64-bit systems) to `double` (which can hold only 52 bits in the mantissa) is intended. They should not impose any runtime overhead.

Some of the promotions from `int` to `meep::integer` might be indeed superfluous. But I decided to be conservative and promoted all variables that came into contact with a `meep::integer`. This might waste a little bit of stack space in some places but I don't consider this a severe problem. With some careful review one could figure out those spurious promotions but at the moment I don't see any benefit in doing this.

stevengj:
Fair enough.

src/mympi.cpp
```diff
@@ -147,10 +147,13 @@ void send(int from, int to, double *data, int size) {
 }
 
 #if MEEP_SINGLE
-void broadcast(int from, realnum *data, int size) {
+void broadcast(int from, realnum *data, meep::integer size) {
 #ifdef HAVE_MPI
   if (size == 0) return;
-  MPI_Bcast(data, size, MPI_FLOAT, from, mycomm);
+  if (size > INT32_MAX){
+ 	  abort("broadcast size exeeds maximum message size");
```
stevengj: inconsistent indentation here.

stevengj: Since you have this check several times, it would be better to write a function:
```c
static void check_size(meep::integer size) {
  if (size > INT32_MAX) abort("MPI message size exceeds maximum");
}
```

matthiaszilk: Sounds good. I include the check into the cast operation by replacing `static_cast<int>(size)` with a separate function `cast_message_size(size)`.

src/mympi.cpp
```diff
@@ -259,10 +287,10 @@ int max_to_all(int in) {
 }
 
 ivec max_to_all(const ivec &pt) {
-  int in[5], out[5];
+  long int  in[5], out[5];
```
stevengj: Seems like there should be a check somewhere that `sizeof(long int) == sizeof(meep::integer)`.

matthiaszilk: As long as `sizeof(long int) >= sizeof(meep::integer)` everything is fine because the data of `pt` is copied. However, on systems with LLP64 data model (are there any except MS Windows?) there will be trouble (maybe also in other places?). 
One way to resolve the issue would be to use `long long int` instead of `long int`. 
Another way would be to use `meep::integer` and define a suiteable MPI data type, e.g. `MPI_MEEP_INTEGER`, that points either to `MPI_INT`, `MPI_LONG` or `MPI_LONG_LONG_INT`, depending on the size of `meep::integer`. Then it would be possible to write
```c++
ivec max_to_all(const ivec &pt) {
    meep::integer  in[5], out[5];
    for (int i=0; i<5; ++i) in[i] = out[i] = pt.in_direction(direction(i));
 #ifdef HAVE_MPI
    MPI_Allreduce(&in,&out,5,MPI_MEEP_INTEGER,MPI_MAX,mycomm);
 #endif
   ivec ptout(pt.dim);
   for (int i=0; i<5; ++i) ptout.set_direction(direction(i), out[i]);
   return ptout;
 }
```
I will try this and see how far I can get with `AC_CHECK_SIZEOF` and the preprocessor.

matthiaszilk:
I added precautions for the case `sizeof(long int) < sizeof(meep::integer)`. Please have a look.

stevengj:
Looking good.  I assume you've tried `make check`?  (It would be nice to set up Travis testing at some point.)

stevengj:
`LIBS="$BLAS_LIBS $LAPACK_LIBS $BLAS_LIBS ...` should be safe enough, if slightly ugly.  But putting it in a separate PR would be good in any case.

stevengj:
Shouldn't the order be `-lblas -llapack -lgsl`?   I thought they have to be sorted in dependency order (i.e. if  A depends on B, then A should come *after* B on the link line).

matthiaszilk:
No, my LAPACK lib is not providing its own BLAS. I checked this. I suspect it is some issue with `libgsl` that is specific to Ubuntu. I think I will try another OS or compile `libgsl` from sources to see if the bug persists.

matthiaszilk:
Yes, I ran `make check`. I always do before pushing any changes. All tests are passed (tested with mpi and mpb).
Today I also set up a 32-bit VM with (Ubuntu 14.04.3, gcc 4.8.2) to make sure everything is fine there, too. Unfortunately on the 32-bit system things do not look as good. Two test are failling: Symmetry and three_d. I think this is not due to the changes of the current PR. On [master](https://github.com/stevengj/meep/commits/master) the tests fail as well. Here are the logs for both branches:
[test-suite_feature_large_grid.txt](https://github.com/stevengj/meep/files/52910/test-suite_feature_large_grid.txt)
[test-suite_master.txt](https://github.com/stevengj/meep/files/52911/test-suite_master.txt)
The symmetry test is passed when run serial or when `eps_compare` in `symmetry.cpp` is increased from `1.0e-9` to `2.0e-9`. Might be just an accuracy issue. The three_d test also fails when run serial. Slightly increasing the tolerance does not help either. Do you have any suggestions how to proceed?

matthiaszilk:
Reducing the optimization level from `-O3` to `-O2` solved the problem. Now all test are passed also under 32-bit. 
After palying with some optimization options I'm quite sure that `-ftree-vectorize` causes the tests to fail.

stevengj:
Was there a measurable performance penalty to turning off auto-vectorization?   (Probably this flag only helps in `step_generic.cpp` and `step_generic_stride1.cpp`, so we could potentially use it only there.)   Perhaps this is a gcc bug?  What gcc version do you have?

matthiaszilk:
I tested mainly with gcc 4.8 and 4.9 and to some minor extent with gcc 5. All versions show the same behavior. The test failures occur when sse2 or avx are activitated either by using `-msse2`, `-mavx` and so on, or by setting an appropriate architecture via the `-march` option.  I tested several configurations for the `-march` flag and can provide the results if you are interested. I did only little benchmarking based on the test suite (total runtime, time stepping time im near2far). On the virtual machine there is no performance penalty when vectorization is turned off. Maybe the speed is limited by the memory bandwidth. However, some testing on a real machine should be perforemed to be really sure. Unfortunately I can do "real" benchmarks only on 64 bit machines at the moment.

stevengj:
#193 is a more minimal version of this change that hopefully addresses the needs for large grids.

PR #27, opened 2016-03-22
added feature to compute flux of far fields in dft_near2far

oskooi:
A new function in dft_near2far class is created to compute the flux (or Poynting vector) of the far fields. The function first computes the far fields and then the flux at each of Nfreq frequencies. The results are displayed in standard output as a comma-delimited array. This function duplicates the far-field generation code from dft_near2far::save_farfields which is slow and not parallelized. In the case where the user wants to both save the far fields and compute its flux, the far-field calculation happens twice and is therefore inefficient. A proper implementation would be to compute and store the far fields once and then to use these stored fields in the two separate functions. This function has been validated by comparing its results to that obtained using post-processing of the far fields.

### DIFF
--- a/src/dft.cpp
+++ b/src/dft.cpp
@@ -492,10 +492,8 @@ double *dft_energy::total() {
   double *Fm = magnetic();
   double *F = new double[Nfreq];
   for (int i = 0; i < Nfreq; ++i) F[i] = Fe[i]+Fm[i];
-  double *Fsum = new double[Nfreq];
-  sum_to_all(F, Fsum, Nfreq);
-  delete[] F;
-  return Fsum;
+  delete[] Fe, Fm;
+  return F;
 }
 
 dft_energy fields::add_dft_energy(const volume_list *where_,
--- a/src/dft.cpp
+++ b/src/dft.cpp
@@ -441,6 +441,151 @@ dft_flux fields::add_dft_flux(const volume_list *where_,
   return dft_flux(cE[0], cH[0], E, H, freq_min, freq_max, Nfreq);
 }
 
+dft_energy::dft_energy(dft_chunk *E_, dft_chunk *H_, 
+		       dft_chunk *D_, dft_chunk *B_, 
+		       double fmin, double fmax, int Nf)
+{
+  if (Nf <= 1) fmin = fmax = (fmin + fmax) * 0.5;
+  freq_min = fmin;
+  Nfreq = Nf;
+  dfreq = Nf <= 1 ? 0.0 : (fmax - fmin) / (Nf - 1);
+  E = E_; H = H_; D = D_; B = B_;
+}
+
+dft_energy::dft_energy(const dft_energy &f) {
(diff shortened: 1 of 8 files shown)


oskooi:
A new feature is added to compute the electric, magnetic, and total energy spectra using the dft fields. This involves computing and storing the DTFT of the E, D, H, and B fields which, due to its lack of parallelization, will likely be slower than the flux feature due to a large number of required fields. The code is mostly adapted from the flux feature. As usual, the dft fields are saved in the HDF5 output file as 1d arrays which we will need to change to a decipherable format in order to facilitate post processing. Also, since no direction property is required when specifying the volume region over which the energy is computed, a default direction of 0 is used which enables the "fields-add-fluxish-stuff" code in libctl/meep.scm.in to be used. A test case should likely be added to the test suite which is not included in this pull request.

src/dft.cpp
```diff
@@ -441,6 +441,151 @@ dft_flux fields::add_dft_flux(const volume_list *where_,
...
+  delete[] F;
+  return Fsum;
+}
+
+double *dft_energy::total() {
+  double *Fe = electric();
+  double *Fm = magnetic();
+  double *F = new double[Nfreq];
+  for (int i = 0; i < Nfreq; ++i) F[i] = Fe[i]+Fm[i];
+  double *Fsum = new double[Nfreq];
```
stevengj: Seems like there's a memory leak here: `Fe` and `Fm` are not freed.

src/dft.cpp
```diff
@@ -441,6 +441,151 @@ dft_flux fields::add_dft_flux(const volume_list *where_,
...
+  sum_to_all(F, Fsum, Nfreq);
```
stevengj: This should not be necessary since there was already `sum_to_all` in `electric` and `magnetic`.

src/near2far.cpp
```diff
@@ -363,6 +364,84 @@ void dft_near2far::save_farfields(const char *fname, const char *prefix,
...
+      F_[i] = 0;
+      for (int j = 0; j < 2; ++j) {
+        double flux_sum = 0;
+	for (int p = 0; p < N; ++p)
+	  flux_sum += real(cE[j][p]*conj(cH[j][p]));
+	F_[i] += flux_sum * (1 - 2*j);
+      }
+      master_printf("near2far-flux:, %0.2g, %0.5g\n",freq_min+i*dfreq,F_[i]);
+    }
+    sum_to_all(F_, F, Nfreq);
```
stevengj: This `sum_to_all` seems redundant with the `sum_to_master` above?  Also, isn't there a memory leak because you're not freeing `EH`?

src/dft.cpp
oskooi: Bug is fixed.

src/dft.cpp
oskooi: Bug is fixed.

src/near2far.cpp
oskooi: Bug is fixed.

stevengj:
Is this PR still relevant?

PR #29, opened 2016-07-21
added float versions to sum_to_master in mympi

oskooi:
added float versions of "sum_to_master" function in src/mympi.cpp and src/meep/mympi.hpp to enable near2far.cpp to work with 32-bit floating precision (previously, it was crashing)


stevengj:
It looks like you mixed together two separate sets of patches?

stevengj:
Is this PR still relevant?

oskooi:
Closed by #30.

PR #36, opened 2016-12-15
add support for multilevel-atomic susceptibility

oskooi:
Support is added for multilevel-atomic susceptibility. This feature has been verified qualitatively with results from the steady-state abinitio laser theory (SALT) based on a 1d laser cavity (A. Cerjan et. al, Optics Express, vol. 20, pp. 474-488, 2012). The implementation builds on previous work which had been commented out with a couple of bug fixes related to the sign of the alpha terms and the update P equation; almost everything else has remained unchanged.

### DIFF
--- a/libctl/meep.scm.in
+++ b/libctl/meep.scm.in
@@ -119,19 +119,20 @@
 ; ****************************************************************
 ; Multilevel-atom nonlinear susceptibilities
 
-;(define-class transition no-parent
-;  (define-property from-level no-default 'integer non-negative?)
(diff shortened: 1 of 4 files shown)


oskooi:
Fixed a bug causing segmentation fault when num_chunks > 1. Thanks to G. Gkantzounis for the bug report.

xj361685640:
Dear Sir，

This request can merge to master now？

It seems that this function is forbidding now.

Thank you indeed!

PR #38, opened 2017-02-22
Material function working with polygons

probstj:
Dear Prof. Johnson,

I have added a material function to meep which allows defining the dielectric function with polygons. For 2D simulations, the dielectric function can be freely defined by adding any amount of non-overlapping polygons and their epsilon value. For 3D simulations, instead of an epsilon value, all added polygons must be assigned a material stack, which defines the dielectric function along the z axis, while the polygons describe the regions on the x-y-plane where these stacks are located. 

It all works with the proper anisotropic averaging done in material_function_for_polygons::eff_chi1inv_row. I made some effort that this runs pretty quick.

I hope you find this useful and could eventually add it to the master node if you reckon other users will find it useful as well.

With very best regards,
Jürgen Probst

### DIFF
--- a/src/anisotropic_averaging.cpp
+++ b/src/anisotropic_averaging.cpp
@@ -260,6 +260,1314 @@ void structure_chunk::set_chi1inv(component c,
   medium.unset_volume();
 }
 
+
+////////////////////////////////////////////////////////////
+//--- Begin: defs for polygon-based material functions ---//
+////////////////////////////////////////////////////////////
+
+// polygon defined by a numpy array with dimensions N*2 (containing the points) and defining an area with a certain epsilon
+simple_polygon::simple_polygon(double* matrix, int dimX, int dimY) {
+    if (dimY != 2) {
+        abort("Array with polygon points should have a shape X,2. The second dimension is not 2.");
+    }
+
+    //make sure last point equals first point. If not, append copy of first point to end:
+    if (dimX > 0 && (matrix[0] != matrix[dimY * (dimX - 1)] || matrix[1] != matrix[dimY * dimX - 1]))
+        this->_number_of_points = dimX + 1;
+    else
+        this->_number_of_points = dimX;
+
+    // copy matrix:
+    _polygon_points = new double*[_number_of_points];
+    for (int i = 0; i < dimX; i++) {
+        _polygon_points[i] = new double[2];
+        _polygon_points[i][0] = matrix[dimY * i];
+        _polygon_points[i][1] = matrix[dimY * i + 1];
+    }
+
+    //if last point does not equal first point -> append copy of first point to end:
+    if (this->_number_of_points == dimX + 1) {
+        _polygon_points[dimX] = new double[2];
+        _polygon_points[dimX][0] = matrix[0];
+        _polygon_points[dimX][1] = matrix[1];
+    }
+};
+
+simple_polygon::simple_polygon(double** matrix, int dimX) {
+    // if this constructor is used, care must be taken that matrix is not deleted
+    // externally. The memory pointed to by matrix will be deleted with ~polygon()
+    // First point of polygon must equal last point.
+
+    //make sure last point equals first point:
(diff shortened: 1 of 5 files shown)


stevengj:
Shouldn't this be added to libctl instead?  That's where the other geometry functions go.  How does this work in 3d?

oskooi:
A more thorough test case beyond just verifying the geometric properties of the polygon objects would involve comparing the results of time-stepping the fields for a polygon and an equivalent structure generated from the material function (e.g., given a vec as input, output the epsilon). Have you tried this yet?

probstj:
Prof. Johnson: 
I added the new material_function_for_polygons to the base meep code inside src/ since then it can be used when accessing meep as a c++ library, which is how I use meep. I understand everything in the base c++ library can also be accessed by the libctl frontend, but not the other way round. I was unsure whether to create a new file for material_function_for_polygons and all its dependencies like polygons, but then decided to add the code to anisotropic_averaging.cpp, since that is also where the other material_function methods are defined. But this could easily be changed.

In 3d, one first defines one or more material stacks, which is a definition of multiple layers by thickness and epsilon value. Then the polygons are added, each linked to a material_stack, which define where along the x-y plane which material_stack is located. I come from a lithography background, which formed the way I designed the material_function in the 3d case. This mimics the way structures are built in lithography, where you can only pattern/mask along the x-y plane, but then depending on following masked etching/deposition steps, you change the stack of layers along z only where polygons where masked in the lithography step. This design also facilitates quick reuse of existing design patterns for lithography for the simulations or vice versa.

probstj:
I have added another test case for polygons. This is actually just a copy of the existing test case 2D_convergence.cpp but working with polygons. I have attached the output of both test cases for comparison. 

Interesting is that 2D_convergence.cpp takes approx. 5 minutes on my computer while 2D_convergence_polygons.cpp takes approx. 1 minute. The thing here is that 2D_convergence.cpp uses the old non-libctl anisotropic averaging with normals calculated with sphere-quad.h which might not be as efficient and exact as the code in libctl. With the polygons, the normal is exact if there is no vertex in the pixel, otherwise it is an average. The mean epsilon and 1/epsilon values are the same with all approaches. 

[2D_convergence_polygons.log.txt](https://github.com/stevengj/meep/files/796454/2D_convergence_polygons.log.txt)
[2D_convergence.log.txt](https://github.com/stevengj/meep/files/796453/2D_convergence.log.txt)

oskooi:
What happens in the case of overlapping polygons? Or when a polygon overlaps a libctlgeom object? Perhaps we should therefore make this a part of libctl.

probstj:
FWIW, making it a part of libctl would be unfortunate, because then users of the c++ library would be left out again. 

Overlapping polygons are not that big of an issue, it won't break anything, just the structure would not be defined as what the user intended, which would be visible in the exported .h5. (If the overlapping polygons have the same eps, only the pixels where two overlapping polygons share a border will be more filled than just the union of the two polygons, i.e. any pixel can always only be at maximum 100% inside a polygon - this case already outputs a warning to the user. If the eps differ, the overlapping parts will have the sum of the chi1=eps-1 of every polygon). 

It is possible to include a check for overlapping polygons when new polygons are added, but is this really necessary? I mean, should we really try to check every user input? For example, it is also not checked whether the structure/sources really obey a given symmetry when some symmetry is exploited. Is it worth it to include these potentially computationally expensive checks every time for things that can very easily be checked by the user? 

I would say the same applies for libctlgeom objects overlapping polygons, but I am not sure, also not about what advantage making the polygons part of libctl would have. I will have a look at the libctl code to gain some insights.

stevengj:
(Or we could make it easier for C++ users to call libctl.)

oskooi:
Adding polygons to libctl and making libctl more callable from C++ is the right approach. This way, libctlgeom is the only place for geometric objects and C++ users can finally access libctlgeom. This would obviously require a major re-organization of Meep, MPB, and libctl. It would be well worth it.

stevengj:
This can be closed — Meep now supports polygonal prisms via libctl.

PR #42, opened 2017-05-07
radiating_2D example working in python

ChristopherHogan:
I'm opening this PR not with the intent to merge it, but to start a discussion.

The good news is that I got a `radiating_2D` example running in python.  Here's a comparison of the output with C++ (FYI, I'm using fairly old hardware).
```
C++ Results
===========
Trying out some physical tests...
time for set_epsilon = 0.00110412 s
final residual = 4.40085e-07
Finished solve_cw after 645 steps and 161 CG iters.
Ratio is 1.98852 from (0.1763 -0.0790262) and (-0.133483 -0.0308781)

Field time usage:
    connnecting chunks: 0.00782013 s
         time stepping: 0.136907 s
         communicating: 0.0405264 s
    Fourier transforming: 0.000184298 s
       everything else: 0.704831 s

Passed radiating source should decay spatially as 1/sqrt(r) in 2D.

Elapsed run time = 0.892151 s

Python Results
=============
Trying out some physical tests...
time for set_epsilon = 0.031812 s
final residual = 4.40085e-07
Finished solve_cw after 645 steps and 161 CG iters.
Ratio is 1.98851552846 from (0.176299844602 (0.176299844602-0.0790261972714j)) and (-0.133483021451 (-0.133483021451-0.0308781319705j))

Field time usage:
    connnecting chunks: 0.00787878 s
         time stepping: 0.201514 s
         communicating: 0.0455053 s
    Fourier transforming: 0.000240326 s
       everything else: 0.696365 s

Passed radiating source should decay spatially as 1/sqrt(r) in 2D.
```
The bad news is that in order to get a python, user-defined callback working I had to modify the C++ to take a `Caller` instance that wraps a function pointer, instead of passing the function pointer directly.

Ideally we would not have to modify the C++ side at all.  I'm still researching other possibilities, but I wanted to have this as a proof of concept for now. Let me know if a solution like this is a possibility, or if you wish to avoid it at all costs.

@stevengj @oskooi @HomerReid

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -88,3 +88,7 @@ tests/latest_output
 # output files
 examples/*.h5
 examples/*.png
+
+# python files
+__pycache__/
+*.pyc
--- a/examples/python/physical.py
+++ b/examples/python/physical.py
@@ -18,7 +18,7 @@
 import meep as mp
 
 
-def one():
+def one(vec):
     return 1.0
 
 
@@ -27,7 +27,7 @@ def radiating_2d(xmax):
     ymax = 3.0
 
     # grid_volume
-    gv = mp.grid_volume.voltwo(xmax, ymax, a)
+    gv = mp.voltwo(xmax, ymax, a)
     s = mp.structure(gv, one, mp.pml(ymax / 3))
 
     f = mp.fields(s)
--- a/examples/python/physical.py
+++ b/examples/python/physical.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2005-2017 Massachusetts Institute of Technology  
+# Copyright (C) 2005-2017 Massachusetts Institute of Technology
 #
 #  This program is free software; you can redistribute it and/or modify
 #  it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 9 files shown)


HomerReid:
Great that you got this working! Nice job.

I agree that it seems unfortunate to need to modify the C++ code. 

Just yesterday I succeeded in compiling `meep.py` myself (partly by referring to the SWIG files you pointed out to me on your branch) so this week I'll study your solution in more detail and try to figure out what the potential alternatives might be.

stevengj:
I guess you did this because there is no `void*` thunk parameter in the callback interface?  But why not just use a global variable?  The material-function callback does not have to be re-entrant.

Alternatively (and this is what we will eventually have to do for subpixel averaging anyway), the `nlopt-python.i` file (or a file included by it) can define its own C++ subclass of `material_function` that can store whatever state it wants.

stevengj:
For a global, couldn't you do something like:

```
PyObject *py_callback = NULL;
static double py_callback_wrap(const meep::vec &v) {
    PyObject *pyv = vec2py(v);
    PyObject *pyret = PyObject_CallFunctionObjArgs(py_callback, pyv, NULL);
    Py_XDECREF(pyv);
    double ret = PyFloat_AsDouble(pyret);
    Py_XDECREF(pyret);
    return pyret;
}

%typemap(in)(double (*f)(const meep::vec &)) {
  $1 = py_callback_wrap;
  py_callback = $input;
  Py_INCREF(py_callback);
}
%typemap(freearg)(double (*f)(const meep::vec &)) {
  Py_XDECREF(py_callback);
}
%typecheck(SWIG_TYPECHECK_POINTER)(double (*f)(const meep::vec &)) {
  $1 = PyCallable_Check($input);
}
```

PR #43, opened 2017-05-12
Autoconf and Travis for SWIG python module build

HomerReid:
+ Created top-level `python` directory with working Makefile.am to build `_meep.so` python module.
+ Modified `configure.ac` to **(a)** enable shared libraries by default, **(b)** detect python installation.
+ Prepared functional `.travis.yml` and achieved successful TravisCI integration with github.
+ In travis builds, the code builds successfully and the successful loading of the meep module in a python session is verified.

To get this working I needed to configure `--without-libctl` in the Travis build. Apparently meep requires a newer version of libctl than is available for `apt-get` fetching, which complicates TravisCI building. 

Options for addressing this:

1. Write complicated `.travis.yml` script to fetch and build sufficiently new version of libctl source.

2. Incorporate libctl into the meep repository, possibly as a submodule (though my experience with submodules is that their merits do not justify the hassle). 

3. Work with package maintainers to get latest libctl versions into package databases.

4. Agree going forward that libctl will be an optional module for meep whose build is not tested via continuous integration.

@stevengj @homerreid @oskooi @ChristopherHogan


.travis.yml
```diff
@@ -0,0 +1,20 @@
...
+    - guile-2.0-dev
+    - libgsl0-dev
+    - libfftw3-dev
+    - gfortran
+    - libhdf5-serial-dev
+    - liblapack-dev
+    - python-numpy
+    - swig
+script:
+  - ./autogen.sh --enable-shared --with-python --without-libctl
```
stevengj: `--with-libctl` will give an error simply because libctl is not installed.

See my travis file for MPB to see how to do it: https://github.com/stevengj/mpb/blob/master/.travis.yml

.travis.yml
```diff
@@ -0,0 +1,20 @@
...
+  - make
+  - cd python
+  - ln -s .libs/_meep.so .
+  - PYTHONPATH=. python -c 'import meep'
```
stevengj: Isn't `.` in the default Python path?

.travis.yml
```diff
@@ -0,0 +1,20 @@
...
+    - libgsl0-dev
+    - libfftw3-dev
+    - gfortran
+    - libhdf5-serial-dev
+    - liblapack-dev
+    - python-numpy
+    - swig
+script:
+  - ./autogen.sh --enable-shared --with-python --without-libctl
+  - make
```
stevengj: should check `make && make check && make install` as in the MPB script.

.travis.yml
stevengj: Also, this should just be a `make check` target in `python/Makefile.am`

.travis.yml
```diff
@@ -0,0 +1,20 @@
...
+  - cd python
+  - ln -s .libs/_meep.so .
+  - PYTHONPATH=. python -c 'import meep'
+  - cd
+  - exit 0
```
stevengj: not necessary

configure.ac
```diff
@@ -14,7 +14,8 @@ AM_CONFIG_HEADER(config.h)
 AC_CONFIG_MACRO_DIR([m4])
 AM_MAINTAINER_MODE
 AC_SUBST(SHARED_VERSION_INFO)
-AM_ENABLE_SHARED(no) dnl shared libs cause too many headaches to be default
+# HR 20170511 need shared libraries for python interfaces
```
stevengj: this is what version control is for — no need to comment on the author of each line.

configure.ac
```diff
@@ -455,6 +456,53 @@ AC_SUBST(MEEP_SUFFIX)
 program_transform_name="s,_,-,g;$program_transform_name"
 
 ##############################################################################
+# HR 20170510 checks for python
+##############################################################################
+AC_ARG_WITH(python,
+	[AC_HELP_STRING([--without-python], [don't compile Python interface])],
+        with_python=$withval,with_python=yes)
+
+
```
stevengj: extra blank lines

configure.ac
```diff
@@ -455,6 +456,53 @@ AC_SUBST(MEEP_SUFFIX)
...
+   AC_MSG_WARN([Python interface requires --enable-shared; disabling])
+   have_python=no
+else
+
+if test "x$with_python" = xno; then
+  have_python=no
+else
+
+AM_PATH_PYTHON([],[have_python=yes],[have_python=no])
+if test $have_python = yes; then
```
stevengj: missing indentation

python/Makefile.am
```diff
@@ -0,0 +1,34 @@
...
+  python_PYTHON = meep.py
+  pyexec_LTLIBRARIES = _meep.la
+endif
+
+if MAINTAINER_MODE
+
+meep-python.cpp:	$(SWIG_SRC) $(HPPFILES)
+	swig -Wall $(AM_CPPFLAGS) -c++ -python -o $@ $(srcdir)/meep.i
+
+meep.py:       		meep-python.cpp
```
stevengj: What is the `meep.py` build rule for?  It doesn't do anything.

stevengj: Oh, nevermind, swig builds it.

python/Makefile.am
```diff
@@ -0,0 +1,34 @@
...
+ $(top_srcdir)/src/meep.hpp               \
+ $(top_srcdir)/src/meep_internals.hpp     \
+ $(top_srcdir)/src/bicgstab.hpp           \
+ $(top_srcdir)/src/meep/vec.hpp           \
+ $(top_srcdir)/src/meep/mympi.hpp
+
+BUILT_SOURCES = meep-python.cpp meep.py
+
+AM_CPPFLAGS = -I$(top_srcdir)/src                   \
+              -I$(top_srcdir)/src/meep              \
```
stevengj: `-Isrc/meep` shouldn't be necessary since we always `#include <meep/vec.hpp>` etcetera.

python/Makefile.am
stevengj: Need an `-outdir` flag in case the build tree is different from the source directory.
```
swig -Wall $(AM_CPPFLAGS) -outdir $(builddir) -c++ -python -o $@ $(srcdir)/meep.i
```

python/Makefile.manual
```diff
@@ -0,0 +1,10 @@
+MEEP=/home/homer/work/Simpetus/meep
+PYTHON=/usr
+CPPFLAGS=-I$(MEEP)/src -I/usr/include/python2.7 -I$(PY_ROOT)/lib/python2.7/dist-packages/numpy/core/include 
+LDFLAGS=-L$(HOME)/codes/lib -L$(MEEP)/src/.libs -L/usr/lib/python2.7
+LIBS="-Wl,-rpath=../src/meep/.libs -lmeep -lgsl -lgslcblas -lopenblas -lfftw3 -lhdf5 -lhdf5_hl -lm -lgfortran -lpython2.7"
+
+all:
+	swig -c++ -python -Wextra $(CPPFLAGS) -o meep_wrap.cxx meep.i
+	$(CXX) $(CPPFLAGS) -c -fPIC meep_wrap.cxx 
+	$(CXX) -shared -fPIC $(LDFLAGS) meep_wrap.o $(LIBS) -o _meep.so
```
stevengj: Doesn't seem like this file should be included in the commit?

python/README
```diff
@@ -0,0 +1,18 @@
+20170511
```
stevengj: No manual version control.

python/README
```diff
@@ -0,0 +1,18 @@
+20170511 
+
+Building with the automake-generated Makefile succeeds, but
```
stevengj: Better to file issues for build problems rather than using notes in a README file like this.

python/meep.i
```diff
@@ -0,0 +1,99 @@
...
+#define SWIG_FILE_WITH_INIT
+#include "meep/vec.hpp"
+#include "meep.hpp"
+using namespace meep;
+%}
+
+%include "numpy.i"
+
+%{
+#if NPY_API_VERSION < 0x00000007
```
stevengj: Note that this is for [NumPy < 1.7](https://github.com/numpy/numpy/commit/5ede7c7bc2e9342269e708fd183c5311917b9169), i.e. numpy versions more than 5 years old; can probably omit this check nowadays?

python/meep.i
```diff
@@ -0,0 +1,99 @@
...
+#endif
+%}
+
+%init %{
+  import_array();
+%}
+
+// TODO: apply necessary numpy typemaps
+
+// Rename python builtins
```
stevengj: See also
```
%include "meep_renames.i"
%include "meep_enum_renames.i"
%include "meep_op_renames.i"
%include "meep_swig_bug_workaround.i"
```
from the libctl directory.  Can we re-use anything?

Couldn't all this `%rename` and `%feature` cruft be put into separate `.i` files?

ChristopherHogan:
Regarding libctl, here is how I was able to build it on Travis.  
```bash
pushd ..
wget http://ab-initio.mit.edu/libctl/libctl-3.2.2.tar.gz
tar xvf libctl-3.2.2.tar.gz
pushd libctl-3.2.2
./configure --prefix=$HOME/libctl --enable-shared
make && make install
popd && popd
```
Then I built meep like this:
```bash
export PATH=$HOME/libctl/bin:$PATH
export CPPFLAGS="-I$HOME/libctl/include $CPPFLAGS"
export LDFLAGS="-L$HOME/libctl/lib $LDFLAGS"
sh autogen.sh --verbose --prefix=$HOME/local --with-libctl=$HOME/libctl/share/libctl
make && make install
```
This is option 1) from your list.  I've also had bad luck with submodules, so I would discourage 2). To make it less complex we can even put all the commands in a script and in `.travis.yml` we just call `- build_libctl.sh`.  The only oddity is that I had to comment out  `test $config = bad && exit 1` at the end of `autogen.sh`.  Even though `config.log` reported that configure exited with status 0, the test was somehow failing and exiting with 1, but despite that, `make && make install` was succeeding anyway.

stevengj:
I agree about avoiding submodules; they are more trouble then they are worth.  In the MPB travis script linked above, I used git to grab a particular commit of libctl.

I would tend to install into something like `$HOME/local`, since if we install harminv too (another optional dependency) then we won't have to add additional directories to CPPFLAGS etc.

Rather than updating the PATH, I would tend to just set `GEN_CTL_IO=$HOME/local/bin/gen-ctl-io`.  Also, it is better style with `configure` to do `configure CPPFLAGS="..."` etcetera rather than setting a `CPPFLAGS` environment variable.  The advantage of passing it as an argument to `configure` is that then it will "remember" CPPFLAGS if the Makefile has to re-run `configure` later (e.g. if you change `configure.ac` in maintainer-mode), whereas you might forget to set the environment variable in a subsequent terminal.

HomerReid:
Thanks for the review.
The latest commit to this branch on my fork has the following.

-- Syntax updates to `configure.ac` and `python/Makefile.am` following Steven's review.
-- Added `tests/pythonMeep.bash` (shell script invoked by `make check` to verify successful loading of python meep module)
-- Removed test of python meep module from `.travis.yml`
-- Updated `.travis.yml` to fetch and build libctl and harminv, then `make && make check && make install`. Harminv is needed for the `aniso_disp` test.
-- Updated detection of GSL libraries in `configure.ac`. The old code seemed not to be adding `-lgslcblas` to the linker command line even upon successfully detecting GSL headers. Without GSL support, most things work, but the `near2far` test fails.
-- All 20 tests (including new `pythonMeep.bash`) now passing in Travis CI.

Similar to what Chris reported, I seem to need `exit 0` as the last line of `.travis.yml` in order to achieve a successful Travis build. Without this line, each individual subprocess returns 0, but the overall Travis script exits with 1 and the build is reported as failing. I don't know what is up with this.

Also, with fetching and building the support codes plus running the full test suite, the Travis build now takes 12+ minutes. This makes it harder on a day-to-day basis to confirm that minor quick updates don't break the build. Maybe we could define a streamlined test suite that tests bare fundamentals and successful compilation, with an optional higher-level test suite that tests extended functionality?

configure.ac
```diff
@@ -455,6 +456,54 @@ AC_SUBST(MEEP_SUFFIX)
...
+        pinc=`echo "import numpy; print (numpy.get_include())" | $PYTHON - 2>/dev/null`
+        AC_MSG_RESULT([${pinc:-unknown}])
+        test -n "$pinc" && PYTHON_INCLUDES="$PYTHON_INCLUDES -I$pinc"
+        CPPFLAGS="$save_CPPFLAGS $PYTHON_INCLUDES"
+        AC_CHECK_HEADER([numpy/arrayobject.h],[],[
+          AC_MSG_WARN([disabling Python wrappers])
+          have_python=no],[#include <Python.h>])jl$A
+      fi
+
+      CPPFLAGS=$save_CPPFLAGS
```
ChristopherHogan: This appears to overwrite the `$PYTHON_INCLUDES` flags we just added, but my assumption is that we only need those flags for the calls to `AC_CHECK_HEADER` and the python includes get added at some later stage.  Is that correct?

stevengj: Yes, they get added in `_meep_la_CPPFLAGS` in python/Makefile.am.   You only want to use these flags when compiling the python plugin, not the pure C++ parts of Meep.

stevengj:
The libctl and harminv commits will get tagged with version on the next release.   I'm postponing that to see if we need any patches to those, especially to libctl, to get the Python stuff working well.

stevengj:
Note that `-lgslcblas` is not needed if we link cblas from somewhere else, e.g. if harminv or MPB is linked to a fast BLAS.    We have to be careful not to link multiple conflicting BLAS libraries.

tests/pythonMeep.bash
```diff
@@ -0,0 +1,8 @@
+#/bin/bash
+
+#
+# test for successful loading of python meep module
+#
+cd ../python
+ln -f -s .libs/_meep.so
+python -c 'import meep'
```
stevengj: This should go directly into the Makefile.am.  The Makefile is already running the shell, so it's a bit perverse to have a separate file for this.

HomerReid: The advantage of doing it this way is that it clarifies that `pythonMeep` is just one of 20 single-word phrases in the `TESTS` definition in `tests/Makefile.am`. I'm concerned that having multiple lines of raw shell code in `Makefile.am` would make it harder to read, and would single out this test from all the others (why is the body of this one in `Makefile.am` instead of a separate file like all others?). I agree that we have here a case of shell invoking shell, but in my view it's worth it to put the `pythonMeep` test on equal footing with the other tests.

Or maybe you are envisioning a way of incorporating the test into `Makefile.am` that would be more transparent than what I am envisioning?

HomerReid: Regarding` -gslcblas`, I was under the impression that this is independent of the low-level BLAS provided by packages such as openblas or atlas. The routines in `gslcblas` have `cblas_` prefixes that aren't present in the routines provided by openblas. All of my codes compile with optional GSL support, and linking with both `-lgslcblas` and `-lopenblas` doesn't seem to cause problems. Am I missing something on this?

configure.ac
```diff
@@ -184,10 +185,10 @@ fi
 ##############################################################################
 # GNU Scientific Library
 
-AC_CHECK_FUNC(cblas_cgemm, [], [AC_CHECK_LIB(gslcblas, cblas_cgemm)])
-
-AC_CHECK_LIB(gsl, gsl_sf_bessel_Jn, [],
-	[AC_MSG_WARN([Missing GNU GSL library...Bessel-function field initialization will not be supported.])])
+AC_CHECK_LIB([m],[cos])
+AC_CHECK_LIB([gslcblas],[cblas_dgemm])
```
stevengj: This line should probably be:
```
AC_CHECK_FUNC(cblas_dgemm, [], [AC_CHECK_LIB(cblas, cblas_dgemm, [], [AC_CHECK_LIB(gslcblas, cblas_dgemm)])])
```

HomerReid: This doesn't work as is, at least not on my system. Linking with `-lgsl` but not `-lgslcblas` fails with error messages of the form

```
/home/homer/codes/lib/libgsl.so: undefined reference to `cblas_dgemm'
```

This is true **even if** I explicitly link to a BLAS library that contains the `cblas_` routines. I have tested this both with openblas and with the standard system BLAS (`/usr/lib/libblas.{a,so}`). Both of these definitely do provide the `cblas` symbols. For example,

```
% nm --dynamic /home/homer/codes/lib/libopenblas.so | grep cblas_dgemm
0000000000093800 T cblas_dgemm
```

But linking with e.g. `... -lgsl -lopenblas ...` fails with the above error messages about not finding `cblas_dgemm` and friends.

On the other hand, if I add `-lgslcblas` then it works, even if I don't remove `-lopenblas` from the command line. What is surprising is that `libgslcblas.so` seems to be providing the same symbols that e.g. openblas provides, for example: 

```
2030 hikari /tmp <> nm --dynamic /usr/lib/libgslcblas.so | grep cblas_dgemm
000000000000e880 T cblas_dgemm
```

The fact that I can link both openblas and gslcblas at the same time despite the fact that they both define the same symbols means that the symbols defined by the latter must be weak symbols somehow, and the fact that the codes in libgsl refuse to accept the openblas versions of these symbols means that there must be some sort of tagging going on in the GSL build. I will look into this. 

For the moment, since we agree that linking to gslcblas works and is harmless except for introducing possibly spurious extra code mass, should we keep things the way they are?

stevengj: I can't reproduce your problem.  On jdj.mit.edu, `gcc -o foo foo.c -lgsl -lopenblas` works where `foo.c` is
```
#include <math.h>
#include <stdio.h>
#include <gsl/gsl_sf_bessel.h>

int main(int argc, char **argv)
{
    printf("%g\n", gsl_sf_bessel_Jn(atoi(argv[1]), atof(argv[2])));
    return 0;
}
```
and `./foo 1 3.4` correctly prints `0.179226`

HomerReid: Thanks for looking into this. The difficulty seems to be related to harminv. Your test program above works for me when built via the command line you suggested. But if I attempt to link the harminv library it fails:

```
% gcc -o foo foo.c -lgsl -lharminv -lopenblas
...
/home/homer/codes/lib/libgsl.so: undefined reference to `cblas_dgemm'
...
```

If I position `-lharminv` after `-lgsl -lopenblas` on the command line then it works:

```
% gcc -o foo foo.c -lgsl -lopenblas -lharminv
% foo 1 3.4
0.179226
```

I'm looking in to why this is happening, but thought I would post just this much in case somebody already knows what's going on.

stevengj: What blas library is harminv linked with?

tests/pythonMeep.bash
stevengj: Ultimately, each of the Python tests should be a `.py` file, not a `.sh` file...

stevengj:
Anyway, this seems fine to merge.

PR #45, opened 2017-05-19
Mp session

HomerReid:
# mpSession: Making libctl functionality available from C++ and python

This is a preliminary proposal for a (slight) reorganization and expansion of the meep code base to facilitate implementation of python module.

The idea is to create a new C++ library named `libmpSession` ("meep-python session") that sits atop `libmeep` to incorporate the added functionality present in a libctl-driven meep session. This functionality then becomes available in both **(a)** C++ codes linked against `libmpSession`, and **(b)** python codes importing the `meep` python module (which now includes both `libmeep` and `libmpSession` via SWIG).

The goal is to ensure that `.ctl` scripts for meep calculations can be ported line-by-line to C++ codes build on libmpSession or python codes after `import meep`; for example, a section of a `.ctl` file like this:

```scheme
 (define-param s 11) ; the size of the computational cell, not including PML
 (define-param dpml 1) ; thickness of PML layers
 (define sxy (+ s (* 2 dpml))) ; cell size, including PML
 (set! geometry-lattice (make lattice (size sxy sxy no-size)))
```

becomes something like

```C
 double s=11.0;
 double dpml=1.0;
 double sxy = s + 2.0*dpml;
 S.set_geometry_lattice(sxy, sxy);
```

in C++, or like

```python
 s=11.0
 dpml=1.0
 sxy = s + 2.0*dpml
 S.set_geometry_lattice(sxy, sxy)
```

in python. (Here `S` is an instance of a new class named `mpSession` provided by `libmpSession`; see below).

# Implementation philosophy 

The design of `libmpSession` follows these guidelines:

-- All C++ code in the `libctl` library---including both native C++ code and C++ code generated by SWIG or by `gen-ctl-io`---is to be used as is, not reimplemented separately in `libmpSession.` (This requires repackaging most code in `meep/libctl` into a new library to be linked by program using `mpSession.`)

-- On the other hand, functionality that is implemented by scheme code in `.scm` files is re-implemented in the form of native C++ code in `libmpSession.`

# Modifications to code base

To facilitate this proposal, the existing meep code base is modified as follows:

-- In the `libctl` subdirectory, the object files that are currently linked to build the `meep` executable are subdivided into two groups: **(1)** `main.cpp`, and **(2)** everything else, including native C++ code, SWIG-generated C++ code, and gen-ctl-io-generated C++ code. Everything in group **(2)** is now put into a library called `libmeep-ctl.` The `meep` executable is built from just `main.cpp` plus this library.

-- I have added a new top-level subdirectory `mpSession` for the implementation of `libmpSession.`

-- The SWIG build of the python interface now includes `libmeep-ctl` and `libmpSession` as well as `libmeep.`

# Examples

As a first example, I have started porting `examples/pw-source.ctl` to C++ (`mpSession/pw-source.cpp`) and python (`python/pw-source.py`).

@ChristopherHogan @oskooi @stevengj


ChristopherHogan:
I brought up most of these points during the meeting Thursday, but I'll repeat them here for Steven's benefit.

I think this PR is an attempt to address many of the same questions I had while working on exposing structure.cpp to python.

**1. Do you imagine the python interface to function similarly to the current implementation?**

 I.e., do you imagine python "control" files that are read by the `meep` executable which contains an embedded python interpreter?  An alternative to this would be to use python as the high level control language and drop into C/C++ only for performance-critical code. I tend to favor calling C from python rather than the other way around. Additionally, if the goal of the project is to gain more adoption by introducing a python API, then it makes sense to expose the top-most control level to python. To that end, I would imagine, at the highest level, a simulation being run from python like this.
```python
def main(args):

    sim = mp.Simulation()

    sim.add_geometry()

    sim.add_sources()

    sim.add_flux_region()

    sim.run()

    sim.display_fluxes()


if __name__ == '__main__':
    sys.exit(main(sys.argv))
```
This looks very similar to the mpSession API that Homer is proposing.  We would have as many sensible defaults as possible, and allow customization through keyword arguments.  One concern is ending up with a python API that resembles C++ too much (a "non-pythonic" API).  But if `mpSession` implements all this with a high-level interface and SWIG wraps it, then hopefully we could end up with something similar to the above.

**2. How much and what parts of libctl should be wrapped/exposed to python, and or ported to python?**

Homer's suggestion here is to rewrite everything in C++.  That makes sense if all the code is performance-critical. The other options are to
* Port the Scheme portions to python.
* Rewrite the whole thing in python (probably not a realistic option).

I think what we do here depends on how much interaction with libctl a python user will require.  If we can hide all that detail away in a C++ `.so`, great, but if we can't, we probably need a C/python mix similar to the current C/Scheme mix. I imagine replacing the Scheme with python would result in performance similar to the current implementation.  For reference, here is what a `Lattice` class would look like in python.
```python
class Lattice(object):

    def __init__(self, basis1=mp.vector3(1, 0, 0), basis2=mp.vector3(0, 1, 0), basis3=mp.vector3(0, 0, 1),
                 size=mp.vector3(1, 1, 1), basis_size=mp.vector3(1, 1, 1)):

        self._process_func1 = mp.unit_vector3
        self._process_func2 = mp.unit_vector3
        self._process_func3 = mp.unit_vector3
        self._basis1 = basis1
        self._basis2 = basis2
        self._basis3 = basis3

        self.size = size
        self.basis_size = basis_size

    @property
    def basis1(self):
        return self._basis1

    @property
    def basis2(self):
        return self._basis2

    @property
    def basis3(self):
        return self._basis3

    @basis1.setter
    def basis1(self, value):
        self._basis1 = self._process_func1(value)

    @basis2.setter
    def basis2(self, value):
        self._basis2 = self._process_func2(value)

    @basis3.setter
    def basis3(self, value):
        self._basis3 = self._process_func3(value)

    def b1(self):
        return mp.vector3_scale(mp.vector3_x(self.basis_size), self.basis1)

    def b2(self):
        return mp.vector3_scale(mp.vector3_y(self.basis_size), self.basis2)

    def b3(self):
        return mp.vector3_scale(mp.vector3_z(self.basis_size), self.basis3)

    def basis(self):
        B = mp.matrix3x3(self.b1(), self.b2(), self.b3())
        if mp.matrix3x3_determinant(B) == 0:
            raise ValueError("lattice basis vectors must be linearly independent!")
        return B

    def metric(self, B):
        return B.T * self.basis()
````

**3. Is the plan to retain `gen-ctl-io` and have it generate `ctl-io.h` and `ctl-io.cpp` from a python specification file?**

The answer to this probably depends on the answers to 1. and 2.

**4. Is it worth rewriting some of the linear algebra (`vector3` and `matrix3x3` routines) in python via numpy/scipy?**

Advantage: Python users will already be familiar with the data structures and API. Everything will "just work" as expected. It should be just as fast as C++ since it's all BLAS underneath. 

Disadvantage: Requires extra effort since the work is already done.

File: configure.ac [hunk 1]
```diff
@@ -455,6 +456,54 @@ AC_SUBST(MEEP_SUFFIX)
...
+
+      if test $have_python = yes; then
+        AC_MSG_CHECKING([for Numpy include directory])
+        pinc=`echo "import numpy; print (numpy.get_include())" | $PYTHON - 2>/dev/null`
+        AC_MSG_RESULT([${pinc:-unknown}])
+        test -n "$pinc" && PYTHON_INCLUDES="$PYTHON_INCLUDES -I$pinc"
+        CPPFLAGS="$save_CPPFLAGS $PYTHON_INCLUDES"
+        AC_CHECK_HEADER([numpy/arrayobject.h],[],[
+          AC_MSG_WARN([disabling Python wrappers])
+          have_python=no],[#include <Python.h>])jl$A
```
ChristopherHogan: There appear to be some extra characters at the end of this line.

mpSession/mpSession.cpp
```diff
@@ -0,0 +1,237 @@
...
+/* structures defined in ctl-io.h; they replicate the behavior  */
+/* of (define-class ...) statements in libctl/xx.scm files, and */
+/* calling them from C++/python codes is the equivalent of      */
+/* saying (make ...) in .ctl scripts.                           */
+/*--------------------------------------------------------------*/
+
+/* (define-class medium material-type ...                       */
+ctlio::material_type *make_dielectric(double epsilon)
+ { 
+   ctlio::material_type *mt = (ctlio::material_type *)malloc(sizeof(ctlio::material_type));
```
ChristopherHogan: What's the reason to prefer `malloc` over `new`, since we're in C++?

ChristopherHogan: I see. It's using C types from ctl-io.h.

configure.ac
HomerReid: Good catch, thanks.

mpSession/mpSession.cpp
HomerReid: I tend to use `malloc` for `struct`s and `new` for `class`es, but to be honest I don't know if this is standard practice.

HomerReid:
Chris, thanks for your detailed comments and questions. The situation was clarified by my meeting with @stevengj yesterday.
 
The tl;dr is that `mpSession` as originally conceived is DOA as expected, but the process of preparing it taught me a lot about how the current code works, and also served to advance the discussion, so I'm filing it a success. I have a revised idea for how to implement it now, which I'll put together over the next days. Hopefully this will be less DOA, and eventually we'll come up with something that lives a few minutes before staggering into inevitable obsolescence :)

My original idea for `mpSession` was that it would link in all C++ code in `meep/libctl` in the form of a library. This would make it guile-dependent, which I thought was necessary to get the functionality implemented by the code in `meep/libctl`. However, Steven says the python module can be made to be guile-independent. One thing that emerged from my meeting with Steven that the native C++ code in `meep/libctl` consists of two separate pieces:

+ generic (meep-independent) geometry primitives in `geom.cpp`
+ meep-specific code in `structure.cpp`

Although it is totally unclear from the structure of the repository, the former of these is actually contained in the `libctlgeom` library provided by libctl, so we can get that functionality into `mpSession` by linking with `libctlgeom,` avoiding a guile dependence. (For unclear reasons, the meep build does not link to `libctlgeom`, but instead *hand-copies the `geom.cpp` file from the `libctl` installation and re-compiles it to link with the meep binary.* I don't know why it's done this way.)

On the other hand, `structure.cpp` **is** needed for `mpSession` and does depend on guile. So to eliminate the guile dependency that will need to be (slightly) rewritten as a separate code file compiled into `mpSession.`

Finally, of course there is the non-native C++ code in `meep/libctl` which executes functionality described in scheme by the `.scm` files. This will need to be rewritten in C++ or python, which is what I started to do with `mpSession.`

Doing things this way means the meep repository will have the same functionality implemented twice in (at least) two different languages, creating headaches for repository maintenance and doubling the work if any of that code ever needs to be changed. But there if no way around this if we are insisting simultaneously on **(a)** having chunks of the `meep` binary executable implemented in scheme, while **(b)** having no guile dependence in the python module.

One thing that I don't quite understand is that Steven's `libctlgeom` actually *does* depend on guile at the moment, so using it instead of the original way I did `mpSession` wouldn't seem to elminate the guile dependency. But I think the idea might be that the guile dependence of `libctlgeom` is minimal and could be removed relatively easily. I'm not sure if I'm understanding that right.

To your questions,

> 1. Do you imagine the python interface to function similarly to the current implementation?
> 
> I.e., do you imagine python "control" files that are read by the meep executable which contains an embedded python interpreter? 

No, the meep executable will not contain an embedded python interpreter. In fact, the meep executable won't be modified at all by the addition of the python module. Instead, there will be a `_meep.so` python module that includes all meep code, and users will import this into their python sessions or standalone python codes. 

> To that end, I would imagine, at the highest level, a simulation being run from python like this.

Yes, what you proposed looks like what I envision a python-driven meep simulation looking like.

> 2. How much and what parts of libctl should be wrapped/exposed to python, and or ported to python?

There is some disagreement on this. 

+ My instinct is to write everything in C++ and SWIG-wrap all of that functionality to make it available in python. Among other things, this has the advantage that we automatically get a full-fledged C++ interface for free in addition to the python interface. (There is already a C++ interface to the *low-level* stuff in meep, but it is crippled by lack of access to the *higher-level* stuff provided by `meep/libctl`). Also,

> I tend to favor calling C from python rather than the other way around.

I agree!

+ On the other hand, Steven's thought is that high-level code wants to be written in a high-level language, and thus that the existing `.scm` code should be ported to python, not C++. Among the arguments for this is that it makes it easier to maintain going forward, which I agree is a point in favor of this approach. 

Of course, as far as end users are concerned it shouldn't *matter* how the python module is implemented, so I don't think we should allow this question to delay progress toward getting something working---once we deliver something that works for end users, we can always go back and refactor to improve code maintainability.

> 3. Is the plan to retain gen-ctl-io and have it generate ctl-io.h and ctl-io.cpp from a python specification file?

No, `gen-ctl-io` will not be involved in the process of building `mpSession` and the python module.

> 4. Is it worth rewriting some of the linear algebra (vector3 and matrix3x3 routines) in python via numpy/scipy?

I think this question is related to your proposed python implementation of the lattice class, and the answer depends on what we think users are going to do with the python interface:

+ If we are envisioning that users would actually want to muck around with the weeds of the lattice implementation---for example, perform geometry operations on the lattice basis vectors---than it might make sense to have that implemented in python. 

+ However, if we assume that users are just going to want to specify the lattice dimensions without worrying about the underlying implementation, then we can just provide get/set functions (`set_lattice_dimensions(Lx, Ly, Lz)` and suchlike) which would be SWIG-translates of C++ functions that manipulate C++ data structures.

My sense would be that the vast majority of users will fall in the second bucket, and moreover that users who fall in the first bucket will be sophisticated enough to be able to handle making python SWIG calls to access C++ data structures as necessary.

 
@oskooi @stevengj @ChristopherHogan

stevengj:
The geometry stuff should certainly be available from C++.  Part of that is finishing the job in fully extricating libctlgeom from Scheme (at least at link time; Scheme will still be required to build it).  Part of this is in extricating libctl/structure.cpp from Scheme.

For the functionality in libctl/meep.scm, I would strongly recommend porting that to Python.  Mainly because it will be a lot less work (less code) that way than porting similar high-level functionality to C++, and be easier to maintain going forward.   For most of those things the low-level C++ interface is fine anyway.

PR #46, opened 2017-05-31
Add working test_physical.py

ChristopherHogan:
* Updates for python user callbacks
* Split vec.i to separate file
* Run python tests from python directory with 'make check'
* Add test_physical.py unit test
* Ignore some python specific files

I've verified that the tests in `test_physical.py` pass, but Travis never gets to run `make check` in the python directory because `near2far` fails.  I noticed the following line in the configure output on Travis:
```
checking for gsl_sf_bessel_Jn in -lgsl... no
configure: WARNING: Missing GNU GSL library...Bessel-function field initialization will not be supported.
```
and `near2far` fails with a message about not being able to find a GSL function.  I was able to fix this on my system (Ubuntu 16.04) by installing `libgsl-dev` (as opposed to `libgsl0-dev`), but Travis uses Ubuntu 12.04, and only `libgsl0-dev` is available. 

@stevengj @oskooi @HomerReid

### DIFF
--- a/python/meep.i
+++ b/python/meep.i
@@ -37,8 +37,7 @@ PyObject *py_callback = NULL;
 
 static PyObject* vec2py(const meep::vec &v) {
 
-    PyObject *py_vec;
-    double x = 0, y = 0, z = 0, r = 0;
+    double x = 0, y = 0, z = 0;
 
     switch (v.dim) {
       case meep::D1:
@@ -56,16 +55,11 @@ static PyObject* vec2py(const meep::vec &v) {
       case meep::Dcyl:
         r = v.r();
         z = v.z();
+        return Py_BuildValue("(ddd)", r, z, 0);
         break;
     }
 
-    if(v.dim == meep::Dcyl) {
-        py_vec = Py_BuildValue("(ddd)", r, z, 0);
-    } else {
-        py_vec = Py_BuildValue("(ddd)", x, y, z);
-    }
-
-    return py_vec;
+    return Py_BuildValue("(ddd)", x, y, z);
 }
 
 static double py_callback_wrap(const meep::vec &v) {
--- a/python/meep.i
+++ b/python/meep.i
@@ -15,29 +15,80 @@
  *  Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
  */
 
-%module(directors="1") meep
+%module meep
 
 %{
 #define SWIG_FILE_WITH_INIT
 #include "meep/vec.hpp"
 #include "meep.hpp"
+
 using namespace meep;
+
 %}
 
 %include "numpy.i"
-
-%{
-#if NPY_API_VERSION < 0x00000007
-#define NPY_ARRAY_C_CONTIGUOUS NPY_C_CONTIGUOUS
-#define NPY_ARRAY_ALIGNED  NPY_ALIGNED
-#endif
-%}
 
 %init %{
   import_array();
 %}
 
-// TODO: apply necessary numpy typemaps
+%{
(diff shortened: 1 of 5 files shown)


stevengj:
I use `libgsl0-dev` on one of our Debian machines and it found `gsl_sf_bessel_Jn` with no problem...  isn't that just the old name for `libgsl-dev`?

ChristopherHogan:
Rebased off master.  The passing python test is now visible in the build log.

python/meep.i
```diff
@@ -15,29 +15,75 @@
...
+        y = v.y();
+        z = v.z();
+        break;
+      case meep::Dcyl:
+        r = v.r();
+        z = v.z();
+        break;
+    }
+
+    PyObject *py_vec = Py_BuildValue("(ddddd)", x, y, z, r, 0.0);
```
stevengj: My suggestion would be to return `(x,y,z)`, or `(r,z,0)` in `Dcyl` (cylindrical).   Returning 5 numbers to user-defined callbacks is exposing internal cruft to users.

python/tests/test_physical.py
```diff
@@ -0,0 +1,100 @@
...
+            f.solve_cw(1e-6)
+        else:
+            while f.time() < 400:
+                f.step()
+
+        # amp1 and amp2 are of type complex
+        amp1 = f.get_field(mp.Ez, self.p1)
+        amp2 = f.get_field(mp.Ez, self.p2)
+
+        if sq_ratio:
```
stevengj: probably this should be just `if self.gv.dim == mp.D2` or whatever the Python equivalent is, since that clarifies the reason for scquaring the amplitude.   e.g.
```py
ratio = abs(amp1) / abs(amp2)
if self.gv.dim == mp.D2:
    ratio = ratio**2 # in 2d, decay is ~1/√r, so square to get 1/r
```

python/meep.i
ChristopherHogan: Fixed.

python/tests/test_physical.py
ChristopherHogan: Fixed.

python/meep.i
```diff
@@ -15,29 +15,80 @@
...
+        break;
+      case meep::D3:
+        x = v.x();
+        y = v.y();
+        z = v.z();
+        break;
+      case meep::Dcyl:
+        r = v.r();
+        z = v.z();
+        break;
```
stevengj: Just put `return Py_BuildValue("(ddd)", r, z, 0);`

python/meep.i
```diff
@@ -15,29 +15,80 @@
...
+    }
+
+    if(v.dim == meep::Dcyl) {
```
stevengj: `return Py_BuildValue("(ddd)", x, y, z);` here

ChristopherHogan:
Refactored vec2py.

python/meep.i
```diff
@@ -15,29 +15,74 @@
...
+      case meep::D3:
+        x = v.x();
+        y = v.y();
+        z = v.z();
+        break;
+      case meep::Dcyl:
+        r = v.r();
+        z = v.z();
+        return Py_BuildValue("(ddd)", r, z, 0);
+        break;
```
stevengj: `r` is no longer declared, hence the Travis failure.  Just do `return Py_BuildValue("(ddd)", v.r(), v.z(), 0);` and skip the `r = ...` and `z = ...`.  (You can also omit `break` given the `return` clause.)

ChristopherHogan: I should have tested locally before committing.  Fixed.

stevengj:
Never mind, I see it.

PR #47, opened 2017-06-02
Mp session v2

HomerReid:
Second attempt at implementing `mpSession.`

-- `libctlgeom-noscheme` is a scheme-independent refactoring of libctlgeom.

-- General idea: `mpSession` will provide all functionality currently available in the libctl-enabled meep binary, so that any meep calculation currently driven by a `.ctl` file can be ported line-by-line to a `.cpp` program.

-- As examples,
   `pw-source.cpp` and `holey-wvg-bands.cpp` are C++ ports of
   `pw-source.ctl` and `holey-wvg-bands.ctl` in `meep/examples.`

-- Then the python module can live atop `mpSession,` making use of its C++ implementations of low-level functionality for e.g. geometry operations, while potentially replacing other functionality (callbacks, etc.) with pythonic alternatives.

@ChristopherHogan @oskooi


stevengj:
libctlgeom in libctl master has now been made completely independent of Scheme (you can link to it without `-lctl` or `-lguile`).  (I'd prefer to avoid forking libctl.)

As we discussed, the next step is to implement a libmeepgeom library which basically implements one function:
```
set_materials_from_geometry(meep::structure *s, geometric_object_list g)
```
based on libctlgeom and meep/libctl/structure.cpp

PR #48, opened 2017-06-02
Updated configure script

HomerReid:
--fixed ordering of gsl, harminv, blas tests in `configure.ac`
--pull harminv from github/stevengj instead of github/homerreid

@ChristopherHogan @oskooi


ChristopherHogan:
👍

PR #52, opened 2017-06-09
Meepgeom

HomerReid:
Initial attempt at libmeep_geom.

The `libmeep_geom` directory contains source code and makefiles that build  `libmeep_geom` as well as a working sample program `pw_source_ll.cpp`, a C++ port of `examples/pw_source.ctl.`

A more interesting sample program is `bend_flux_ll.cpp`, a C++ port of `examples/bend_flux.ctl`. 

@ChristopherHogan @oskooi

### DIFF
--- a/libmeep_geom/material_data.hpp
+++ b/libmeep_geom/material_data.hpp
@@ -1,0 +1,101 @@
+/* Copyright (C) 2005-2015 Massachusetts Institute of Technology
+%
+%  This program is free software; you can redistribute it and/or modify
+%  it under the terms of the GNU General Public License as published by
+%  the Free Software Foundation; either version 2, or (at your option)
+%  any later version.
+%
+%  This program is distributed in the hope that it will be useful,
+%  but WITHOUT ANY WARRANTY; without even the implied warranty of
+%  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+%  GNU General Public License for more details.
+%
+%  You should have received a copy of the GNU General Public License
+%  along with this program; if not, write to the Free Software Foundation,
+%  Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+*/
+#ifndef MATERIAL_DATA_H
+#define MATERIAL_DATA_H
+
+/***************************************************************/
+/* the void *data field in geometric_object_struct points to   */
+/* a material_data structure, defined below.                   */
+/***************************************************************/
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <math.h>
+
+#include <meep.hpp>
+#include <ctlgeom.h>
+
+#include "meep.hpp"
+
+namespace meep_geom {
+
+typedef struct susceptibility_struct {
+  vector3 sigma_offdiag;
+  vector3 sigma_diag;
+  double frequency;
+  double gamma;
+  double noise_amp;
+  bool drude;
+  bool is_self;
+} susceptibility;
+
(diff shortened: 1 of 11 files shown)


libmeep_geom/material_data.hpp
```diff
@@ -0,0 +1,101 @@
...
+  vector3 E_chi2_diag;
+  vector3 E_chi3_diag;
+  vector3 H_chi2_diag;
+  vector3 H_chi3_diag;
+  vector3 D_conductivity_diag;
+  vector3 B_conductivity_diag;
+} medium_struct;
+
+typedef struct material_type_list
+ { material_type *items;
```
ChristopherHogan: Brace position in this typedef, the previous, and the next are not consistent.

libmeep_geom/meep_geom.cpp
```diff
@@ -0,0 +1,1540 @@
...
+
+static meep::realnum *epsilon_data = NULL;
+static int epsilon_dims[3] = {0,0,0};
+
+void read_epsilon_file(const char *eps_input_file)
+{
+  delete[] epsilon_data; epsilon_data = NULL;
+  epsilon_dims[0] = epsilon_dims[1] = epsilon_dims[2] = 1;
+  if (eps_input_file && eps_input_file[0]) { // file specified
+    char *fname = new char[strlen(eps_input_file)+1];
```
ChristopherHogan: Where is this memory freed?  Do you think `std::string` would be safer and less error prone?

ChristopherHogan:
Since a lot of this is coming from C it would be nice to update it to be more object oriented (i.e, classes with methods instead of structs and functions that take pointers to them).  That would align more with what the Python interface will be trying to accomplish.  On the other hand, if it isn't broken, don't fix it.  I would also suggest more consistent formatting.  Sometimes we see `var=val;`, other times `var = val;`.  Sometimes we have
```
func() {
  body
}
```
other times
```
func()
{
  body
}
```
Function prototype parameters sometimes indent to the level of the previous line's param, and sometimes they indent farther.

Also, you're probably aware of this, but Travis is currently failing with "config.status: error: cannot find input file: `Makefile.in'".

HomerReid:
Thanks for looking into this, and for the headsup about the build failing.

I've actually created a separate new branch to disentangle the libmeep_geom stuff from the ReadTheDocs stuff, so I will be closing this pull request and deleting this branch. I will make a pull request for the new branch once I've gotten a couple of successful tests in there.

To your points about style, I totally agree. The code that I assembled into libmeep_geom was a hodgepodge of native and machine-generated C code from a few different sources. It wasn't particularly well organized or stylistically consistent. Also, it retains some of the kluges left over from the need to play well with guile, and some global variables and other things that should probably be removed. I figured we'd worry about prettifying and consolidating the style once it had been established that the code was even going to remain going forward. On the other hand, almost all of this code is taken directly from the existing libctl/meep code base which has worked fine for many years, cf. your point about not fixing what ain't broke.

As for memory leaks, I didn't look carefully at this (the string-allocation code you referenced comes verbatim from the existing meep/libctl/structure.cpp file), and ultimately it would be good to bypass this global-variable approach and do something nicer about reading in epsilon files anyway. I have never been very meticulous about freeing all my string allocations in numerical codes, since it's such a tiny memory leak compared to the large arrays and matrices I'm usually working with---I would say it's important to avoid buffer overruns and other security hazards, but not crucial to deallocate all memory before the OS does so anyway when the process terminates. But, obviously I'm not up to date with current best corporate practices on this sort of thing.

PR #53, opened 2017-06-11
Ctl c io only

HomerReid:
It looks like `ctl-c-io-only` is a new variable added to libctl to facilitate the creation of guile-free libctlgeom. It is instantiated in `libctl/utils/geom.scm` when building libctlgeom, but isn't instantiated by any of the files in the `meep/libctl` folder used to build the meep binary. The result is that the meep binary does build, but aborts with an error as soon as it is executed, i.e.

```
% meep/examples <> ../../meep-installation/bin/meep < pw-source.ctl 
ERROR: Unbound variable: ctl-io-c-only?
```

I think the correct place to put this is `libctl/meep.scm.in`, and I have verified that with this fix the meep binary correctly runs the example codes.

@ChristopherHogan @oskooi


stevengj:
This should be fixed in the source (libctl) rather than fixing the symptoms.   (`ctl-io-c-only?` was supposed to be an internal hack that libctl-using programs wouldn't have to deal with.)

stevengj:
Fixed by https://github.com/stevengj/libctl/commit/9a2f274a1999fd5243c4a09ad05f91b32535c859

PR #54, opened 2017-06-11
point_in_objectp working

ChristopherHogan:
This PR isn't to merge, but to request feedback for the path I'm taking. I can successfully write
```python
some_point in some_sphere
```
in python and it will call `point_in_objectp(some_point, some_object)`, returning the correct result, but something seems off about this approach.  Does it make sense to build up a pure python `Sphere` just to tear it apart in the typemap?  It seems like we should just wrap `ctlgeom-types.h` and then do something like
```
point_in_objectp(meep.vector3(0, 0, 0), meep.sphere(center=meep.vector3(0, 0, 0), radius=2))
```
Then we wouldn't even need a typemap because the python objects would be SWIG proxy objects with types that correspond to the C types.  I think I'm still unclear on exactly which header files we want to create python wrappers for.

@stevengj @HomerReid @oskooi

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ addons:
     - swig
 script:
   - git clone https://github.com/stevengj/libctl libctl-src
-  - (cd libctl-src && git checkout 7d51710e63952d343ce8f241cf92b5cbfd5f1ada && sh autogen.sh --prefix=$HOME/local --enable-shared && make && make install)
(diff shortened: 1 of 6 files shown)


stevengj:
My concern is that SWIG wouldn't create nice Python proxy objects for the C types in `ctlgeom-types.h`, so we won't have a nice Python API unless we make our "proxy objects" manually as you are doing here.

ChristopherHogan:
Okay.  I'll continue with this approach.  Thanks.

PR #56, opened 2017-06-16
Meep geom

HomerReid:
Replaces previous `meepgeom` PR. 

Adds new folder `libmeep_geom` to top-level source tree with three working tests (`pw-source-ll.cpp`, bend-flux-ll.cpp`, `ring-ll.cpp`).

@oskooi @christopherhogan


HomerReid:
In the most recent commit, I have 

+ renamed `meep_geom` to `meepgeom`

+ removed `bend-flux-ll` from the test suite, as it was failing in the travis build. I wasn't able to debug exactly why this is, but I'm guessing it's because the code tries to write and then read back a largish HDF5 file which make not work in the travis build space. For now the test suite contains two tests which are passing in travis.

libmeepgeom/meepgeom.cpp
```diff
@@ -23,39 +23,55 @@ namespace meep_geom {
...
-material_data vacuum_material_data{
- material_data::MEDIUM, 0, 0, &vacuum_medium
-};
+material_type vacuum;
+material_type air;
+medium_struct vacuum_medium;
+material_data vacuum_material_data;
+bool materials_initialized = false;
+
+void initialize_materials()
```
stevengj: Why does this need to be a function?  Can't you just use a static initializer `const material_type vacuum = {....}`?

libmeepgeom/meepgeom.cpp
```diff
@@ -23,39 +23,55 @@ namespace meep_geom {
...
+   
+  vacuum_material_data.which_subclass = material_data::MEDIUM;
+  vacuum_material_data.user_func=0;
+  vacuum_material_data.user_data=0;
+  vacuum_material_data.medium = &vacuum_medium;
 
-material_type vacuum { (void *)&vacuum_material_data };
-material_type air = vacuum;
+  vacuum.data=(void *)&vacuum_material_data;
+  air = vacuum;
```
stevengj: I wouldn't bother with the the `air` synonym in the C++ interface.

libmeepgeom/meepgeom.cpp
HomerReid: No, this is what was preventing the build from working on Travis. Previously I had all the initializations done the way you suggested, but these static structure initializations require the `-std=c++11` compiler flag, which is not supported by the compiler versions used by Travis.

stevengj: Static structure initializations can be done even in 1990 ANSI C.  You just can't use named fields.  i.e.
```c
const medium_struct vacuum_medium = {{1,1,1},{0,0,0},{1,1,1},{0,0,0}};
const material_data vacuum_material = {MEDIUM,NULL,NULL,&vacuum_medium};
const material_type vacuum = {(void*) &vacuum_material};
```
(Note that the unspecified extra fields in `vacuum_medium` are initialized to 0, per the C/C++ standard.)

HomerReid: OK, in the latest commit (a1117e609d4ab4ccf54ff79d831cc314f3963e33) I've eliminated the `initialize_materials()` function in favor of static structure initializers.

libmeepgeom/meepgeom.cpp
HomerReid: OK, I removed it.

PR #58, opened 2017-06-17
Build and test python 3.6 in Travis

ChristopherHogan:
When building locally, autotools will use whichever python it finds first on your PATH.  Also, you can force it to use a python of your choice by adding `PYTHON=/path/to/bin/python` to the environment before running `autogen.sh`.
 
I also added libmeepgeom files that we don't need in source control to `.gitignore`.
@stevengj @oskooi @HomerReid

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,6 @@ script:
   - (CPPFLAGS=-I${HOME}/local/include LDFLAGS=-L${HOME}/local/lib GEN_CTL_IO=${HOME}/local/bin/gen-ctl-io sh autogen.sh --verbose --enable-maintainer-mode --prefix=$HOME/local --with-libctl=$HOME/local/share/libctl)
   - make && make check && make install
   # Build and test pymeep with python 3.6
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -10,10 +10,17 @@ addons:
     - liblapack-dev
     - python-numpy
     - swig
+before_script:
+  - wget https://repo.continuum.io/miniconda/Miniconda3-4.3.14-Linux-x86_64.sh
```
stevengj: Do you really need to install Miniconda for this?  Can't you just add `python3-numpy` to the Ubuntu dependencies and then configure with `PYTHON=python3` below?

ChristopherHogan: Miniconda isn't necessary, but I think it's a good idea to make sure we're compatible with it, and we'll need it eventually when we have a conda package.  Plus, it gives us flexibility to test against any package versions we want using lightweight conda environments.  Also, it looks like the default pythons on Ubuntu 12.04 (which is what Travis uses) are 2.7.3 and 3.2.3, so Miniconda is an easy way to test more recent versions.

stevengj: Okay.

PR #65, opened 2017-07-01
Python geometric object classes

ChristopherHogan:
It seems good to get the code review process going, so I went ahead and implemented the rest of the error handling.  This PR includes

* Typemaps and Python classes for all geometric objects and materials.
* Tests for these objects.
* Python versions of Homer's ring-ll.cpp and bend-flux-ll.cpp.
* Support for Python exceptions raised form C++ if a typemap fails.

It does not include the following:

* ~~I don't have test_bend_flux.py in the list of tests to run because it takes about 20 minutes and the output is not correct.  I'll replace this once we have more precise test cases.~~
* ~~I need a better test for the CompoundGemotricObject typemap.~~
* ~~I haven't implemented the sublclasses of Susceptibility.~~
* ~~I haven't done anything for "post-processed" properties (e.g., unit-vector3).~~
* geom.py will eventually be contained within the meep package.
* ~~Exceptions are not being raised properly on Python 3.  I need to dig into the unicode/string stuff a little more.~~

@stevengj @oskooi @HomerReid


ChristopherHogan:
I've fixed everything we talked about yesterday. This PR is ready for review.

python/Makefile.am
```diff
@@ -1,26 +1,33 @@
...
 _meep_la_SOURCES = meep-python.cpp
-_meep_la_LIBADD = $(top_builddir)/src/libmeep@MEEP_SUFFIX@.la $(PYTHON_LIBS)
+_meep_la_LIBADD = $(top_builddir)/src/libmeep@MEEP_SUFFIX@.la -L$(top_srcdir)/libmeepgeom/.libs $(PYTHON_LIBS) -lctlgeom -lmeepgeom
 _meep_la_LDFLAGS = -module -version-info @SHARED_VERSION_INFO@
 _meep_la_CPPFLAGS = $(PYTHON_INCLUDES) $(AM_CPPFLAGS)
 
 TEST_DIR = tests
-TESTS = $(TEST_DIR)/test_physical.py
+TESTS =                               \
+	$(TEST_DIR)/test_cyl_ellipsoid.py \
```
stevengj: wrong indentation

python/Makefile.am
```diff
@@ -1,26 +1,33 @@
...
 
 CLEANFILES = $(BUILT_SOURCES) meep-python.h
 
 AM_CPPFLAGS = -I$(top_srcdir)/src                   \
+              -I$(top_srcdir)/libmeepgeom           \
               -I$(top_builddir) # for config.h
 
 _meep_la_SOURCES = meep-python.cpp
-_meep_la_LIBADD = $(top_builddir)/src/libmeep@MEEP_SUFFIX@.la $(PYTHON_LIBS)
+_meep_la_LIBADD = $(top_builddir)/src/libmeep@MEEP_SUFFIX@.la -L$(top_srcdir)/libmeepgeom/.libs $(PYTHON_LIBS) -lctlgeom -lmeepgeom
```
stevengj: Should be in `top_builddir`, not `top_srcdir` (in case you are building out-of-tree).

Shouldn't it just be `$(top_builddir)/src/libmeepgeom.la` and omit the `-lmeepgeom`?

python/geom.py
```diff
@@ -0,0 +1,173 @@
+import math
+import meep as mp
+
+
+def non_negative(x):
```
stevengj: Seems clearer to just check `x >= 0` than to define a function.

python/geom.py
```diff
@@ -0,0 +1,173 @@
...
+
+class Vector3(object):
+
+    def __init__(self, x=0.0, y=0.0, z=0.0):
+        self.x = float(x)
+        self.y = float(y)
+        self.z = float(z)
+
+    def __eq__(self, other):
+        return self.x == other.x and self.y == other.y and self.z == other.z
```
stevengj: Eventually, we'll want the usual vector addition, subtraction, `* scalar`, etcetera operations, but that can be another PR.

python/geom.py
```diff
@@ -0,0 +1,173 @@
...
+        self.radius = radius
+        super(Sphere, self).__init__(**kwargs)
+
+    @property
+    def radius(self):
+        return self._radius
+
+    @radius.setter
+    def radius(self, val):
+        if not val or non_negative(val):
```
stevengj: What is the `not val` check for?  Isn't it sufficient to check `val >= 0`?

python/geom.py
```diff
@@ -0,0 +1,173 @@
...
+            self._radius = val
+        else:
+            raise ValueError("Radius cannot be negative. Got {}".format(val))
+
+
+class Cylinder(GeometricObject):
+
+    def __init__(self, axis=Vector3(0, 0, 1), radius=1.0, height=1.0, **kwargs):
```
stevengj: `radius` shouldn't have a default value.  For height, the default value should probably be `Inf`?

python/geom.py
```diff
@@ -0,0 +1,173 @@
...
+    def radius(self):
+        return self._radius
+
+    @property
+    def height(self):
+        return self._height
+
+    @radius.setter
+    def radius(self, val):
+        if non_negative(val) or not val:
```
stevengj: Since you have this check a lot, maybe it would be better to have a `check_nonnegative(name, val)` function that does:
```py
if val < 0:
    raise ValueError("{} cannot be negative: {}".format(name, val))
```
and then you can just do `self._radius = check_nonnegative("Cylinder.radius", val)`

python/geom.py
```diff
@@ -0,0 +1,173 @@
...
+    def height(self, val):
+        if non_negative(val) or not val:
+            self._height = val
+        else:
+            raise ValueError("Cylinder.height cannot be negative: {}".format(val))
+
+
+class Wedge(Cylinder):
+
+    def __init__(self, wedge_angle=8 * math.atan(1), wedge_start=Vector3(1, 0, 0), **kwargs):
```
stevengj: `2*math.pi`?

python/meep.i
```diff
@@ -81,6 +84,78 @@ static double py_callback_wrap(const meep::vec &v) {
...
+}
+
+// Typemap suite for GEOMETRIC_OBJECT
+
+%typecheck(SWIG_TYPECHECK_POINTER) GEOMETRIC_OBJECT {
+    PyObject *geom_mod = PyImport_ImportModule("geom");
+    PyObject *geom_dict = PyModule_GetDict(geom_mod);
+    PyObject *py_sphere = PyDict_GetItemString(geom_dict, "GeometricObject");
+    $1 = PyObject_IsInstance($input, py_sphere);
+    Py_XDECREF(geom_mod);
```
stevengj: Don't you need to decref the other two objects?  And isn't `py_sphere` a misnomer here, maybe you should rename to `py_geometricobject`?

I don't think you need `geom_dict`.  You can do
```c
py_geometricobject = PyObject_GetAttrString(geom_mod, "GeometricObject");
```
directly.

To speed this up you could potentially cache `py_geometricobject` in a global.  That is, define a function:
```
static PyObject *py_geometricobject(void) {
    static PyObject *geometricobject = NULL;
    if (geometric_object == NULL) {
        PyObject *geom = PyImport_ImportModule("geom");
        geometricobject = PyObject_GetAttrString(geom_mod, "GeometricObject");
        Py_XDECREF(geom);
    }
    return geometricobject;
}
```
and then just do `PyObject_IsInstance($input, py_geometricobject())`.

python/meep.i
```diff
@@ -81,6 +84,78 @@ static double py_callback_wrap(const meep::vec &v) {
...
+%typemap(freearg) GEOMETRIC_OBJECT {
+    if($1.subclass.sphere_data || $1.subclass.cylinder_data || $1.subclass.block_data) {
+        geometric_object_destroy($1);
+    }
+}
+
+// Typemap suite for boolean
+
+%typemap(out) boolean {
+    long b = $1 == 0 ? 0 : 1;
```
stevengj: This isn't necessary.  You can just do `PyBool_FromLong($1)` and it will be `true` if `$1 != 0` because that is the [truth value](https://docs.python.org/2.4/lib/truth.html) of an integer.

python/typemap_utils.cpp
```diff
@@ -0,0 +1,427 @@
...
+    return 1;
+}
+
+static std::string py_class_name_as_string(PyObject *po) {
+    PyObject *py_type = PyObject_Type(po);
+    PyObject *name = PyObject_GetAttrString(py_type, "__name__");
+
+    char *bytes = PyObject_ToCharPtr(name);
+
+    std::string class_name(bytes);
```
stevengj: This makes a copy of the string.  You need to `free(bytes)` to avoid a memory leak here.

python/Makefile.am
ChristopherHogan: Fixed

python/Makefile.am
ChristopherHogan: Fixed

python/geom.py
ChristopherHogan: Fixed

python/geom.py
ChristopherHogan: Originally I was allowing `None`, but forgot to change this after I made the arguments required. Fixed.

python/geom.py
ChristopherHogan: Fixed

python/geom.py
ChristopherHogan: Fixed.

python/geom.py
ChristopherHogan: Fixed.

python/meep.i
ChristopherHogan: I wasn't decrementing the the other two objects because those are weak references.  Using your solution eliminates the need for them anyway. Fixed.

python/meep.i
ChristopherHogan: Fixed.

python/typemap_utils.cpp
ChristopherHogan: I tried freeing `bytes` and saw a crash, then I noticed that the [documentation](https://docs.python.org/2/c-api/string.html#c.PyString_AsString) says this memory "must not be deallocated." It's shared with the interpreter, so python handles it.

ChristopherHogan:
I've addressed all comments, and made a note to add the additional `Vector3` operations in a later PR.

python/geom.py
stevengj: (Even then, `None >= 0` returns `False` in Python.)

PR #67, opened 2017-07-09
Libmeepgeom tests

HomerReid:
This is not quite ready to be merged, but I'm using the PR for discussion purposes.

I made progress on identifying and fixing what was preventing the proper detection of objects. In particular, the `.h5` output produced by `cyl-ellipsoid-ll` for just the material geometry itself (i.e. the `eps-00000000.h5` file) now agrees with that produced by `meep-test.ctl,` the scheme version of the same code. There were several issues, not all particularly interesting, but the last one may require some thought.

+ Some global geometry variables in `libctlgeom` needed to be initialized, as Steven suggested.

+ I had been using `HUGE_VAL` (which evaluates to `inf`) for large numbers like heights or *z*-coordinates in 2D geometries. This was causing problems in cases where it got multiplied against `0.0`, yielding `nan`. This was one of the factors confounding proper object detection. I finally realized that the libctl symbol `infinity` actually evaluates to a large *finite* number, so I did something similar for libmeepgeom.

+ A more significant issue is that the current structure of `libmeepgeom` complicates the handling of symmetries. The issue seems to relate to the order in which symmetries and geometric objects are defined: 

    - In a libctl run, the user makes a sequence of statements like `set! geometry,` `set! symmetries`, `set! pml-layers` which basically just register information in libctl data structures without actually instantiating a `meep::structure`. After defining all of that stuff in any arbitrary order, the user calls `run-until` or similar, and only then is the `structure` instantiated with all information (geometry, symmetries, objects, etc.) available from the start.

    + On the other hand, the way `libmeepgeom` is designed is that the user first creates a `structure` with symmetries and PML in place, and then subsequently adds the objects via `set_materials_from_geometry.` This rearrangement of the ordering seems to screw up proper detection of objects---with a 4-fold rotational symmetry in place the cylinder and ellipsoid seemed to be reduced to just 1/4 their correct size.
 
For the time being, I have simply disabled symmetries in all the `libmeepgeom` test codes. This allows the calculations to proceed and the `eps.00000.h5` files to be properly created.

On the other hand, the actual field output differs between codes trying to do the same thing in `.ctl` files and `libmeepgeom` codes. For example, here's the `Ez` component at the end of the `meep-test` (`cyl-ellipsoid-ll`) simulation as produced by the `.ctl` version (top) and the `libmeepgeom` version (bottom): 

meep-test-ez-000004600

ez-000004600

This looks to me like the PML is not getting properly turned on for the upper and right walls of the computational cell. The two calls in question here are 

+ libctl version: `make pml (thickness 1)`

+ libmeepgeom version: `pml(1.0)`

In other words, in both cases only the thickness is getting specified, with the direction and other parameters left to default settings. I spent some time trying to figure out if the `libctl` and bare `meep` defaults differ on this, and will continue looking into that unless the above images ring a different bell with anybody.

@oskooi @stevengj @ChristopherHogan


libmeepgeom/cyl-ellipsoid-ll.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+  // (set! sources (list (make source (src (make gaussian-src (frequency 1) (fwidth 0.1)))
+                     //  (center 0 0 0) (component src-cmpt))))
+  fields f(&the_structure);
+  double fcen = 1.0;
+  double df   = 0.1;
+  gaussian_src_time src(fcen, df);
+  vec src_point = vec(0.0, 0.0);
+  vec src_size  = vec(10.0, 10.0);
+  f.add_volume_source(src_cmpt, src, volume(src_point, src_size));
```
HomerReid: Thanks! That solves the problem---here's what my field output looks like now:

ez-000004600

libmeepgeom/cyl-ellipsoid-ll.cpp
```diff
@@ -97,25 +144,50 @@ int main(int argc, char *argv[])
...
-  f.add_volume_source(src_cmpt, src, volume(src_point, src_size));
+  f.add_point_source(src_cmpt, src, src_point);
+
+  // first test: write permittivity to HDF5 file and
+  // compare with contents of reference file
+  if ( am_really_master() )
+   { f.output_hdf5(Dielectric, f.total_volume());
+     #define DATAFILE "eps-000000000.h5"
+     #define DATASET  "eps"
+     if (compare_hdf5_datasets(DATAFILE, DATASET, eps_ref_file, DATASET))
```
stevengj: don't use preprocessor symbols for "local variables" ... use `const char x[] = "..."`, or in this case I would just paste in the strings directly.

HomerReid: ok, i did that.

HomerReid:
Update: There is now one fully working unit test for libmeepgeom, `libmeepgeom/cyl-ellipsoid-ll` invoked by `make check`. It checks two things: **(1)** binary HDF5 dataset produced by `fields::output_hdf5` agrees entry-by-entry (to single precision) with reference `.h5` file produced by the libctl version of this code, **(2)** field components at a given place and time agree numerically with reference values produced by the libctl version of the code.

On my local machine I notice that `make check` fails in the `python` subdirectory when I build after configuring with `--with-mpi.`

ChristopherHogan:
I haven't done anything with MPI for python yet.  Is there a way to disable it until I get to it?

PR #68, opened 2017-07-11
Automatic release creation on tag

ChristopherHogan:
Something like this should work for releases.  Travis will deploy to Github releases when a new tag is pushed to the repo.  
```bash
git tag 1.4
git push origin --tags
```
I created the secure api key using the Travis client.  I assume Steven will have to create a key that is linked to his repo and account.  Installation instructions are [here](https://github.com/travis-ci/travis.rb#installation).  Creating the key is as easy as `travis setup releases`.  This changes the `.travis.yml` file though, so I would do something like
```bash
mv .travis.yml .travis.yml.bak
travis setup releases
<copy api key from generated .travis.yml to .travis.yml.bak>
mv .travis.yml.bak .travis.yml
```
@stevengj @HomerReid @oskooi

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,4 +23,15 @@ script:
   - make && make check && make install
   # Build and test pymeep with python 3.6
   - (CPPFLAGS=-I${HOME}/local/include LDFLAGS=-L${HOME}/local/lib GEN_CTL_IO=${HOME}/local/bin/gen-ctl-io PYTHON=${HOME}/miniconda3/bin/python sh autogen.sh --verbose --enable-maintainer-mode --prefix=$HOME/local --with-libctl=$HOME/local/share/libctl)
-  - cd python && make clean && make && make check && make install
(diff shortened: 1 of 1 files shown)


stevengj:
It seems weird to put a "secure api key" in a public file on a repo.  Wouldn't that allow anyone to push releases?

ChristopherHogan:
From what I understand, that api key is encrypted and requires the private key generated by travis to decrypt.  That's why I think you have to generate your own key using the method above (`travis setup releases` does the encryption automatically).  I think if you merged this and tried to push a tag it wouldn't create a release because that api key is tied to my account.

stevengj:
I decided to revert this.  It seems much easier just to do `make dist` locally and drag the resulting `.tar.gz` file to the upload box on github when [creating a new release](https://github.com/stevengj/meep/releases/new).  (I just did this for a couple of other repos today and it is painless.)

PR #69, opened 2017-07-11
Improve python package structure

ChristopherHogan:
Currently all the python files get installed into `site-packages`, but they should be contained in their own `meep` package.  The structure should be
```
site-packages/
    meep/
        __init__.py
        geom.py
        _meep.so
```
This way, any `meep` class can be accessed like
```python
from meep import vec as vec
from meep.geom import Sphere as Sphere
```
or
```python
import meep as mp
import meep.geom as gm
mp.vec()
gm.Sphere()
```
The extra targets in `Makefile.am` are so local builds have the same directory structure as the installed library.  There's probably a better "automake" way of doing this that I'm unaware of.

@stevengj @HomerReid @oskooi

### DIFF
--- a/python/Makefile.am
+++ b/python/Makefile.am
@@ -49,7 +49,7 @@ __init__.py: meep.py
 
 pymeep: .libs/_meep.so __init__.py
 	test -d meep || mkdir meep
-	cp __init__.py geom.py .libs/_meep.so meep
+	cp __init__.py $(srcdir)/geom.py .libs/_meep.so meep
 
 all-local: pymeep
 
--- a/python/Makefile.am
+++ b/python/Makefile.am
@@ -30,7 +30,9 @@ TESTS =                               \
 
 TEST_EXTENSIONS = .py
 PY_LOG_COMPILER = $(PYTHON)
-AM_TESTS_ENVIRONMENT = PYTHONPATH=$(abs_top_builddir)/python:$PYTHONPATH; export PYTHONPATH;
+AM_TESTS_ENVIRONMENT =
+    PYTHONPATH=$(abs_top_builddir)/python:$$PYTHONPATH \
+    export PYTHONPATH;
 
 if WITH_PYTHON
   pkgpython_PYTHON = geom.py __init__.py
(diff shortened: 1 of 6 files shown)


python/Makefile.am
```diff
@@ -33,16 +33,28 @@ PY_LOG_COMPILER = $(PYTHON)
...
 
-meep.py:       		meep-python.cpp
+meep.py: meep-python.cpp
+
+__init__.py: meep.py
+	cp $< $@
+
+pymeep: .libs/_meep.so __init__.py
+	test -d meep || mkdir meep
+	cp __init__.py geom.py .libs/_meep.so meep
```
stevengj: Probably you need some `$(srcdir)` and `$(builddir)`s in here.

This is to support out-of-tree builds, and it would be a good idea to update travis so that we test this.  The way you do it is:
```
cd meep
mkdir build
cd build
../configure ...options...
make && make check && make install
```
instead of configuring and making in the top-level source directory.

ChristopherHogan:
This works now.  Since I changed Travis to do out-of-tree builds, I had to modify `libmeepgeom/cyl-ellipsoid-ll.cpp` so it could find `cyl-ellipsoid-eps-ref.h5` regardless of whether the test is run in or out of the source tree.
@HomerReid

HomerReid:
This looks fine to me. 

I had envisioned that the specification of where to find the data file would happen in `Makefile.am`, which is why I made the file path a command-line argument. I was envisioning that the `Makefile.am` would invoke the test by saying `cyl-ellipsoid-ll --eps_ref_file $(datadir)/cyl-ellipsoid-eps-ref.h5` or similar. 

But your solution seems equivalent, and I don't see a reason to prefer mine, so I'm happy to go with yours.

HomerReid:
Actually, it seems command-line arguments to `make check` codes must be specified in a global environment variable (`TEST_FLAGS`) that gets passed to *all* test codes, so there might not be a way to specify environment variables for just *one* test. Of course we could write a little shell script to do it, but that gets cumbersome, so it seems your solution is preferred.

PR #75, opened 2017-07-28
Disable python without shared

HomerReid:
Modify `configure.ac` to process `--disable-shared` option (which had previously been ignored) and disable compilation of python interface if present. 
@oskooi @ChristopherHogan


configure.ac
```diff
@@ -14,8 +14,6 @@ AM_CONFIG_HEADER(config.h)
 AC_CONFIG_MACRO_DIR([m4])
 AM_MAINTAINER_MODE
 AC_SUBST(SHARED_VERSION_INFO)
-# need shared libraries for python interfaces
-AM_ENABLE_SHARED
```
stevengj: You can change this to `AC_ENABLE_SHARED(no)` to make the default static-only (you can still do `--enable-shared`), and you'll need to change Travis to use that flag.   But I think we should leave it as `AC_ENABLE_SHARED` (building shared libraries) and just disable Python...

stevengj: (Note that `AM_ENABLE_SHARED` is the old spelling; nowadays this is called `AC_ENABLE_SHARED`, apparently.)

stevengj:
I think we should change `--with-python` to `no` by default, i.e. do:
```
AC_ARG_WITH(python,
	[AC_HELP_STRING([--with-python], [compile Python interface])],
        with_python=$withval,with_python=no)
```
(Note that the `.travis.yml` file should do `--with-python`.)

stevengj:
pythonMeep.bash is failing...

stevengj:
Could it be that you need `--enable-shared` for libctl too?

PR #76, opened 2017-07-28
Read the docs base url fix

HomerReid:
Rearrange files in `doc/docs` to ensure that `http://meep.readthedocs.io/en/latest` points to the head of the meep documentation tree with no need to specify any additional filename.
@oskooi @ChristopherHogan


HomerReid:
Note that, with this commit, the home pages for the meep and mpb documentation trees have migrated to

http://meep.readthedocs.io/en/latest
http://mpb.readthedocs.io/en/latest

I already corrected internal links within the doc trees, but if the old URLs (which had `Meep` or `MPB` at the tail end) were used anywhere else those need to be updated.

@oskooi @ChristopherHogan

PR #77, opened 2017-07-29
Add typemap for SourceTime and src_time

ChristopherHogan:
This turned out to be complicated.  I thought the tyepmap would be as simple as 
```
$1 = PyObject_GetAttrString(pyobj, "swigobj");
```
but that returns a `PyObject`, so I had to manually do the conversion from `PyObject` to C++ object that SWIG normally does, plus still allow a SWIG proxy object.  Users can now pass a subclass of `meep.source.SourceTime`, or `meep.src_time`.
@stevengj @HomerReid @oskooi

### DIFF
--- a/python/meep.i
+++ b/python/meep.i
@@ -45,6 +45,7 @@ extern boolean point_in_objectp(vector3 p, GEOMETRIC_OBJECT o);
 PyObject *py_callback = NULL;
 
 static PyObject *py_geometric_object();
+static PyObject *py_source_time_object();
 static PyObject* vec2py(const meep::vec &v);
 static double py_callback_wrap(const meep::vec &v);
 static int pyv3_to_v3(PyObject *po, vector3 *v);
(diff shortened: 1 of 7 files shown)


ChristopherHogan:
I just thought of a much nicer way to do this.

ChristopherHogan:
The other option is to do something like
```
%feature("pythonprepend") meep::fields::add_volume_source %{
    new_args = [arg.swigobj if isinstance(arg, mp.source.SourceTime) else arg for arg in args]
    args = new_args
%}
```
which is much simpler.  The disadvantage is that we would have to do this manually for every function that takes a `src_time` as an argument (and for any other high-level Python object that stores a `swigobj`).  I think I prefer the typemap, even though it's more complicated, because we can write it once and it will work for every case.  Reopening.

python/meep.i
```diff
@@ -152,6 +153,36 @@ static int py_list_to_gobj_list(PyObject *po, geometric_object_list *l);
...
+    $1 = py_source_time || swig_src_time;
+}
+
+%typemap(in) const meep::src_time & {
+    PyObject *swig_obj = NULL;
+    void *tmp_ptr = 0;
+    int tmp_res = 0;
+
+    if(PyObject_IsInstance($input, py_source_time_object())) {
+        swig_obj = PyObject_GetAttrString($input, "swigobj");
```
stevengj: This returns a new reference, so you need to `Py_DECREF(swig_obj)` after the `tmp_res = ...`

However, `swig_obj = $input` does *not* create a new reference, so you will need a `Py_INCREF(swig_obj)` there to balance out the decref.

ChristopherHogan: Good catch.  Fixed.

PR #81, opened 2017-08-12
Refactor test_cyl_ellipsoid.py and add assertions

ChristopherHogan:
Here's an updated, working `test_cyl_ellipsoid.py`. 

The bug I was encountering was introduced by my `GaussianSource` class.  I failed to notice that in `(make gaussian-src (frequency fcen) (fwidth df))`, `fwidth` changes the `width` to `1/df`. 

Is `fwidth=True` the sensible default in `GaussianSource.__init__()`?

@stevengj @HomerReid @oskooi

### DIFF
--- a/python/source.py
+++ b/python/source.py
@@ -41,13 +41,13 @@ class ContinuousSource(SourceTime):
 
 class GaussianSource(SourceTime):
 
-    def __init__(self, frequency, width, start_time=0, cutoff=5.0):
+    def __init__(self, frequency, width, fwidth=True, start_time=0, cutoff=5.0):
         super(GaussianSource, self).__init__()
         self.frequency = frequency
-        self.width = width
-        self.start_time = 0
+        self.width = 1 / width if fwidth else width
+        self.start_time = start_time
         self.cutoff = cutoff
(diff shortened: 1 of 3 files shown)


python/source.py
```diff
@@ -41,13 +41,13 @@ def __init__(self, frequency, start_time=0, end_time=float('inf'), width=0, cuto
 
 class GaussianSource(SourceTime):
 
-    def __init__(self, frequency, width, start_time=0, cutoff=5.0):
+    def __init__(self, frequency, width, fwidth=True, start_time=0, cutoff=5.0):
         super(GaussianSource, self).__init__()
         self.frequency = frequency
-        self.width = width
-        self.start_time = 0
+        self.width = 1 / width if fwidth else width
```
stevengj: I would make `width=0, fwidth=0` a numeric default, and then set `self.width = 1 / fwidth if fwidth > 0 else width`, so that you can pass either `width` or `fwidth`.

(We could modify the gaussian source in the C++ code to support `width=0` by converting it to a width of the timestep Δt.)

stevengj: Or alternatively it would be more consistent to use `width=0, fwidth=float('inf')` and maybe set `self.width = max(width, 1/fwidth)`

ChristopherHogan: Fixed.

PR #82, opened 2017-08-12
Initial high level python API preview

ChristopherHogan:
This isn't very polished yet, but I thought it would be good to get feedback on the API now that I have something working.

Working high-level examples are in `meep/python/examples`:
* `cyl-ellipsoid.py`
* `ring.py`

I've manually verified that the results are the same as the scheme examples, but I need some automated tests.

`simulation.py`: Contains the main `Simulation` class that maintains the state of the simulation and implements the functionality from `meep.scm`. I will add the missing features as I port the remaining examples.

Many functions in `Simulation` have a 'closure' dictionary. This is ugly, but is required to get closures to work correctly for python 2 (python 3 added a `nonlocal` keyword to fix this).  I'd like to come up with a nicer solution in the future.

Python 2 requires the `funcsigs` module at the moment, but I plan on removing that dependency.

@stevengj @HomerReid @oskooi


python/examples/cyl-ellipsoid.py
```diff
@@ -0,0 +1,46 @@
...
+    if src_cmpt == mp.Hz:
+        symmetries = [Mirror(mp.X, -1), Mirror(mp.Y, -1)]
+
+    sim = Simulation(cell_size=Vector3(10, 10, no_size),
+                     geometry=[c, e],
+                     pml_layers=[Pml(1.0)],
+                     sources=[sources],
+                     symmetries=symmetries,
+                     resolution=100,
+                     eps_averaging=False)
```
oskooi: Why is epsilon averaging being turned off? It is on by default and should remain so for this test.

python/examples/ring.py
```diff
@@ -0,0 +1,57 @@
...
+    w = 1  # width of waveguide
+    r = 1  # inner radius of ring
+    pad = 4  # padding between waveguide and edge of PML
+    dpml = 2  # thickness of PML
+    sxy = 2 * (r + w + pad + dpml)  # cell size
+
+    # Create a ring waveguide by two overlapping cylinders - later objects
+    # take precedence over earlier objects, so we put the outer cylinder first.
+    # and the inner (air) cylinder second.
+    dielectric = Medium(epsilon_diag=epsilon(n))
```
oskooi: This should be `epsilon(n*n)`.

python/examples/cyl-ellipsoid.py
```diff
@@ -0,0 +1,46 @@
...
+
+import meep as mp
+from meep.geom import Cylinder, epsilon, Ellipsoid, Vector3, Medium
+from meep.simulation import Mirror, Pml, Simulation, no_size
+from meep.source import GaussianSource, Source
+
+
+def main():
+
+    c = Cylinder(radius=3, material=Medium(epsilon_diag=epsilon(3.5)))
```
oskooi: The 'index' is 3.5 and the permittivity (i.e. epsilon) value is the square of the index. This should therefore be `epsilon(3.5*3.5)`.

python/examples/ring.py
```diff
@@ -0,0 +1,57 @@
...
+                     geometry=[c1, c2],
+                     sources=[src],
+                     resolution=10,
+                     symmetries=[Mirror(mp.Y)],
+                     pml_layers=[Pml(dpml)])
+
+    sim.run(
+        sim.at_beginning(sim.output_epsilon),
+        sim.after_sources(sim.harminv(mp.Ez, Vector3(r + 0.1), fcen, df)),
+        until=300, sources=True
```
oskooi: Why is `sources=True` included? It should be true by default and not need to be explicitly defined in the `run` function.

python/examples/cyl-ellipsoid.py
```diff
@@ -0,0 +1,46 @@
+from __future__ import division
+
+import meep as mp
+from meep.geom import Cylinder, epsilon, Ellipsoid, Vector3, Medium
+from meep.simulation import Mirror, Pml, Simulation, no_size
+from meep.source import GaussianSource, Source
+
```
oskooi: The `geom`, `simulation`, and `source` objects should all be loaded by default. It is cumbersome to import them one-by-one like this.

python/examples/cyl-ellipsoid.py
```diff
@@ -0,0 +1,46 @@
...
+    src_cmpt = mp.Hz
+    sources = Source(src=GaussianSource(1, 0.1), component=src_cmpt, center=Vector3())
+
+    if src_cmpt == mp.Ez:
+        symmetries = [Mirror(mp.X), Mirror(mp.Y)]
+
+    if src_cmpt == mp.Hz:
+        symmetries = [Mirror(mp.X, -1), Mirror(mp.Y, -1)]
+
+    sim = Simulation(cell_size=Vector3(10, 10, no_size),
```
stevengj: Can we just allow `0` here rather than `no_size`?  I've come to dislike `no-size` in the Scheme interface, which was an early hack that never went away.

python/examples/ring.py
stevengj: Maybe that's the equivalent of `run-sources+`?  Maybe would be clearer to use `until_after_sources=300` or similar?

python/simulation.py
```diff
@@ -0,0 +1,689 @@
...
+    from inspect import signature
+except ImportError:
+    from funcsigs import signature
+
+CYLINDRICAL = -2
+no_size = 1e-20
+
+
+def get_num_args(func):
+    sig = signature(func)
```
stevengj: How is this different from `return func.func_code.co_argcount`?

python/simulation.py
```diff
@@ -0,0 +1,689 @@
...
+        vec1 = py_v3_to_vec(self.dims, v1)
+        vec2 = py_v3_to_vec(self.dims, v2)
+
+        self.swigobj = mp.volume(vec1, vec2)
+
+
+class Simulation(object):
+
+    def __init__(self, cell_size, geometry, sources, resolution, eps_averaging=True,
+                 dimensions=2, pml_layers=[], symmetries=[], verbose=False):
```
stevengj: Maybe this should be called `absorbers` or `boundary_layers` since it nowadays includes non-PML layers.

python/simulation.py
```diff
@@ -0,0 +1,689 @@
...
+        else:
+            return self.filename_prefix
+
+    def _eval_step_func(self, func, todo):
+        num_args = get_num_args(func)
+        if num_args == 0:
+            if todo == 'step':
+                func()
+        else:
+            func(todo)
```
stevengj: Unlike Scheme (where everything is a global), probably the step functions need to be passed `self` (the current `Simulation` object) so that they have access to the fields etcetera.

python/examples/ring.py
ChristopherHogan: The idea was to have just one public `run` function that dispatches to the specialized run functions via keyword arguments.  If the user supplies an `until` argument, `Simulation._run_until` is called.  If the user supplies `sources=True`, then `Simulation._run_sources` is called, and if both are supplied, then `Simulation._run_sources_until` is called.  Would you prefer it to be the same as in scheme, or would clearer argument names help, as Steven suggests?

oskooi: Steven's approach is fine.

python/examples/ring.py
ChristopherHogan: My `epsilon` function in `geom.py` squared the argument passed to it, which is why this worked.  I've aligned it with scheme now.

python/examples/cyl-ellipsoid.py
ChristopherHogan: I added the `index` function to align with scheme.

python/simulation.py
ChristopherHogan: `signature` takes `self` into account and doesn't include it since it's a bound argument. I've removed the functools dependency by doing this for python 2
```python
def get_num_args(func):
    argc = func.func_code.co_argcount
    if 'self' in func.func_code.co_varnames:
        argc -= 1
    return argc
```

python/simulation.py
ChristopherHogan: Done.

python/examples/cyl-ellipsoid.py
ChristopherHogan: That was left over from debugging.  Fixed.

python/simulation.py
ChristopherHogan: Are you ok with requiring a 'placeholder' arg for every user-defined step function? I wanted to make the following argument combinations possible for a step function:
1. 0 arguments.
2. 1 argument - the instance of the `Simulation` class.
3. 1 argument - `todo`, which can be "step" or "finish".
4. Both 3 and 4.

The issue is that we can't tell at runtime which of `2` or `3` the user has in mind when their function takes 1 argument. One way around this is requiring one argument for every step function, which will be the `Simulation` instance, even if its unused, and if there is a second argument, we know it's `todo`.  Another option would be some kind of flag. Thoughts?

python/examples/cyl-ellipsoid.py
```diff
@@ -1,14 +1,14 @@
...
-from meep.geom import Cylinder, epsilon, Ellipsoid, Vector3, Medium
+from meep.geom import Cylinder, Ellipsoid, index, Vector3, Medium
 from meep.simulation import Mirror, Pml, Simulation, no_size
 from meep.source import GaussianSource, Source
 
 
 def main():
 
-    c = Cylinder(radius=3, material=Medium(epsilon_diag=epsilon(3.5)))
+    c = Cylinder(radius=3, material=Medium(epsilon_diag=index(3.5)))
```
stevengj: It would be better to be able to just write `Medium(epsilon=12)` or `Medium(index=3.5)`.   You should only ever specify `epsilon_diag` if you actually want an anisotropic material, e.g. `Medium(epsilon_diag=(12,13,11))`.

python/examples/cyl-ellipsoid.py
ChristopherHogan: Fixed.

python/examples/cyl-ellipsoid.py
ChristopherHogan: Fixed.

python/simulation.py
stevengj: I would be fine with requiring every step function to take at least one argument, the simulation instance.

The `todo` argument should probably be a Python enum, not a string.

ChristopherHogan: Done. I'll talk to you about the enum during the meeting.

python/examples/cyl-ellipsoid.py
ChristopherHogan: Done.

ChristopherHogan:
While finishing up the remaining tasks for this PR, I ran into a few issues.

## 1. Remove step functions from `Simulation` class

Many of the step functions depend on an instance of `Simulation` because they call `fields` methods.  After pulling the step functions out of `Simulation`, I would still need to pass a `Simulation` instance to them. The annoying thing about having the step functions as methods is that the user has to preface them with the instance name.
```python
sim.run(
        at_beginning(output_epsilon),
        after_sources(harminv(mp.Ez, mp.Vector3(r + 0.1), fcen, df))
)
```
would be much better than
```python
sim.run(
        sim.at_beginning(sim.output_epsilon),
        sim.after_sources(sim.harminv(mp.Ez, mp.Vector3(r + 0.1), fcen, df)),
```
One workaround would be keyword args like this
```python
sim.run(
    at_beginning="output_epsilon",
    after_sources={"harminv": (mp.Ez, mp.Vector3(r + 0.1), fcen, df)}
)
```

## 2. Require 2 arguments for step functions

This turned out to be difficult because of the way the `todo` argument is handled, and the solution wasn't immediately apparent to me. Currently step functions require 1 argument (the `Simulation` instance), and a second `todo` argument is optional.

## 3. Extract harminv into a separate class

I was able to do this, but the `Harminv` class needs a `fields` instance for `_get_field_point` and `fields.time()`, as well as some helpers and data members from the `Simulation`.  Hence, `Harminv` requires a `Simulation` to be passed to it, which isn't very clean.

## 4. High level tests

I successfully converted `test_ring.py` and `test_cyl_ellipsoid.py` to high level versions. Is the accuracy of harminv in `test_ring.py` what you would expect?  I could probably use some help getting better assertions there (e.g., using `rel_tol` like in `test_cyl_ellipsoid.py`).

@stevengj @HomerReid @oskooi

stevengj:
I don't understand the problem with point 1.  Since all step functions take a `Simulation` object anyway, then why is it a problem to move things like `after_sources` out of the simulation object?

python/tests/test_holy_wvg_bands.py
```diff
@@ -0,0 +1,69 @@
...
+        self.assertAlmostEqual(all_freqs[18][0].real, 0.19782709203322993)
+        self.assertAlmostEqual(all_freqs[18][0].imag, -0.0013233828667934015)
+        self.assertAlmostEqual(all_freqs[19][0].real, 0.1927618763491877)
+        self.assertAlmostEqual(all_freqs[19][0].imag, 0.001034260690735336)
+        self.assertAlmostEqual(all_freqs[20][0].real, 0.19335527231544278)
+        self.assertAlmostEqual(all_freqs[20][0].imag, 4.6649450258959025e-4)
+
+    def test_fields_at_kx(self):
+        self.sim.k_point = mp.Vector3(3.5)
+        h = mp.Harminv(self.sim, mp.Hz, mp.Vector3(0.1234), self.fcen, self.df)
```
stevengj: I don't understand why the `Harminv` constructor needs to take a simulation argument.   It is passed the simulation object during timestepping, shouldn't that be sufficient?

python/simulation.py
```diff
@@ -0,0 +1,783 @@
...
+
+class Harminv(object):
+
+    def __init__(self, sim, c, pt, fcen, df, mxbands=None):
+        self.c = c
+        self.pt = pt
+        self.fcen = fcen
+        self.df = df
+        self.mxbands = mxbands
+        self.harminv_data = []
```
stevengj: the `harminv_` prefix seems redundant here, since it is already namespaced inside the `Harminv` class

python/simulation.py
```diff
@@ -0,0 +1,783 @@
...
+            '_dt': 0,
+            '_c': self.c,
+            '_pt': self.pt,
+            '_fcen': self.fcen,
+            '_df': self.df,
+            '_maxbands': self.mxbands,
+            'data': self.harminv_data,
+            'dt': self.harminv_data_dt,
+            'results': self.harminv_results,
+        }
```
stevengj: I don't understand why the dictionary trick is needed here; it seems like something that should be done only when it is actually necessary.

python/simulation.py
```diff
@@ -0,0 +1,783 @@
...
+
+        def _harm(sim):
+            closure['data'] = closure['_data']
+            closure['dt'] = closure['_dt']
+            if closure['_maxbands'] is None or closure['_maxbands'] == 0:
+                mb = 100
+            else:
+                mb = closure['_maxbands']
+
+            closure['results'].extend(
```
stevengj: `extend` seems wrong here... shouldn't the new results replace any old ones?

python/simulation.py
```diff
@@ -0,0 +1,783 @@
...
+
+    def __init__(self, sim, c, pt, fcen, df, mxbands=None):
+        self.c = c
+        self.pt = pt
+        self.fcen = fcen
+        self.df = df
+        self.mxbands = mxbands
+        self.harminv_data = []
+        self.harminv_data_dt = 0
+        self.harminv_results = []
```
stevengj: the results shold be in some kind of user-accessible format.  Maybe a `harminv_results` should be called `modes` (for resonant modes) and should be in some kind of `Mode` object so that you can do `modes[1].Q` etcetera to access the different quantities.

python/simulation.py

python/simulation.py

python/simulation.py

python/simulation.py
ChristopherHogan: Done.

ChristopherHogan:
I believe I've addressed all the issues brought up at the last meeting.

### `Harminv` class
* Removed `Simulation` instance from `Harminv` class.
* Removed need for user to call `Harminv` instance (they can just pass it to a step function).
* Made `Mode.freq` return the real part of the frequency.  I'm still storing the imaginary part as `Mode.freq_imag` because I needed it in several places and it seemed silly to recalculate it from `Mode.Q`. 
### Step functions
* Removed all step functions from the `Simulation` class.  This  simplified `_eval_step_function`, allowing me to remove the `is_bound` checks. However, it doesn't help much with the examples because now instead of 
```python
sim.run(
    sim.at_beginning(),
    sim.at_end(),
    sim.at_every()
)
```
We have
```python
sim.run(
    mp.at_beginning(),
    mp.at_end(),
    mp.at_every()
)
```
unless the user does
```python
from meep import at_beginning, at_end, at_every

sim.run(
    at_beginning(),
    at_end(),
    at_every()
)
```
### Tests
* (Reluctantly) removed `tests_` prefix from all python files in `python/tests`.
### TODO (I'll get to these Sunday)
* `tests/test_bend_flux.py` is currently not being run.  I will update it to a high-level version, rename it, and add it to the `TESTS` list.
* Compare `output_epsilon` for `ring.py` with scheme.

python/simulation.py
```diff
@@ -131,7 +131,7 @@ def __init__(self, center, size=Vector3(), dims=2,):
         self.swigobj = mp.volume(vec1, vec2)
 
 
-Mode = namedtuple('Mode', ['freq', 'Q', 'amp', 'err'])
+Mode = namedtuple('Mode', ['freq', 'freq_imag', 'Q', 'amp', 'err'])
```
stevengj: `freq_imag` could be renamed `decay`, since it is called the decay rate in the harminv documentation.

ChristopherHogan: Done

ChristopherHogan:
The `output_epsilon` results from `ring.py` and `ring.ctl` appear to be the same.
```bash
chris@ubuntu:$ h5diff ring-eps-000000.00.h5 ../libctl/ring-eps-000000.00.h5 
chris@ubuntu:$ h5ls -v ring-eps-000000.00.h5 
Opened "ring-eps-000000.00.h5" with sec2 driver.
eps                      Dataset {160/160, 160/160}
    Location:  1:800
    Links:     1
    Storage:   204800 logical bytes, 204800 allocated bytes, 100.00% utilization
    Type:      native double
chris@ubuntu:$ h5ls -v ../libctl/ring-eps-000000.00.h5 
Opened "../libctl/ring-eps-000000.00.h5" with sec2 driver.
eps                      Dataset {160/160, 160/160}
    Location:  1:800
    Links:     1
    Storage:   204800 logical bytes, 204800 allocated bytes, 100.00% utilization
    Type:      native double
```

stevengj:
Good, so if there are differences, they are due to something else... sources or boundary conditions?

stevengj:
Is it running for the same number of time steps?

ChristopherHogan:
Yes, same time steps.  For both python and scheme I see
```
run 0 finished at t = 400.0 (8000 timesteps)
...
run 1 finished at t = 406.7 (8134 timesteps)
```
I'll double check sources and boundary conditions.  I verified that increasing resolution gave me greater accuracy too.

ChristopherHogan:
I think I've found the source of the issue. The data passed to `do_harminv` in scheme has 6001 items, but only 6000 in python.  The data at indices 0:5999 in python corresponds to 1:6000 in scheme. Scheme calls `get-field-point` once more than python does, and the extra value is at the very beginning of the list. 

I verified that manually prepending this extra value to the python data gives accuracy up to 15 decimal places (instead of 3).  I figured I'd post my progress here in case it rings any bells with anyone.

ChristopherHogan:
I solved the issue.  I was missing an `=` in `after_sources` which caused harminv to start half a second later in python than in scheme.

ChristopherHogan:
The harminv results of `ring.py` are equal up to 15 decimal places on my machine, but only up to 5 on travis.  Any idea why that might be?

python/examples/ring.py
```diff
@@ -0,0 +1,52 @@
...
+    dpml = 2  # thickness of PML
+    sxy = 2 * (r + w + pad + dpml)  # cell size
+
+    # Create a ring waveguide by two overlapping cylinders - later objects
+    # take precedence over earlier objects, so we put the outer cylinder first.
+    # and the inner (air) cylinder second.
+    dielectric = mp.Medium(epsilon=n * n)
+    air = mp.Medium()
+
+    c1 = mp.Cylinder(r + w, material=dielectric)
```
oskooi: Why not `radius=r+w` rather than just `r+w` since there is also a `height` attribute which needs to be disambiguated.

python/tests/holy_wvg_bands.py
```diff
@@ -0,0 +1,75 @@
+import unittest
+import meep as mp
+from meep import after_sources
+
+
+class TestHolyWvgBands(unittest.TestCase):
+
```
oskooi: Need to rename this test to `holey_wvg_bands.py`, `TestHoleyWvgBands`, etc.

python/tests/holy_wvg_bands.py
```diff
@@ -0,0 +1,75 @@
+import unittest
+import meep as mp
+from meep import after_sources
+
```
oskooi: This line is redundant.

oskooi:
I did a `git clone` from your `chogan/simulation` PR branch and ran a `make check`: all tests passed except for `ring.py`. Are you sure that all necessary changes have been committed to your branch?

ChristopherHogan:
That's strange. Travis is green right now on this PR. I verified that my local branch is the same as this PR.

When you run `meep/examples/ring.ctl` do you get the same output for `harminv0` as I've put in lines 53-57 of `python/tests/ring.py`?

ChristopherHogan:
For reference, here is what I get for the scheme `ring.ctl`. I hardcoded the `harminv0` results into `python/tests/ring.py`.
```
(mp) chris@ubuntu:libctl (chogan/simulation)*$ ./meep ../../examples/ring.ctl 
-----------
Initializing structure...
Working in 2D dimensions.
Computational cell is 16 x 16 x 0 with resolution 10
     cylinder, center = (0,0,0)
          radius 2, height 1e+20, axis (0, 0, 1)
          dielectric constant epsilon diagonal = (11.56,11.56,11.56)
     cylinder, center = (0,0,0)
          radius 1, height 1e+20, axis (0, 0, 1)
time for set_epsilon = 0.139094 s
-----------
creating output file "./ring-eps-000000.00.h5"...
Meep progress: 154.4/400.0 = 38.6% done in 4.0s, 6.4s to go
on time step 3145 (time=157.25), 0.00127196 s/step
Meep progress: 311.8/400.0 = 78.0% done in 8.0s, 2.3s to go
on time step 6291 (time=314.55), 0.00127151 s/step
harminv0:, frequency, imag. freq., Q, |amp|, amplitude, error
harminv0:, 0.11810131514715808, -7.315132416234067e-4, 80.72397629129884, 0.0034126763443671944, -0.0030495166730153483-0.0015319294671768792i, 1.0381385854126664e-5+0.0i
harminv0:, 0.1471626812839631, -2.3254886360785353e-4, 316.4123853387714, 0.028643397080539017, 0.019308635941190238-0.02115705022927364i, 7.308034193445086e-7+0.0i
harminv0:, 0.17524671265311115, -5.227797025346263e-5, 1676.1047894882997, 0.007211539425378899, -8.130068213316621e-4-0.0071655649318293395i, 1.943699068820815e-7+0.0i
run 0 finished at t = 400.0 (8000 timesteps)
creating output file "./ring-ez-000400.35.h5"...
creating output file "./ring-ez-000400.70.h5"...
creating output file "./ring-ez-000401.05.h5"...
creating output file "./ring-ez-000401.40.h5"...
creating output file "./ring-ez-000401.75.h5"...
creating output file "./ring-ez-000402.10.h5"...
creating output file "./ring-ez-000402.45.h5"...
creating output file "./ring-ez-000402.80.h5"...
creating output file "./ring-ez-000403.15.h5"...
creating output file "./ring-ez-000403.50.h5"...
creating output file "./ring-ez-000403.85.h5"...
creating output file "./ring-ez-000404.20.h5"...
creating output file "./ring-ez-000404.55.h5"...
creating output file "./ring-ez-000404.90.h5"...
creating output file "./ring-ez-000405.25.h5"...
creating output file "./ring-ez-000405.60.h5"...
creating output file "./ring-ez-000405.95.h5"...
creating output file "./ring-ez-000406.30.h5"...
creating output file "./ring-ez-000406.65.h5"...
run 1 finished at t = 406.70000000000005 (8134 timesteps)

Elapsed run time = 11.9118 s

```

ChristopherHogan:
I also tried a fresh clone and build in a new directory and `python/tests/ring.py` passed as it currently is in this PR. Here are the steps I took.
```bash
mkdir test
cd test
git clone https://github.com/stevengj/meep
cd meep
git checkout -b ChristopherHogan-chogan/simulation
git pull https://github.com/ChristopherHogan/meep.git chogan/simulation
export CPPFLAGS="-I${HOME}/local/include -I/usr/include/hdf5/serial ${CPPFLAGS}"
export LDFLAGS="-L${HOME}/local/lib -L/usr/lib/x86_64-linux-gnu/hdf5/serial"
export GEN_CTL_IO="${HOME}/local/bin/gen-ctl-io"
sh autogen.sh --enable-debug --enable-maintainer-mode --enable-shared --with-python --prefix=$HOME/local --with-libctl=$HOME/local/share/libctl
mkdir build
make distclean
cd build
../configure --enable-debug --enable-maintainer-mode --enable-shared --with-python --prefix=$HOME/local --with-libctl=$HOME/local/share/libctl
make && make install
cd python
make check TESTS=../../python/tests/ring.py
>>>PASS: tests/ring.py
```

oskooi:
Followed your instructions but it's still failing on my EC2 Ubuntu instance:

```
Ran 1 test in 5.373s

FAILED (failures=1)
Meep progress: 398.1/400.0 = 99.5% done in 4.0s, 0.0s to go
harminv0:, frequency, imag., freq., Q, |amp|, amplitude, error
harminv0:, 0.118101879017, -0.000732066501362, 80.6633542152, 0.00341448782045, (-0.0030503227893-0.00153435913562j), (1.02562230693e-05+0j)
harminv0:, 0.147162759134, -0.000232562115525, 316.394522818, 0.0286437485749, (0.0193073321123-0.0211587159139j), (7.29738949177e-07+0j)
harminv0:, 0.175246717461, -5.23183804979e-05, 1674.81022724, 0.00721188715646, (-0.000813060635121-0.00716590878822j), (1.95828271497e-07+0j)
```

oskooi:
The discrepancy is likely due to `harminv`'s two dependencies BLAS and LAPACK. I used Ubuntu's precompiled packages `libblas-dev` and `liblapack-dev` on the EC2 `c4.large` instance (Intel Xeon 2.9 GHz). Both version numbers are 3.6.0-2ubuntu2. `harminv` and Meep's other dependencies were installed based on the following [steps](https://www.mail-archive.com/meep-discuss@ab-initio.mit.edu/msg05819.html). Perhaps you can try reproducing the `ring.py` failure.

ChristopherHogan:
FYI, the bash script you linked to is missing the untar and `cd` steps for fftw. Also, h5utils is missing a `make` in `sudo 'make' install`.

oskooi:
Thanks for pointing this out. I've updated the scripts with the fixes.

ChristopherHogan:
I set up a docker image with your environment and was able to reproduce the failure. My environment is the same except I'm using serial hdf5, serial fftw, and no mpi.  Also I installed guile via `apt` instead of compiling it myself, but it's version 2.0.11.

I think the issue is that we're obtaining output from `ring.ctl` on one platform, and expecting the results to be exactly the same on all other platforms.

What we should be doing is running both the scheme and python versions on platform X, and testing that both versions produce the same results on platform X.

The other option is to relax the tolerance.

oskooi:
Since the Scheme and Python versions are producing the same output for the ring example, we can go ahead and reduce the tolerance.

python/examples/ring.py
ChristopherHogan: Radius is the only positional argument since there's no default, so it will always be first.  I added `radius=` for clarity though.

python/tests/holy_wvg_bands.py
ChristopherHogan: Fixed.

python/tests/holy_wvg_bands.py
ChristopherHogan: Removed.

ChristopherHogan:
All tests should pass now on your EC2 instance.

PR #83, opened 2017-08-14
Disable python without shared

HomerReid:
Modify `configure.ac` so that the defaults are `--enable-shared --without-python.`
Disable python build if `--enable-shared` is not present.
Modify `.travis.yml` command-line arguments to include `--with-python` for building with python checks.
(This replaces a previous pull request.)
@oskooi @ChristopherHogan


stevengj:
Homer, in the future please do a force-push to the old branch, rather than closing PRs and opening new ones.

PR #84, opened 2017-08-14
Reentrant absorbers in meep geom

HomerReid:
Add support for absorbing layers in `libmeepgeom` in a reentrant way.
See usage example in `libmeepgeom/absorber-1d-ll.cpp.`
(Replaces an earlier pull request.)
@oskooi @ChristopherHogan


HomerReid:
Introduced a name (`absorber_list`) for the libmeepgeom data structure formerly referenced by void pointer. 
See `libmeepgeom/absorber-1d-ll.cpp` for usage example.

PR #85, opened 2017-08-14
Son of fix doc search

HomerReid:
Fixed an issue that was preventing search from working.  
@oskooi @ChristopherHogan


stevengj:
Great, seems to be working now.

PR #86, opened 2017-08-14
harminv calculations with libmeepgeom

HomerReid:
In `libmeepgeom/ring-ll.cpp` I have implemented a harminv calculation that replicates what is done in the `examples/ring.ctl` script. I confirmed that the results for the extracted frequencies and amplitudes are correct and added this to the list of things checked by the `ring-ll` test. (This has the drawback of slowing down the test suite, as it's a costly calculation.)

It might be worthwhile to offer python users a more streamlined interface---there could be a single library function that both **(a)** timesteps for a sufficiently long time to assemble an array of field data for passage to harminv, and then **(b)** calls harminv and simply returns the results, so that the caller doesn't have to hassle with creating the field data array and futzing with harminv parameters.


libmeepgeom/ring-ll.cpp
```diff
@@ -151,6 +151,9 @@ int main(int argc, char *argv[])
 
   master_printf("all harminv results match reference values\n");
 
+  // this seems to be necessary to prevent failures
+  all_wait();
```
stevengj: That worries me, since `f.step()` already should have a barrier.

stevengj:
For Python, I would prefer following the technique similar to the Scheme interface: just pass a `harminv(...)` function to `simulation.run(...)`, where `harminv` creates a callable object that stores the array etcetera and outputs the frequencies at the end.

libmeepgeom/ring-ll.cpp
HomerReid: Without this line, the results are erratic: sometimes everything works fine, sometimes the master process produces proper output but one of the other processes aborts with an MPI error message, and sometimes the abort happens before the master process ever produces output.

Could there be something non-MPI-friendly in harminv?

stevengj: Normally all serial code (like in harminv) is "MPI friendly", because each MPI process is a separate process with its own address space.

HomerReid:
Closing this as the functionality has been superseded by Chris's work on the python side.

PR #87, opened 2017-08-14
mpiPython V2P0

HomerReid:
Added code to `python/meep.i` to autodetect MPI support in meep library and do appropriate initialization if present.

In this case, multiprocessor python meep jobs can be launched by saying e.g.

`mpiexec -n 8 python test_cyl_ellipsoid.py`

To account for the possibility that `.py` tests may be run in parallel, we should generally use `master_printf()` instead of simply `printf()` for output statements, etc. 

I also modified `configure.ac` to check for the presence of the python MPI module (mpi4py), but I didn't implement anything graceful for the case in which it is not found.

The travis build is failing for the same mysterious reason as before: the python tests all succeed on the first pass, but then travis goes back and rebuilds/relinks and the tests all fail on this second pass. I'm still trying to figure this out, but in the meantime this PR may be useful for anybody who wants to accelerate python runs via parallelism. 

@oskooi @ChristopherHogan


python/meep.i
```diff
@@ -199,10 +200,26 @@ static int py_list_to_gobj_list(PyObject *po, geometric_object_list *l);
...
 %ignore inveps_func;
+
+// test for mpi and do requisite setup if present
+%pythoncode %{
+
+if with_mpi():
+    try:
+        from mpi4py import MPI
+    except ImportError:
+        print '\n**\n** failed to load python MPI module (mpi4py)\n**\n'
```
ChristopherHogan: The `print` statements will need parentheses to support python 3.  `print('\n**\n* failed ...')`.  Maybe that's why it's failing on travis.

.travis.yml
```diff
@@ -23,6 +23,6 @@ script:
   - make distclean
   - mkdir build && pushd build
   - (CPPFLAGS=-I${HOME}/local/include LDFLAGS=-L${HOME}/local/lib GEN_CTL_IO=${HOME}/local/bin/gen-ctl-io ../configure --enable-maintainer-mode --prefix=$HOME/local --with-libctl=$HOME/local/share/libctl && make && make check && make install)
-  # Build and test pymeep with python 3.6
+  # Build and test pymeep with python 3.6j
```
ChristopherHogan: Extra `j`.

ChristopherHogan:
Is there any downside to having travis always use MPI so the tests run faster?

stevengj:
Will the tests run faster?  Since Travis is presumably virtualizing some machine, I would think that it gives us the same number of cycles no matter how many virtual processors that we use?

python/tests/test_bend_flux.py
```diff
@@ -118,6 +118,7 @@ def main(args):
 
     bend_flux(True)
     bend_flux(False)
+    mp.all_wait()
```
stevengj: All of these `all_wait` calls should not be necessary—Meep functions (like flux computation) do synchronization exactly where it is needed), and in general aren't a style I would recommend (for the most part, Meep users shouldn't need to care whether their scripts run in serial or in parallel).

python/tests/test_bend_flux.py
```diff
@@ -118,6 +118,7 @@ def main(args):
...
 
 if __name__ == '__main__':
     sys.exit(main(sys.argv))
```
stevengj: Explicitly calling `exit` seems weird to me (and is not really right for MPI either).

python/meep.i
HomerReid: Thanks, fixed.

python/tests/test_bend_flux.py
HomerReid: I removed them.

HomerReid:
`make check` should now be working for all tests (including in python/tests) whether built `--with-mpi` or without. In the former case the python tests are executed under `mpirun` and they do run faster with the parallelism. 

For the travis build, the tests in the top-level `tests` directory (tests of `libmeep` or `libmeep_mpi`) are succeeding whether built `--with-mpi` or without. (I can't tell if there is any speedup in the former case.)

However, for the travis build the tests in `python/tests' are failing when built `--with-mpi`, so in `.travis.yml` I have just turned off MPI support for the python test phase.

.travis.yml
```diff
@@ -21,10 +21,12 @@ script:
   - (cd libctl-src && git checkout master && sh autogen.sh --prefix=$HOME/local --enable-shared && make CFLAGS="-W" CXXFLAGS="-W" && make install)
   - git clone https://github.com/stevengj/harminv
   - (cd harminv && git checkout c221b2bcbaaa761f683aa5e2c6fa7efbbecdca1f && sh autogen.sh --prefix=$HOME/local --enable-shared && make && make install)
+  # test meep build with mpi, without python, omitting test suite
   - (CPPFLAGS=-I${HOME}/local/include LDFLAGS=-L${HOME}/local/lib GEN_CTL_IO=${HOME}/local/bin/gen-ctl-io sh autogen.sh --verbose --enable-maintainer-mode --prefix=$HOME/local --with-libctl=$HOME/local/share/libctl --with-mpi)
+  # test meep build without mpi, without python, with test suite
```
stevengj: It doesn't seem worthwhile to run the tests without python and mpi, and then with python but without mpi, because the latter tests are a strict superset of the former.  Just do the latter.

HomerReid: Done.
I had to remove `near2far` from the travis test suite because it takes too long and the travis build always times out.

stevengj:
`make TESTS=....tests to run.... check` in tests directory to run a subset.

ChristopherHogan:
It would be nice to keep the out-of-tree builds and the python 2 testing.

stevengj:
Handy rebase reference:
```
git checkout master
git pull origin master
git checkout mpiPythonV2P0
git rebase master
```

HomerReid:
Update: This is finally working now.

As a reminder, this PR contains the following improvements:

-- Python module is correctly built when configuring `--with-mpi.` Python-driven parallel meep runs may be launched by saying e.g. `mpirun -np 8 python ring.py.`

-- The travis build matrix involves 4 simultaneous build/test runs, namely
    --A: python 2.7,      without mpi,   first half of core test suite 
    --B: python 3.4,      without-mpi,   second half of core test suite
    --C: python 2.7,      with-mpi,        first half of core test suite 
    --D: python 3.4,      with-mpi,        second half of core test suite

-- I have divided the 18 tests in the core-library test suite (the one invoked by running `make check` in the top-level `tests` directory) into two sets of 9 tests, each set taking roughly as long as the other to run. This reduces the total travis build time to 10-15 minutes including libmeepgeom and python tests.

PR #89, opened 2017-08-25
Consolidate gaussian src time constructors

HomerReid:
Modify one of the two `gaussian_src_time` constructors in `libmeep` so that values of `last_source_time()` now agree.


HomerReid:
Made appropriate modifications to tests. All tests now passing.

PR #95, opened 2017-09-09
Add holey-wvg-cavity.py example and test

ChristopherHogan:
This includes everything necessary for `holey-wvg-cavity.py` to run correctly.

I observed an inconsistency with the tutorial and my local results.  When I run the python **and scheme** versions of `holey-wvg-cavity`, followed by `h5topng -Zc dkbluered holey-wvg-cavity-hz-slice.h5`, the result is this image
holey-wvg-cavity-hz-slice
However, the tutorial shows this image
holey-wvg-cavity-hz-slice
Maybe something in the scheme example changed and uploading the new image was overlooked?

@stevengj @oskooi @HomerReid

### DIFF
--- a/python/Makefile.am
+++ b/python/Makefile.am
@@ -26,6 +26,7 @@ TESTS =                                    \
     $(TEST_DIR)/cyl_ellipsoid.py      \
     $(TEST_DIR)/geom.py               \
     $(TEST_DIR)/holey_wvg_bands.py    \
+    $(TEST_DIR)/holey_wvg_cavity.py   \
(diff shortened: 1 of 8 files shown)


ChristopherHogan:
I removed `arith_sequence` and added `force_complex_fields` as a `Simulation` constructor argument.  I don't see a way around checking for a `Vector3` in `interpolate`. `numpy.linspace` takes scalar start and stop arguments, so I have to call it on v.x, v.y, and v.z, which means I have to know if `v` is a Vector3.

PR #98, opened 2017-09-24
Add 3rd-harm-1d.py example and test

ChristopherHogan:
`set_materials_from_geometry` now takes a `default_material` parameter.

This also includes a couple requests from Ardavan: 
* Rename Pml to PML
* Add `wavelength` param to sources

@stevengj @oskooi @HomerReid

### DIFF
--- a/python/meep.i
+++ b/python/meep.i
@@ -47,6 +47,7 @@ PyObject *py_callback = NULL;
 
 static PyObject *py_geometric_object();
 static PyObject *py_source_time_object();
+static PyObject *py_material_object();
 static PyObject* vec2py(const meep::vec &v);
 static double py_callback_wrap(const meep::vec &v);
 static int pyv3_to_v3(PyObject *po, vector3 *v);
@@ -282,6 +283,16 @@ PyObject *py_do_harminv(PyObject *vals, double dt, double f_min, double f_max, i
     }
 
   delete $1;
+}
+
+%typecheck(SWIG_TYPECHECK_POINTER) material_type _default_material {
+    $1 = PyObject_IsInstance($input, py_material_object());
+}
+
+%typemap(in) material_type _default_material {
+    if(!pymaterial_to_material($input, &$1)) {
+        SWIG_fail;
+    }
 }
 
 // Rename python builtins
--- a/python/meep.i
+++ b/python/meep.i
@@ -285,11 +285,11 @@ PyObject *py_do_harminv(PyObject *vals, double dt, double f_min, double f_max, i
   delete $1;
 }
 
-%typecheck(SWIG_TYPECHECK_POINTER) material_type _default_material {
+%typecheck(SWIG_TYPECHECK_POINTER) material_type {
     $1 = PyObject_IsInstance($input, py_material_object());
 }
 
-%typemap(in) material_type _default_material {
+%typemap(in) material_type {
     if(!pymaterial_to_material($input, &$1)) {
         SWIG_fail;
     }
--- a/python/meep.i
+++ b/python/meep.i
@@ -305,6 +305,8 @@ PyObject *py_do_harminv(PyObject *vals, double dt, double f_min, double f_max, i
 
 %rename(get_field_from_comp) meep::fields::get_field(component, const vec &) const;
 
+%rename(_pml) meep::pml;
+
 // TODO:  Fix these with a typemap when necessary
 %feature("immutable") meep::fields_chunk::connections;
 %feature("immutable") meep::fields_chunk::num_connections;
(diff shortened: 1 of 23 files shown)


oskooi:
Actually, it might be more convenient to use "pml" (all lower case) than "PML".

ChristopherHogan:
Well, lowercase `mp.pml` is the SWIG-wrapped C++ class, so that would cause a name conflict. The python convention for class names is CapitalizedCamelCase, which is why I went with `mp.Pml` originally.

ChristopherHogan:
But if you really want lowercase `mp.pml` for the python calss we can rename the SWIG wrapper `pml` to `_pml`.

oskooi:
PML is an acronym for [perfectly matched layer](https://en.wikipedia.org/wiki/Perfectly_matched_layer). The object should therefore be named either "pml" or "PML". Lower case is preferred.

ChristopherHogan:
I renamed the Python class to `pml` and the SWIG wrapped `pml` to `_pml`.

stevengj:
Since [Python convention is to capitalize words in class names](https://www.python.org/dev/peps/pep-0008/#class-names), and each letter in "PML" stands for a word, I would think that it would be more Pythonic to use `PML` for the class name.

See also the [discussion here](https://stackoverflow.com/questions/2853531/how-do-you-pep-8-name-a-class-whose-name-is-an-acronym); e.g. PEP 8 explicitly gives `HTTPServerError` as a class-name example.

python/meep.i
```diff
@@ -284,6 +285,16 @@ PyObject *py_do_harminv(PyObject *vals, double dt, double f_min, double f_max, i
   delete $1;
 }
 
+%typecheck(SWIG_TYPECHECK_POINTER) material_type _default_material {
```
stevengj: Shouldn't this typecheck etc. apply to any `material_type` parameter, not limited to those named `_default_material`?

python/simulation.py
```diff
@@ -33,9 +33,9 @@ def get_num_args(func):
 def py_v3_to_vec(dims, v3):
     if dims == 1:
         if v3.x == v3.y == v3.z:
-            return mp.vec(dims, v3.x)
+            return mp.vec_from_dim(0, v3.x)
         else:
-            return mp.vec(dims)
+            return mp.vec(v3.z)
```
stevengj: What's the difference between `vec_from_dim(0, x)` and `vec(x)`?     And what's the reason for the `v3.x == v3.y == v3.z` check here?

python/source.py
```diff
@@ -30,26 +30,34 @@ def __init__(self, is_integrated=False):
...
 class ContinuousSource(SourceTime):
 
-    def __init__(self, frequency, start_time=0, end_time=1.0e20, width=0, cutoff=3.0):
+    def __init__(self, frequency=None, start_time=0, end_time=1.0e20, width=0, cutoff=3.0, wavelength=None):
+        if frequency is None and wavelength is None:
+            raise ValueError("Must set either frequency or wavelength in {}.".format(self.__class__.__name__))
+
         super(ContinuousSource, self).__init__()
-        self.frequency = float(frequency)
+        self.frequency = 1 / wavelength if wavelength else frequency
```
stevengj: It doesn't need to be `float(frequency)`?  Hopefully SWIG does the conversion for you if `frequency` is e.g. an integer?

python/typemap_utils.cpp
```diff
@@ -53,6 +53,16 @@ static PyObject *py_meep_src_time_object() {
     return src_time;
 }
 
+static PyObject *py_material_object() {
+    static PyObject *material_object = NULL;
+    if(material_object == NULL) {
```
stevengj: most of the rest of the code puts a space between `if` and the paren

python/meep.i
ChristopherHogan: Fixed.

python/simulation.py
ChristopherHogan: I'm not sure what I was thinking WRT the equality statement. I removed it.

`vec_from_dim` is a renaming of `vec(ndim, double)`.  I did this because python can't distinguish between an enum and an int.  If I create a `vec` with `mp.vec(0, 1)`, intending a 2D vec (even though I didn't explicitly pass doubles), SWIG will call `vec(ndim, double)`, which creates a 1D `vec`. Having `vec_from_dim` eliminates the possibility of that mistake.

python/source.py
ChristopherHogan: Fixed.

python/typemap_utils.cpp
ChristopherHogan: Fixed (in entire file).

PR #100, opened 2017-09-29
rename "pml" to "PML" in Python

stevengj:
Following PEP 8 and discussion in #98.  cc @ChristopherHogan

### DIFF
--- a/python/tests/physical.py
+++ b/python/tests/physical.py
@@ -34,7 +34,7 @@ class TestPhysical(unittest.TestCase):
 
         w = 0.30
 
-        s = mp.structure(self.gv, one, mp._pml(self.ymax / 3.0))
+        s = mp.structure(self.gv, one, mp.pml(self.ymax / 3.0))
 
         f = mp.fields(s)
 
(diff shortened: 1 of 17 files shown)


python/tests/physical.py
```diff
@@ -34,7 +34,7 @@ def radiating_base(self, sq_ratio, solve_cw=True):
 
         w = 0.30
 
-        s = mp.structure(self.gv, one, mp._pml(self.ymax / 3.0))
+        s = mp.structure(self.gv, one, mp.pml(self.ymax / 3.0))
```
stevengj: @ChristopherHogan, should this test be using `mp.PML` (i.e. the high-level object) nowadays?

python/tests/source.py
```diff
@@ -57,7 +57,7 @@ def dummy_eps(v):
         gv = mp.voltwo(16, 16, 10)
         gv.center_origin()
         sym = mp.mirror(mp.Y, gv)
-        the_structure = mp.structure(gv, dummy_eps, mp._pml(2), sym)
+        the_structure = mp.structure(gv, dummy_eps, mp.pml(2), sym)
```
stevengj: also `mp.PML` here?

python/tests/physical.py
ChristopherHogan: This is a low level test, so rewriting the whole thing in the high level API is on my todo list. You can leave it as is for now.

python/tests/source.py
ChristopherHogan: This test creates low level objects to test typemaps, so I think `mp.pml` is fine.

PR #103, opened 2017-10-02
new Python tutorial plus minor fixes to markdown pages in documentation

oskooi:
The first Python tutorial involving the straight and bent waveguides as well as resonant modes of a ring resonator are added to the ReadTheDocs documentation tree. These are based on existing scripts found in  the `python/examples` subdirectory. Also included are minor fixes to the markdown pages in the documentation involving syntax highlighting and a few other things.


doc/docs/C++_Tutorial.md
```diff
@@ -198,26 +198,26 @@ In order to compile your code and link against the Meep library, you must link i
 
 (1) After compiling the Meep package following the instructions elsewhere, place foo.cpp in the `tests/` subdirectory, cd to the same directory, and invoke:
 
-```
+```html
```
stevengj: `html` is wrong here ... this is shell code, so it should probably be `sh`

doc/docs/Download.md
```diff
@@ -34,11 +34,4 @@ You can also install the [parallel version of Meep](http://packages.debian.org/t
...
 ```
 sudo apt-get install meep-mpi-default
-```
-
-Python User Interface
-----------------
-
-A native Python user interface for Meep is currently under development by [Simpetus](http://www.simpetuscloud.com).
-
-A non-native version was developed by the [Photonics Research Group](http://photonics.intec.ugent.be/) at Ghent University, in particular thanks to Emmanuel Lambert, Martin Fiers, Shavkat Nizamov, Martijn Tassaert, Peter Bienstman, and Wim Bogaerts. This [project](http://f.dominec.eu/meep/) is now maintained by Filip Dominec.
```
stevengj: Maybe "official" and "unofficial" rather than "native" and "nonnative"

doc/docs/C++_Tutorial.md
oskooi: Unfortunately `sh` introduces unnecessary highlighting for certain code blocks. `text` produces non-highlighted text but contains no text boxes with sliders; this sometimes gives rise to sentences which awkwardly extend beyond the margins. 

Among the available [language options](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml), `html` produced no or minimal highlighting for text (i.e., non Scheme/Python/Shell). Even though this label seems inappropriate, it produces the right visual layout.

stevengj: Why not just use no label at all, then?  Seems better to use no highlighting at all than highlighting for the wrong language.

oskooi: No label also produces unnecessary highlighting. In any event, I have remove the `html` labels.

PR #105, opened 2017-10-07
Initial high level interface to get_array_slice

ChristopherHogan:
* Updated `examples/cavity_arrayslice.py` to use the higher level `get_array`
* Aligned the formatting of `examples/cavity_arrayslice.py` with the other python examples.
* Added `tests/cavity_arrayslice.py`

Questions:

* Do we want an `ArraySlice` class so the user can get/plot multiple slices, each with its own `dim_sizes`? Currently, each `Simulation` can only track one slice at a time.
* Do we want higher level wrappers around the plot functions, or should we leave that to the user?  `ArraySlice.plot()` might be nice.

If anyone is wondering about the `copy` on line 597 in simulation.py, my preemptive answer is that it seemed nicer than doing something like this:

```python
dimx = self.dim_sizes[0] if self.dim_sizes[0] else 1
dimy = self.dim_sizes[1] if self.dim_sizes[1] else 1
dimz = self.dim_sizes[2] if self.dim_sizes[2] else 1
n = dimx * dimy * dimz
```
@stevengj @HomerReid @oskooi


python/simulation.py
```diff
@@ -590,13 +590,25 @@ def output_component(self, c, h5file=None):
                 self.last_eps_filename = nm
             self.output_h5_hook(nm)
 
+    def get_array(self, center, size, component=mp.Ez):
+        self.dim_sizes = np.zeros(3, dtype=np.int32)
```
stevengj: I don't understand why this is a field of `self`.  Why can't each call to `get_array` allocate new arrays to store the sizes?

Also, there should be an optional parameter `arr=None` so that the user can pass in a pre-allocated array (the code should check that it is the correct size and type … in most cases it would be obtained from a previous call to `get_array`).

stevengj:
I think we should leave plotting to the user

ChristopherHogan:
* Accepts user arrays
* Added complex support
* Support derived components
* Fixed typemaps.
* I think Steven wanted the C++ functions with `slice_length` params completely removed. Since I would have to repeat the work those functions do in typemaps, it seemed convenient to keep them and just get rid of the `slice_length` arguments.  I can move them to `python/typemap_utils.cpp` if you don't want them exposed at all.
* Added test cases. Note that the complex test cases don't verify any data yet; they just check that the array is complex and the right size.

ChristopherHogan:
I rebased this after #122 was merged and now all the tests pass.

PR #106, opened 2017-10-08
Add ring-cyl.py example and test

ChristopherHogan:
@stevengj @HomerReid @oskooi


stevengj:
Needs a rebase.

ChristopherHogan:
* User `Volume`s now get converted to cylindrical if necessary. 
* `geometry` and `sources` default to `[]`.

Is there anywhere you can think of besides `in_volume` where the user might pass in a `Volume`?

stevengj:
In Scheme, you pass a volume object to `electric-energy-in-box` etcetera, to `integrate-field-function` etc., and to `output-farfields `.

PR #114, opened 2017-10-19
User defined materials

HomerReid:
First pass at adding user-defined materials in libmeepgeom.


libmeepgeom/meepgeom.cpp
```diff
@@ -676,10 +661,17 @@ bool geom_epsilon::get_material_pt(material_type &material, const meep::vec &r)
...
-    // material = eval_material_func(md->user_material_func,md->user_data,p);
+  else if (md->which_subclass == material_data::MATERIAL_FUNCTION) { 
+
+    // is *THIS* what I want to do here? This can't be right...
+    material_data *new_material_data = (material_data *)malloc(sizeof(material_data));
+    new_material_data->which_subclass = material_data::MEDIUM;
+    new_material_data->user_func = 0;
+    new_material_data->user_data = 0;
+    new_material_data->medium = md->user_func(p, md->user_data);
+    material.data = (void *)new_material_data;
```
stevengj: No, I think it should be something like:
```jl
material. which_subclass = material_data::MEDIUM;
material.medium = md->user_func(p, md->user_data);
destroy_material = true;
```

HomerReid: The bit that I still can't figure out here is that `material_type_destroy` is an empty function in libctlgeom. It doesn't deallocate anything. So how are all of these `medium` structures newly allocated by `user_func` ever going to get freed up? Am I just overlooking something in all the flipping back and forth between hierarchies of data structures in different code files and different source trees?

HomerReid:
Added a line to `material_type_destroy()` to free the `medium` structure if it has been allocated.

I'm still unclear on the vision here--for example, it looks like trying to free the `medium` structure is the *wrong* thing to do if `material_type_destroy()` is invoked from one of the branches of `get_material_pt()` that calls `epsilon_file_material()`. Also, why again does the user function allocate a new structure each time it is called, instead of filling in an existing structure passed to it? I think this has something to do with interfacing to scheme and/or python, but it seems to add complexity. Maybe things will be clarified by switching to a more C++-friendly approach using classes and polymorphism, for which I will submit a separate PR. 

@oskooi @ChristopherHogan

HomerReid:
Made modifications discussed in teleconference; all tests passing including new `user-defined-material` test.

libmeepgeom/meepgeom.cpp
```diff
@@ -289,38 +250,52 @@ static geom_box gv2box(const meep::volume &v)
...
-// TODO rename this to something more descriptive like
-//      is_user_defined_material?
 static bool is_variable(material_type mt)
 {
-  return (mt->which_subclass == material_data::MATERIAL_FUNCTION);
+  return (mt->which_subclass == material_data::MATERIAL_USER);
 }
 static bool is_variable(void* md) { return is_variable((material_type) md); }
 
 static bool is_self(material_type md)
```
stevengj: This shouldn't be called `is_self` any more?

libmeepgeom/meepgeom.cpp
```diff
@@ -1209,50 +1133,68 @@ static double get_chi2(meep::component c, const medium_struct *m) {
...
+bool geom_epsilon::has_chi(meep::component c, int p)
 {
   medium_struct *mm;
 
   for (int i = 0; i < geometry.num_items; ++i)
-   if (    is_medium(geometry.items[i].material, &mm)
-        && get_chi2(c, mm) != 0
-      ) return true;
+   if ( is_medium_or_user(geometry.items[i].material, &mm) )
+    if (get_chi(c, mm, p)!=0)
```
stevengj: Would be clearer to just remove the `!= 0` since `get_chi` returns a boolean (0/1) value.

libmeepgeom/meepgeom.cpp
```diff
@@ -289,38 +250,52 @@ static geom_box gv2box(const meep::volume &v)
...
+
+static bool is_medium(void* md, medium_struct **m) 
+ { return is_medium((material_type) md, m); }
+
+static bool is_medium_or_user(material_type md, medium_struct **m)
+{ 
+  if (    md->which_subclass == material_data::MEDIUM
+       || md->which_subclass == material_data::MATERIAL_USER
+     )
+   { *m = &(md->medium);
```
stevengj: I don't understand how this works.  If it is `MATERIAL_USER`, don't you need to call the user function to get the material?

libmeepgeom/meepgeom.cpp
```diff
@@ -416,19 +369,23 @@ static meep::realnum linear_interpolate(
 static void epsilon_file_material(material_data *md, vector3 p)
```
stevengj: Since you can now assign this material to an object, maybe `p` should be offset by the center of the object when computing ε.  (So the file is almost like a "texture" applied to the object?)

libmeepgeom/meepgeom.cpp
HomerReid: Fixed, good point.

libmeepgeom/meepgeom.cpp
HomerReid: `get_chi` returns a `double` value. (Almost all of the code here is taken directly from the original; I just eliminated some redundancy between the chi2/chi3 cases.)

libmeepgeom/meepgeom.cpp
HomerReid: The idea was that this function, like `material_epsmu`, should only ever be called *after* a call to `get_material_pt` and could thus rely on `md->medium` having been initialized.

But, to eliminate any assumptions, I modified `make_user_material` to call the user's function (at the origin) to fill in `md->medium` before returning the new `material_data` it creates. Thus, as long as user-defined materials are always created by calling `make_user_material,` the `md->medium` structure will always have been filled in by the user's function and will thus accurately reflect whether or not the material has nonlinearities (and, eventually, perhaps susceptibilities).

libmeepgeom/meepgeom.cpp
HomerReid: At the moment, file-based materials are representations of entire geometries, not of individual objects---if two or more objects exist in the original geometry, they are all flattened into a single spatially-varying background. I think I can picture what you have in mind, but wouldn't this require a little more sophistication in the interface? For example, we would want to check that the material file specified for an object describes a grid whose dimensions are commensurate with the object's bounding box? If you have a specific application in mind we could think about how to define appropriate conventions.

stevengj:
Great, thanks.  @ChristopherHogan, is there a Python API to the epsilon-file functionality in this PR (where you can assign a different file to each object/material)?

PR #122, opened 2017-11-03
Array slice fix

HomerReid:
Previously the contribution of each chunk to the slice array was getting written to the same piece of the array. Added a notion of "slice_chunk_offset" so that each chunk writes to its own slot in the array.


HomerReid:
Fixed issues associated with array_slice strides and offsets.
@oskooi @ChristopherHogan

PR #124, opened 2017-11-05
Use global Vector3 for material function

ChristopherHogan:
The change from returning a `tuple` to a `Vector3` in #118 resulted in `set_epsilon` being about 10x slower.  This PR implements Steven's idea of keeping a global `Vector3` instead of creating it every time the material function is called.  It mitigates the performance issue substantially, but this version is still about 25% slower for `set_epsilon` than the original `tuple` version.
@stevengj @oskooi @HomerReid


stevengj:
It would be good to have a test case for `set_epsilon`, by the way.

PR #127, opened 2017-11-11
Add python ports of all remaining examples

ChristopherHogan:
* Add metal-cavity-ldos.py
* Add parallel-wvgs-force.py
* Add test ldos.py
* Add test force.py
* I innstalled MPB locally and tested `fields.add_eigenmode_source` manually, but I didn't set MPB up in travis. I assume we'll want to do that at some point.

@stevengj @oskooi @HomerReid

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ dist: trusty
 sudo: false
 
 ##################################################
-# required linux packages and python packages 
+# required linux packages and python packages
 ##################################################
 addons:
   apt:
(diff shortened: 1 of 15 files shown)


ChristopherHogan:
* Added wvg-src.py and test.
* Added MPB to travis.
* Added `output_png`

ChristopherHogan:
* Added pw-source.py and sanity test
* Added support for user amplitude functions

PR #128, opened 2017-11-17
don't put sources at non-owned points

### LINKED ISSUE #23: LDOS in 2D
I tried comparing LDOS computed in meep with analytical value in vacuum, but could not get the two results agree.

Analytically, LDOS in 2D = 2*\omega/(\pi c^2) \* Tr[ Im [G] ], G = dyadic Greens function of the vacuum. In 2D, Im[G] = 1/4, for all three diagonal components. 

I use a continuous-wave point source at a fixed frequency to compute the LDOS at that frequency and at that location(I could use other temporal sources to get LDOS at several frequencies in a single FDTD run, but for now, I just want to compare the numerical with analytical results and for simplicity I choose a continuous-wave source).

I am using meep as a c++ library and use dft_ldos object with update and ldos functions to compute the LDOS.

The result is off by an order of magnitude.

What is the unit of LDOS in 2D in meep?

stevengj:
Hopefully, this fixes #23.

### DIFF
--- a/src/sources.cpp
+++ b/src/sources.cpp
@@ -124,7 +124,7 @@ complex<double> continuous_src_time::dipole(double time) const
   else {
     double ts = (time - start_time) / width - slowness;
     double te = (end_time - time) / width - slowness;
-    
+
     return exp(complex<double>(0,-1.0) * (2*pi)*freq*time) * amp
       * (1.0 + tanh(ts))  // goes from 0 to 2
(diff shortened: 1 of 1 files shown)


oskooi:
This is working. I verified using a simple test case involving placing the ldos/source point at the origin of 2d vacuum surrounded by PML. The ldos value is now unchanged even when intersecting 2 chunks.

PR #133, opened 2017-11-23
local density of states (LDOS) and optical forces Python tutorials for documentation

oskooi:
Add new Python tutorials on local density of states (LDOS) and optical forces to the documentation along with other minor fixes and tweaks.


oskooi:
The `make check` failure after the second commit is puzzling since this PR only relates to the documentation.The travis logs indicate that the failing test is `python/tests/wvg_src.py` which is addressed by #131 (not yet merged). Not sure why this failure did not occur in #127 when this test first appeared.

PR #135, opened 2017-12-06
Add missing functionality from meep.scm

ChristopherHogan:
@stevengj @oskooi

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,15 +67,15 @@ script:
   - make
 #
   - pushd tests
-  - make check TESTS=${CORETESTS} VERBOSE=1 || cat test-suite.log
+  - make check TESTS=${CORETESTS} VERBOSE=1 || (cat test-suite.log && exit 1)
   - popd
 #
   - pushd libmeepgeom
-  - make check || cat test-suite.log
+  - make check || (cat test-suite.log && exit 1)
   - popd
 #
   - pushd python
-  - make check || cat test-suite.log
+  - make check || (cat test-suite.log && exit 1)
   - popd
 #
   - popd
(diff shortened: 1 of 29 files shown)


ChristopherHogan:
Added the remaining missing functionality from meep.scm and tests.

ChristopherHogan:
* Fixed field function tests.
* Added `ALL_COMPONENTS`
* Replaced `if mp.comm.Get_rank() != 0` with `if not mp.am_master()`

This is green now.

ChristopherHogan:
I added some memory leak fixes for `material_type` to this PR. I verified that the current pymeep leaks memory by monitoring `top` while running scripts like
```python
while True:
    b = mp.Block(size=mp.Vector3(1, 1, 1), center=mp.Vector3())
    mp.point_in_objectp(mp.Vector3(), b)
```
Commit ef8d931 fixes the issue. This will probably be cleaned up once `material_data` and `medium_struct` become proper classes.

ChristopherHogan:
Travis is currently failing on master but reporting the build as passing. 

Adding `set -e` made the builds fail fast, which is what we wanted. However, that caused the `cat test-suite.log` code to be skipped, so I changed the `make check` lines to `make check || cat test-suite.log`. However, that line exits with 0 which makes travis think everything is ok.  The (hopefully final) solution is 
```bash
make check || (cat test-suite.log && exit 1)
```

ChristopherHogan:
* Added `ring-mode-overlap.py` example that uses `integrate2_field_function`.
* Added test for `integrate2_field_function` to `tests/field_functions.py`
* Fixed bug in `py_field_func_wrap` revealed by `integrate2` test case.

PR #139, opened 2017-12-07
Only print on master node in MPI mode

ChristopherHogan:
* Removed the `mpirun` flags that are specific to OpenMPI (anaconda uses MPICH).
* Python 2 files that call `print` in MPI mode will need `from __future__ import print_function` at the top.

@stevengj @oskooi @HomerReid

### DIFF
--- a/python/meep.i
+++ b/python/meep.i
@@ -577,11 +577,4 @@ extern boolean point_in_objectp(vector3 p, GEOMETRIC_OBJECT o);
                 import sys
                 saved_stdout = sys.stdout
                 sys.stdout = open(os.devnull, 'w')
-
-            import atexit
-
-            @atexit.register
-            def close_dev_null():
-                if comm.Get_rank() != 0:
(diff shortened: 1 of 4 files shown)


python/meep.i
```diff
@@ -571,4 +571,10 @@ extern boolean point_in_objectp(vector3 p, GEOMETRIC_OBJECT o);
             # this variable reference is needed for lazy initialization of MPI
             comm = MPI.COMM_WORLD
             master_printf('\n**\n** successfully loaded python MPI module (mpi4py)\n**\n')
+
+            if comm.Get_rank() != 0:
+                import sys
+                saved_stdout = sys.stdout
+                sys.stdout = None
```
stevengj: This might break some code that expects `stdout` to be an IO stream?   Seems better to set it to `open(os.devnull, 'w')`.

ChristopherHogan: Changed.

PR #147, opened 2017-12-15
Allign with docs and use parallel make in travis

ChristopherHogan:
* public `run_k_points`
* Simulation cell_size and resolution defaults


ChristopherHogan:
* Removed default values for `cell_size` and `resolution`.

This is green now.

Issue #12, opened 2015-02-06
Lorentzian check reports instability even when stable


FilipDominec:
I would like to open the broad question of FDTD numerical stability. By trial-and-error I isolated two independent criteria of how a definition of a material can make the simulation unstable (the third mechanism of unstable PML for grazing incidence wave has been resolved in MEEP 1.2). In the treatise that follows, I tried to describe these two criteria.

In last 3 years I tested different materials thoroughly and these rules seem to hold so far (though I have no theory why).

What does not always hold, however, is the corresponding function `lorentzian_unstable' (in src/susceptibility.cpp). MEEP checks for the position of the oscillator pole in the frequency complex plane, based on some mathematics that I admit not to understand yet, and sometimes reports false positives. In the cases near the edge of stability it reports instability and aborts.
I experimentally verified that by disabling the function, recompiling and running the same simulation again, no instability developed even in very long time.

Of course, there are also false negatives, due to the second criterion: The simulation goes unstable whenever the permittivity at high frequencies is too low.

I suggest that first of all, the `lorentzian_unstable' should be changed to a warning only, it should be called only once after the initialisation (not in n-times every step), and no matter what the outcome, the simulation should continue. In the future, we should thoroughly test out the hypothesis attached below. This starts to be a bit experimental research instead of programming, but with so many people computing some plasmonics etc, it is important to make it clear.

Looking forward to your comments!
Filip
----

![fdtd_stability1](https://cloud.githubusercontent.com/assets/511950/6085296/c57879f8-ae38-11e4-88c5-95d7ddfb0910.png)
![fdtd_stability2](https://cloud.githubusercontent.com/assets/511950/6085297/c8ad5102-ae38-11e4-8cc0-c7b15f929e3d.png)

stevengj:
@FilipDominec, wouldn't the normal way to proceed be to perform a Von Neumann stability analysis for a homogeneous material? Trial-and-error does not inspire confidence, although it can be a useful first step.

I haven't thought about this for a while, but the `lorentzian_unstable` is probably overly conservative because it considers only the P equation by itself, not coupled to Maxwell.   It is supposed to be just a standard z-transform analysis of a linear-ODE discretization, although of course it's possible there is a bug.

Because the check is performed outside the `LOOP_OVER_VOL_OWNED`, there is no significant performance penalty to performing it on any iteration, and it is easier than adding a flag to the structure to record whether the check has been performed.

FilipDominec:
[...]

PR #150, opened 2017-12-19
replace stability error with warning message in Lorentzian susceptibilities

oskooi:
This fixes a bug due to rounding-error in Lorentzian susceptibilities which causes erratic behavior when identifying unstable Lorentzian poles as reported by Ian Sage on the [mailing list](https://www.mail-archive.com/meep-discuss@ab-initio.mit.edu/msg05891.html).

### DIFF
--- a/src/susceptibility.cpp
+++ b/src/susceptibility.cpp
@@ -157,7 +157,7 @@ static bool lorentzian_unstable(double omega_0, double gamma, double dt) {
   double w = 2*pi*omega_0, g = 2*pi*gamma;
   double g2 = g*dt/2, w2 = (w*dt)*(w*dt);
   double b = (1 - w2/2) / (1 + g2), c = (1 - g2) / (1 + g2);
-  return b*b > c && float(2*b*b - c + 2*fabs(b)*sqrt(b*b - c)) > 1;
+  return b*b > c && 2*b*b - c + 2*fabs(b)*sqrt(b*b - c) > 1;
 }
 
 #define SWAP(t,a,b) { t SWAP_temp = a; a = b; b = SWAP_temp; }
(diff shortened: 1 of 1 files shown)


stevengj:
See #12.   The real problem here is not a rounding error, but rather that this stability check is apparently too conservative.  Unfortunately, no one has gone through the analysis (or found the relevant paper) to figure out the "real" stability criterion.

My suggestion is to simply replace the error here with a warning (saying that the simulation is potentially unstable, and if you see divergences you may want to reduce the courant number.

stevengj:
Alternatively, the right way to compute this accurately is to replace
`2*b*b - c + 2*fabs(b)*sqrt(b*b - c)` with `c*c / (2*b*b - c - 2*fabs(b)*sqrt(b*b - c))`, which computes the same thing in exact arithmetic but is much more accurate when `c` is very small compared to `b`.   Let's just use this more accurate expression for now, as the more conservative change.

(Just throwing away half the significant digits, with `float`, is not the right fix.)

src/susceptibility.cpp
```diff
@@ -179,7 +179,7 @@ void lorentzian_susceptibility::update_P
 
   if (!no_omega_0_denominator && gamma >= 0
       && lorentzian_unstable(omega_0, gamma, dt))
-    abort("Lorentzian pole at too high a frequency %g for stability with dt = %g: reduce the Courant factor, increase the resolution, or use a different dielectric model\n", omega_0, dt);
+    master_printf("WARNING: simulation is potentially unstable as the Lorentzian pole is at too high a frequency %g for stability with dt = %g. In the event the fields diverge: reduce the Courant factor, increase the resolution, or use a different dielectric model\n", omega_0, dt);
```
stevengj: I don't think we want to print out a warning on every timestep.

stevengj: Maybe we should just omit the warning entirely.  We don't warn about other potential instabilities.

PR #156, opened 2017-12-28
Support user defined materials in python

ChristopherHogan:
This needs a better test. I've done manual verification with the output of `libmeepgeom/user-defined-material.cpp` using `h5diff`.  I also need to update the docs once #155 is merged.
@stevengj @oskooi @HomerReid


ChristopherHogan:
* Material function now takes 1 arg and returns a `Medium`.
* Added optimized path for `epsilon_func`.
* Added proper tests for both `material_function` and `epsilon_func` both as `default_material` and in `GeometricObjects`.
* Added `epsilon_file_input` support (no test).
* Updated docs with material function info.

PR #162, opened 2018-01-02
Get epsilon_input_file working and add test

ChristopherHogan:
The `epsilon_file_material` function in `libctl/structure.cpp` and `libmeepgeom/meepgeom.cpp` uses `geometry_lattice.size` to compute the arguments to `linear_interpolate`, which it then calls. 

My understanding is that the code
```cpp
switch (s->gv.dim)
{ case meep::D1:   dimensions=1;
                     size.z = gv.nz()/resolution;
                     break;
      case meep::D2:   dimensions=2;
                     size.x = gv.nx()/resolution;
                     size.y = gv.ny()/resolution;
                     break;
      case meep::D3:   dimensions=3;
                     size.x = gv.nx()/resolution;
                     size.y = gv.ny()/resolution;
                     size.z = gv.nz()/resolution;
                     break;
      case meep::Dcyl: dimensions= CYLINDRICAL;
                     size.x = gv.nr()/resolution;
                     size.z = gv.nz()/resolution;
                     break;
 };
geometry_lattice.size = size;
```
in `set_materials_from_geometry` is meant to reconstruct the `cell_size` that is passed to the `Simulation` from Python. However, at this point symmetries have already been processed (because we need a `structure` before calling `set_materials_from_geom`), which means `gv.halve()` was called which makes `geometry_lattice.size` differ in scheme and python in `epsilon_file_material`.

This fix passes the `cell_size` to `set_materials_from_geometry`, but it would be better to reconstruct the cell size from the grid volume.  Is there a way to do that after symmetries have been applied?
@

### DIFF
--- a/libmeepgeom/meepgeom.cpp
+++ b/libmeepgeom/meepgeom.cpp
@@ -1480,7 +1480,7 @@ void set_materials_from_geometry(meep::structure *s,
 
   dimensions=3;
   vector3 size = {0.0,0.0,0.0};
-  switch (s->gv.dim)
+  switch (s->user_volume.dim)
    { case meep::D1:   dimensions=1;
                       size.z = s->user_volume.nz()/resolution;
                       break;
--- a/libmeepgeom/meepgeom.cpp
+++ b/libmeepgeom/meepgeom.cpp
@@ -1458,8 +1458,7 @@ void add_absorbing_layer(absorber_list alist,
 /***************************************************************/
 /***************************************************************/
 /***************************************************************/
-void set_materials_from_geometry(vector3 size,
-                                 meep::structure *s,
+void set_materials_from_geometry(meep::structure *s,
                                  geometric_object_list g,
                                  bool use_anisotropic_averaging,
                                  double tol,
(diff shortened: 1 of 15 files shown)


ChristopherHogan:
@oskooi @HomerReid

ChristopherHogan:
I'm now passing `None` instead of `dummy_eps` to the `structure` constructor, which is a great improvement, but I still have the problem with symmetries and the grid volume.  We have to either
1. Pass the original cell size to `set_materials_from_geometry`, or
2. Somehow calculate the original cell size after symmetries have been applied and the grid volume has been `halve`ed. You started to mention some "fold" function, but we got off topic and forgot to come back to this.

libmeepgeom/meepgeom.cpp
```diff
@@ -1464,7 +1464,8 @@ void add_absorbing_layer(absorber_list alist,
 /***************************************************************/
 /***************************************************************/
 /***************************************************************/
-void set_materials_from_geometry(meep::structure *s,
+void set_materials_from_geometry(vector3 size,
+                                 meep::structure *s,
```
stevengj: Can't you get `size` from `s->user_volume` (the grid `volume` before symmetry reduction) or `s->v` (the corresponding `volume`)?

ChristopherHogan: `s->user_volume` works. Thanks.  This is green now and should be ready to merge.

PR #163, opened 2018-01-04
Replace h5utils with python libraries

ChristopherHogan:
This is my initial work in replacing h5utils with matplotlib. I got an animation working with matplotlib but it looks a little off (at least before it hits the bend) and will probably need some tweaking.
ez
Note that saving the gif to disk with `anim.save('tutorial-ez.gif', writer='imagemagick')` took 2 hours on my machine.  Additionally, the `y=-3.5` slice looks slightly off, and I selected the y index by trial and error.
tutorial-ez-y-slice

It seems to me that if we really want to replace h5utils with matplotlib, we will need a nice API that mimics the h5utils interface.  As this would be time consuming, I wanted to make sure everyone is on the same page before I continue this work.
@oskooi @stevengj @HomerReid


ChristopherHogan:
Added an example of using `get_array`.

PR #164, opened 2018-01-04
Refactor geom_epsilon::eff_chi1inv_row for MPB

ChristopherHogan:
This is so MPB can get the whole matrix back (via `eff_chi1inv_matrix`) instead of just a row.
@stevengj @oskooi @HomerReid


ChristopherHogan:
* `eff_chi1inv_row` now calls `eff_chi1inv_matrix` and extracts the rows.
* Removed intermediate `meps_inv`.

PR #166, opened 2018-01-05
no longer install MPI version with -mpi suffix

stevengj:
As discussed, there is no real reason to install the parallel version as `meep-mpi`.  It can be installed simply as `meep`, *replacing* the serial version, because if you run an MPI program without `mpirun` it is equivalent to `mpirun -np 1`.


stevengj:
Failing test is a pypi timeout.

PR #177, opened 2018-01-18
 Revise pympi startup message

HomerReid:
Revise startup message printed by MPI-enabled python meep module to mimic what is printed by the libctl meep binary, e.g.. `Using MPI version 3.0, 8 processes.`


stevengj:
Shouldn’t it only print from the master process?

PR #182, opened 2018-01-22
Fit user volumes to Simulation dimensions

ChristopherHogan:
Prevent user from having to specify `dims` argument to `Volume`.
@stevengj @oskooi @HomerReid

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -128,13 +128,6 @@ class Volume(object):
 
         self.swigobj = mp.volume(vec1, vec2)
 
-    # To prevent the user from having to specify `dims` and `is_cylindrical`
-    # to Volumes they create, the library will adjust them appropriately based
-    # on the settings in the Simulation instance. This method must be called on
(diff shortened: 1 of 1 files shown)


python/simulation.py
```diff
@@ -645,7 +649,7 @@ def get_farfield(self, f, v):
         return mp._get_farfield(f, py_v3_to_vec(self.dimensions, v, is_cylindrical=self.is_cylindrical))
 
     def output_farfields(self, near2far, fname, where, resolution):
-        vol = where.to_cylindrical() if self.is_cylindrical else where
+        vol = where._fit_to_simulation(self.dimensions, self.is_cylindrical)
```
stevengj: why not have this be a method of Simulation so that you don't need to pass `self.*`?

i.e. do `vol = _fixvolume(where)` where `_fixvolume(v)` is a simulation method that calls `Volume(v.center, v.size, dims=self.dimensions, is_cylindrical=self.is_cylindrical)`

PR #183, opened 2018-01-23
Fix source distribution and run make distcheck in Travis

ChristopherHogan:
These are the changes required for `make distcheck` to work. Travis will now run `make distcheck` so we can catch distribution bugs earlier. However, since that would double the build times, we only want Travis to do this once per day, not every commit. After this is green and approved, but before merging, I'll set Travis to only run `make distcheck`  if `[ "${TRAVIS_EVENT_TYPE}" == cron ]`. Then Steven can add a daily cron job in the Travis settings.
@stevengj @oskooi @HomerReid

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,10 +76,11 @@ script:
 #
   - pushd python
   - make ${MKCHECKFLAGS} check || (cat test-suite.log && exit 1)
+  - make clean
   - popd
 #
   # Only run distcheck in nightly builds because it doubles the build time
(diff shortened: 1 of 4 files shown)


stevengj:
Maybe we could do `make distcheck` *instead* of `make check`, since the former runs the latter?

ChristopherHogan:
I thought about that, but the nice thing about running `make check` is that it lets us split the `CORETESTS` over the different builds instead of running them all for both python versions. Is there a way to have the best of both worlds?

stevengj:
Can't we use a `CORETESTS` environment variable?

ChristopherHogan:
How do we tell `make distcheck` that `TESTS` should equal `CORETESTS`, but only when it's running `make check` in `meep/tests`?

PR #191, opened 2018-01-30
Python MPB interface

### LINKED ISSUE #180: meep segfaults
How do I debug this (on Ubuntu 16, with latest meep version)?

kbriggs> ./try_meep_01.py
-----------
Initializing structure...
Segmentation fault (core dumped)

ChristopherHogan:
Adds MPB-based mode-solver support from Python.

Eventually, we might want to refactor this into MPB itself, but that will require refactoring libmeepgeom into a meep-independent part, so for now we will require users to install Meep in order to get the Python MPB interface.

Incidentally, this PR fixes #180.


ChristopherHogan:
## Description of Changes by File
### .travis.yml
Add `h5py` and `scipy` as dependencies.  `Scipy` is only used in `mpb-tutorial.py` as replacements for some of the optimization functions in `libctl/base/math-utils.scm`. These packages probably need to be checked in `configure`.

### Makefile.am
Build `libpympb` if we have `MPB` and `libctlgeom`.

### configure.ac
* Make a `WITH_MPB` variable available in automake.
* Make a `@LIBCTLGEOM_LIBS@` variable available in automake.

### libmeepgeom/material_data.hpp
* Convert `susceptibility_list`, `medium_struct`, and `material_data` into C++ style structs with default values set in a constructor.  This is what fixes #180.

### libmeepgeom/meepgeom.cpp
* Removed `vacuum_medium` because the default constructor gives us that.
* Remove `static` keyword from a couple functions so they can be called from `libpympb`.

### libpympb/Makefile.am
Builds `libpympb.so`, which is the C++ portion of MPB that python calls into.

### libpympb/pympb.cpp and pympb.hpp
Contains the `mode_solver` class, which holds all the MPB state. This class is wrapped by SWIG, and all MPB state is accessed from Python through an instance of this class. Most of the code in here is taken from MPB.
* `dielectric_function` is what's passed to `set_maxwell_dielectric`. 
* `mode_solver::init` corresponds to `init_params` in `mpb.c`. It's called in the constructor though, so the Python API just creates a `mode_solver` and calls `solve_kpoint` for each kpoint. The `get_Xfield` functions all take arrays and a size so we can pass a numpy array to them as an output argument. The numpy arrays are created in the python API layer with the help of `get_dims`.

### python /Makefile.am
* Build the MPB interface if `WITH_MPB` is defined.

### python/examples/mpb-tutorial.py
A port of `tutorial.ctl` from MPB. This demonstrates all the functionality currently supported.

### python/geom.py
Add the `Matrix` and `Lattice` Python classes, along with `geometric_object_lattice_duplicates` for the "Point defect state" tutorial.

### python/meep.i
I put the common typemap functions used by both pymeep and pympb in `typemap_utils.cpp`.  The typemap helpers specific to each module are defined in their respectiv ".i" file.

### python/mpb.i
* Wraps the `mode_solver` class
* Defines some typemap helpers specific to pympb
* Adds typmaps for getting data from/to the `mode_solver`.

### python/solver.py
Defines the main Python object, `ModeSolver`, that wraps the `mode_solver`. Analogous to `simulation.py` and `Simulation`. Most of the code is ported from `mpb.scm`. The `output_*_field` functions were ported from `output_field_to_file` in `mpb/mpb/fields.c`.

### python/tests/data/tutorial-*.h5
Reference data collected from scheme runs (setting `deterministic=true` and passing a NULL `mean_epsilon_func` to `set_maxwell_dielectric`). This data is read by the unit tests and compared to the python results.

### python/tests/geom.py
Add test case for #180 

### python/tests/mpb.py
Unit tests for all MPB functionality.  There is a test for every section of the tutorial in `mpb-tutorial.py` comparing results obtained from scheme. The `setUp` function registers a function to clean up all the generated h5 files after the tests complete.

The only test that's questionable is `test_output_efield_z`. It doesn't have the accuracy that the rest do. I've spent hours in the debugger trying to track it down, but haven't been successful. 

### python/typemap_utils.cpp
Refactored to include the functions that are used in both `meep.i` and `mpb.i`.

@stevengj @oskooi @HomerReid

PR #194, opened 2018-02-02
Portable sed commands

ChristopherHogan:
Travis didn't catch this because it doesn't use SWIG 3.0.12. Creating a backup file seems to be the only portable way of doing it.
@stevengj @oskooi


stevengj:
This only affects `--enable-maintainer-mode`, right?

It would be good to add a [`clean-local` target](https://www.gnu.org/software/automake/manual/html_node/Clean.html) to remove the `.bak` file.

ChristopherHogan:
> This only affects --enable-maintainer-mode, right?

Correct.
> It would be good to add a clean-local target to remove the .bak file.

Done.

PR #202, opened 2018-02-08
Fix 2d Absorbers

### LINKED ISSUE #198: Absorbing boundaries in Python
Absorbing boundaries don't seem to be working correctly in the Python interface. (I'm running the Feb 2 version installed via Anaconda.) They give the same result as no boundaries.

Here's a simple working example (point source at the centre of a 2D box):
````
import meep as mp

source = mp.Source(src = mp.GaussianSource(frequency = 0.1, fwidth = 0.1),
				   component = mp.Hz,
				   center = mp.Vector3())

sim = mp.Simulation(cell_size = mp.Vector3(20, 20, mp.inf),
					resolution = 10,
					sources = [source],
					dimensions = 2,
					boundary_layers = [mp.Absorber(5)])


sim.run(mp.at_beginning(mp.output_epsilon),
		mp.to_appended("hz", mp.at_every(5.0, mp.output_hfield_z)),
		mp.in_volume(mp.Volume(mp.Vector3()), mp.to_appended("hz-src", mp.output_hfield_z)),
		until_after_sources = 1000)
````

The wave just bounces around like a resonant cavity. The results seem to be identical if I set `boundary_layers = []`. On the other hand, setting `boundary_layers = [mp.PML(5)]` gives the expected absorbing boundary behaviour.

ChristopherHogan:
Fixes #198. The `strength` property in scheme is marked "obsolete" so I didn't add it to the Python `Absorber`. Should it be removed from C++?
@stevengj @HomerReid @oskooi

### DIFF
--- a/libmeepgeom/meepgeom.cpp
+++ b/libmeepgeom/meepgeom.cpp
@@ -1436,14 +1436,13 @@ void destroy_absorber_list(absorber_list alist)
 
 void add_absorbing_layer(absorber_list alist,
                          double thickness, int direction, int side,
-                         double strength, double R_asymptotic, double mean_stretch,
+                         double R_asymptotic, double mean_stretch,
                          meep::pml_profile_func func, void *func_data)
 {
   absorber myabsorber;
   myabsorber.thickness=thickness;
   myabsorber.direction=direction;
   myabsorber.side=side;
-  myabsorber.strength=strength;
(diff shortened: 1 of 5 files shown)


stevengj:
Probably remove it from C++.

stevengj:
The point is that this parameter is redundant with `R-asymptotic` — there is no point in having both.

PR #210, opened 2018-02-19
Add convenience version of set_materials_from_geometry

ChristopherHogan:
This makes it easier to update the geometry or default_material from a step function.
Requries #209. 
@stevengj @oskooi


ChristopherHogan:
The issue was that calling `fields::remove_susceptibilities` called `fields_chunk::changing_structure`, which recreated the chunks even though they were supposed to be shared. The commit in this PR titled "Fields need to know if chunks are shared" works around the issue by copying the `shared_chunks` variable from the `structure`.

Initially you said to call `fields::remove_susceptibilites` from python and `structure::remove_susceptibilites` in `set_materials_from_geometry`.  However, now that the `fields` and `structure` can share chunks, isn't it only necessary to call it on the `structure` because then the `fields` will see the updated chunks? If that's the case, we don't really need the "Fields need to know if chunks are shared" commit (without this commit, the test passes if I omit calling `fields.remove_susceptibilites` in `simulation.py:ModeSovler.set_materials`).

PR #212, opened 2018-02-26
Lower MPB test accuracy

ChristopherHogan:
This should make the MPB tests less dependent on a specific BLAS/LAPACK implementation.
* Collected all test data with tolerance = 1e-12. 
* Added `fix_field_phase` calls to all tests that check fields.
* Reduced accepted precision of tests to 1e-3.
* Fix bug in filename creation
@stevengj @oskooi


ChristopherHogan:
Replace by #217

PR #214, opened 2018-02-27
Return complex numbers in flux routines

ChristopherHogan:
Ardavan implemented the C++ changes and I fixed scheme and python.

There were some compiler warnings about some of the `fprintf` calls in `wvg_taper.cpp` so I changed it to print the real part. I'm not sure if that's what we want or not.

@stevengj @oskooi @HomerReid


stevengj:
What is this change for? In what case do you need the imaginary part?

oskooi:
The phase is necessary to compute [scattering parameters](https://en.wikipedia.org/wiki/Scattering_parameters#Two-Port_S-Parameters), etc., particularly when modeling component devices for integrated circuits.

stevengj:
Don’t those use the phase of the fields, not the flux? Since the Poynting flux involves E*xH, the phase of the fields is lost anyway.

oskooi:
The phase of the reflection/transmission coefficients is necessary to develop compact models of component devices for circuit simulators (i.e., via [Verilog-A](https://en.wikipedia.org/wiki/Verilog-A), etc.). For reference including several examples, see Section 9.3.2 of the book [Silicon Photonics Design](https://www.amazon.com/Silicon-Photonics-Design-Devices-Systems/dp/1107085454). Another important application for the phase is when computing the transmission coefficient of circularly-polarized light (i.e., for metasurface or Pancharatnam lenses used in liquid-crystal displays, etc).

stevengj:
As I said, the phase of the transmission/scattering coefficients is the phase of the fields, *not* the phase of the Poynting vector.  This PR is not what you want.

stevengj:
Note that Homer’s mode expansion code already returns the complex amplitude, which is what you want. 

(For scattering coefficients etc you need the modal expansion anyway unless it is single-mode.)

oskooi:
Most structures of interest are single mode. The only advantage to using the mode-expansion approach in this case is that it doesn't require a separate normalization run which can be costly. Otherwise, the complex reflection/transmission coefficients from the flux calculation are exactly equivalent to the scattering parameters.

stevengj:
I'm getting tired of saying the same thing over and over: the phase of E*×H (the phase of the flux) is not the same thing as the phase of the fields or modes.  E*×H is not the complex transmission coefficient.

PR #215, opened 2018-02-28
Fix mpb in installation package

ChristopherHogan:
`make install` wasn't installing the pympb stuff correctly.
@stevengj @oskooi

### DIFF
--- a/python/Makefile.am
+++ b/python/Makefile.am
@@ -70,11 +70,19 @@ TEST_EXTENSIONS = .py
 PY_LOG_COMPILER = $(RUNCODE) $(PYTHON)
 TESTS_ENVIRONMENT = export PYTHONPATH=$(abs_top_builddir)/python:$$PYTHONPATH;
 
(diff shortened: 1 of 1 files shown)


python/Makefile.am
```diff
@@ -70,12 +70,20 @@ TEST_EXTENSIONS = .py
...
+  pkgpyexec_LTLIBRARIES = _meep.la
+  pympb_DATA = mpb.py solver.py
+  pympb_LTLIBRARIES = _mpb.la
 endif
 
+install-data-hook:
+	mv $(DESTDIR)$(pkgpythondir)/mpb/mpb.py $(DESTDIR)$(pkgpythondir)/mpb/__init__.py
+
+uninstall-hook:
+	rm $(DESTDIR)$(pkgpythondir)/mpb/__init__.py
```
stevengj: Should be `rm -f`

ChristopherHogan: Done

PR #226, opened 2018-03-09
rehauled flux/mode overlap calculation, added HDF5 output routines for all DFT structures

HomerReid:
…


src/meep.hpp
```diff
@@ -1490,30 +1508,41 @@ class fields {
...
+                                             component c_conjugate=Ex,
+                                             const volume *where=0);
+  
+  // output DFT fields to HDF5 file
+  void output_dft_components(dft_chunk **chunklists, int num_chunklists,
+                             const char *HDF5FileName, const volume *where=0);
+
+  void output_dft_flux(dft_flux flux, const char *HDF5FileName, const volume *where=0);
+  void output_dft_force(dft_force force, const char *HDF5FileName, const volume *where=0);
+  void output_dft_near2far(dft_near2far n2f, const char *HDF5FileName, const volume *where=0);
```
stevengj: Just rename these to `output_dft`, overloaded on the first argument.

Add a new `dft_fields` object that just takes the frequencies and a list of field components (`Ex`, `Hz`, etcetera) to Fourier transform in some volume, and support these in `output_dft`.

oskooi:
A possible test for the output DFT routine involves a comparison of the fundamental mode of a [ring resonator](https://github.com/stevengj/meep/blob/master/python/examples/ring.py) using two different methods/sources: (1) a pulse excitation with output DFT called at the end of the time stepping when the fields have sufficiently decayed away (at only the center/mode frequency), and (2) a continuous-wave source using the [solve_cw](http://meep.readthedocs.io/en/latest/Python_User_Interface/#frequency-domain-solver) routine with the existing [field output routines](http://meep.readthedocs.io/en/latest/Python_User_Interface/#output-functions) called after the iterative solver has converged. The test would verify that the L2 norm of the difference in the fields converges to zero as the resolution is increased.

libmeepgeom/dft-fields.cpp
```diff
@@ -0,0 +1,116 @@
...
+     while( f.round_time() < f.last_source_time() + 100.0 )
+      f.step();
+
+     f.output_dft(dftFlux,   "dft_flux");
+     f.output_dft(dftFields, "dft_fields");
+   }
+  else
+   { 
+     f.add_point_source(Ez, continuous_src_time(fcen), x0);
+     f.solve_cw();
```
oskooi: For better convergence, you may want to use a larger value of L (one of the parameters used in the [biCGSTAB iterative solver](http://meep.readthedocs.io/en/latest/Python_User_Interface/#frequency-domain-solver)), e.g., f.solve_cw(1e-8, 10000, 10). Default is 2. Benchmarking this type of problem has shown that larger L values can also increase the accuracy of the results.

libmeepgeom/dft-fields.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+     while( f.round_time() < f.last_source_time() + 100.0 )
+      f.step();
+
+     f.output_dft(dftFlux,   "dft_flux");
+     f.output_dft(dftFields, "dft_fields");
+   }
+  else
+   { 
+     f.add_point_source(Ez, continuous_src_time(fcen), x0);
```
oskooi: Consider adding a 'width' parameter to the CW source which describes the temporal width of the smoothing, e.g., f.add_point_source(Ez, continuous_src_time(fcen, 1/df), x0). Default is 0. This will reduce high-frequency oscillations due to its abrupt turn on which may improve convergence.

libmeepgeom/dft-fields.cpp
```diff
@@ -0,0 +1,116 @@
...
+  geometric_object_list g={ 2, objects };
+  meep_geom::set_materials_from_geometry(&the_structure, g);
+  fields f(&the_structure);
+  f.step(); // single timestep to trigger internal initialization
+  f.output_hdf5(Dielectric, f.v);
+
+  /***************************************************************/
+  /***************************************************************/
+  /***************************************************************/
+  double fcen = 0.15;  // ; pulse center frequency
```
oskooi: Note that, as shown in the [tutorial example](http://meep.readthedocs.io/en/latest/Python_Tutorials/Basics/#modes-of-a-ring-resonator), the fundamental mode (with three-fold symmetry) for this structure is at a frequency of 0.118. The next higher-order modes are at 0.147 and 0.175. The convergence properties of the frequency-domain solver may be improved by setting the CW source frequency to one of these modes.

src/meep.hpp
```diff
@@ -1488,31 +1522,52 @@ class fields {
   dft_flux add_dft_flux(const volume_list *where,
 			double freq_min, double freq_max, int Nfreq);
 
+  dft_fields add_dft_fields(component *components, int num_components,
+                            const volume where,
+                            double freq_min, double freq_max, int Nfreq);
+  dft_fields add_dft_fields(int *components, int num_components,
```
ChristopherHogan: If this override is only useful for python, you could add it to `meep.i` using the [extend](http://www.swig.org/Doc3.0/SWIGDocumentation.html#Python_nn43) directive instead of including it in the library.

The other option, since `num_components` will never be very big (right?), would be to accept a list instead of a numpy array.  Then the typemaps would look like this:

```
%typemap(in) (meep::component *components, int num_components) {
    if (!PyList_Check($input)) {
        PyErr_SetString(PyExc_ValueError, "Expected a list");
        SWIG_fail;
    }
    $2 = PyList_Size($input);
    $1 = new meep::component[$2];

    for (Py_ssize_t i = 0; i < $2; i++) {
        $1[i] = (meep::component)PyInteger_AsLong(PyList_GetItem(cs, i));
    }
}

%typemap(freearg) (meep::component *components, int num_components) {
    delete[] $1;
}
```

ChristopherHogan:
It would be nice to have a test that calls `add_dft_fields` from python. I can add that if you want.

HomerReid:
Updated the `dft-fields.cpp` test code following suggestions of @oskooi. After normalizing both DFT and CW datasets by maximum field amplitude, the results look visually identical, i.e. here's a plot of Ez magnitude.

dftvscw

On the other hand, the test of decreasing L2 norm between the datasets as the resolution increases seems to be more subtle, and this would also be a fairly expensive calculation to run on e.g. travis, so I'm not sure we want to include that in the unit test suite.

oskooi:
Have you also verified that the results match for the parallel case when there are multiple chunks? Even if we don't add this to the test suite, it would be worth adding to the new tutorial for the frequency-domain solver (#232) as an example of the output DFT routines.

libmeepgeom/dft-fields.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+/***************************************************************/
+/***************************************************************/
+/***************************************************************/
+int main(int argc, char *argv[])
+{
+  initialize mpi(argc, argv);
+
+#define NUM_RESOLUTIONS 3    
+  double resolutions[NUM_RESOLUTIONS]={10.0, 20.0, 40.0};
```
oskooi: Since the range of resolutions up to 40 makes this test too time intensive and thus impractical to be included in the test suite, can the same trend of decreasing error with resolution be demonstrated with a smaller range (i.e., 8, 12, 16) which executes much faster?

src/meep.hpp
HomerReid: Thanks! I put this into `meep.i.` I changed `cs` to `$input` in the line containing `PyInteger_AsLong.`

HomerReid:
Fixed compilation issue in new `dft-fields` test that had caused travis failure. Tests now passing.

PR #229, opened 2018-03-09
Add high level modal_volume_in_box and solve_cw

ChristopherHogan:
@stevengj @oskooi


stevengj:
A separate PR can add docs and a test for solve_cw

PR #233, opened 2018-03-14
Initial mpb-data support for python

ChristopherHogan:
* `mpb_data.py`: Includes functionality of `mpb-data` but accepts and returns numpy arrays instead of dealing with h5 files.
* `mpb_data_file_io.py`: Exact port of `mpb-data` to python. I'll probably end up removing this unless anyone thinks it will be useful.
* `mpb_data_analysis.py`: Example of using `mpb_data.py` that follows the "Data Analysis Tutorial" from the MPB docs.

TODO
* Get correct Bloch wavevectors and lattice (these were obtained from the h5 file before).
* Handle cvector case (some datasets are processed multiple times and I'm not sure why).

@stevengj @oskooi @HomerReid


ChristopherHogan:
This is ready now.

* Wrote triple loop in C
* Fixed all get_*field functions to return correct shape
* Adding working test for `handle_cvector`.
* `MPBData.convert()` takes a kpoint as argument, and the constructor takes a lattice via `ModeSolver.get_lattice()`.
* `python/examples/mpb_data_analysis.py` is the equivalent of the data analysis tutorial (the tri-rods portion).

I still need to investigate subclassing `np.ndarray` to store the lattice.

PR #234, opened 2018-03-15
Initial support for Hermitian epsilon

ChristopherHogan:
This turned out to be pretty complicated.

TODO
* Should/can meep use complex mu/eps offdiagonals, or is it only for MPB?
* Need tests.
* Typemaps aren't quite right yet.

@stevengj @oskooi @HomerReid


ChristopherHogan:
This has a working test now. For meep, imaginary parts of `mu_offdiag` and `epsilon_offdiag` are silently ignored. Would you prefer to print a warning, or abort the simulation?

stevengj:
Probably should abort the simulation if the off-diag imaginary parts are nonzero.

ChristopherHogan:
Replaced by #243

PR #238, opened 2018-03-19
Update susceptibility.cpp

fesc3555:
Added checks in update_P: check, whether there is a diagonal element before you add contributions from off-diagonal elements. This avoids instabilities at boundaries for non-diagonal sigma.

### DIFF
--- a/src/susceptibility.cpp
+++ b/src/susceptibility.cpp
@@ -208,7 +208,9 @@ void lorentzian_susceptibility::update_P
 	SWAP(const realnum *, s1, s2);
       }
       if (s1 && s2) { // 3x3 anisotropic
-	LOOP_OVER_VOL_OWNED(gv, c, i) {
+	LOOP_OVER_VOL_OWNED(gv, c, i) if(s[i]!=0) {
+		//"if" to avoid instabilities coming from yee-grid shifts of off-diagnal elements
++		//FIXME: this prevents materials with pure off-diagonal sigma from working
 	  realnum pcur = p[i];
 	  p[i] = gamma1inv * (pcur * (2 - omega0dtsqr_denom)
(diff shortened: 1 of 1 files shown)


src/susceptibility.cpp
```diff
@@ -208,24 +208,30 @@ void lorentzian_susceptibility::update_P
 	SWAP(const realnum *, s1, s2);
       }
       if (s1 && s2) { // 3x3 anisotropic
-	LOOP_OVER_VOL_OWNED(gv, c, i) {
+	LOOP_OVER_VOL_OWNED(gv, c, i) if(s[i]!=0) {
+		//"if" to avoid instabilities coming from yee-grid shifts of off-diagnal elements
++		//FIXME: this prevents materials with pure off-diagonal sigma from working
```
stevengj: Wouldn't a purely off-diagonal sigma matrix be indefinite, and hence lead to both loss and gain?

src/susceptibility.cpp
```diff
@@ -208,24 +208,30 @@ void lorentzian_susceptibility::update_P
...
 	  realnum pcur = p[i];
 	  p[i] = gamma1inv * (pcur * (2 - omega0dtsqr_denom)
 			      - gamma1 * pp[i]
 			      + omega0dtsqr * (s[i] * w[i]
 					       + OFFDIAG(s1,w1,is1,is)
 					       + OFFDIAG(s2,w2,is2,is)));
 	  pp[i] = pcur;
+	}else{
+	  p[i]=0; //seems unnecessary, but no harm is done here
```
stevengj: Delete the `else` clause, since `p[i]` is initialized to zero anyway.

stevengj:

PR #244, opened 2018-03-22
Refactor get_array

ChristopherHogan:
Replace `center` and `size` arguments to `get_array` with a `meep.Volume`.
@stevengj @oskooi @HomerReid

### DIFF
--- a/doc/docs/Python_Tutorials/Basics.md
+++ b/doc/docs/Python_Tutorials/Basics.md
@@ -101,7 +101,7 @@ import matplotlib.pyplot as plt
 We will first create an image of the dielectric function $\varepsilon$. This involves obtaining a slice of the data using the [`get_array`](../Python_User_Interface/#array-slices) routine which outputs to a NumPy array and then display the results.
 
 ```py
-eps_data = sim.get_array(mp.Volume(mp.Vector3(), size=cell), mp.Dielectric)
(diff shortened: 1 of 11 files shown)


stevengj:
Could all the functions that accept a `Volume` also accept a named tuple `(center=..., size=...)`, just by adding an appropriate `Volume` constructor or typemap?

stevengj:
Or just have the functions that accept a volume optionally take `center` and `size` keywords instead.

ChristopherHogan:
* All functions accepting a volume now accept an optional `center` and `size`. 
* Updated tests and docs
* Fixed a couple examples that had errors.

doc/docs/Python_Tutorials/Basics.md
```diff
@@ -209,7 +209,7 @@ vals = []
 def get_slice(sim):
     center = mp.Vector3(0, -3.5)
     size = mp.Vector3(16, 0)
-    vals.append(sim.get_array(center=center, size=size, component=mp.Ez))
+    vals.append(sim.get_array(mp.Volume(center, size=size), component=mp.Ez))
```
stevengj: Using `center=x, size=y` in the tutorial seems better, since it is shorter than constructing a `Volume` object.

ChristopherHogan:
* Reverted tutorial code for `get_array` to use `center` and `size` again instead of `Volume`.

PR #248, opened 2018-03-23
High level get_eigenmode_coefficients

ChristopherHogan:
This will require a documentation change, but I think Ardavan is working on that documentation now, so it will need his PR first.
@stevengj @oskooi @HomerReid

### DIFF
--- a/python/examples/mode-decomposition.py
+++ b/python/examples/mode-decomposition.py
@@ -76,20 +76,16 @@ def main(args):
                         sources=sources,
                         symmetries=symmetries)
 
-    xm = -0.5*sx + dpml + 0.5*Lw; # x-coordinate of monitor
-    mflux = sim.add_flux(fcen, 0, 1, mp.FluxRegion(center=mp.Vector3(xm,0), size=mp.Vector3(0,sy-2*dpml)));
+    xm = -0.5*sx + dpml + 0.5*Lw  # x-coordinate of monitor
+    mflux = sim.add_eigenmode(fcen, 0, 1, mp.FluxRegion(center=mp.Vector3(xm,0), size=mp.Vector3(0,sy-2*dpml)))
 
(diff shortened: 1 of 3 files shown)


ChristopherHogan:
When called from Python, `get_eigenmode_coefficients` is unable to infer the direction, resulting in
```
meep: cannot determine normal direction in get_eigenmode_coefficients
```
Looking at @HomerReid 's `wvg-taper.cpp` (which has been deleted from the repository, so I had to look at an old commit), he calls the overload of `add_dft_flux` that takes a `direction` as the first argument. Then, `get_eigenmode_coefficients` gets the direction via `dft_flux::normal_direction`. 

However, in Python and Scheme, the overload of `add_dft_flux` that takes a `volume_list` as the first parameter is called, which returns a `dft_flux` with `NO_DIRECTION` as the direction. This happens in `_add_fluxish_stuff`.

```python
def _add_fluxish_stuff(self, add_dft_stuff, fcen, df, nfreq, stufflist):
        vol_list = None

        for s in stufflist:
            v = Volume(center=s.center, size=s.size, dims=self.dimensions,
                       is_cylindrical=self.is_cylindrical)
            d0 = s.direction
            d = self.fields.normal_direction(v.swigobj) if d0 < 0 else d0
            c = mp.direction_component(mp.Sx, d)
            v2 = Volume(center=s.center, size=s.size, dims=self.dimensions,
                        is_cylindrical=self.is_cylindrical).swigobj
            vol_list = mp.make_volume_list(v2, c, s.weight, vol_list)

        stuff = add_dft_stuff(vol_list, fcen - df / 2, fcen + df / 2, nfreq)
        vol_list.__swig_destroy__(vol_list)
        return stuff
```
Note that the direction, `d`, is available, but `stuff.normal_direction` is `NO_DIRECTION`. How can we make this work?

PR #250, opened 2018-03-26
Small typo corrections

scimax:
- I guess the argument of $\epsilon$ should be $\omega$
- I'm not a native speaker but something is wrong in the second addition


doc/docs/Materials.md
```diff
@@ -52,7 +52,7 @@ which is equivalent to the Lorentzian model except that the $\omega_n^2$ term ha
 Numerical Stability
 -------------------
 
-If a Lorentzian resonance $\omega_n$ is specified at too high a frequency relative to the time discretization $\Delta t$, the simulation becomes unstable. Essentially, the problem is that $\mathbf{P}_n$ oscillates too fast compared with the time discretization for the discretization to work properly. If this happens, there are three workarounds: increase the resolution which increases the resolution in both space and time, decrease the Courant factor which decreases $\Delta t$ compared to $\Delta x$, or use a different model function for your dielectric response.
+If a Lorentzian resonance $\omega_n$ is specified at too high frequency relative to the time discretization $\Delta t$, the simulation becomes unstable. Essentially, the problem is that $\mathbf{P}_n$ oscillates too fast compared with the time discretization for the discretization to work properly. If this happens, there are three workarounds: increase the resolution which increases the resolution in both space and time, decrease the Courant factor which decreases $\Delta t$ compared to $\Delta x$, or use a different model function for your dielectric response.
```
stevengj: The original "too high a frequency" is correct.  (Google "too high a")

stevengj:
Right, the argument should be ω.

(Probably better to just use Unicode here and write ε(ω) directly, rather than using `$...$`.)

doc/docs/Materials.md
scimax: Thank you!

stevengj: Can you revert that part of the PR so that I can merge it?

PR #258, opened 2018-03-28
docs fixes and tweaks

oskooi:


doc/docs/C++_Tutorial.md
```diff
@@ -60,7 +60,7 @@ double eps(const vec &p) {
 }
 ```
 
-which returns the dielectric function $\varepsilon(\mathbf{x})$ which is just a 2$\times$3 rectangle of $\varepsilon$=12 in the upper-left corner. Unlike in the Scheme interface, by default the origin of the coordinate system is at the *corner* of the computational cell.
```
stevengj: Here is another case where you used unicode in a LaTeX equation; it isn't safe to rely on this rendering correctly because it is not allowed in "standard" LaTeX.

doc/docs/C++_Tutorial.md
```diff
@@ -120,7 +120,7 @@ int main(int argc, char **argv) {
  fields f(&s);
 ```
 
-Note the constructor for the `grid_volume` class in 1d which takes as parameters the size of the cell and the resolution. The sources in Meep are excitations of the polarization vector in $\mathbf{D}=\varepsilon\mathbf{E}+\mathbf{P}$. The polarization can be any one of the six cartesian or cylindrical fields. There are a variety of sources including dipole and current sources, gaussian pulses and a continuous wave sources. For now, we use a gaussian source centered at the midgap frequency with a narrow width, along with the start time and end time of the source.
+Note the constructor for the `grid_volume` class in 1d which takes as parameters the size of the cell and the resolution. The sources in Meep are excitations of the polarization vector in $\mathbf{D}=ε\mathbf{E}+\mathbf{P}$. The polarization can be any one of the six cartesian or cylindrical fields. There are a variety of sources including dipole and current sources, gaussian pulses and a continuous wave sources. For now, we use a gaussian source centered at the midgap frequency with a narrow width, along with the start time and end time of the source.
```
stevengj: also here.

doc/docs/C++_Tutorial.md
```diff
@@ -167,7 +167,7 @@ The `get_field` function does exactly that, obtains the value of the field at a
...
 ```
 
-The integer returned by `harminv` is the number of mode frequencies obtained from the input data `p`. The particular call to `harminv` included passing the arrays by value and telling `harminv` to look for frequencies within 20% of the mid gap frequency up to a maximum of 5 bands. At this point, the necessary information to compute the quality has been stored in the `freq_re` and `freq_im` arrays and we compute the quality factor using the formula, $Q=-\omega_r/2\omega_i$. All that is left to do, is to print these results with the quality factor:
+The integer returned by `harminv` is the number of mode frequencies obtained from the input data `p`. The particular call to `harminv` included passing the arrays by value and telling `harminv` to look for frequencies within 20% of the mid gap frequency up to a maximum of 5 bands. At this point, the necessary information to compute the quality has been stored in the `freq_re` and `freq_im` arrays and we compute the quality factor using the formula, $Q=-ω_r/2ω_i$. All that is left to do, is to print these results with the quality factor:
```
stevengj: and here.... probably other places too

PR #261, opened 2018-03-29
structure dump/load functions

stevengj:
This adds `structure::dump(filename)` and `structure::load(filename)` functions to dump the raw ε data to disk and load it back, for doing multiple simulations with the same materials but different sources etcetera.

Requires that the dump/load simulations have the same chunks (i.e. the same grid, number of processors, and PML).

Currently only stores ε and μ, and not nonlinear coefficients or polarizability.

Also makes the `h5file` API support 64-bit sizes.

(Missing tests, but I tried a quick test by hand and it seems to be functioning as intended.)


stevengj:
In Python, maybe `sim.dump_structure(filename)` and `sim.load_structure(filename)`.

PR #262, opened 2018-03-29
Only run mpb tests serially

ChristopherHogan:
Since PyMPB doesn't support mpi, we should always run the tests serially.
@stevengj @oskooi

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,7 +63,7 @@ script:
   - export MEEP_VERSION=$(./configure -V | grep meep | awk '{print $3}')
   - mkdir -p build && pushd build
   - ../configure --enable-maintainer-mode --prefix=$HOME/local --with-libctl=$HOME/local/share/libctl ${MPICONF}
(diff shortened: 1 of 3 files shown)


stevengj:
When MPB creates the `h5file`, pass `false` for the `parallel` argument of the constructor.  That will allow the file to be opened and read only from the master process.

ChristopherHogan:
Passing `parallel=false` doesn't work. The problem is in `h5file.cpp:369`:
```C
  if (!parallel) {
    *rank = broadcast(0, *rank);
    broadcast(0, dims, *rank);
    size_t N = 1;
    for (int i = 0; i < *rank; ++i)
      N *= dims[i];
    if (!am_master())
      data = new realnum[N];
    broadcast(0, data, N);
  }
```
Since the calls to `MPI_Bcast` are collective communication calls, they involve synchronization points, and since only the master process runs this code (with the changes in this PR), it waits and never makes it out of this block. Maybe I can find a way for autotools to always run the mpb.py test without MPI.

ChristopherHogan:
I'm just skipping the MPB tests in MPI builds for now. We could add a travis step for parallel builds where we manually run python/tests/mpb.py serially if you think it's necessary.

PR #264, opened 2018-03-30
Improve options to multiply Bloch phase

ChristopherHogan:
You had requested that `MPBArray` hold a flag that `MPBData` could check to determine whether or not `multiply_bloch_phase` has been called on the input array. The issue there is that `MPBData` would need a `ModeSolver` instance to call `multiply_bloch_phase`.

This adds two ways to control the Bloch phase.
1. `get_*field` has a `bloch_phase` flag that can be set to true but is false by default.
2. `multiply_bloch_phase` can take an optional numpy array. If called with no argument, it does its work on whatever's in `curfield`.

If `MPBData` gets an array that hasn't been multiplied by the Bloch phase, it prints an error message explaining how to do that.

@stevengj @oskooi

### DIFF
--- a/libpympb/pympb.cpp
+++ b/libpympb/pympb.cpp
@@ -1889,9 +1889,11 @@ cvector3 mode_solver::get_field_point(vector3 p) {
   return F;
 }
 
-void mode_solver::multiply_bloch_phase() {
+void mode_solver::multiply_bloch_phase(std::complex<double> *cdata) {
 
   std::vector<mpb_real> kvector = get_output_k();
+
+  scalar_complex *data = cdata ? (scalar_complex*)cdata : (scalar_complex*)curfield;
 
   int dims[] = {mdata->nx, mdata->ny, mdata->nz};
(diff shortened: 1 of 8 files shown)


stevengj:
In your `map_data` routine you should be able to do the equivalent of `multiply_bloch_phase`.  In particular, add a parameter `int multiply_bloch_phase` to `map_data` indicating whether you want `map_data` to do this.  Then, immediately after the [MODF_POSITIVE lines](https://github.com/stevengj/meep/blob/ecba723c77c49727eea8e2da16b149295c10c0d3/libpympb/pympb.cpp#L2592-L2594)  (*before* `ADJ_POINT`), do:
```c
if (multiply_bloch_phase) {
    xi += x; yi += y; zi += z;
}
```

ChristopherHogan:
* Now multiplying Bloch phase in `map_data` (if necessary).
* The `get_*field` functions multiply by the Bloch phase by default now.

I think this is ready.

PR #265, opened 2018-04-02
MPB Additions

ChristopherHogan:
* `optimize_grid_size` is true by default
* Added `allow_negative_epsilon`
* Use Python's `current_k` and remove C++ `get_cur_kvector`
* Add ability to use `eigensolver_davdison`.
* Allow passing `eigensolver_nwork` and `eigensolver_block_size` from Python to C++

@stevengj @oskooi


ChristopherHogan:
Removed `eigensolver_davidson`. This should be ready now.

PR #266, opened 2018-04-02
Add load_structure and dump_structure with test

ChristopherHogan:
@stevengj @oskooi

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -255,7 +255,8 @@ class Simulation(object):
                  accurate_fields_near_cylorigin=False,
                  filename_prefix='',
                  output_volume=None,
-                 output_single_precision=False):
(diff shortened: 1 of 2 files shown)


ChristopherHogan:
The MPI version of the test I wrote hangs on `structure::dump`. The master process is waiting on `structure_dump.cpp:60` in `partial_sum_to_all`, but the other process is stuck in `h5file.cpp:127` at `begin_critical_section`. Any advice?

stevengj:
Just add the following line after the `h5file` is constructed:
```c
CHECK(file.ok(), "error opening HDF5 dump file");
```
in both `dump` and `load`.

ChristopherHogan:
After rebasing, the tests are now passing.

PR #270, opened 2018-04-05
Fix non Hermitian epsilon case

### LINKED ISSUE #269: Build error (problem in pympb.cpp) "Cannot convert cnumber to mpb_real {aka double}"
Hi,

I'm encountering a new issue when building meep from source.  I get the following error message when at the meep "make" step:

`pympb.cpp: In member function ‘void py_mpb::mode_solver::material_epsmu(meep_geom::material_type, symmetric_matrix*, symmetric_matrix*, bool)’:
pympb.cpp:563:20: error: cannot convert ‘cnumber’ to ‘mpb_real {aka double}’ in assignment
         epsmu->m01 = md->medium.epsilon_offdiag.x;`

The compilation then aborts.  Any idea what might be going on?  The build script I had was working with previous versions of meep.  Below is my build script:

Thank you!

```#!/bin/bash

set -e

RPATH_FLAGS="-Wl,-rpath,/usr/local/lib:/usr/lib/x86_64-linux-gnu/hdf5/openmpi"
MY_LDFLAGS="-L/usr/local/lib -L/usr/lib/x86_64-linux-gnu/hdf5/openmpi ${RPATH_FLAGS}"
MY_CPPFLAGS="-I/usr/local/include -I/usr/include/hdf5/openmpi"

export CC=mpicc
export CXX=mpic++

cd ~/install
git clone https://github.com/stevengj/mpb.git
cd mpb/
sh autogen.sh --enable-shared CC=mpicc CXX=mpic++ LDFLAGS="${MY_LDFLAGS}" CPPFLAGS="${MY_CPPFLAGS}"
make && sudo make install

sudo pip install --upgrade pip
pip install --no-cache-dir mpi4py
export HDF5_MPI="ON"
pip install --no-binary=h5py h5py

cd ~/install
git clone https://github.com/stevengj/meep.git
cd meep/
sh autogen.sh --enable-shared --with-mpi PYTHON=/home/dkita/anaconda2/bin/python \
    CC=mpicc CXX=mpic++ LDFLAGS="${MY_LDFLAGS}" CPPFLAGS="${MY_CPPFLAGS}"
make && sudo make install

ChristopherHogan:
Fixes #269. Only use the real part of `<epsilon|mu>_offdiag` when MPB wasn't built with `WITH_HERMITIAN_EPSILON`. I recall doing this before, but it must have been undone by a rebase.
@stevengj @oskooi


stevengj:
It seems like we should throw an error if nonzero imaginary parts are passed in this case?

ChristopherHogan:
Added checks for imaginary parts of epsilon and mu offdiags when not using Hermitian epsilon. For mediums, the check is performed once for the default material and once for each geometric object's material. User and file materials are checked on every `material_epsmu` call.

PR #271, opened 2018-04-06
Fixed Harminv result column headers

shanham:
The output from Harminv writes 7 column headers when there should only be 6. Fixed by combining the two incorrect column headers 'imag.' and 'freq.' into a single column header 'imag. freq.'.


stevengj:
LGTM.

PR #273, opened 2018-04-08
add support for multilevel susceptibility

oskooi:
For now, this is an experimental feature. Several bugs were corrected in the original implementation which had been disabled. A [tutorial example](http://meep.readthedocs.io/en/latest/Scheme_Tutorials/Multilevel_Atomic_Susceptibility/) is already available in the documentation. We still need to add Python wrappers similar to those in Scheme.


stevengj:
Did you ever fix the agreement with @acerjan's code?

oskooi:
Not yet; we are still working on that. This PR provides the basic functionality which we will be using for further testing. The main challenge has been getting the units/constants to be similar when comparing the oscillator and bloch equations with SALT. The underlying update equations in this PR seem to be correct.

One further note: the example added in this PR (multilevel-atom.ctl) works for the serial/uniprocessor version but crashes for the parallel case with a segmentation fault. The cause has been tricky to track down with gdb. Any ideas?

acerjan:
I'm happy to help with the algorithm side, to the best of my ability. And I slightly disagree with @oskooi above, the underlying update equations are close, but we need to add a term or two to see exact agreement, as per the notes/plots i sent a few months ago, attached here.
[osc_eqns_comparison.pdf](https://github.com/stevengj/meep/files/1890944/osc_eqns_comparison.pdf)

PR #274, opened 2018-04-09
Add Diamond example from the data analysis tutorial

ChristopherHogan:
* Make vector field arrays 4d
* Add diamond example from the data analysis tutorial and a test. The example still needs plotting code, but I think Ardavan said he would take care of that.
* Add `get_dpwr` and `get_bpwr` (only the `output_` versions were available previously).

@stevengj @oskooi


python/solver.py
```diff
@@ -266,6 +257,24 @@ def _get_field(self, f, band, bloch_phase):
...
+        arr = np.zeros(np.prod(dims))
+        self.mode_solver.get_curfield(arr)
+        arr = np.reshape(arr, dims)
+
+        return MPBArray(arr, self.get_lattice(), self.current_k, bloch_phase=bloch_phase)
+
+    def get_dpwr(self, band):
+        self.get_dfield(band)
+        self.compute_field_energy()
+        return self.get_curfield_as_array()
```
stevengj: Shouldn’t Bloch_phase be false here?

ChristopherHogan: Yes, you're right. Fixed. I also fixed `get_tot_pwr`. It was multiplying by the Bloch phase too.

PR #276, opened 2018-04-10
Don't multiply Poynting vector by Bloch phase

ChristopherHogan:
I realized this was wrong while going over MPB documentation.
@stevengj @oskooi


python/solver.py
```diff
@@ -518,7 +516,8 @@ def _output_vector_field(self, curfield_type, fname_prefix, output_k, component)
             f['Bloch wavevector'] = np.array(output_k)
             self._write_lattice_vectors(f)
 
-            self.mode_solver.multiply_bloch_phase()
+            if curfield_type != 'v':
+                self.mode_solver.multiply_bloch_phase()
```
stevengj: Right, this corresponds to https://github.com/stevengj/mpb/blob/0fa695b03b642c423b5a1b262c2804d2418f62ef/mpb/fields.c#L714-L723

python/solver.py
```diff
@@ -178,8 +178,8 @@ def multiply_bloch_phase(self, arr):
         return np.reshape(arr, dims)
 
     def get_poynting(self, which_band):
-        e = self.get_efield(which_band).ravel()
-        h = self.get_hfield(which_band).ravel()
+        e = self.get_efield(which_band, False).ravel()
+        h = self.get_hfield(which_band, False).ravel()
```
stevengj: It actually doesn't matter if you multiply by the Bloch phase here or not, since it will end up cancelling in the Poynting vector.  But we might as well pass `False` here to avoid doing unnecessary work.

PR #279, opened 2018-04-12
Support Unicode strings in Python 2

### LINKED ISSUE #278: typechecking error in meep/simulation.py
Python meep seems to return a type-error when you set `epsilon_input_file` with a string (`epsilon_input_file="somefile.h5"`).  The type-error happens at the `set_materials_from_geometry()` call in meep/simulation.py (line 396).  However, casting it to a string fixes this:

`epsilon_input_file=str("somefile.h5")`.

I think this might be fixed, if line [394](https://github.com/stevengj/meep/blob/7998e6dc96eac709211f923128021bddbb9b7ee1/python/simulation.py) of `simulation.py` is changed from:

`self.default_material = self.epsilon_input_file`

to:

`self.default_material = str(self.epsilon_input_file)`

ChristopherHogan:
Fixes #278 
@stevengj @oskooi

### DIFF
--- a/python/typemap_utils.cpp
+++ b/python/typemap_utils.cpp
@@ -25,7 +25,7 @@
     #define PyInteger_FromLong(n) PyLong_FromLong(n)
 #else
     #define PyObject_ToCharPtr(n) PyString_AsString(n)
-    #define IsPyString(n) PyString_Check(n)
+    #define IsPyString(n) PyString_Check(n) || PyUnicode_Check(n)
     #define PyInteger_Check(n) PyInt_Check(n)
     #define PyInteger_AsLong(n) PyInt_AsLong(n)
     #define PyInteger_FromLong(n) PyInt_FromLong(n)
--- a/python/typemap_utils.cpp
+++ b/python/typemap_utils.cpp
@@ -24,7 +24,7 @@
     #define PyInteger_AsLong(n) PyLong_AsLong(n)
     #define PyInteger_FromLong(n) PyLong_FromLong(n)
 #else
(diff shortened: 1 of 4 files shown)


stevengj:
I don't quite understand this.  If the user passes `3`, do we really want to treat it as the file `"3"` instead of throwing a type error?

Maybe do this conversion only if `isinstance(self.epsilon_input_file, unicode)` in Python 2, if SWIG isn't handling Unicode strings properly in Python 2?

In principle, we should convert in the other direction, since filenames can be unicode.  On everything except Windows, we can handle this on the C++ side by making sure that Python Unicode strings are converted to UTF8-encoded text in C++ — according to the SWIG docs this is the [default](http://www.swig.org/Doc3.0/Python.html#Python_nn77).  This won't work on Windows, but handling Unicode filenames on Windows is not worth the trouble.

python/typemap_utils.cpp
```diff
@@ -25,7 +25,7 @@
     #define PyInteger_FromLong(n) PyLong_FromLong(n)
 #else
     #define PyObject_ToCharPtr(n) PyString_AsString(n)
```
stevengj: Note that `PyString_AsString` "computes the default encoding of string and operates on that" according to the manual, whereas we probably should use UTF-8 (which will work on everything except Windows, and on Windows will work for strings containing only ASCII characters)?

To do this, you might need to call `PyUnicode_AsUTF8String` on Unicode strings to convert them to strings in the utf-8 encoding, and then call `PyString_AsString` on the result.

ChristopherHogan:
Updated. How does that look?

PR #280, opened 2018-04-13
Fix memory errors

ChristopherHogan:
* Set `susceptibility_list.items` to NULL after deleting them to avoid double free in Python  typemap
* Initialize user material's medium so Python typemap doesn't try to free susceptibilities.
@stevengj @oskooi @HomerReid


stevengj:
Thanks. I was getting a warning about a double free of volume_list in the Scheme interface, I wonder if a related fix is needed in the type maps there.

PR #284, opened 2018-04-17
Use Bloch vector to set initial-guess k vector for `get_eigenmode_coefficients` in periodic geometries

HomerReid:
Use Bloch-vector components to set initial-guess k vector for `get_eigenmode_coefficients` in periodic geometries.

### DIFF
--- a/src/mpb.cpp
+++ b/src/mpb.cpp
@@ -615,6 +615,16 @@ void fields::get_eigenmode_coefficients(dft_flux flux,
   if (d==NO_DIRECTION)
    abort("cannot determine normal direction in get_eigenmode_coefficients");
 
+  // if the flux region extends over the full computational grid and we are bloch-periodic 
+  // in any direction, set the corresponding component of the eigenmode initial-guess 
+  // k-vector to be the (real part of the) bloch vector in that direction. otherwise just 
+  // set the initial-guess k component to zero.
+  vec kpoint(0.0,0.0,0.0);
+  if ( where == this->v )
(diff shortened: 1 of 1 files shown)


src/mpb.cpp
```diff
@@ -615,6 +615,16 @@ void fields::get_eigenmode_coefficients(dft_flux flux,
   if (d==NO_DIRECTION)
    abort("cannot determine normal direction in get_eigenmode_coefficients");
 
+  // if the flux region extends over the full computational grid and we are bloch-periodic 
+  // in any direction, set the corresponding component of the eigenmode initial-guess 
+  // k-vector to be the (real part of the) bloch vector in that direction. otherwise just 
+  // set the initial-guess k component to zero.
+  vec kpoint(0.0,0.0,0.0);
+  if ( where == this->v )
```
stevengj: This condition seems too strong.  The algorithm should be:  Set `kpoint` in any direction that is (a) periodic and (b) `where` *in that direction* is the size of the computational cell.

Something like:
```c
FOR_DIRECTIONS(d)
  if (boundaries[High][d]==Periodic && boundaries[Low][d]==Periodic &&
      float(where.in_direction_min(d)) == float(v.in_direction_min(d)) &&
      float(where.in_direction_max(d)) == float(v.in_direction_max(d)))
    kpoint.set_direction(d, real(this->k[d]));
```
where I've rounded to `float` for comparison to reduce the sensitivity to rounding errors.

PR #293, opened 2018-04-19
fix memory leak in dft_flux/fields

stevengj:
Also works towards symmetry support in the mode coefficient calculation (though that is not fully working yet).


ChristopherHogan:
Rejoice! After the `eigenmode_amplitude` bugfix, valgrind reports no errors.

stevengj:
Now `test_wvg_src` is failing, presumably because the discretization has changed slightly?

I'm also not sure it's quite correct in a Bloch-periodic simulation where "wrapping" around should pick up a Bloch phase factor.  (Since any error is only in edge pixels, presumably it will still converge with resolution, but it may hurt the convergence rate.)

ChristopherHogan:
If you do decide this is the right fix, the new reference values from scheme to fix the python `wvg_src` test would be:
```python
self.assertAlmostEqual(flux1 / -1.8, 3.823251619670293e-5)
self.assertAlmostEqual(flux2 / 1.8, 1.724792740529718)
```

PR #307, opened 2018-04-27
Add test for susceptibilities in user materials

ChristopherHogan:
@stevengj @oskooi

### DIFF
--- a/python/tests/material_dispersion.py
+++ b/python/tests/material_dispersion.py
@@ -7,23 +7,19 @@ import meep as mp
 
 class TestMaterialDispersion(unittest.TestCase):
 
-    def setUp(self):
+    def test_material_dispersion_with_user_material(self):
         susceptibilities = [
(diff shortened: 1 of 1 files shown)


python/tests/material_dispersion.py
```diff
@@ -13,12 +13,17 @@ def setUp(self):
             mp.LorentzianSusceptibility(frequency=0.5, gamma=0.1, sigma=2e-5)
         ]
 
-        default_material = mp.Medium(epsilon=2.25, E_susceptibilities=susceptibilities)
+        self.default_material = mp.Medium(epsilon=2.25, E_susceptibilities=susceptibilities)
+
```
stevengj: Set the default material to air so that we are sure that the susceptibility comes from the user function

ChristopherHogan: I removed the old version.

PR #315, opened 2018-05-03
Only restore fields that were backed up during synchronized_magnetic

### LINKED ISSUE #309: Magnetic synchronization causes segmentation fault
I'm trying to run one of the MEEP examples and output a synchronized magnetic field.  Whenever I try to call synchronized_magnetic(), I get a segmentation fault.  The code that I'm trying to run is:

```
import meep as mp

cell = mp.Vector3(16, 8, 0)

geometry = [mp.Block(mp.Vector3(1e20, 1, 1e20),
                     center=mp.Vector3(0, 0),

sources = [mp.Source(mp.ContinuousSource(frequency=0.15),
                     component=mp.Ez,
                     center=mp.Vector3(-7,0))]

pml_layers = [mp.PML(1.0)]

resolution = 10

                    boundary_layers=pml_layers,
                    geometry=geometry,
                    sources=sources,
                    resolution=resolution)

sim.run(mp.synchronized_magnetic(mp.output_bfield_y),
        until=10)
```

This works fine if I omit the mp.synchronized_magnetic() call.

I installed MEEP from conda using the instructions in the documentation.  I'm using Ubuntu 16.04.  The exact error message is:

```
python example.py
-----------
Initializing structure...
Working in 2D dimensions.
Computational cell is 16 x 8 x 0 with resolution 10
     block, center = (0,0,0)
          size (1e+20,1,1e+20)
time for set_epsilon = 0.0905111 s
-----------
creating output file "./example-by-000000.00.h5"...
Segmentation fault (core dumped)
```

Any advice would be appreciated.  Is there anyway to get more debug information?

ChristopherHogan:
I'm not sure if this retains the correct semantics, but it seems correct logically: we only want to restore a field if it was previously backed up. Fixes #309, where `RESTORE(f_u)` expanded to
```C
if (f_u[c][cmp])
    memcpy(f_u[c][cmp], NULL, gv.ntot()*sizeof(realnum));
```
and caused a crash (a NULL argument to `memcpy` is undefined behavior).
@stevengj @oskooi

### DIFF
--- a/python/tests/simulation.py
+++ b/python/tests/simulation.py
@@ -372,6 +372,31 @@ class TestSimulation(unittest.TestCase):
         np.testing.assert_allclose(energy, energy_arr)
         np.testing.assert_allclose(efield, efield_arr)
 
+    def test_synchronized_magnetic(self):
+        # Issue 309
+        cell = mp.Vector3(16, 8, 0)
+
+        geometry = [mp.Block(mp.Vector3(1e20, 1, 1e20),
+                             center=mp.Vector3(0, 0),
+                             material=mp.Medium(epsilon=12))]
+
+        sources = [mp.Source(mp.ContinuousSource(frequency=0.15),
+                             component=mp.Ez,
+                             center=mp.Vector3(-7, 0))]
+
+        pml_layers = [mp.PML(1.0)]
+        resolution = 10
+
+        sim = mp.Simulation(
+            cell_size=cell,
+            boundary_layers=pml_layers,
+            geometry=geometry,
+            sources=sources,
+            resolution=resolution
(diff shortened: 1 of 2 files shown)


stevengj:
It looks like this bug was introduced by 0c02923fa3ff36e0814f664b2fbca6a6f3792705 … in µ=1 regions it doesn't backup H in cases where H points to the same array as B, so the f_backup array for H will never be allocated.

I'm not sure why this didn't crash a long time ago…?

stevengj:
Actually, no, it did check `f_backup`.   It must be something to do with the lazy allocation of `f_u`...

PR #318, opened 2018-05-04
Diffraction eigenmodes

HomerReid:
Attempt at automatic reverse-engineering to detect when user is trying to compute plane-wave diffraction coefficients by calling `get_eigenmode_coefficients.`


stevengj:
Replaced by #329.

PR #324, opened 2018-05-04
Get swig from environment if present

ChristopherHogan:
This is useful for testing different SWIG versions.
@stevengj @oskooi


stevengj:
The right thing to do is probably to integrate this with a configure check.  i.e. add an `AC_CHECK_PROG(SWIG, swig)` to the `configure script, which will automatically add a `SWIG` variable.  This way if you do `./configure SWIG=....` then it will remember it until you re-run configure.

PR #326, opened 2018-05-09
Validate material frequencies.

### LINKED ISSUE #319: validity check for the materials library
Since the material fits are only valid in given range of wavelengths (and assume a particular distance unit), it would be good to print a warning if the user attempts use them outside this range.   To do this, we could:

1. Add an optional field to the Python `Medium` data structure (not sent to C++) which indicates a range of valid frequencies.  By default, this would be None, but the materials library would set it.

2. Before the simulation runs, the Python would call a `simulation.check_material_frequencies` function that would print a warning if any source frequency (± bandwidth) or DFT frequency in the simulation falls outside the valid frequencies for any of the materials in the geometry or extra_materials list.

ChristopherHogan:
Fixes #319.
To enable validation for the materials library, each `Medium` will need to be passed a `FreqRange` namedtuple via the `valid_freq_range` keyword argument. For example:
```python
mat = mp.Medium(epsilon=12, valid_freq_range=mp.FreqRange(min=2, max=8))
```
I can take care of this if someone gives me a list of valid frequencies for each material.

To check DFT frequencies, it was necessary to store all dft objects created by the user (via `add_flux`, etc.) in `Simulation.dft_objects`. I couldn't find another way to get that information.
@stevengj @oskooi

### DIFF
--- a/python/geom.py
+++ b/python/geom.py
@@ -4,11 +4,15 @@ import functools
 import math
 import numbers
 import operator
+from collections import namedtuple
 from copy import deepcopy
 from numbers import Number
 
 import numpy as np
 import meep as mp
+
+
+FreqRange = namedtuple('FreqRange', ['min', 'max'])
 
 
 def check_nonnegative(prop, val):
@@ -151,7 +155,8 @@ class Medium(object):
                  E_chi2=None,
                  E_chi3=None,
                  H_chi2=None,
-                 H_chi3=None):
+                 H_chi3=None,
+                 valid_freq_range=None):
 
         if epsilon:
             epsilon_diag = Vector3(epsilon, epsilon, epsilon)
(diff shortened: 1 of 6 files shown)


oskooi:
Each material in the library has its valid wavelength range indicated (e.g., for [silicon](https://github.com/stevengj/meep/blob/master/python/examples/materials_library.py#L7), it is 0.4 - 1.0 μm). To convert the wavelength to frequency involves an inversion (e.g., a wavelength of 0.4 μm corresponds to a frequency of 1/0.4=2.5).

ChristopherHogan:
Added valid ranges to the materials in `materials_library.py`.

ChristopherHogan:
Reverted changes to`materials_library.py` to avoid merge conflicts with #325.

Issue #321, opened 2018-05-04
get_dft_array and output_dft for zero-thickness regions


stevengj:
If the region is zero thickness in any direction, you might expect the output array to be 1 pixel thick in that direction, but in fact it is 2 pixels thick because of how Meep does interpolation.

Solution: for any direction with zero thickness, change it to be 1 pixel thick.  Take the current 2-pixel thick array and postprocess it (before returning it to the user or writing it to a file) and simply sum the values in the 2 pixels.  (This is the correct thing to do because they have already been pre-multiplied by the interpolation weights as computed by loop_in_chunks.)

stevengj:
Closed by #330.

PR #330, opened 2018-05-11
Zero thickness dft array output

HomerReid:
Initial stab at addressing #321


HomerReid:
For DFT fields defined on a volume of zero thickness in one or more directions, the `dft_array` routines now collapse the two-pixel-thick regions in those dimensions to one-pixel thick regions. Calling conventions, etc. are unchanged. I tested that it behaved properly in a few 2D and 3D situations involving DFT flux regions of zero thickness in various directions.

Similarly, when the `output_dft` routines for DFT fields on volumes of zero thickness are called, the raw DFT file output is now bypassed in favor of fetching the fields in array form (with two-pixel thick regions for zero-width dimensions properly condensed to single-pixel thickness) and then exporting the compressed array to HDF5. I verified that this works for a couple of cases. Unfortunately, after writing the HDF5 file it seems to be core dumping from my python script. I haven't been able to track this down.

ChristopherHogan:
In case you were unaware, you can usually get a backtrace by running the script through gdb:
```bash
$ gdb python
(gdb) run path/to/script.py
```
Fee free to send me the script if you want help troubleshooting.

HomerReid:
Thanks, yes, I've been trying this and also running in valgrind. The difficulty is that valgrind produces a massive torrent of errors, almost all of which seem to be referencing python code. At first I thought it might have something to do with python reference counting and the complications of passing C++-allocated objects back to python codes. However, my C++ code does some allocation and deletion, but only on the C++ side, and the allocated/deleted objects are not passed back to python, so it's not clear why I'm getting so many python-related errors. 

I am wondering if it might have to do with using the raw, low-level `h5file::write` routine instead of the higher-level routines like the ones in `h5fields.` In this case the full array to be written to disk is available on a single core, so I don't need the HDF5 export to happen on all processors, and in fact I have that code enclosed in a `if (am_master()){}` block. But it seems this might be problematic, and I will next try using the higher-level interface.

ChristopherHogan:
> valgrind produces a massive torrent of errors

If you use Python 3.6, you can change the allocator with an environment variable. That plus the python suppressions file [here](https://github.com/python/cpython/blob/master/Misc/valgrind-python.supp) makes the output much easier to parse.  I do something like:
```bash
PYTHONMALLOC=malloc valgrind --track-origins=yes --suppressions=/path/to/valigrind-python.supp python script.py
```
More info [here](https://github.com/python/cpython/blob/master/Misc/README.valgrind).

HomerReid:
Updated to fix a subtlety that caused freezing on some multiprocessor runs for HDF5 output of DFT fields on volumes with zero-thickness dimensions. The issue was that, depending on chunking, processors may not all agree on the number of DFT frequencies, which was problematic for the handling of zero-thickness dimensions in DFT arrays.

src/dft.cpp
```diff
@@ -1063,13 +979,27 @@ void fields::output_dft_components(dft_chunk **chunklists, int num_chunklists,
...
-  //h5file *f = (have_empty_dims && am_master()) ? open_h5file(HDF5FileName,h5file::WRITE) : 0;
-  h5file::access_mode mode = h5file::WRITE;
+  h5file *file=0;
+  if ( have_empty_dims && am_master() )
+   { char filename[100];
+     snprintf(filename,100,"%s%s",HDF5FileName,strstr("%.h5",HDF5FileName) ? "" : ".h5");
+     file = new h5file(filename,h5file::WRITE, false /*parallel*/ );
+   }
+  if (have_empty_dims)
+   NumFreqs = (int)max_to_all( (double)NumFreqs ); // subtle!
```
stevengj: Note that there is a `int max_to_all(int in)` method

HomerReid: My mistake. This call started out as a call to `max_to_master`, for which there doesn't appear to be an `int`-valued version. For micro-revisions like this, in the interest of speeding the pace of development I really encourage you to go ahead and make the revisions yourself.

PR #335, opened 2018-05-14
Fix range of valid frequencies

ChristopherHogan:
I guess it should be `frequency  +/- 0.5 * bandwidth` instead of `frequency +/- bandwidth`.
@stevengj @oskooi

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -368,15 +368,19 @@ class Simulation(object):
             dft_freqs.append(dftf.freq_min)
             dft_freqs.append(dftf.freq_min + dftf.Nfreq * dftf.dfreq)
 
-        warn_fmt = "{} frequency {} is out of material's range of {}-{}"
+        warn_src = ('Note: your sources include frequencies outside the range of validity of the ' +
+                    'material models. This is fine as long as you eventually only look at outputs ' +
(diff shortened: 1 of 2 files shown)


stevengj:
I guess, though I'm not sure how much it matters:

* If you look at the spectrum of the source, it is something like a gaussian peak, and `bandwidth` is related to the width.  But there are nonzero "tails" outside of this bandwidth.

* It is actually perfectly fine to have a source that extends outside the valid bandwidth of the materials, as long as you don't actually *use* those frequencies.  e.g. you might use a very broad-bandwidth source (a very short pulse), but only output the dft-flux in a valid range of frequencies, or excite a resonant mode at a valid frequency, and this is fine.

Because of the latter issue, we might want to issue a different warning for sources outside the valid range of frequencies, to make it clear that this is okay as long as you eventually only look at quantities from valid frequencies.

stevengj:
Example warning text for source outside the valid bandwidth:

"Note: your source includes frequencies outside the range of validity of the material models.  This is fine as long as you eventually only look at outputs (fluxes, resonant modes, etc.) at valid frequencies."

ChristopherHogan:
I updated the warning text. I'll add the Harminv check in a separate PR.

PR #336, opened 2018-05-14
Added Si3N4 to materials library

DerekK88:


oskooi:
Thanks for this. For completeness, please also add Si<sub>3</sub>N<sub>4</sub> to the Scheme [materials library](https://github.com/stevengj/meep/blob/master/scheme/examples/materials-library.scm).

Note: there is another fit for Si<sub>3</sub>N<sub>4</sub> from the [Horiba ellipsometry notes](http://www.horiba.com/fileadmin/uploads/Scientific/Downloads/OpticalSchool_CN/TN/ellipsometer/Lorentz_Dispersion_Model.pdf) for the visible wavelength range of  0.22 - 0.83 μm. Also, there is a fit for non-stoichiometric silicon nitride (SiN). I will add these as well in a separate PR.

We will soon add an output routine for dispersive materials which will facilitate a comparison of different fits.

DerekK88:
Ok, just added this in PR #337.  I haven't looked into the differences between the Horiba material data vs the data provided by Philipp 1973, but the wavelength range seems to include the Horiba data as a subset (0.207-1.24 μm).  

The output routine you mention sounds quite helpful!

oskooi:
PR #339, which was recently merged, added Si<sub>3</sub>N<sub>4</sub> (based on the Horiba data which is valid for 0.23 - 0.83 μm) to the [materials library](https://github.com/stevengj/meep/blob/master/python/examples/materials_library.py#L791-L799). In order to now merge this PR, please rename the variables (i.e., to Si3N4_IR; IR is for infrared since this data is fit to 0.20 to 1.24 μm) and also add the `valid_freq_range` parameter to the object.

DerekK88:
Ok, changed labeling to Si3N4_VISNIR (perhaps the Horiba data can be Si3N4_VIS?).  I also added material data from [K. Luke et al.](https://refractiveindex.info/?shelf=main&book=Si3N4&page=Luke) which should be relevant to people working at telecom wavelengths and a bit further in the IR

luspock:
Hi, I am considering using the Si3N4 material in the NIR wavelength.  I am confused with parameters in the model of Si3N4_NIR in file: [materials.py](https://github.com/NanoComp/meep/blob/master/python/materials.py). I have checked the data from [ref.](https://refractiveindex.info/?shelf=main&book=Si3N4&page=Luke)
image
Should Si3N4_NIR_frq1 and Si3N4_NIR_frq2 be 1/(0.1353406 * um_scale) and 1/(1239.842 * um_scale) instead of 0.1353406 and 1239.842?
The [file](https://github.com/NanoComp/meep/blob/master/scheme/materials.scm) for Scheme is fine.

oskooi:
This is indeed a bug which is fixed by #768. Thanks for pointing this out.

PR #340, opened 2018-05-16
Compute fragment statistics

ChristopherHogan:
### Workflow

Before initializing the structure, Python calls `mp.compute_fragment_stats`, which is the high-level entry point for splitting the cell into boxes and calculating statistics for each box.

#### `mp.compute_fragment_stats`

1. Initializes libctl.
2. Splits the computational cell into boxes (defaulting to size 10 per dimension, but controllable via the optional `box_size` argument).
3. Initializes a list of `fragment_stats` (the data structure that tracks the statistics for each box).
4. Calls the `compute_stats` and `compute_dft_stats` method on each `fragments_stats` instance in the list.
5. Returns the list of `fragment_stats` back to python as a list of SWIG objects.

See `python/tests/fragment_stats.py` for examples.

#### `fragment_stats::compute_stats` calls

1. `count_anisotropic_pixels`
2. `count_nonlinear_pixels`
3. `count_susceptibility_pixels`
4. `count_nonzero_conductivity_pixels`
 
### TODO

~~1. Support cylindrical coordinates. Should I convert to Cartesian to create the boxes?~~
~~2. Calculate DFT statistics.~~
~~3. Handle eps file materials, user materials, and perfect metal.~~


ChristopherHogan:
This is ready for review. I reworked the dft stuff so that the API doesn't change. The SWIG objects are wrapped by Python objects, and the fields are only initialized and the `add_*` method called when the user needs to access a property of the C++ object, or it's passed to a method expecting a SWIG-wrapped C++ object. I'm not sure I'm handling the `geometry_center` correctly.

ChristopherHogan:
* Empty DFT pixels now count as 2 instead of 1.
* Removed `is_line_or_plane`.
* Adjusted tests to account for 2 changes above.

This should be ready now.

ChristopherHogan:
Rebased to fix merge conflicts.

PR #352, opened 2018-05-22
update docs

oskooi:
Updates to the Tutorial/Basics example on Angular Reflectance Spectrum as well as Tutorial/Material Dispersion. 

We will need to add a new test to the Python suite based on the recently-added example [refl-angular.py](https://github.com/stevengj/meep/blob/master/python/examples/refl-angular.py) which is updated in this PR.


python/examples/bend-flux.py
```diff
@@ -99,6 +99,6 @@
 plt.plot(wl,Ts,'ro-',label='transmittance')
 plt.plot(wl,1-Rs-Ts,'go-',label='loss')
 plt.axis([5.0, 10.0, 0, 1])
-plt.xlabel("wavelength (um)")
+plt.xlabel("wavelength (μm)")
```
ChristopherHogan: Unicode literals will break in Python 2. If you add this line to the top of the file:
```python
# -*- coding: utf-8 -*-
```
then it should work in python 2 and 3.

oskooi: Thanks for the suggestion. Fixed.

PR #360, opened 2018-05-30
Fix bug with coeffs size in get_eigenmode_coefficients

ChristopherHogan:
Fixes #359.
@stevengj @oskooi @HomerReid

PR #361, opened 2018-05-30
update mode-decomposition tutorial and other fixes to docs

oskooi:


stevengj:
Will have to be updated again after #360, presumably.  Or should this be merged first?

oskooi:
This should be merged before #360. Afte merging, #360 will have a conflict in [python/examples/mode-decomposition.py](https://github.com/stevengj/meep/blob/master/python/examples/mode-decomposition.py) which can be resolved. Also, [python/tests/mode_coeffs.py](https://github.com/stevengj/meep/blob/master/python/tests/mode_coeffs.py) will need to be updated.

Issue #363, opened 2018-06-01
Output K vector with get_eigenmode_coefficients()


smartalecH:
Would it be too difficult (or even practical) to modify the `get_eigenmode_coefficients()` python function to also output the computed k vector magnitude and group velocity? It's often nice to have the accompanying mode index and group index for S Parameter sweeps. Currently, I just run a separate analysis using the MPB python interface, but this seems overly redundant.

Obtaining the group velocity, for example, seems trivial since we could pass an array into `self.fields.get_eigenmode_coefficients()` (rather than `None`). Would it be straightforward to pull out the corresponding k vector as well?

stevengj:
It would be straightforward to modify the code to return this information, I agree.

smartalecH:
I'm working on a fix. Retrieving the group velocity through both the C++ and python interfaces are rather easy. Retrieving the corresponding k vector is a bit more challenging. In the C interface, I simply call the `get_k()` function and store the resulting `vec` object in an array -- similar to what's done with `coeffs` and `vgrp`. Passing this preallocated array from the python interface to the C++ interface using SWIG is the tricky part.

Is there a recommended way to pass a list of preallocated Vector3 objects generated in python to the C++ interface? For some reason, numpy doesn't like generating a numpy array of Vector3 objects, so that doesn't seem like an easy solution to me.

Alternatively, I could extract out the individual x,y and z components of the k vector and store those in a simple array. It's probably cleaner to return the original `vec` /  `Vector3` object, however.

PR #368, opened 2018-06-07
Eigenmode k vector

HomerReid:
In response to #363, I modified the C++ prototype for `get_eigenmode_coefficients` to accept a new optional parameter called `kpoints.` If non-null, this should be a pointer to a caller-allocated array of type `meep::vec` with length `num_bands * num_freqs`. On return the k-vector for the mode with band index #`nb` and frequency #`nf` will be stored in `kpoints[nb*num_freqs + nf]`.

If `kpoints` is NULL (the default), it is ignored. Existing code calling `get_eigenmode_coefficients` will work with this updated version, except for cases where a `user_kpoint_func` is specified; in that case the call to `get_eigenmode_coefficients` will need to be modified with an extra `0` argument in the list of arguments before `user_kpoint_func.`

For the python interface, I don't know how convenient it is to wrap a routine accepting an optional parameter of type meep::vec*. If this is inconvenient, an alternative would be to change the optional `kpoints` argument to have type `double *`, in which case it would be a caller-allocated array of length `3*num_bands*num_freqs` with the components of the k-vector for mode {`nb`, `nf`} living in `kpoints[3*(nb*num_freqs + nf) + 0,1,2].` Maybe such a calling convention would be easier to accommodate in the high-level interface.


src/mpb.cpp
```diff
@@ -640,6 +645,7 @@ void fields::get_eigenmode_coefficients(dft_flux flux,
 
       double vg=get_group_velocity(mode_data);
       if (vgrp) vgrp[nb*num_freqs + nf]=vg;
+      if (kpoints) kpoints[nb*num_freqs + nf]=get_k(mode_data);
```
stevengj: This is in the reciprocal basis, I think. You need to multiply by G (probably when mode_data is initialized?) to convert to Cartesian

ChristopherHogan:
#374 is trying to do the same thing as this PR, correct?  I guess the 2 need to sync up. #374 handles the python typemaps, whereas this PR does not. #374 just needs some adjustments to the tests to account for the new interface. I think #374 needs converted the cartesian conversion discussed in this PR though. So either this PR needs the python stuff from #374 or #374 needs the cartesian conversion stuff from this.

ChristopherHogan:
Also, it's easy to forget that any change to `get_eigenmode_coefficients` requires updates to the following files:
* `python/examples/mode-decomposition.py`
* `python/tests/mode_coeffs.py`
* `doc/docs/Python_User_Interface.md`
* `doc/docs/Python_Tutorials/Mode_Decomposition.md`
* `doc/docs/Mode_Decomposition.md`
@smartalecH

smartalecH:
> #374 is trying to do the same thing as this PR, correct?

Yes, that's right.

> So either this PR needs the python stuff from #374 or #374 needs the cartesian conversion stuff from this.

It's probably more convenient to use this PR, rather than #374. It's easy to copy over the python interface. I like how this implementation creates an array of `vecs` rather than an array of `doubles` (like I did). We can either return a python list of `Vector3` objects, or convert it in the python interface to a numpy array.

I had trouble passing the `vec` type using SWIG, however, which is why I opted for the doubles. @chogan, do you have a typmap for a `vec` array?

ChristopherHogan:
> do you have a typmap for a vec array?

We'd have to write one that is similar to [_get_farfield](https://github.com/ChristopherHogan/meep/blob/master/python/meep.i#L282) but uses [vec2py](https://github.com/ChristopherHogan/meep/blob/master/python/typemap_utils.cpp#L68) instead of `PyComplex_FromDoubles`. The issue is that `vec2py` includes an optimization that only makes it suitable for `Vector3`s used in callback functions, so we'd need to create a version that creates and returns a new object on each call. Currently it creates an object once and mutates the parameters when called, which I don't think would work if you want to create a list of distinct instances. It would be easy to add a `new_object` flag to `vec2py` that defaults to `false`.

stevengj:
@smartalecH, would you want to take a shot at merging this with your PR?   i.e. start with Homer's branch, add a version of your Python interface, and then make a new PR?

stevengj:
Actually, I think we'll first have @HomerReid update this PR to add a second method for `get_eigenmode_coefficients` that matches the old calling convention.   This way the old Python interface and test will continue to work.   Once that is green, we'll merge that, and maybe @smartalecH can create a PR for a Python interface to the new functionality.

smartalecH:
Yes, I can update the Python interface once the C implementation is updated and passes all checks.

PR #369, opened 2018-06-07
Replace blocks with prism in mode-decomposition.py

ChristopherHogan:
This was a request from Ardavan.
@stevengj @oskooi


DerekK88:
As a meep user, thank you!  Would it be possible to get the prism type added to the Python user interface as well (either in this or perhaps a separate PR)?

ChristopherHogan:
The prism type for Python was merged in #345.

DerekK88:
Ah, I was referring to the documentation for the [Python user interface](https://meep.readthedocs.io/en/latest/Python_User_Interface/) (specifically [Python_User_Interface.md](https://github.com/stevengj/meep/blob/master/doc/docs/Python_User_Interface.md)).  Now that the prism type is added it would be great for users to know the correct function inputs.

Thanks for all these great additions to meep!

ChristopherHogan:
Oh, I see. Yes, we'll get documentation up very soon.

PR #374, opened 2018-06-08
Added kvec and vgrp output for get_eigenmode_coefficient

smartalecH:
This is my solution to my feature request (#363). The C interface returns an array of doubles, similar to what's done with the group velocities. The python interface returns a 3D numpy array where the 1st column is the band, the second column is frequency, and the third column is the dimension (i.e mp.X, mp.Y, mp.Z).

After reviewing #368, I think my array is in the reciprocal basis as well. I can work on a fix.

### DIFF
--- a/python/examples/materials_library.py
+++ b/python/examples/materials_library.py
@@ -1,40 +1,12 @@
 # Materials Library
 
 import meep as mp
-import numpy as np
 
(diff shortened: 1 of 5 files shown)


smartalecH:
Officially fixed in #396.

PR #381, opened 2018-06-13
Sync python material dispersion doc to example

kitchenknif:
The second example in the docs doesn't run, updated it to follow the code in the examples.


stevengj:
Thanks!

PR #388, opened 2018-06-20
Get source amplitude from hdf5 file

### LINKED ISSUE #372: source amplitude from HDF5 file
The sources already allow you to specify an `amp_fun` that gives an arbitrary function of space for the amplitude.   Might be nice to layer a higher-level function on top of this that reads in a grid of values from an HDF5 file and creates an `amp_fun` that interpolates into this grid.  (Much like the epsilon_file feature.)

This could be put as a helper method into the `fields` object, I guess, a new method of `add_volume_source`, where you pass a filename and a dataset name instead of `A`.   (You would still pass an `amp` parameter to allow you to multiply the values in the file by an overall scale factor.)

ChristopherHogan:
Fixes #372 
* Move `linear_interpolate` to `libmeep` to avoid multiple definitions, and to use it in `sources.cpp`. The definition is currently in `fields.cpp`, but there might be a better place for it.
* Link `libmeep.so` with `libctlgeom.so` so that `src/sources.cpp:amp_file_func` can access `geometry_lattice` and `geometry_center`.  
* Allow passing a filename and a dataset name to a `mp.Source`.
* Added overload of `add_volume_source` that creates an amplitude function that reads from an hdf5 file. It expects both a `.re` and `.im` dataset.
* Added basic test that runs a simulation with an amplitude function that always returns `1+1j`, then runs the same simulation that gets the amplitude from an hdf5 file of all ones (2, 10x10 datasets for real and imaginary), and compares a field point. A more thorough test is needed (Ardavan mentioned a ray tracing example).

@stevengj @oskooi


ChristopherHogan:
Updates:
* Create `add_volume_source` that takes array and size
* Error checking for real and imag dims sizes
* Print name of missing dataset on error
* Base linear interpolation on `where_` instead of geometry_lattice
* Don't need to link with libctlgeom

PR #392, opened 2018-06-26
Python support for GDSII import

ChristopherHogan:
* Add `libGDSII` build to Travis.
* Add typemaps to convert C++ prism objects to Python.
* Update `python/tests/bend_flux.py` to run with both manually built prisms and prisms imported from `bend-flux.gds`. The test is only accurate to 2 decimal places because 1) the GDSII version doesn't extend off the cell to simulate infinity and 2) the GDSII version's height is 0.

@HomerReid, could you make sure `typemap_utils.cpp:gobj_to_py_obj` looks right? I have to transform the vertices based on the `centroid` and `m_p2c`. Also I hard-code the axis to `{0, 0, 1}`. Is there a way to reconstruct it, or is it always unit z?

@stevengj @oskooi


python/typemap_utils.cpp
```diff
@@ -639,3 +772,65 @@ static int py_list_to_gobj_list(PyObject *po, geometric_object_list *l) {
...
+        PyObject *geom_mod = get_geom_mod();
+        PyObject *prism_class = PyObject_GetAttrString(geom_mod, "Prism");
+
+        int num_verts = gobj->subclass.prism_data->vertices.num_items;
+        prism *prsm = gobj->subclass.prism_data;
+
+        PyObject *py_verts = PyList_New(num_verts);
+        for (int i = 0; i < num_verts; ++i) {
+            vector3 vp = prsm->vertices.items[i];
+            vector3 prism_v3 = vector3_plus(prsm->centroid, matrix3x3_vector3_mult(prsm->m_p2c, vp));
```
HomerReid: This is correct, but I encapsulated this operation into a subroutine in libctl---you can just say

```
   vector3 prism_v3 = prism_coordinate_p2c(prsm, vp)
```
Here 'c2p' stands for 'cartesian -> prism.' There is also `prism_coordinate_c2p` for converting in the other direction, and `prism_vector_{p2c,c2p}` for converting vectors instead of coordinates (i.e. skip the translation step and just do the rotation).

PR #396, opened 2018-06-29
Return group velocity and kpoints from get_eigenmode_coefficients

### LINKED ISSUE #363: Output K vector with get_eigenmode_coefficients()

ChristopherHogan:
Fixes #363.
I went ahead and implemented this since it was a bit tricky.
@stevengj @oskooi @HomerReid @smartalecH


ChristopherHogan:
This is rebased and green.

PR #402, opened 2018-07-02
Add function to get meep::vec from iterable

woodenToaster:
@stevengj @oskooi

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -2119,3 +2119,11 @@ def interpolate(n, nums):
             res.extend(np.linspace(low, high, n + 1, endpoint=False).tolist())
 
     return res + [nums[-1]]
+
+
+def meep_vec(it):
+    try:
+        return mp.vec(*[i for i in it])
+    except (TypeError, NotImplementedError):
+        print("Expected an iterable with three or fewer floating point values.")
+        raise
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -38,7 +38,7 @@ def vec(*args):
     try:
         # Check for vec(iterable)
         return mp._vec(*args[0])
-    except (TypeError, NotImplementedError):
+    except (TypeError, NotImplementedError, IndexError):
         try:
             # Check for vec(x, [y, [z]])
             return mp._vec(*args)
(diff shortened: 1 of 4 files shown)


python/simulation.py
```diff
@@ -2107,3 +2107,11 @@ def interpolate(n, nums):
             res.extend(np.linspace(low, high, n + 1, endpoint=False).tolist())
 
     return res + [nums[-1]]
+
+
+def meep_vec(it):
+    try:
+        return mp.vec(*[i for i in it])
```
stevengj: Can't you just do `mp.vec(*it)`?

python/simulation.py
```diff
@@ -2107,3 +2107,11 @@ def interpolate(n, nums):
             res.extend(np.linspace(low, high, n + 1, endpoint=False).tolist())
 
     return res + [nums[-1]]
+
+
+def meep_vec(it):
```
stevengj: the name `meep_vec` kind of rubs me the wrong way, since the `meep_` prefix is redundant with the module name.

My ideal interface would be something that simply dispatched on the argument type, e.g `mp.vec(1,2,3)` and `mp.vec([1,2,3])` would both work.

python/simulation.py
ChristopherHogan: This wasn't working before, but turned out to be a symptom of a different error. Fixed.

python/simulation.py
ChristopherHogan: Updated to your ideal interface.

python/simulation.py
```diff
@@ -34,6 +34,20 @@ def get_num_args(func):
     return func.__code__.co_argcount
 
 
+def vec(*args):
+    try:
+        # Check for vec(iterable)
+        return mp._vec(*args[0])
```
stevengj: This will silently ignore other arguments if the first argument is iterable.

Shouldn't it try `mp._vec(*args[0])` if `arg.len == 1` and `mp._vec(*args)` otherwise?

ChristopherHogan: Updated to handle all malformed arguments.

PR #403, opened 2018-07-03
import before_time into meep.py

Chuban:
The `before_time` step modifier is not available in the Python interface.  This is fixed by adding the import to `meep.i`.


ChristopherHogan:
:+1:

PR #409, opened 2018-07-09
Add docs for source amplitude file

ChristopherHogan:
@stevengj @oskooi


stevengj:
Looks good.  But it would be good to remove the dataset argument and instead support filename:dataset

PR #411, opened 2018-07-10
Add warning about MKL in conda environment

ChristopherHogan:
Fixes stevengj/mpb#64.
@stevengj @oskooi


soamaven:
Just to clarify, for anyone who finds this, this only applies to the precompiled conda packages, but not if one is compiling source against the `mkl` libraries, correct? Would I encounter the same `SEGFAULT`s when calling MPB? I haven't needed to yet, but expect to be working up to it.

ChristopherHogan:
Correct, this only applies to the conda packages, which are currently built against OpenBLAS. MKL works fine if Meep and all dependencies are built with and linked to it.

PR #417, opened 2018-07-12
updates to ensure correct calculation of eigenmode coefficients with symmetries

HomerReid:
The only API modification is the addition of a new method `fields::add_mode_monitor` with the same calling convention as `add_dft_flux.` This instantiates a `dft_flux` object in which symmetry reduction is turned off. Existing code that calls `add_dft_flux` to create a flux region for subsequent passage to `get_eigenmode_coefficients` should now call `add_mode_monitor` instead.

### DIFF
--- a/src/mpb.cpp
+++ b/src/mpb.cpp
@@ -230,10 +230,10 @@ void *fields::get_eigenmode(double omega_src,
   // in any direction, set the corresponding component of the eigenmode initial-guess
   // k-vector to be the (real part of the) bloch vector in that direction.
   vec kpoint(_kpoint);
-  LOOP_OVER_DIRECTIONS(v.dim, d)
-    if (float(eig_vol.in_direction(d) == float(v.in_direction(d))))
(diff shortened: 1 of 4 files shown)


src/mpb.cpp
```diff
@@ -634,8 +634,8 @@ void fields::get_eigenmode_coefficients(dft_flux flux,
   if (d==NO_DIRECTION)
    abort("cannot determine normal direction in get_eigenmode_coefficients");
 
-  if (S.multiplicity() > 1)
-   abort("symmetries are not yet supported in get_eigenmode_coefficients");
+  //if (S.multiplicity() > 1)
```
stevengj: Can we save the value of `use_symmetry` in the `dft_flux` object and still abort `if (use_symmetry && S.multiplicity() > 1)`, i.e. in case an ordinary `add_dft_flux` was used?

stevengj:
cc @ChristopherHogan for a Python API.

PR #420, opened 2018-07-16
Method Renaming

ChristopherHogan:
* `Simulation.init_fields` -> `Simulation.init_sim`
* `Simulation.add_eigenmode` -> `Simulation.add_mode_monitor`
@stevengj @oskooi


ChristopherHogan:
I've left the old versions with deprecation warnings. See lines 751 and 1102 of `simulation.py`.

PR #423, opened 2018-07-17
Add Python interface for add_mode_monitor

ChristopherHogan:
This is work in progress of the Python side of #417.

The comment in #417 says to replace all calls to `fields::add_dft_flux` with `fields::add_mode_monitor`. But what if the simulation has symmetries. Don't we then want to call `fields::add_dft_flux`? What's the point of adding support for symmetry if it's always turned off? Here I call `fields::add_dft_flux` (with `symmetry` defaulting to `true`) if the simulation contains symmetries, otherwise I call `fields::add_mode_monitor`. Is that right?
@stevengj @HomerReid @oskooi

### DIFF
--- a/python/tests/mode_coeffs.py
+++ b/python/tests/mode_coeffs.py
@@ -49,6 +49,7 @@ class TestModeCoeffs(unittest.TestCase):
                             boundary_layers=boundary_layers,
                             geometry=geometry,
                             sources=sources,
+                            symmetries=[mp.Mirror(mp.Y)] if mode_num == 1 else []
                             )
 
         xm = 0.5*sx - dpml  # x-coordinate of monitor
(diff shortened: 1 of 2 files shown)


ChristopherHogan:
I think I answered my question. It seems that `add_mode_monitor` always takes symmetries into account. I was confused by the `use_symmetry=false`. I added a `mp.Mirror(mp.Y)` symmetry to the `mode_coeffs.py` test and it works when `mode_num == 1`, but fails when `mode_num == 2`. Is there something about mode 2 that makes the symmetry invalid, or would you expect that to work?

Strangely, running the test for mode 1 followed by mode 2 causes " test 2: |mode coeff|^2 = power" to fail, but running the test for mode 2 only causes "# test 1: coefficient of excited mode >> coeffs of all other modes" to fail.

python/tests/mode_coeffs.py
```diff
@@ -49,6 +49,7 @@ def run_mode_coeffs(self, mode_num, kpoint_func):
                             boundary_layers=boundary_layers,
                             geometry=geometry,
                             sources=sources,
+                            symmetries=[mp.Mirror(mp.Y)] if mode_num == 1 else [mp.Mirror(mp.Y, phase=-1)]
```
stevengj: `mode_num % 2 == 1`

ChristopherHogan: Fixed.

PR #427, opened 2018-07-23
Symmetry in dft arrays

### LINKED ISSUE #421: bug: get_dft_array with symmetry objects
There is a bug in the [`get_dft_array`](http://meep.readthedocs.io/en/latest/Python_User_Interface/#array-slices) routine when used with symmetry objects. This is demonstrated in the following example of a ring resonator which is adapted from the [frequency-domain solver tutorial](http://meep.readthedocs.io/en/latest/Python_Tutorials/Frequency_Domain_Solver/).

```py
import meep as mp
import numpy as np

n = 3.4
w = 1
r = 1
pad = 4
dpml = 2

sxy = 2*(r+w+pad+dpml)
vol = mp.Volume(mp.Vector3(), size=mp.Vector3(sxy-2*dpml,sxy-2*dpml))

c2 = mp.Cylinder(radius=r)

fcen = 0.118
df = 0.08

                    geometry=[c1,c2],
                    sources=[src],
                    resolution=10,
                    symmetries=[mp.Mirror(mp.Y)],

ez_dat = np.absolute(sim.get_dft_array(dfts, mp.Ez, 0))

plt.figure(dpi=100)
plt.imshow(ez_dat.transpose(), interpolation='spline36', cmap='Reds')
plt.axis('off')
plt.show()
```

This simulation script produces the following image of the ring-resonator mode:

ring_sym

When the mirror-symmetry object is removed, the resulting image is correct:

ring_nosym

HomerReid:
Fixes #421.

### DIFF
--- a/src/dft.cpp
+++ b/src/dft.cpp
@@ -671,7 +671,7 @@ cdouble dft_chunk::process_dft_component(int rank, direction *ds,
          // where NI = number of points in Ith direction.
          int idx2=0;
          for (int i=rank-1, stride=1; i>=0; stride*=array_count[i--])
-          idx2 += stride * iloc.in_direction(ds[i]) / 2;
+          idx2 += stride * (iloc.in_direction(ds[i]) / 2);
 
          field_array[idx2] = (retain_dV_and_interp_weights ? w : 1.0) * dft_val;
        }
--- a/src/dft.cpp
+++ b/src/dft.cpp
@@ -666,13 +666,29 @@ cdouble dft_chunk::process_dft_component(int rank, direction *ds,
          buffer[idx2] = reim ? imag(val) : real(val);
        }
       else if (field_array)
-       { int idx2 = ((((array_offset[0] + array_offset[1] + array_offset[2])
+       { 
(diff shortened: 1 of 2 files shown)


src/dft.cpp
```diff
@@ -666,14 +659,21 @@ cdouble dft_chunk::process_dft_component(int rank, direction *ds,
...
+         iloc -= min_corner;                   // iloc <-- 2*(indices of point in DFT array)
+
+         // the index of point n1 or (n1,n2) or (n1,n2,n3) in a 1D, 2D, or 3D array is
+         // (for a 1D array) n1
+         // (for a 2D array) n2 + n1*N2
+         // (for a 3D array) n3 + n2*N3 + n1*N2*N3
+         // where NI = number of points in Ith direction.
+         int idx2=0;
+         for (int i=rank-1, stride=1; i>=0; stride*=array_count[i--])
+          idx2 += stride * iloc.in_direction(ds[i]) / 2;
```
stevengj: Looks like it should be `stride * (iloc.in_direction(ds[i]) / 2);

PR #428, opened 2018-07-23
Docs on chunking and symmetries

### LINKED ISSUE #405: documentation regarding implementation of symmetries
Among the more useful things to include on the [developer information page](https://meep.readthedocs.io/en/latest/C++_Developer_Information/) would be a discussion of how symmetries are implemented. For example, the following code snippet from [`dft.cpp`](https://github.com/stevengj/meep/blob/master/src/dft.cpp) is somewhat inscrutable to the novice (or even intermediate) MEEP hacker:

```c++
   ivec isS = S.transform(is, sn) + shift;
   ivec ieS = S.transform(ie, sn) + shift;

   ivec permute(zero_ivec(fc->gv.dim));
   for (int i = 0; i < 3; ++i)
    permute.set_direction(fc->gv.yucky_direction(i), i);
   permute = S.transform_unshifted(permute, sn);
   LOOP_OVER_DIRECTIONS(permute.dim, d)
    permute.set_direction(d, abs(permute.in_direction(d)));
```

What is `permute`? What is the difference between `S.transform` and `S.transform_unshifted`? 
 
Similarly, from [`loop_in_chunks.cpp`](https://github.com/stevengj/meep/blob/master/src/loop_in_chunks.cpp):

```C++
  // loop over symmetry transformations of the chunks:
  for (int sn = 0; sn < (use_symmetry ? S.multiplicity() : 1); ++sn) {
    component cS = S.transform(cgrid, -sn);
    ivec iyee_cS(S.transform_unshifted(iyee_c, -sn));

    volume gvS = S.transform(gv.surroundings(), sn);
    vec L(gv.dim);
    ivec iL(gv.dim);

    // n.b. we can't just S.transform(lattice_vector,sn), 'cause of signs
      direction dS = S.transform(d, -sn).d;
      L.set_direction(d, fabs(lattice_vector(dS).in_direction(dS)));
      iL.set_direction(d, iabs(ilattice_vector(dS).in_direction(dS)));
    }
```

What is the meaning of the comment ending in "`cause of signs"? 

There are comments sprinkled throughout the code base from which it may be possible to reconstruct the  ideas of the symmetry implementation, but a few sentences describing the big picture could save a lot of time.

HomerReid:
Fixes #405. I decided to go ahead and write the documentation myself. Here is a PDF printout of what the online documentation will look like.
[ChunksAndSymmetry.pdf](https://github.com/stevengj/meep/files/2217522/ChunksAndSymmetry.pdf)


doc/docs/ChunksAndSymmetry.md
```diff
@@ -0,0 +1,306 @@
...
+
+To use this routine, you will first write a "chunk-loop function"
+which carries out some operation involving grid points and (optionally)
+field components at grid points. Then you pass your routine to
+`loop_in_chunks` with some additional arguments customizing the
+type of loop you want (see below). Your loop function will then be
+called once for every `chunk` in the problem---including both
+chunks whose fields are present in memory, and those 
+whose aren't due to being eliminated by symmetry---with a long list
+arguments describing the chunk in question.
```
ChristopherHogan: list "of" arguments?

doc/docs/ChunksAndSymmetry.md
```diff
@@ -0,0 +1,306 @@
...
+
+  // loop over all grid points in chunk
+  LOOP_OVER_IVECS(fc->gv, is, ie, idx)
+   { 
+     // get grid indices and coordinates of parent point
+     IVEC_LOOP_ILOC(gv, iparent);  // grid indices
+     IVEC_LOOP_LOC(gv, rparent);   // cartesian coordinates
+
+     // apply symmetry transform to get grid indices and coordinates of child point
+     ichild = S.transform(iparent, sn) + shift;
```
stevengj: `ivec ichild` and `vec rchild`

doc/docs/ChunksAndSymmetry.md
```diff
@@ -0,0 +1,306 @@
...
+     rchild = S.transform(rparent, sn) + rshift;
+
```
stevengj: Should probably also document how to get a field at the child point (e.g. in h5fields.cpp).   Suppose you are looping over the field component `cchild = Ex` at a bunch of grid points, and we want to get the value of this.  We need to compute:

* `component cparent = S.transform(cchild, -sn)` — get the corresponding parent component in this chunk.  This should be the same as `cgrid` if you passed `cchild` to `loop_in_chunks`.
* `complex<double> phase = shift_phase * S.phase_shift(cparent, sn)` — phase shift to multiply by `cparent` component to get back the `cchild` field with the correct phase.  (e.g. this could be `-1` for an odd symmetry mirror plane, a bloch phase for bloch-periodic boundary conditions, etcetera).
* `complex<double> fieldvalue = fc->f[cparent][0] ? (fc->f[cparent][1] ? complex<double>(fc->f[cparent][0][idx], fc->f[cparent][1][idx]) : fc->f[cparent][0][idx]) : 0;`   Note that `fc->f[cparent][0]` points to an array of real parts, and `fc->f[cparent][1]` points to an array of imaginary parts, both indexed by `idx`.   The imaginary-part pointer will be NULL if the field stored is purely real, and the real part is NULL if the field is not stored at all (because it is identically zero).

There's also the complication of how you do the averaging if you loop over the pixel centers by passing `cgrid == Centered` to `loop_in_chunks`.  In general you need to average any field component (e.g. `Ex`) over four points to get the value at the center of a pixel.  There is a function called `fc->gv.yee2cent_offsets` that you can call to get the indexing offsets for this, e.g. as used as `h5fields`.

doc/docs/ChunksAndSymmetry.md
HomerReid: Good catch! Thanks for the eagle eye.

doc/docs/ChunksAndSymmetry.md
HomerReid: fixed

doc/docs/ChunksAndSymmetry.md
HomerReid: This is sufficiently complicated to warrant abstraction in the form of a utility routine. I added one, called `get_field_components`, to `loop_in_chunks.cpp` and updated the sample code in the documentation appropriately. Punting for now on the hassle of evaluating permittivity/permeability, which would have significantly complicated things.

src/loop_in_chunks.cpp
```diff
@@ -180,9 +180,51 @@ integral = f_i [-w0^3/3 + (i+1) w0^2/2
...
+/* The 6th argument is the integer idx defined inside the body */
+/* of a LOOP_OVER_IVECS.                                       */
+/* components is a caller-created list of components; the      */
+/* return value is an array of the same length storing the     */
+/* values of those components at the grid point.               */
+/***************************************************************/
+vector<complex<double> > get_field_components(fields_chunk *fc, component cgrid,
+                                              complex<double> shift_phase,
+                                              const symmetry &S, int sn,
+                                              int idx, vector<component> components)
```
stevengj: This really should be refactored into two routines, one since the call to `S.transform`, `S.phase_shift`, and `yee2cent_offsets` can happen once outside the ivec loop.   That is, there should be one routine that sets up a data structure or class for field computations (outside the loop) and one that applies it to `fc->f` (inside the loop).

HomerReid: Yes, I considered this optimization, but it would make the code harder to read and would yield only modest speedup, at least for field components. (This optimization would be more essential for permittivity and permeability, which is why I omitted them here.) 

On the spectrum from "easy to read but not speed-optimized" versus "replete with specialized tweaks and accelerations that obfuscate what's going on," meep is already massively weighted toward the latter extreme, and the whole point of this documentation and this utility routine are to lower the entry barrier for new developers to become familiar with basic functionality. If there comes a time when the slightly suboptimal implementation of `get_field_components` actually slows something down in real life, I'll happily implement a more complicated speed-optimized version.

src/loop_in_chunks.cpp
```diff
@@ -185,44 +185,49 @@ typedef complex<double> cdouble;
...
-  vector<complex<double> > fvals;
-
-  for (size_t nc=0; nc<components.size(); nc++)
-   { 
-     // get symmetry-parent component and yee-grid offsets
-     component cparent = S.transform(components[nc], -sn);
+field_component_data create_field_component_data(fields_chunk *fc, component cgrid,
+                                                 complex<double> shift_phase,
+                                                 const symmetry &S, int sn,
+                                                 vector <component> components)
```
stevengj: It feels odd to have a `create_foo(args...)` function in C++ rather than having a `foo` class with a `foo(args...)` constructor.

stevengj:
Updated.

@HomerReid, as general suggestion when submitting pull requests to any project, you should always strive to write code that is stylistically similar to the surrounding code.

PR #429, opened 2018-07-23
Adjust tolerance for eigenmode_amplitude test

ChristopherHogan:
Ardavan reported that this test failed on his machine. Is accuracy to 4 places reasonable for `eigenmode_amplitude`?
@stevengj @oskooi


stevengj:
Should be fine.

Issue #393, opened 2018-06-27
Varying Amplitudes with get_eigenmode_coefficients()


smartalecH:
It seems that the `get_eigenmode_coefficients()` function depends on the location of the monitor. I produced a simple example (similar to #380) where I varied the `y` location of a 2D waveguide (y=0, y=2 and y=4). As the waveguide moves further from the origin, the extracted forward coefficient's relative amplitude for the left monitor decreases (center plot). The right plot shows all three cases normalized.

![eigamps](https://user-images.githubusercontent.com/24902086/41988189-9ade3bda-79f8-11e8-97ee-74c09d3aeacb.png)

Since even the normalized plot is slightly off, it makes me wonder if something goes wrong before the overlap integral.

I have a feeling that the `get_eigenmode` C routine is producing slightly different field amplitudes for each of the three scenarios. Maybe I could produce another example that really exaggerates the differences. It also might be useful to generate a python wrapper for the C routine to debug this possiblity. This functionality would be useful for several other applications as well.

[debug_get_eig_amp.py.zip](https://github.com/stevengj/meep/files/2142184/debug_get_eig_amp.py.zip)

ChristopherHogan:
`get_eigenmode` is already callable from python via `sim.fields.get_eigenmode`. However, there is no typemap to convert the returned `void*` to a Python version of `eigenmode_data`, so the results are currently useless.  What do you want to do with the result? I.e., which fields from the `eigenmode_data` struct would be useful for debugging? I ask because it's a lot of work to convert everything (especially the `maxwell_data`).

HomerReid:
Probably all that's needed is for python users to be able to call `eigenmode_amplitude,` i.e. something like 

````

mode = get_eigenmode(...)

vec EvalPoint=[0.7, -0.2 , 0.3] # coordinates of evaluation point
ExAtEvalPoint = eigenmode_amplitude(mode, EvalPoint, meep.Ex)
HzAtEvalPoint = eigenmode_amplitude(mode, EvalPoint, meep.Hz)

```

Since `eigenmode_amplitude` expects its first input in the form of a `void *`, it should be possible to pass the return value of `get_eigenmode` directly without any need for python code to look inside the structure? I suspect this might already work, or could be make to work with very minor tweaks, not requiring wrapping the full complexity of the `eigenmode_data` struct.

ChristopherHogan:
[...]

PR #431, opened 2018-07-24
update normalization of eigenmode field amplitudes to match that of MPB

HomerReid:
Addresses the issue raised at the end of the discussion of #393.


stevengj:
@smartalecH, does this fix the match with MPB from #393?

stevengj:
Needs an update to the `wvg_src.py` test since the normalization has changed.

PR #439, opened 2018-07-30
Add ModeRegion alias for FluxRegion

ChristopherHogan:
Requested by Ardavan.
@stevengj @oskooi

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -162,6 +162,15 @@ class FluxRegion(object):
         self.weight = complex(weight)
 
 
+class ModeRegion(object):
+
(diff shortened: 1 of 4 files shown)


python/simulation.py
```diff
@@ -162,6 +162,15 @@ def __init__(self, center, size=Vector3(), direction=mp.AUTOMATIC, weight=1.0):
         self.weight = complex(weight)
 
 
+class ModeRegion(object):
```
stevengj: This seems like a separate class, not an alias. What’s wrong with just ModeRegion=FluxRegion?

PR #443, opened 2018-08-01
fix GDSII error introduced by #441

stevengj:
#441 makes the `center` argument to the `Prism` object actually shift the vertices, and otherwise the `center` (which is currently ignored in the libctl prism code) is computed from the vertices.   This means that when we are converting C geometric objects to Python we shouldn't pass the `center` argument in order to leave the vertices as-is.


stevengj:
CI failure is an unrelated Travis hiccup.

PR #459, opened 2018-08-08
add prism objects to documentation of Scheme interface

oskooi:
The prism objects are supported in Scheme but were not documented.

### DIFF
--- a/doc/docs/Scheme_User_Interface.md
+++ b/doc/docs/Scheme_User_Interface.md
@@ -251,7 +251,7 @@ Instead of `material-func`, you can use `epsilon-func`: give it a function of po
 
 Dispersive dielectric and magnetic materials, above, are specified via a list of objects that are subclasses of type `susceptibility`.
 
-**`susceptibility`**
+### susceptibility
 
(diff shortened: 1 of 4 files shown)


doc/docs/Scheme_User_Interface.md
```diff
@@ -359,11 +359,35 @@ The lengths of the block edges along each of its three axes. Not really a 3-vect
...
+—
+The prism thickness, extruded in the direction of `axis`. `mp.inf` can be used for infinite height.
+
+**`axis` [`vector3`]**
+—
+The axis perpendicular to the prism. Defaults to `(0,0,1)`.
+
+**`center` [`vector3`]**
+—
+If `center` is *not* specified, then the coordinates of the `vertices` define the *bottom* of the prism
```
ChristopherHogan: I don't think this is true until we add the equivalent of #441 for scheme. I guess that can be accomplished via `define-post-processed-property`? The API will be a bit different though because `center` is a required argument in scheme (inherited from `geometric-object`).

oskooi: Thanks for pointing this out. I have removed the reference to `center` in the Scheme documentation. It would be good to add support for this in order to be consistent with the Python interface.

PR #462, opened 2018-08-10
Account for symmetries in fragment stats

ChristopherHogan:
Only supports `Mirror` symmetries for now.
@stevengj @oskooi


ChristopherHogan:
No longer halving the dft pixels for symmetry. This is green now.

Issue #463, opened 2018-08-11
bug: load_minus_flux_data in parallel simulations


oskooi:
There is a bug in `load_minus_flux_data` when used in parallel simulations. This is demonstrated in the following simulation script which is adapted from the [waveguie taper](http://meep.readthedocs.io/en/latest/Python_Tutorials/Mode_Decomposition/#reflectance-of-a-waveguide-taper) tutorial example. The simulation consists of two parts: (1) a straight waveguide to compute the incident flux due to the source, and (2) the waveguide taper to compute the reflected flux. In both cases, the flux values obtained via `get_fluxes` can be directly compared with those obtained by squaring the mode coefficients via `get_eigenmode_coefficients` (since the waveguide is single mode). This provides a way to validate the results. The incident and reflected flux computed using the two approaches are displayed as part of the output in the lines prefaced by `incident:` and `refl:`. When run using a single process, the flux values are as expected. However the results are different when run with 2 or more processes.

**Simulation Script**
```py
import meep as mp

resolution = 20

w1 = 1            # width of waveguide 1                                                                                                         
w2 = 2            # width of waveguide 2                                                                                                         
Lw = 10           # length of waveguide 1/2                                                                                                      
Lt = 4            # length of taper                                                                                                              

Si = mp.Medium(epsilon=12.0)

dair = 3.0
dpml = 5.0

sx = dpml+Lw+dpml
sy = dpml+dair+w2+dair+dpml
cell_size = mp.Vector3(sx,sy,0)

prism_x = sx+1
half_w1 = 0.5*w1
half_w2 = 0.5*w2

# straight waveguide                                                                                                                             
vertices = [mp.Vector3(-prism_x,half_w1),
            mp.Vector3(prism_x,half_w1),
            mp.Vector3(prism_x,-half_w1),
            mp.Vector3(-prism_x,-half_w1)]

boundary_layers = [mp.PML(dpml)]

# mode wavelength                                                                                                                                
lcen = 6.67
# mode frequency                                                                                                                                 
fcen = 1/lcen

sources = [mp.EigenModeSource(src=mp.GaussianSource(fcen,fwidth=0.2*fcen),
                              component=mp.Ez,
                              size=mp.Vector3(0,sy-2*dpml,0),
                              center=mp.Vector3(-0.5*sx+dpml+0.2*Lw,0,0),
                              eig_match_freq=True,
                              eig_parity=mp.ODD_Z+mp.EVEN_Y)]

[...]

PR #465, opened 2018-08-14
WIP: Fix bugs in _get_dft_data and _load_dft_data

ChristopherHogan:
We can't use `dft_chunks_Ntotal` here because it counts twice the number of dft points since it separates real and complex for writing to hdf5.

~~With the first reproducer from #463, I now get the same results when saving/loading the flux to hdf5 via `save_flux` and `load_minus_flux`, and keeping the flux in a numpy array via `get_flux_data` and `load_minus_flux_data`.~~

~~However, I still see an inconsistency. Running Ardavan's reproducer serially, both the hdf5 and numpy versions give~~
```
taper:, 0.00140979, 0.00144195
```
~~But when running it with 2 processes, both the hdf5 and numpy versions give~~
```
taper:, 0.00125602, -3.40803651
``` 
~~Thus, this PR doesn't completely solve the problem yet, but makes the numpy version incorrect in a way that's consistent with the hdf5 version.~~
#463 turned out not to be a bug.
@stevengj @oskooi

### DIFF
--- a/python/meep.i
+++ b/python/meep.i
@@ -377,50 +377,44 @@ PyObject *_get_dft_array(meep::fields *f, dft_type dft, meep::component c, int n
     return py_arr;
 }
 
-size_t _get_dft_data_size(meep::dft_chunk *dft_chunks) {
-    size_t n = 0;
-    for (dft_chunk *cur = dft_chunks; cur; cur = cur->next_in_dft)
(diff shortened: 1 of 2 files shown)


ChristopherHogan:
Refactored. This is green now.

PR #472, opened 2018-08-17
Fix filename_prefix bug

ChristopherHogan:
This aligns with scheme and the documentation, which states
```
If you don't want to use any prefix, then you should set filename_prefix to False.
```
@stevengj @oskooi

### DIFF
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -780,13 +780,13 @@ Miscellaneous Functions
 
 ### Output File Names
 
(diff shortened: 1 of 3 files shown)


ChristopherHogan:
Well, the problem is that I'm using two "falsey" things to convey two different meanings.
* `''` is the default, which means use the filename as the prefix
* `False` means use `''` as the prefix

How about
* `None`: default (filename as prefix)
* Any string (including `''`): that string as prefix

ChristopherHogan:
This is updated and green.

Issue #415, opened 2018-07-11
Enhancement: Medium Evaluation


smartalecH:
Along with #398, it would be nice to simply evaluate a Medium at a given frequency point and return the complex permittivity. This would be useful for verifying that a given dispersion profile really is what you intend. It's all too easy to incorrectly convert one dispersion format to Meep's...

This would also enable the same dispersion material definitions within MPB and the frequency domain solver.

stevengj:
Yes, that would be nice.  We've also discussed adding a `frequency` argument to `output-epsilon` or similar in order to output the dielectric profile evaluated at a given frequency.

smartalecH:
I agree that would also be useful.

Do the eigenmode source and eigenmode expansion features currently support material dispersion? If not, would this be a viable way to enable that feature?

smartalecH:
If this hasn't been assigned to anyone, I'm willing to work on this feature (I need it for my research).

stevengj:
Since all the susceptibilities are of the form `χ = σ f(ω)` (where `f` is a scalar and `σ` can be a tensor), we basically just need to give `f(ω)`.   Probably should start with a virtual method `double susceptibility::chi1(double freq, double sigma=1)` in C++ with a corresponding method in `lorentzian_susceptibility` that returns `σ f(ω)` for a scalar `σ` that defaults to `1`.

smartalecH:
Sorry this is taking so long. I'm still learning to balance my various research projects with a newborn...

I have a few questions regarding your preferred implementation:

Would it be practical to create a member function within `medium_struct` that loops through each susceptibility within its corresponding `E_susceptibilities` list, summing up each call to `susceptibility::chi1`? i.e. `double medium_struct::get_chi1(double freq)` could cleanly and natively account for multiple susceptibilities and for predefined `sigmas`. This function could even return a tensor for anisotropic materials (i.e `σ` is a tensor).

This approach seems to encourage a more object-oriented description of the `Medium` class in both python and C++. It could also (temporarily) enable dispersion within the eigenmode source/ deomposition features by adding it to the `meep_mpb_eps()` function (I'm looking at a possible alternative to the `get_chi1inv ()` function).

Or would you prefer to leave everythin to the user and stick strictly with the `susceptibility::chi1` member function? This approach seems easier to implement with the swig interface since the susceptibilities are explicately defined, while the python `Medium` class doesn't appear to have an explicit mapping to the C++ backend.

[...]

PR #475, opened 2018-08-22
Medium Evaluations

smartalecH:
A first attempt at adding medium evaluation functionality (see #415).

Added the function, `chi1`, that evaluates a susceptibility at a given frequency point. An optional conductivity term can be passed (defaults to 1).

Once we add more features, I'll work on a SWIG interface.


src/susceptibility.cpp
```diff
@@ -274,6 +279,16 @@ realnum *lorentzian_susceptibility::cinternal_notowned_ptr(
   return d->P[c][cmp] + n;
 }
 
+double lorentzian_susceptibility::chi1(double freq, double sigma) {
+  if (no_omega_0_denominator){
+    // Drude model
+    return sigma * omega_0*omega_0 / ( -freq*freq - sqrt(-1)*gamma*freq);
```
stevengj: I would do `std::complex<double>(-freq*freq, gamma*freq)` here rather than `sqrt(-1)`.

stevengj:
Merge conflict.  Can we have a test?

stevengj:
Fixed the merge conflict.  A test might be tricky to implement before we have a higher-level interface that gets ε(x,ω) or similar.

smartalecH:
Thanks for the conflict fix. I agree, it makes more sense to implement a test with slightly more functionality.

On that note, which direction would you like to go from here?

From a python perspective, it would be great to be able to do the following:
```
from meep.materials import SiO2
freq = np.linspace(1,2,100)
eps  = SiO2.getEps(freq)
```

where `getEps` is obviously a member function of the `Medium()` class and loops through all of the defined susceptibilites and returns the relative permitivity (similar to what you proposed). 

If we develop some simple C++ functions to do this and simply wrap them, I'm hoping we can use the same C++ functions to build mode sources and perform eigenmode decomposition with dispersive materials.

stevengj:
I agree that it would be nice to have this higher-level functionality.

We might end up having to copy these χ(ω) formulas to more than one place, however, because when you have a `Medium` object in Python you haven't yet instantiated a susceptibility object.

smartalecH:
As suggested, I copied the formulas over directly to the Python interface. I compared the MEEP output with the corresponding refractiveindex.info database, and they appear to align really well.

materialeval

Loading the refractive index data is as easy as the following:
```
    from meep.materials import Ag, Au, Cu, Si
    wavelength = np.linspace(1/f0,1/f1,1000)
    freq = 1/wavelength
    eps  = Au.get_eps(freq)
```
`get_eps()` outputs a list of `Vector3` objects to work with nonisotropic materials. It takes into account the diagonal of the permittivity tensor and the conductivity tensor.

I've attached the script and data used to generate the above plot.
[Archive.zip](https://github.com/stevengj/meep/files/2380764/Archive.zip)

smartalecH:
My latest commit modifies the `get_chi1inv` routine to account for possible `structure_chunk` susceptibilities (so long as the susceptibility has a `chi1` function defined). This new functionality enables dispersion compensation within the mode source and mode expansion features for "arbitrarily" dielectric materials only (i.e Si, SiO2, etc.).

In order to enable all materials, including metals, MPB's eigenmodemode solver would need to change slightly.

I'm working on a number of tests to verify:

**Test 1**
I create a simulation domain with a uniform dispersive medium and find the eigenmode k vectors for a variety of omegas for an arbitrary subdomain. With the calculated k vectors and the corresponding omegas, I find the effective index, which simply corresponds to Silicon's index of refraction (since there is no geometric dispersion). I compare it to the analytic refractive index profile (labeled 'Python'). As seen below, the results align well.

sidispersion
sio2dispersion

**Test 2**
I form a typcial silicon photonic waveguide (500nm width, 220 nm thickness) and find the effective index across the frequency band. I compare the results with a corresponding Lumerical simulation. While the results don't align perfectly, these kinds of simulations are typically sensitive to the mesh type and resolution. 

effectiveindex

I'm working on extracting the S parameters of a silicon photonic device from the SIEPIC toolbox (like the DC tutorial) and comparing them to the previously calculated results. This is my ultimate goal, and should attract the attention of several PIC designers. I'm debugging some weird issues that arise with higher resolutions.

cc @stevengj @HomerReid

stevengj:

stevengj:
In the short term, can we at least pull out the Python portion of this PR as a separate PR that we can merge?

smartalecH:
Working on C implementation in a separate pull request. Will update with more details.

PR #478, opened 2018-08-24
Add new mode_decomposition test

ChristopherHogan:
Ardavan requested a new test based on the updated `mode-decomposition.py` tutorial. Should this replace the existing `python/tests/mode_coeffs.py` or does it make sense to keep both?  
@stevengj @oskooi


stevengj:
TODO: separate PR removing the old test.

PR #479, opened 2018-08-24
Add materials_library.py to python package

ChristopherHogan:
This makes using the materials library more convenient and should help solve some recent confusion on the mailing list.
@stevengj @oskooi


stevengj:
Todo: rename to `meep.materials`.

PR #483, opened 2018-08-29
Include materials library in scheme

ChristopherHogan:
Also renamed it to `materials.scm`.
@stevengj @oskooi


ChristopherHogan:
This is rebased and green.

PR #488, opened 2018-09-04
Fix bug in structure::dump

ChristopherHogan:
Ardavan found a bug when dumping dispersive materials.

The code starting at `structure_dump.cpp:200` is trying to find the component and direction for each non-null sigma, which I call `sigma_cd`. Since this data is the same in each chunk that has non-null sigmas, I only need to find the data once, then make sure every processor has access to it. However, if multiple processors find the data, `sum_to_all` will not give us the correct values. Switching the reduce operator to "bitwise or" gives us the desired behavior.

Also fixed a hard-coded `E_stuff` to properly loop through field types.
@stevengj @oskooi

### DIFF
--- a/src/meep/mympi.hpp
+++ b/src/meep/mympi.hpp
@@ -87,6 +87,7 @@ void sum_to_all(const size_t *in, size_t *out, int size);
 void sum_to_master(const size_t *in, size_t *out, int size);
 bool or_to_all(bool in);
 void or_to_all(const int *in, int *out, int size);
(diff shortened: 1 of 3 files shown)


oskooi:
When writing the epsilon to a file via `dump_structure`, would it be possible to also print a line to standard output indicating so (i.e. `creating epsilon output file "./my-epsilon.h5"`, similar in format to the output of e.g., `mp.output_epsilon`)?

PR #489, opened 2018-09-04
Clarification in load dump structure docs

ChristopherHogan:
Ardavan asked for clarification on the following 2 points:

> 1) If symmetries were used in the original run which involved using dump_structure to create the epsilon file, must they also be used in the load_structure run?

 > 2) Does specifying load_structure in the Simulation constructor also support superimposing additional geometric objects using the geometry object?

I've attempted to address these in the docs, but I may be misunderstanding something.
@stevengj @oskooi

### DIFF
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -1152,7 +1152,7 @@ Scale the Fourier-transformed fields in `near2far` by the complex number `s`. e.
 
 ### Load and Dump Structure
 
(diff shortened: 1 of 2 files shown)


stevengj:
Loading ε and μ from the file overwrites anything that was set from the `geometry`, no?

I guess you could load ε and μ from the file but still use the geometry parameter for nonlinear properties, but I don't know how useful that would be.

ChristopherHogan:
You're right, it does overwrite everything. I've updated the docs. I also added some logging that Ardavan requested in a comment to #488.

PR #493, opened 2018-09-05
Add docs for libGDSII support

### LINKED ISSUE #487: missing support for GDSII in configure script, installation instructions, and Conda packages
The recently-added tutorial example for the GDSII import feature involving a [bidirectional coupler](https://meep.readthedocs.io/en/latest/Python_Tutorials/BidirectionalCoupler/) requires that `Meep must be configured/compiled with libGDSII for get_GDSII_prisms`. Currently, there doesn't seem to be a flag involving GDSII in the configure script for either libctl or Meep (or rather one that is documented and part of the `man` or `help` pages). Once this is resolved, it would be useful to also update the [installation instructions](https://meep.readthedocs.io/en/latest/Build_From_Source/) as well as the [Conda packages](https://meep.readthedocs.io/en/latest/Installation/#conda-packages) to include support for GDSII by default.

ChristopherHogan:
Closes #487.
@stevengj @oskooi @HomerReid

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -430,22 +430,10 @@ AC_SUBST(LIBCTLGEOM_LIBS)
 
 ##############################################################################
 # check for libGDSII
-AC_ARG_WITH(libGDSII, [AC_HELP_STRING([--without-libGDSII],[do not enable GDSII file support])], ok=$withval, ok=yes)
-LIBGDSII_LIBS=""
-if test "$ok" = "yes"; then
-    AC_CHECK_HEADER(libGDSII.h, [have_gdsii=maybe], [have_gdsii=no])
-    if test $have_gdsii = maybe; then
-        AC_CHECK_LIB(GDSII, libGDSIIExists, have_gdsii=yes, have_gdsii=no)
-    fi
-    if test "$have_gdsii" = "no"; then
-        AC_MSG_WARN([Disabling GDSII file support because libGDSII was not found])
-    else
-        AC_DEFINE(HAVE_LIBGDSII, 1, [Define if we have & link libGDSII])
-        LIBGDSII_LIBS="-lGDSII"
-    fi
-
+AC_CHECK_HEADER(libGDSII.h, [have_gdsii=maybe], [have_gdsii=no])
+if test $have_gdsii = maybe; then
+  AC_CHECK_LIB(GDSII, libGDSIIExists)
 fi
-AC_SUBST(LIBGDSII_LIBS)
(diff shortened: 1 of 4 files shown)


ChristopherHogan:
* Added configure behavior requested in #487. GDSII support is enabled if `configure` finds the library unless `--without-libGDSII` is specified.
* Added proper #define for `HAVE_LIBGDSII` and linked `libmeepgeom` with `libGDSII`. I'm not sure how this was ever working.

configure.ac
```diff
@@ -430,10 +430,22 @@ AC_SUBST(LIBCTLGEOM_LIBS)
...
+LIBGDSII_LIBS=""
+if test "$ok" = "yes"; then
+    AC_CHECK_HEADER(libGDSII.h, [have_gdsii=maybe], [have_gdsii=no])
+    if test $have_gdsii = maybe; then
+        AC_CHECK_LIB(GDSII, libGDSIIExists, have_gdsii=yes, have_gdsii=no)
+    fi
+    if test "$have_gdsii" = "no"; then
+        AC_MSG_WARN([Disabling GDSII file support because libGDSII was not found])
+    else
+        AC_DEFINE(HAVE_LIBGDSII, 1, [Define if we have & link libGDSII])
```
HomerReid: The `#define HAVE_LIBGDSII 1` and `LIBS+=-lGDSII` steps are automatically taken care of by the `AC_CHECK_LIB` macro.

HomerReid:
Thanks for adding the documentation. I don't think the modifications to `configure.ac` are needed---the `#define HAVE_LIBGDSII 1` and `LIBS+=-lGDSII` steps are already handled by the default behavior of `AC_CHECK_LIB` on success (admittedly, the autotools are byzantine and hard to figure out). With the existing `configure.ac` I've had no trouble configuring and building against `libGDSII` as long as the header and library binary are findable.

configure.ac
ChristopherHogan: It only takes those steps if `action-if-found` is not present. I added my own `action-if-found` (have_gdsii=yes) while trying to add Ardavan's request for the `--without` flag. That made it necessary to take the extra steps manually (though there's probably a better way). I guess the real question is exactly what the `--with/--without` flags and default behavior should be. I never got an answer in #487.

stevengj:
Let's revert the changes to configure.ac in this PR as discussed in https://github.com/stevengj/meep/issues/487#issuecomment-419528512

PR #499, opened 2018-09-07
include exp(ikx) in MPB mode amplitudes

stevengj:
Fixes the problem mentioned in https://github.com/stevengj/meep/issues/486#issuecomment-419548696

### DIFF
--- a/src/mpb.cpp
+++ b/src/mpb.cpp
@@ -98,6 +98,8 @@ typedef struct eigenmode_data
    double group_velocity;
  } eigenmode_data;
 
+#define TWOPI 6.2831853071795864769252867665590057683943388
+
 /*******************************************************************/
 /* compute position-dependent amplitude for eigenmode source       */
 /*  (similar to the routine formerly called meep_mpb_A)            */
@@ -133,8 +135,13 @@ complex<double> eigenmode_amplitude(void *vedata, const vec &p,
   int nz = n[2];
   double r[3] = {0,0,0};
   vec p0(p - center);
-  LOOP_OVER_DIRECTIONS(p.dim, d)
-   r[d%3] = p0.in_direction(d) / s[d%3] + 0.5;
+  double phase = 0;
+  LOOP_OVER_DIRECTIONS(p.dim, d) {
+    double pd = p0.in_direction(d);
+    int i = d%3;
+    phase += edata->Gk[i] * pd; // k dot p
+    r[i] = pd / s[i] + 0.5;
+  }
   double rx = r[0], ry = r[1], rz = r[2];
 
   /* linearly interpolate the amplitude from MPB at point p */
@@ -173,7 +180,7 @@ complex<double> eigenmode_amplitude(void *vedata, const vec &p,
 #undef D
 
   return (complex<double>(double(real(ret)), double(imag(ret)))
-	  * amp_func(p));
+	  * amp_func(p)) * std::polar(1.0, TWOPI*phase);
 }
 
 /***************************************************************/
--- a/src/mpb.cpp
+++ b/src/mpb.cpp
@@ -149,14 +149,19 @@ complex<double> eigenmode_amplitude(void *vedata, const vec &p,
   double dx, dy, dz;
 
   /* get the point corresponding to r in the epsilon array grid: */
-  x = pmod( int(rx * nx), nx );
-  y = pmod( int(ry * ny), ny );
-  z = pmod( int(rz * nz), nz );
+  x = int(rx * nx);
+  y = int(ry * ny);
+  z = int(rz * nz);
 
   /* get the difference between (x,y,z) and the actual point */
   dx = rx * nx - x;
   dy = ry * ny - y;
   dz = rz * nz - z;
+
+  /* wrap around to 0..n-1, assuming periodic boundaries */
+  x = pmod( x, nx );
+  y = pmod( x, ny );
+  z = pmod( x, nz );
 
   /* get the other closest point in the grid, with periodic boundaries: */
   x2 = pmod( (dx >= 0.0 ? x + 1 : x - 1), nx );
--- a/src/mpb.cpp
+++ b/src/mpb.cpp
@@ -160,8 +160,8 @@ complex<double> eigenmode_amplitude(void *vedata, const vec &p,
 
   /* wrap around to 0..n-1, assuming periodic boundaries */
   x = pmod( x, nx );
-  y = pmod( x, ny );
-  z = pmod( x, nz );
+  y = pmod( y, ny );
+  z = pmod( z, nz );
 
   /* get the other closest point in the grid, with periodic boundaries: */
   x2 = pmod( (dx >= 0.0 ? x + 1 : x - 1), nx );
@@ -461,7 +461,7 @@ void *fields::get_eigenmode(double omega_src,
   /*--------------------------------------------------------------*/
   /* do a second round of post-processing to tabulate E-fields   -*/
   /* on a (separate) internal storage buffer.  (Previously       -*/
-  /* there was only one internal buffer which held either E-field */
+  /* there was only one internal buffer which held either E-field *
   /* or H-field data, but this is inconvenient for cases in which */
   /* you want the E and H fields of an eigenmode simultaneously.) */
   /*--------------------------------------------------------------*/
--- a/python/tests/mode_coeffs.py
+++ b/python/tests/mode_coeffs.py
@@ -98,8 +98,8 @@ class TestModeCoeffs(unittest.TestCase):
         eval_point = mp.Vector3(0.7, -0.2, 0.3)
         ex_at_eval_point = emdata.amplitude(eval_point, mp.Ex)
         hz_at_eval_point = emdata.amplitude(eval_point, mp.Hz)
-        self.assertAlmostEqual(ex_at_eval_point, 2.4827383502368017e-05 + 0.16095103621219753j, places=2)
-        self.assertAlmostEqual(hz_at_eval_point, 1.1381993864995965-3.7330585912764835e-06j, places=2)
+        self.assertAlmostEqual(ex_at_eval_point, 0.10591314723115094+0.12458333138964374j, places=2)
+        self.assertAlmostEqual(hz_at_eval_point, 0.8681206797182762-0.737695596449452j, places=2)
 
     def test_kpoint_func(self):
 
--- a/python/tests/wvg_src.py
+++ b/python/tests/wvg_src.py
@@ -41,8 +41,8 @@ class TestWvgSrc(unittest.TestCase):
         flux1 = self.sim.flux_in_box(mp.X, mp.Volume(center=mp.Vector3(-6.0), size=mp.Vector3(1.8, 6)))
         flux2 = self.sim.flux_in_box(mp.X, mp.Volume(center=mp.Vector3(6.0), size=mp.Vector3(1.8, 6)))
 
-        self.assertAlmostEqual(flux1 / -1.8, 7.087977151878242e-05)
-        self.assertAlmostEqual(flux2 / 1.8, 0.2898407385370755)
+        self.assertAlmostEqual(flux1 / -1.8, 7.125912849303947e-5)
+        self.assertAlmostEqual(flux2 / 1.8, 0.2898407833272916)
 
 if __name__ == '__main__':
     unittest.main()
--- a/python/tests/wvg_src.py
+++ b/python/tests/wvg_src.py
@@ -41,8 +41,8 @@ class TestWvgSrc(unittest.TestCase):
         flux1 = self.sim.flux_in_box(mp.X, mp.Volume(center=mp.Vector3(-6.0), size=mp.Vector3(1.8, 6)))
         flux2 = self.sim.flux_in_box(mp.X, mp.Volume(center=mp.Vector3(6.0), size=mp.Vector3(1.8, 6)))
 
-        self.assertAlmostEqual(flux1 / -1.8, 6.372086032775481e-06)
-        self.assertAlmostEqual(flux2 / 1.8, 0.2874654567549529)
+        self.assertAlmostEqual(flux1 / -1.8, 7.087977151878242e-05)
+        self.assertAlmostEqual(flux2 / 1.8, 0.2898407385370755)
 
 if __name__ == '__main__':
     unittest.main()


oskooi:
Failing tests:

**python/tests/mode_coeffs.log**
```
FAILED (failures=1)

FAIL: test_modes (__main__.TestModeCoeffs)
  File "./tests/mode_coeffs.py", line 99, in test_modes
    self.assertAlmostEqual(ex_at_eval_point, 2.4827383502368017e-05 + 0.16095103621219753j, places=2)
AssertionError: (0.1059133133272661+0.12458350200050837j) != (2.4827383502368017e-05+0.16095103621219753j) within 2 places

```

**python/tests/wvg_src.log**
```
FAIL: test_wvg_src (__main__.TestWvgSrc)
  File "./tests/wvg_src.py", line 44, in test_wvg_src
    self.assertAlmostEqual(flux1 / -1.8, 6.372086032775481e-06)
AssertionError: 7.087973785172087e-05 != 6.372086032775481e-06 within 7 places

Ran 1 test in 0.448s

FAILED (failures=1)

FAIL: test_wvg_src (__main__.TestWvgSrc)
  File "./tests/wvg_src.py", line 44, in test_wvg_src
    self.assertAlmostEqual(flux1 / -1.8, 6.372086032775481e-06)
AssertionError: 7.087973785172087e-05 != 6.372086032775481e-06 within 7 places

```

stevengj:
Looks like the test failure isn't a problem.  There is a slight change in the output in this test case because the waveguide source plane interpolates to sources at points ±1 pixel in the waveguide direction, and these points now pick up very slightly different phases and hence change the result slightly.  It should converge to the same value as before with increasing resolution, and does not increase the discretization error.

stevengj:
Two things:

1. Look at the modes with `get_eigemode` and see if they are planewaves with the expected ky

2. Try creating the source with an eigenmode source (which should also pick up the ky from the boundary conditions automatically).  If you do this in an empty computational cell and just look at the fields it is producing, it should be clear whether it is producing the desired planewave or not.

oskooi:
Trying out the second test described above which involves using an eigenmode source to create a planewave in homogeneous media at normal and oblique incidence, the results are incorrect even with the fixes in this PR.

The following script involves a 2d simulation with PML boundaries in _x_ and Bloch-periodic boundaries in _y_. A pulsed source with E<sub>z</sub> polarization is positioned near the -_x_ boundary and there is a line monitor (spanning the entire length of the computational cell along _y_) to compute the Fourier-transformed E<sub>z</sub> fields near the +_x_ boundary. The test involves verifying that the real part of the monitor fields at a single frequency (same as the source) is a sinusoid consistent with a planewave. Four different angles of the incident source are used: 0°, 10.7°, 22.9°, and 38.3°. The field plots are shown below.

```py
import meep as mp
import math

resolution = 25        # pixels/μm                                                                                                                                             

dpml = 2               # PML thickness                                                                                                                                         
dmat = 10              # length of material                                                                                                                                    

sx = dpml+dmat+dpml
sy = 10

pml_layers = [mp.Absorber(thickness=dpml,direction=mp.X)]

fcen = 1               # center frequency (wavelength = 1 μm)                                                                                                                  
df = 0.01              # frequency width                                                                                                                                       

ng = 1.5
glass = mp.Medium(index=ng)

# angle of incident planewave; CCW about Y axis, 0 degrees along +X axis                                                                                                       
theta_in = math.radians(10.7)

# k (in vacuum) with correct length (plane of incidence: XY)                                                                                                                   
k = mp.Vector3(math.cos(theta_in),math.sin(theta_in),0).scale(fcen)

eig_parity=mp.ODD_Z
if theta_in == 0:
  k = mp.Vector3(0,0,0)
  eig_parity += mp.EVEN_Y

src_pt = mp.Vector3(-0.5*sx+dpml+0.2*dmat,0,0)
sources = [mp.EigenModeSource(mp.GaussianSource(fcen,fwidth=df),
                              center=src_pt,
                              size=mp.Vector3(0,sy,0),
                              eig_parity=eig_parity)]

                    cell_size=cell_size,
                    boundary_layers=pml_layers,
                    k_point=k,
                    default_material=glass,
                    sources=sources)

mon_pt = mp.Vector3(0.5*sx-dpml-0.2*dmat,0,0)
mon_dft = sim.add_dft_fields([mp.Ez], fcen, fcen, 1, center=mon_pt, size=mp.Vector3(0,sy,0))

sim.run(until_after_sources=1000)

mode_ez = sim.get_dft_array(mon_dft, mp.Ez, 0)

import numpy as np

plt.figure(dpi=150)
y = np.linspace(-0.5*sy,0.5*sy,sy*resolution+2)
plt.plot(y,np.real(mode_ez),'bo-')
plt.xlabel("y coordinate (μm)")
plt.ylabel("real {Ez}")
plt.title("θ={:.1f}°".format(math.degrees(theta_in)))
plt.show()
```

ez_eigsrc_pr499_vary_theta

The following field plots are the results from today's master repo which differ more noticeably from the expected planewave profiles.

ez_eigsrc_vary_theta

For comparison, we replace the eigenmode source with an amplitude function via the following lines:

```py
# angle of incident planewave; CCW about Y axis, 0 degrees along +X axis                                                                                                       
theta_in = math.radians(10.7)

# k (in vacuum) with correct length (plane of incidence: XY)                                                                                                                   
k = mp.Vector3(math.cos(theta_in),math.sin(theta_in),0).scale(fcen)
if theta_in == 0:
  k = mp.Vector3(0,0,0)

def pw_amp(k,x0):
  def _pw_amp(x):
    return cmath.exp(1j*2*math.pi*ng*k.dot(x+x0))
  return _pw_amp

src_pt = mp.Vector3(-0.5*sx+dpml+0.2*dmat,0,0)
                     center=src_pt,
                     size=mp.Vector3(0,sy,0),
                     component=mp.Ez,
                     amp_func=pw_amp(k,src_pt))]
```

This produces the expect results.

ez_ampfunc_vary_theta

oskooi:
It's puzzling why the normally-incident case, which doesn't involve the phase at all, is producing results which are way off considering that the [binary grating tutorial example](https://meep.readthedocs.io/en/latest/Python_Tutorials/Mode_Decomposition/#diffraction-spectrum-of-a-binary-grating) which involved `get_eigenmode_coefficients` produced results which were correct.

oskooi:
Looks like there is a bug in a boundary pixel. The following script uses `get_eigenmode` to compute planewaves at four different `kpoint`s (based on the same four angles from above). The plot of the real E<sub>z</sub> fields at a cross section (identical to the line monitor region in the above script) in all cases (including a `k_point` of `mp.Vector3(0,0,0)` for "normal incidence") is correct except at a *single* pixel at the boundary.

```py
import meep as mp
import math

resolution = 25        # pixels/μm                                                                                                                                                          

dpml = 2               # PML thickness                                                                                                                                                      
dmat = 10              # length of material                                                                                                                                                 

sx = dpml+dmat+dpml
sy = 10

pml_layers = [mp.Absorber(thickness=dpml,direction=mp.X)]

fcen = 1               # center frequency (wavelength = 1 μm)                                                                                                                               
df = 0.01              # frequency width                                                                                                                                                    

ng = 1.5
glass = mp.Medium(index=ng)

# angle of incident planewave; CCW about Y axis, 0 degrees along +X axis                                                                                                                    
theta_in = math.radians(10.7)

# k (in vacuum) with correct length (plane of incidence: XY)                                                                                                                                
k = mp.Vector3(math.cos(theta_in),math.sin(theta_in),0).scale(fcen)

symmetries = []
eig_parity = mp.ODD_Z
if theta_in == 0:
  k = mp.Vector3(0,0,0)
  eig_parity += mp.EVEN_Y

                    cell_size=cell_size,
                    boundary_layers=pml_layers,
                    k_point=k,
                    symmetries=symmetries,
                    default_material=glass)

sim.init_sim()

mon_pt = mp.Vector3(0.5*sx-dpml-0.2*dmat,0,0)
mon_EigenmodeData = sim.get_eigenmode(fcen, mp.X, mp.Volume(center=mon_pt,size=mp.Vector3(0,sy,0)), 1, k, parity=eig_parity, resolution=resolution, eigensolver_tol=1e-13, verbose=True)

import numpy as np

plt.figure(dpi=150)
y = np.linspace(-0.5*sy,0.5*sy,sy*resolution+2)
Ez_real = [np.real(mon_EigenmodeData.amplitude(mp.Vector3(0,yc,0),mp.Ez)) for yc in y]
plt.plot(y,Ez_real,'bo-')
plt.xlabel("y coordinate (μm)")
plt.ylabel("real {Ez}")
plt.title("θ={:.1f}°".format(math.degrees(theta_in)))
plt.show()
```

ez_geteigenmode_vary_theta

This could explain why the results using the eigenmode source looked strange.

Also, the following warnings occurred multiple times in the MPB output. This may be another indication of something wrong with the fields at the boundaries (and why, as demonstrated above, the `eig_tolerance` had to be reduced significantly to improve accuracy).

```
...
linmin: WARNING: XTOL TEST SATISFIED
...
linmin: WARNING: ROUNDING ERRORS PREVENT PROGRESS
...
```

stevengj:
I don't think the 1-pixel error is related — that should be fixed, but should be causing only a small contribution to the error.

oskooi:
The boundary-pixel bug has been fixed. The results from `get_eigenmode` now disagree with the expected/analytic results by a factor of `ng` (the index of the default material which is `1.5`). The figure below shows the real and imaginary parts of E<sub>z</sub> (which are exactly 90° out of phase) for two angles: 10.7° and 22.9°. The wavelengths are `5.38` and `2.55` which is exactly equivalent to the *vacuum* wavelength: `1/k.y`. The expected results are `1/(ng*k.y)` which are `3.59` and `1.70` (matching the results from the `amp_func` source shown above). As a test, setting the default material to `ng = 1` produced *identical* results as `ng = 1.5`. Thus, the default index is not being passed to MPB. The fix should be straightforward.

ez_real_imag_geteigenmode

oskooi:
One important additional note: while today's boundary-pixel bug fix finally yields correct results for non-zero `k` (neglecting the remaining bug involving the default material described in the previous comment), the results are still incorrect for zero `k` as shown below.

ezr_geteigenmode_t0 0_pr499

stevengj:
I think there may be a misunderstanding here: the ky in the boundary conditions must match the ky used for the source (e.g. when you do it manually with an `amp_func`) — they do not differ by a factor of `n`.    (The medium then affects kx for a given ω — this is precisely the source of Snell's law.)

oskooi:
```py
import meep as mp
import cmath
import math

resolution = 25        # pixels/μm                                                                                                                                                                                  

dpml = 2               # PML thickness                                                                                                                                                                              
dmat = 10              # length of material                                                                                                                                                                         

sx = dpml+dmat+dpml
sy = 10

pml_layers = [mp.Absorber(thickness=dpml,direction=mp.X)]

fcen = 1               # center frequency (wavelength = 1 μm)                                                                                                                                                       
df = 0.01              # frequency width                                                                                                                                                                            

ng = 1.5
glass = mp.Medium(index=ng)

# angle of incident planewave; CCW about Y axis, 0 degrees along +X axis                                                                                                                                            
theta_in = math.radians(10.7)

# k (in vacuum) with correct length (plane of incidence: XY)                                                                                                                                                        
k = mp.Vector3(math.cos(theta_in),math.sin(theta_in),0).scale(fcen)*ng
if theta_in == 0:
  k = mp.Vector3(0,0,0)

def pw_amp(k,x0):
  def _pw_amp(x):
  return _pw_amp

src_pt = mp.Vector3(-0.5*sx+dpml+0.2*dmat,0,0)
                     center=src_pt,
                     size=mp.Vector3(0,sy,0),
                     component=mp.Ez,
                     amp_func=pw_amp(k,src_pt))]

                    cell_size=cell_size,
                    boundary_layers=pml_layers,
                    k_point=k,
                    default_material=glass,
                    sources=sources)

mon_pt = mp.Vector3(0.5*sx-dpml-0.2*dmat,0,0)
mon_dft = sim.add_dft_fields([mp.Ez], fcen, fcen, 1, center=mon_pt, size=mp.Vector3(0,sy,0))

sim.run(until_after_sources=1000)

mode_ez = sim.get_dft_array(mon_dft, mp.Ez, 0)

import numpy as np

plt.figure(dpi=150)
y = np.linspace(-0.5*sy,0.5*sy,sy*resolution+2)
plt.plot(y,np.real(mode_ez),'bo-')
plt.xlabel("y coordinate (μm)")
plt.ylabel("real {Ez}")
plt.title("θ={:.1f}°".format(math.degrees(theta_in)))
plt.show()
```

produces

ez_ampfunc_t10 7_oblique

oskooi:
```py
import meep as mp
import math

resolution = 25        # pixels/μm                                                                                                                                                                                  

dpml = 2               # PML thickness                                                                                                                                                                              
dmat = 10              # length of material                                                                                                                                                                         

sx = dpml+dmat+dpml
sy = 10

pml_layers = [mp.Absorber(thickness=dpml,direction=mp.X)]

fcen = 1               # center frequency (wavelength = 1 μm)                                                                                                                                                       
df = 0.01              # frequency width                                                                                                                                                                            

ng = 1.5
glass = mp.Medium(index=ng)

# angle of incident planewave; CCW about Y axis, 0 degrees along +X axis                                                                                                                                            
theta_in = math.radians(0)

# k (in vacuum) with correct length (plane of incidence: XY)                                                                                                                                                        
k = mp.Vector3(math.cos(theta_in),math.sin(theta_in),0).scale(fcen)*ng

symmetries = []
eig_parity = mp.ODD_Z
if theta_in == 0:
  k = mp.Vector3(0,0,0)
  eig_parity += mp.EVEN_Y

                    cell_size=cell_size,
                    boundary_layers=pml_layers,
                    k_point=k,
                    symmetries=symmetries,
                    default_material=glass)

sim.init_sim()

mon_pt = mp.Vector3(0.5*sx-dpml-0.2*dmat,0,0)
mon_EigenmodeData = sim.get_eigenmode(fcen, mp.X, mp.Volume(center=mon_pt,size=mp.Vector3(0,sy,0)), 1, k, parity=eig_parity, resolution=resolution, eigensolver_tol=1e-13, verbose=True)

import numpy as np

y = np.linspace(-0.5*sy,0.5*sy,sy*resolution+2)
Ez_data = [mon_EigenmodeData.amplitude(mp.Vector3(0,yc,0),mp.Ez) for yc in y]
Ezr = np.real(Ez_data)
Ezi = np.imag(Ez_data)

plt.figure(dpi=150)
plt.plot(y,Ezr,'bo-')
plt.xlabel("y coordinate (μm)")
plt.ylabel("real {Ez}")
plt.title("θ={:.1f}°".format(math.degrees(theta_in)))
plt.savefig("ezr_geteigenmode_t{:.1f}_PR499.png".format(math.degrees(theta_in)))
plt.show()

if k.y != 0:
  idx = np.where(Ezr < 0.9995*np.amin(Ezr))
  print("ezr-wavelength:, {:.5f}, {:.5f}".format(1/k.y,y[idx[0][-1]]-y[idx[0][0]]))

plt.figure(dpi=150)
plt.plot(y,Ezi,'bo-')
plt.xlabel("y coordinate (μm)")
plt.ylabel("imag {Ez}")
plt.title("θ={:.1f}°".format(math.degrees(theta_in)))
plt.savefig("ezi_geteigenmode_t{:.1f}_PR499.png".format(math.degrees(theta_in)))
plt.show()

if k.y != 0:
  idx = np.where(Ezi < 0.9995*np.amin(Ezi))
  print("ezi-wavelength:, {:.5f}, {:.5f}".format(1/k.y,y[idx[0][-1]]-y[idx[0][0]]))
```
ezi_geteigenmode_t0 0_pr499

ezr_geteigenmode_t0 0_pr499

src/mpb.cpp
```diff
@@ -133,24 +135,34 @@ complex<double> eigenmode_amplitude(void *vedata, const vec &p,
...
 
   /* get the difference between (x,y,z) and the actual point */
   dx = rx * nx - x;
   dy = ry * ny - y;
   dz = rz * nz - z;
 
+  /* wrap around to 0..n-1, assuming periodic boundaries */
+  x = pmod( x, nx );
+  y = pmod( y, ny );
+  z = pmod( z, nz );
```
oskooi: Don't these lines (162-164) conflict with line 170 further below in this file: `x = x % nx; y = y % ny; z = z % nz;`?

src/mpb.cpp
```diff
@@ -173,7 +185,7 @@ complex<double> eigenmode_amplitude(void *vedata, const vec &p,
 #undef D
 
   return (complex<double>(double(real(ret)), double(imag(ret)))
-	  * amp_func(p));
+	  * amp_func(p)) * std::polar(1.0, TWOPI*phase);
```
oskooi: It turns out that by simply negating the phase (i.e., `-TWOPI*phase`) the results from `get_eigenmode` (based on the latest commit `whoops`) are once again sinusoidal. A side-by-side comparison of the results from the `amp_func` source and `get_eigenmode` are shown below for 0°, 10.7°, 22.9°. The wavelength of the planewaves from `get_eigenmode` for non-zero `k` are *exactly* double those from the `amp_func` source.

Perhaps an extra factor of `2` was included somewhere when it should not have been?

ez_real_ampfunc_geteigenmode_vary_theta

oskooi: Removing `std::polar(1.0, TWOPI*phase)` from this line produces results which match the expected results from the `amp_func` source. This is shown in the figure below. It does not seem obvious why removing the `phase` yields the correct results but it does. The rest of this PR simply fixes the 1-pixel boundary bug. However, even with this fix, `python/tests/mode_coeffs.py` is still failing: the computed and expected results differ by much more than a few decimal places.

ez_real_ampfunc_geteigenmode_vary_theta2

ChristopherHogan:
The failing test Ardavan referred to above is specifically testing the `EigenmodeData.amplitude` function, which calls `meep::eigenmode_amplitude`, which was changed slightly in this PR, so the results will be different.

When I originally wrote the test, I was just ensuring that the result of `EigenmodeData.amplitude` was the same as the result I got from calling `meep::eigenmode_amplitude` in a convoluted way through SWIG objects, and that the high level interface didn't introduce any errors. We probably need a more robust test that's based on theoretical results. Any suggestions?

stevengj:
I think the answer is correct here, you just aren't asking it for the right mode.

Your boundary conditions specify ky=0.2785.  This means that you want any solution where the phase from y=0 to y=sy=10 is exp(2πi ky 10).   But there are many planewave angles (diffraction orders that satisfy this).

The one you are hoping to get is (kx,ky) = (1.47392,0.2785), corresponding to your 10.6° angle.  But you are asking it for the **first** mode, i.e. the **lowest** diffraction order, and the lowest diffraction order at this frequency is actually at a much smaller angle.

The one you want in this case is the 6th mode.  If you ask it for that mode, with this PR it looks like what you expect.

oskooi:
The two tests which are failing (`python/tests/mode_coeffs.py` and `python/tests/wvg_src.py`) require changes to the hard-coded assert values due to the modifications in this PR to `eigenmode_amplitude`:

**mode_coeffs.py**
```
diff --git a/python/tests/mode_coeffs.py b/python/tests/mode_coeffs.py
index 7a27bb1..f9aa4f8 100644
--- a/python/tests/mode_coeffs.py
+++ b/python/tests/mode_coeffs.py
@@ -98,8 +98,8 @@ class TestModeCoeffs(unittest.TestCase):
         eval_point = mp.Vector3(0.7, -0.2, 0.3)
-        self.assertAlmostEqual(ex_at_eval_point, 2.4827383502368017e-05 + 0.16095103621219753j, places=2)
-        self.assertAlmostEqual(hz_at_eval_point, 1.1381993864995965-3.7330585912764835e-06j, places=2)
+        self.assertAlmostEqual(ex_at_eval_point, 0.10591315784691106+0.12458329847477048j, places=2)
+        self.assertAlmostEqual(hz_at_eval_point, 0.868120843295042-0.7376957400298366j, places=2)
```
**wvg_src.py**
```
diff --git a/python/tests/wvg_src.py b/python/tests/wvg_src.py
index ff15761..3d24dab 100644
--- a/python/tests/wvg_src.py
+++ b/python/tests/wvg_src.py
@@ -41,8 +41,8 @@ class TestWvgSrc(unittest.TestCase):
         flux1 = self.sim.flux_in_box(mp.X, mp.Volume(center=mp.Vector3(-6.0), size=mp.Vector3(1.8, 6)))
         flux2 = self.sim.flux_in_box(mp.X, mp.Volume(center=mp.Vector3(6.0), size=mp.Vector3(1.8, 6)))
 
-        self.assertAlmostEqual(flux1 / -1.8, 6.372086032775481e-06)
-        self.assertAlmostEqual(flux2 / 1.8, 0.2874654567549529)
+        self.assertAlmostEqual(flux1 / -1.8, 7.125911840794114e-05)
+        self.assertAlmostEqual(flux2 / 1.8, 0.28984078367708505)
```

ChristopherHogan:
@oskooi, I think `mode_coeffs.py` is okay in this PR. The travis logs only show that `python/tests/wvg_src.py` is failing, not `python/tests/mode_coeffs.py`, and the diff you posted for `mode_coeffs.py` seems to be based off a different branch. This branch already has values similar to the change you suggest. The `wvg_src.py` diff also seems to be based off a different branch, but the values you end up with are similar to what I get when running `scheme/examples/wvg-src.ctl` as a reference:

I think the only change required is

```diff
diff --git a/python/tests/wvg_src.py b/python/tests/wvg_src.py
index 2ff0a72..99d3d77 100644
--- a/python/tests/wvg_src.py
+++ b/python/tests/wvg_src.py
@@ -41,8 +41,8 @@ class TestWvgSrc(unittest.TestCase):
         flux1 = self.sim.flux_in_box(mp.X, mp.Volume(center=mp.Vector3(-6.0), size=mp.Vector3(1.8, 6)))
         flux2 = self.sim.flux_in_box(mp.X, mp.Volume(center=mp.Vector3(6.0), size=mp.Vector3(1.8, 6)))
 
-        self.assertAlmostEqual(flux1 / -1.8, 7.087977151878242e-05)
-        self.assertAlmostEqual(flux2 / 1.8, 0.2898407385370755)
+        self.assertAlmostEqual(flux1 / -1.8, 7.125912849303947e-5)
+        self.assertAlmostEqual(flux2 / 1.8, 0.2898407833272916)

```

stevengj:
Note that the k=0 mode your plot is actually nearly constant to about 6 digits, which is what is to be expected given the eigensolver tolerance.  See #535

PR #500, opened 2018-09-08
Working and verified version of multilevel atom susceptibility.

acerjan:
This is the version of the code that is verified with SALT, as we discussed in the issue.

### DIFF
--- a/scheme/examples/multilevel-atom.ctl
+++ b/scheme/examples/multilevel-atom.ctl
@@ -1,0 +1,84 @@
+;; This file realizes a 1D, one-sided Fabry-Perot laser, as described in Fig. 2 of Cerjan et al.,
+;; Opt. Express 20, 474 (2012).
+
+;; Cavity definitions
+(set-param! resolution 400)
+(define-param ncav 1.5)                          ; cavity refractive index
+(define-param Lcav 1)                            ; cavity length
+(define-param dpad 1)                            ; padding thickness
+(define-param dpml 1)                            ; PML thickness
+(define-param sz (+ Lcav dpad dpml))
+(set! geometry-lattice (make lattice (size no-size no-size sz)))
+(set! dimensions 1)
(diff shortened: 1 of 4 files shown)


stevengj:
Great!  Ideally you can also include documentation in the PR, including specification of precisely what equations are implemented and the definitions of the parameters.  (Note that the markdown documentation can include LaTeX equations.)

acerjan:
I'm working on the documentation now, and I can put together a separate PR, sure. it took a while to get mkdocs running correctly.

I understand conceptually what you want by a regression test, but I'll need a bit more guidance as to where this goes in the package. Can you point me in the direction of the other tests?

oskooi:
The C++ tests are in [tests/](https://github.com/stevengj/meep/tree/master/tests) and the Python tests in [python/tests/](https://github.com/stevengj/meep/tree/master/python/tests). A Python test would be sufficient but we need to add the front end first (which we are working on now). @ChristopherHogan can help you port the existing [Scheme multilevel example](https://github.com/stevengj/meep/blob/master/scheme/examples/multilevel-atom.ctl) to Python as a unit test.

oskooi:
For reference, here are the commands to install and run `mkdocs` in a separate Conda environment for viewing the Meep documentation via a local web server:

```
conda create -n md -c conda-forge mkdocs python-markdown-math
source activate md
pip install --user mkdocs-material
cd /path/to/meep/
mkdocs serve
source deactivate
```

acerjan:
For the documentation, I've written the general theory along with some verification plots into Materials.md. Are there additional locations where you'd like for me to provide documentation? Will you be writing documentation for the scheme and python interfaces?

oskooi:
Please also update the [Scheme tutorial example](https://meep.readthedocs.io/en/latest/Scheme_Tutorials/Multilevel_Atomic_Susceptibility/) as necessary with additional information based on the recent changes. We will port this tutorial to Python and also provide documentation for the interface.

PR #507, opened 2018-09-13
change normalization of eigenmode sources to yield unit input power

### LINKED ISSUE #470: normalize eigenmode source to unit power
It would be possible to change the normalization of the eigenmode source so that the eigenmodes are launched with unit power (optionally multiplied by a user specified complex amplitude).  This way you wouldn't have to do a separate normalization run.

(Note that this could also be used for planewave sources, just via an eigenmode source in a uniform region.)

One drawback of this would be that it would only be unit power up to discretization error, i.e. it would converge to unit power, but I'm not sure that would be a big deal.  (For the specific case of a planewave source, in principle you could solve for the exact discrete-space unit-power planewave, as reviewed in section 4.2.2 of https://arxiv.org/abs/1301.5366).

(Suggested by Colton Bukowsky on meep-discuss.)

HomerReid:
Fixes #470. Renormalize the amplitude of eigenmode sources to yield unit power flux.

The power flux depends on the frequency width of the source. I added a `width` parameter to `fields::get_eigenmode` with a default value of 0. (If `width==0` the amplitude renormalization is skipped.) This requires updating calling conventions in a couple of other places: 

(1) Within `add_eigenmode_source,` the frequency width can in principle be fetched from the existing `src` input parameter, so the calling convention doesn't need to change. However, the frequency width is a private data field in `gaussian_src_time` and `continuous_src_time,` and there were no getter functions. I added a virtual getter routine `get_width()` to the parent `src_time` class, which returns 0 by default and is overridden by `gaussian_src_time` and `continuous_src_time`.

(2) I updated the `_get_eigenmode` python routine to accept a new optional `width=0.0` argument. (Again, passing `width=0.0` to `get_eigenmode` yields behavior unchanged from before this PR.) If a user calls `get_eigenmode` directly from python and wants the amplitude renormalization` the user must specify a nonzero value for `width`. 

(3) No modification is needed for `get_eigenmode_coefficients.` The overall normalization of the eigenmodes cancels out of this calculation anyway, so it suffices to pass `width=0.0` yielding behavior equivalent to before.

### DIFF
--- a/python/meep.i
+++ b/python/meep.i
@@ -450,10 +450,10 @@ struct py_eigenmode_data {
 
 py_eigenmode_data _get_eigenmode(meep::fields *f, double omega_src, meep::direction d, const meep::volume where,
                                  const meep::volume eig_vol, int band_num, const meep::vec &_kpoint,
-                                 bool match_frequency, int parity, double resolution, double eigensolver_tol,
-                                 bool verbose) {
+                                 bool match_frequency, int parity, double resolution, double eigensolver_tol, 
+                                 double width, bool verbose) {
     void *data = f->get_eigenmode(omega_src, d, where, eig_vol, band_num, _kpoint, match_frequency,
-                                    parity, resolution, eigensolver_tol, verbose);
+                                    parity, resolution, eigensolver_tol, width, verbose);
     meep::eigenmode_data *emdata = (meep::eigenmode_data *)data;
 
     py_eigenmode_data result = {};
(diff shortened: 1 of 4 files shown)


stevengj:
I think the mode normalization should not depend upon the time profile.  You need to know the whole time-profile Fourier transform anyway to use this, so you might as well include the time-profile normalization there too.

In particular, the source current looks like mode(x)*p(t), where p(t) is the time profile.   My suggestion is:

* Normalize mode(x) to be unit power, just like in the mode-coefficient code.  This is independent of p(t), whether it is Gaussian or whatever.

* If you have a CW source, then it should result in unit power.

* If you have a Gaussian or some other source, you take this unit power and multiply by |p̂(ω)|² to get the expected spectrum (modulo the discretization error), where p̂(ω) is a Fourier transform of p(t) with an appropriate normalization.   We will update the manual to provide p̂(ω) for the Gaussian source (and maybe have a Python method for this).

stevengj:

PR #508, opened 2018-09-13
Add get_center_and_size

### LINKED ISSUE #506: Python API for GDSII regions
The recently added [GDSII import tutorial example](https://meep.readthedocs.io/en/latest/Python_Tutorials/BidirectionalCoupler/) involves a utility function `get_center_and_size` which returns the `center` and `size` of a C++ `meep::volume` object returned by `get_GDSII_volume`. Given that `get_center_and_size` is necessary for parsing rectilinear objects (i.e., sources and monitors) which can be part of the GDSII file itself, it would be convenient to add this function to the Python interface.

ChristopherHogan:
Closes #506. The original implementation used `x` for 1d center and size, but I thought 1d vecs used `z` so I changed it. Is that correct? 
@stevengj @oskooi


ChristopherHogan:
Rebased and green.

PR #517, opened 2018-09-19
Documentation for multilevel atom susceptibility

acerjan:
Finally found some time to write up some documentation on the multilevel atom susceptibility, adding text to Materials.md and rewriting Scheme_Tutorials/Multilevel_Atom_Susceptibility.md. I did not write the corresponding documentation for python. 

I have one other question for @oskooi, that should be reflected in the documentation, but at the moment I don't understand how it works, namely, how does one specify the vectorial nature of sigma in Scheme? I see how this is being done in python, but both of these are maybe a bit wrong. If (sigma sigma-21) in scheme results in a gain medium that can couple to Ex, Ey, and Ez, (as appears to be explicitly stated in the python example) we should change this to show how to specify sigma to only couple to Ex (and change python to also only couple to Ex). In our current example we're saved because Ey and Ez are zero, and there's nothing birefringent about the system. But, in general these gain media have an orientation, and we need to force the user to specify this, because it matters. 

Also, I saw that @ChristopherHogan added a test using the python example, so it sounds like I'm off the hook for that?


oskooi:
The polarization can be defined for a single field component as Meep supports anisotropic (real-symmetric) σ. The `sigma` defined for each `susceptibility` object is actually `sigma-diag` (see [scheme/meep.scm.in:#L49](https://github.com/stevengj/meep/blob/master/scheme/meep.scm.in#L49)). Thus, we just need to change `(sigma 1)` in [scheme/examples/multilevel-atom.ctl#L61](https://github.com/stevengj/meep/blob/master/scheme/examples/multilevel-atom.ctl#L61) (and also in the tutorial) to `(sigma-diag 1 0 0)`. A comment regarding this point in the tutorial and related file would also be useful. Perhaps you can make this change in this PR as an extra commit? We will make the changes to the Python version in a separate PR.

oskooi:
The failing test here can be safely ignored since it is related to a timeout issue in Travis which frequently (and inexplicably) arises.

acerjan:
Ok, so you're saying that in Line 61 of multilevel-atom.ctl it needs to be (sigma-diag 1 0 0), but that (sigma sigma-21) in line 63 stays the same? in other words, sigma really gets calculated as something like (sigma sigma-21) * (sigma-diag 1 0 0)?

If so, sure, i can fix this, change the documentation, and remake the commit.

oskooi:
Yes, that's correct: the two `sigma` terms are multiplied together.

acerjan:
Ok, both python and scheme examples are updated with the discussion about sigma, as is the documentation for the scheme example. Someone will still need to write the Python tutorial, but otherwise maybe this feature is all set to go? (still somewhat hard to believe...)

stevengj:
Thanks for writing this, Alex!

PR #522, opened 2018-09-24
Guard config.h in HAVE_CONFIG_H.

postmasters:
When we cannot run `configure.sh` to generate `config.h`, this will
allow us to build by providing the defines via CXXFLAGS and friends.

### DIFF
--- a/libpympb/pympb.cpp
+++ b/libpympb/pympb.cpp
@@ -4,7 +4,9 @@
 #include <cstddef>
 #include <iostream>
 
-#include "config.h"
+#ifdef HAVE_CONFIG_H
+#  include "config.h"
+#endif
 #include "pympb.hpp"
 #include "mpb/scalar.h"
 #include "meep/mympi.hpp"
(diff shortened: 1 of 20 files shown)


stevengj:
Is it really practical to define all of the things manually?  Even if you are doing this, wouldn't it be easier to edit `config.h` from another machine?

If we have to do this, it seems preferable to put the `#ifdef HAVE_CONFIG_H` into `config.h` itself if that is possible, so that we don't need to edit every file that includes `config.h`.

jeinarsson:
FYI I accidentally ran into one inconsistency the other week because config.h _is_ actually guarded in [GDSIIgeom.cpp](https://github.com/NanoComp/meep/blob/4bafe6e4d28fffeac7dca393a495564bae2189e2/src/GDSIIgeom.cpp#L28). Perhaps that should go, too.

stevengj:
Note: [this stackoverflow answer](https://stackoverflow.com/a/67930642/3662120) shows how to add include guards directly into the header file.

PR #525, opened 2018-09-25
comment out custom_dir in docs

oskooi:
mkdocs, which is used to build the documentation, on [readthedocs](https://meep.readthedocs.io/en/latest/) seems to have suddenly stopped supporting `custom_dir`. This is a temporary solution in order to re-enable the docs for the time being while a solution is sought.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -9,7 +9,7 @@ site_dir: 'doc/site'
 
 theme:
    name: readthedocs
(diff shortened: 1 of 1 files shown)


oskooi:
Yes. @HomerReid put the custom theme together. Unfortunately, it doesn't seem to play well with recent versions of mkdocs for reasons we are still looking into. In the meantime, we will just have to use the default theme.

PR #533, opened 2018-10-02
Add missing parameter in add-source

### LINKED ISSUE #532: amp-func and eigenmode-source
In the user reference, it states that an amp-func can be specified for an eigenmode-source, but I'm having a couple of questions / issues with this.

1) is this functionality currently working? I've successfully specified amp-func for an ordinary source, but when I change to an eigenmode-source, meep quits with the message "ERROR: In procedure meep-fields-add-eigenmode-source: No matching method for generic function `meep_fields_add_eigenmode_source'"

2) When solving for an eigenmode-source with a non-zero eig-kpoint, does the mode that MPB returns, u_(n,k) (r), contain the factor of e^(i k r)? Or is it only the Bloch mode u_(n,k) (r)? This is not clear in the user reference.

Thanks!

ChristopherHogan:
Fixes #532.
The `eigenmode-source` `amp-func` branch was missing the `eig-match-freq?` parameter.
@stevengj @oskooi

### DIFF
--- a/scheme/meep.scm.in
+++ b/scheme/meep.scm.in
@@ -416,6 +416,7 @@
 	       (volume (center ecen) (size esz))
 	       (object-property-value s 'eig-band)
 	       (object-property-value s 'eig-kpoint)
+	       (object-property-value s 'eig-match-freq?)
 	       (object-property-value s 'eig-parity)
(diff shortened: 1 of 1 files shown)


acerjan:
Thank you!

PR #539, opened 2018-10-03
python test for binary grating with oblique planewave source

oskooi:
Add a test to the Python suite based on the [binary grating](https://meep.readthedocs.io/en/latest/Python_Tutorials/Mode_Decomposition/#diffraction-spectrum-of-a-binary-grating) tutorial example which verifies that the total reflectance and transmittance from all diffracted orders for an oblique planewave source match their corresponding Poynting flux values. Should require a runtime of less than a minute.


ChristopherHogan:
I restarted the travis builds since #546 is merged.

ChristopherHogan:
Probably needs a rebase to run with the #546 changes.

oskooi:
This is ready to be merged.

PR #538, opened 2018-10-03
python test for kdom

oskooi:
Add a test for kdom to the Python suite. Requires just a few seconds to run.

PR #529, opened 2018-09-28
tutorial example for binary grating at oblique incidence

oskooi:
Closes #486.

This should be merged after #499. We will also need to add a new test.

PR #540, opened 2018-10-03
Travis is skipping non-documentation builds

ChristopherHogan:
Builds were unintentionally skipped for #538 and #539 as a result of #529. This may take some trial and error since it's difficult to test. I'll keep an eye on all PR builds for a while.
@stevengj @oskooi


ChristopherHogan:
This is complex enough that I propose reverting #529, and just using the built-in travis functionality to skip PRs that don't change any code. Adding `[skip travis]` to the commit message will skip the build. This is much more flexible and less error prone.

PR #542, opened 2018-10-03
minor tutorial typos

soamaven:
came across these when copying/pasting code from docs


stevengj:
Thanks!

stevengj:
In the future, it would be good to have a way to automatically sync the documentation with the python/examples files.

PR #543, opened 2018-10-03
Fix syntax error in multilevel atom example

ChristopherHogan:
The error was introduced in #521.
@stevengj @oskooi


ovidiopr:
Hi,

Could you please explain in more detail the definition of the transition parameters? Following the paper by Cerjan [Opt. Express 20 (2012) 474-488] I was able to figure out how to convert omega_a and gamma_perp but I don't understand why you take theta=1 and N0=37. Also, it would be nice if you can explain how to do the full conversion starting with experimental parameters.

Thank you very much and best regards,

Ovidio

PD: An example for a four-level atom would also be very helpful to me (and probably others).

stevengj:
They are described in detail in: https://arxiv.org/abs/2007.09329

ovidiopr:
Thank you very much! I will check it.

PR #545, opened 2018-10-04
Increase multilevel_atom.py resolution for MKL convergance

ChristopherHogan:
This test needs higher resolution for the MKL version to pass. I verified that the `DGETRF` and `DGETRI` calls produce the same results with and without MKL.
@stevengj @oskooi


ChristopherHogan:
Restarted the travis builds since #546 is merged.

PR #546, opened 2018-10-04
Revert 528

ChristopherHogan:
As explained in #540, it seems more flexible and reliable to use `[skip travis]` in commit messages instead of trying to automate it, as long as Ardavan is ok with it. 
@stevengj @oskooi


stevengj:
In general you write [ci skip], which skips all tests (Travis or other systems we might enable in the future)

Issue #502, opened 2018-09-09
get_array clarification


smartalecH:
While I use the `get_array` function in the Python API quite frequently, I can't seem to predict its behavior 100% of the time.

For example, if I pass a (1D) volume that is 4 units long with a predefined simulation resolution of 10, I would expect `get_array` to return an array of length 40. Instead, it returns an array of length 42.

Now, If I pass that same volume on a simulation domain with a resolution of 15, it returns an array of length 63.

I'm assuming this has something to do with Meep's averaging (grid interpolation), but I feel like the results should be much more predictable. The documentation is a little vague in this regard.

Am I missing something obvious here?

stevengj:
This depends sensitively on whether the volume lies exactly on a pixel boundary or somewhere in between, which depends on the resolution etc.

stevengj:
It might be useful to include some metadata about the array axes in the return value.

ChristopherHogan:
`xarray` has a 2 dependencies that require hdf5, which means I would have to build and maintain those conda packages with parallel hdf5 support.  I would prefer to avoid that if possible until conda-forge can support a parallel hdf5 package. I'm not 100 percent clear on what is needed. Is it something that can be done with a subclass of `numpy.ndarray`?

stevengj:
An alternative would be to add a `xaxis, yaxis, zaxis = simulation.volume_axes(volume)` function that returned 3 1d arrays (or 3 ranges) corresponding to the coordinates along each axis for a given volume (for `get_array` etcetera).

ChristopherHogan:
This should be fixed by #655.

PR #552, opened 2018-10-09
Get array metadata

HomerReid:
This PR adds functionality that I needed for implementing adjoint sensitivity analysis, but which also addresses #502.

It provides new functions called `get_array_metadata` and `get_dft_array_metadata` that can be called for a `Volume` to obtain information useful for interpreting the arrays returned by `get_array` or `get_dft_array` for that volume.

More specifically, suppose `V` is a `Volume,` and suppose a call to `get_array` or `get_dft_array` for `V` returns an array `F` with dimensions `(N1,...,ND)`. 

Then `get_array_metadata(V)` returns an array `xyzw` of dimensions `(N1,...,ND,4)` defined as follows:

+ if `F[n1,...,nd]` is the (scalar) value of the field at a given point in the array, 

+ then `xyzw[n1,...,nd]` is an array of length 4 whose entries `x,y,z,w` are the cartesian coordinates `(x,y,z`) of the grid point and the corresponding integration/interpolation weight `w`. 

`get_dft_array_metadata` is identical to `get_array_metadata`, with the exception that empty dimensions are collapsed, as is done for DFT arrays but not time-domain field arrays.

Returning this entire array is overkill for the purposes of #502, which really only needs information on axis extents. However, for some purposes it may be useful to have the full array. Examples:

+ The columns of the `xyzw` array are in the right form for passage as grid axes to plotting routines like `plot_surface`, obviating calls to `meshgrid.`

```
from matplotlib.pyplot import plt
from mpl_toolkits.mplot3d import Axes3D

xyzw=sim.get_array_metadata(my_volume)
EzArray=sim.get_dft_array(my_volume,mp.Ez,0)

plt.figure()
ax=plt.axes(projection='3d')
ax.plot_surface(xyzw[:,:,0], xyzw[:,:,1], np.abs(EZ), rstride=1, cstride=1, cmap='RdBu')
```

meshplot

+ Having the `xyzw` array makes it easy to evaluate integrals over the `Volume` in which the integrand involves both the fields plus the value of some external function of spatial coordinates:

```

xyzw=sim.get_array_metadata(my_volume)
EzArray=sim.get_dft_array(my_volume,mp.Ez,0)

it=np.nditer(EzArray,flags=['multi_index'])
integral=0.0
while not it.finished:
    ii=it.multi_index
    (p,w)=(mp.Vector3(*tuple(xyzw[ii][0:3])), xyzw[ii][3])
    integral += w*EzArray[ii]*user_function(p)
return integral

```


stevengj:
My preference would be for `x,y,z,w = get_array_coords(myvolume)` that returns 3 1d arrays `x,y,z` and one 3d array `w`.  (If you want the 3d array version of `x,y,z` you would call `numpy.meshgrid`.)

stevengj:
That is, the C++ `get_array_slice_dimensions` function would take four array arguments (x,y,z,w) and would fill in the data.  Python would allocate these four arrays via numpy, and then return them as a tuple.

src/loop_in_chunks.cpp
```diff
@@ -520,4 +530,69 @@ void fields::loop_in_chunks(field_chunkloop chunkloop, void *chunkloop_data,
...
+  LOOP_OVER_DIRECTIONS(gv.dim, d)
+   array_size *= (1 + iabs(ie.in_direction(d)-is.in_direction(d))/2);
+  std::vector<double> xyzw(4*array_size,0.0);
+  int np=0;
+  LOOP_OVER_IVECS(gv, is, ie, idx)
+   { 
+     IVEC_LOOP_LOC(gv, loc);
+     double w = IVEC_LOOP_WEIGHT(s0, s1, e0, e1, dV0 + dV1 * loop_i2);
+
+     if (has_direction(gv.dim,X)) xyzw[4*np+0] = loc.x();
```
stevengj: Instead, if you allocate a 1d `x` array, do:
```c
IVEC_LOOP_ILOC(gv, iloc);

if (has_direction(gv.dim,X)) xyzw[(iloc.in_direction(X) - is.in_direction(X))/2] = loc.x();
```
(This overwrites the `x` array multiple times, but computational cost is irrelevant here.)

stevengj:
`get_array_slice_dimensions` should take a flag on whether to collapse empty dimensions: if `true`, then every empty `where` dimension should become 1 pixel thick.

HomerReid:
OK, I redid it per your specification.

```python
>>> dft_flux = sim.add_flux(...)
>>> (x,y,z,w)=sim.get_array_metadata(vol=dft_flux.where, collapse=False)
>>> np.shape(x)
(2,)
>>> np.shape(y)
(62,)
>>> np.shape(w)
(2,62)

>>> (x,y,z,w)=sim.get_dft_array_metadata(dft=dft_flux)  # 'dft_array' --> empty dimensions collapsed automatically
>>> np.shape(x)
(1,)
>>> np.shape(y)
(62,)
>>> np.shape(w)
(62,)
```

This PR also fixes an issue (in `meep.i:_get_dft_array`) that caused the python `get_dft_array` routine to crash when the DFT array in question was empty (for example, if the requested field component vanishes identically by symmetry). In such cases `get_dft_array` now returns an array of rank 0. For some purposes it might be better to convert this to an array of the expected size and rank with all zero entries.

ChristopherHogan:
Unfortunately the green here is a false positive. I'm seeing

```
../../../tests/array-slice-ll.cpp: In function ‘int main(int, char**)’:
../../../tests/array-slice-ll.cpp:234:59: error: no matching function for call to ‘meep::fields::get_array_slice_dimensions(meep::volume&, size_t [1], meep::direction [1])’
      rank=f.get_array_slice_dimensions(v1d, dims1D, dirs1D);
                                                           ^
In file included from ../../../tests/array-slice-ll.cpp:9:0:
../../../src/meep.hpp:1430:7: note: candidate: int meep::fields::get_array_slice_dimensions(const meep::volume&, size_t*, int*, bool, void*)
   int get_array_slice_dimensions(const volume &where, size_t dims[3],
       ^
../../../src/meep.hpp:1430:7: note:   no known conversion for argument 3 from ‘meep::direction [1]’ to ‘int*’
../../../tests/array-slice-ll.cpp:241:59: error: no matching function for call to ‘meep::fields::get_array_slice_dimensions(meep::volume&, size_t [2], meep::direction [2])’
      rank=f.get_array_slice_dimensions(v2d, dims2D, dirs2D);
                                                           ^
```

If you close and reopen the PR it should pick up #651 which fixes the Travis issue and will show failed builds correctly.

ChristopherHogan:
I'll work on rebasing this and fixing the SWIG issue with passing a `direction` array and open a new PR with these changes.

PR #554, opened 2018-10-10
Fix failing test

ChristopherHogan:
The `binary_grating.py` test fails occasionally. Should we expect better accuracy for this?
@stevengj @oskooi

### DIFF
--- a/src/sources.cpp
+++ b/src/sources.cpp
@@ -18,6 +18,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <math.h>
+#include <memory.h>
 #include <complex>
 
 #include "meep.hpp"
@@ -279,6 +280,7 @@ static void src_vol_chunkloop(fields_chunk *fc, int ichunk, component c,
   LOOP_OVER_DIRECTIONS(is.dim, d)
     npts *= (ie.in_direction(d) - is.in_direction(d)) / 2 + 1;
   ptrdiff_t *index_array = new ptrdiff_t[npts];
+  memset(index_array, 0, sizeof(ptrdiff_t) * npts);
   complex<double> *amps_array = new complex<double>[npts];
 
   complex<double> amp = data->amp * conj(shift_phase);
--- a/src/sources.cpp
+++ b/src/sources.cpp
@@ -18,7 +18,6 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <math.h>
-#include <memory.h>
 #include <complex>
 
 #include "meep.hpp"
@@ -280,7 +279,6 @@ static void src_vol_chunkloop(fields_chunk *fc, int ichunk, component c,
   LOOP_OVER_DIRECTIONS(is.dim, d)
     npts *= (ie.in_direction(d) - is.in_direction(d)) / 2 + 1;
   ptrdiff_t *index_array = new ptrdiff_t[npts];
-  memset(index_array, 0, sizeof(ptrdiff_t) * npts);
   complex<double> *amps_array = new complex<double>[npts];
 
   complex<double> amp = data->amp * conj(shift_phase);
--- a/.travis.yml
+++ b/.travis.yml
@@ -171,7 +171,7 @@ after_success:
   - if [[ "${RUN_COVERAGE}" = "1" ]]; then cd ${TRAVIS_BUILD_DIR} && coveralls; fi
 
 after_script:
-  - BUILD_PREFIX=${TRAVIS_BUILD_DIR}/build/meep-${MEEP_VERSION}/_build
+  - BUILD_PREFIX=${TRAVIS_BUILD_DIR}/build/meep-${MEEP_VERSION}/_build/sub
   - MEEP_LOG=${BUILD_PREFIX}/tests/test-suite.log
   - if [[ -e ${MEEP_LOG} ]]; then cat ${MEEP_LOG}; fi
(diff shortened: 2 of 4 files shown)


ChristopherHogan:
Valgrind reported an error in `binary_grating.py`:
```
==11932== Conditional jump or move depends on uninitialised value(s)
==11932==    at 0x852B9B9: meep::src_vol::operator==(meep::src_vol const&) const (meep_internals.hpp:65)
==11932==    by 0x8528611: meep::src_vol::add_to(meep::src_vol*) (sources.cpp:187)
==11932==    by 0x8529C6B: meep::src_vol_chunkloop(meep::fields_chunk*, int, meep::component, meep::ivec, meep::ivec, meep::vec, meep::vec, meep::vec, meep::vec, double, double, meep::ivec, std::complex<double>, meep::symmetry const&, int, void*) (sources.cpp:315)
==11932==    by 0x850B958: meep::fields::loop_in_chunks(void (*)(meep::fields_chunk*, int, meep::component, meep::ivec, meep::ivec, meep::vec, meep::vec, meep::vec, meep::vec, double, double, meep::ivec, std::complex<double>, meep::symmetry const&, int, void*), void*, meep::volume const&, meep::component, bool, bool) (loop_in_chunks.cpp:507)
==11932==    by 0x852AD55: meep::fields::add_volume_source(meep::component, meep::src_time const&, meep::volume const&, std::complex<double> (*)(meep::vec const&), std::complex<double>) (sources.cpp:449)
...
==11932==  Uninitialised value was created by a heap allocation
==11932==    at 0x4C2E87B: operator new[](unsigned long) (vg_replace_malloc.c:423)
==11932==    by 0x8528DE5: meep::src_vol_chunkloop(meep::fields_chunk*, int, meep::component, meep::ivec, meep::ivec, meep::vec, meep::vec, meep::vec, meep::vec, double, double, meep::ivec, std::complex<double>, meep::symmetry const&, int, void*) (sources.cpp:281)
==11932==    by 0x850B958: meep::fields::loop_in_chunks(void (*)(meep::fields_chunk*, int, meep::component, meep::ivec, meep::ivec, meep::vec, meep::vec, meep::vec, meep::vec, double, double, meep::ivec, std::complex<double>, meep::symmetry const&, int, void*), void*, meep::volume const&, meep::component, bool, bool) (loop_in_chunks.cpp:507)
==11932==    by 0x852AD55: meep::fields::add_volume_source(meep::component, meep::src_time const&, meep::volume const&, std::complex<double> (*)(meep::vec const&), std::complex<double>) (sources.cpp:449)
...
```
Initializing `index_array` to 0 silenced the error, but is that correct, or does this signify a bug in `LOOP_OVER_IVECS` in `sources.cpp:290`?

src/sources.cpp
```diff
@@ -279,6 +280,7 @@ static void src_vol_chunkloop(fields_chunk *fc, int ichunk, component c,
   LOOP_OVER_DIRECTIONS(is.dim, d)
     npts *= (ie.in_direction(d) - is.in_direction(d)) / 2 + 1;
   ptrdiff_t *index_array = new ptrdiff_t[npts];
+  memset(index_array, 0, sizeof(ptrdiff_t) * npts);
```
stevengj: This doesn't seem like the right fix.  All `idx_vol` points of `index_array` and `amps_array` are initialized.   The problem is in `src_vol::==`.

ChristopherHogan:
Fixing `src_vol::operator==` took care of the valgrind issue. However, I'm still seeing random failures of `python/tests/binary_grating.py`. This PR is green, but my own branch failed with 

```
FAIL: test_kdom (__main__.TestEigCoeffs)
  File "../../../python/tests/binary_grating.py", line 133, in test_kdom
    self.run_binary_grating_oblique(10.7)
  File "../../../python/tests/binary_grating.py", line 128, in run_binary_grating_oblique
    self.assertAlmostEqual(Rsum,Rflux,places=2)
AssertionError: 0.08403537987182948 != 0.07806410478645255 within 2 places
```
Is setting reducing expected accuracy to 1 decimal acceptable?

stevengj:
1 digit seems awfully low; it would be good to get to the bottom of this inconsistency.

stevengj:
Let' me merge this for now, since it is a clear improvement, and we can discuss further improvements in the test reliability elsewhere.

oskooi:
Two observations which might help debug this issue.

(1) `python/tests/binary_grating.py` is the *only* test to fail on my local machine when run as part of `make check` with the following error:
```
  File "./tests/binary_grating.py", line 133, in test_kdom
    self.run_binary_grating_oblique(10.7)
  File "./tests/binary_grating.py", line 128, in run_binary_grating_oblique
    self.assertAlmostEqual(Rsum,Rflux,places=2)
AssertionError: 0.08389643694611333 != 0.07806410478645255 within 2 places
```
However, when this test is run *outside* of `make check` it passes (and consistently so). This might suggest that a particular setting in the Makefile could be responsible for the test failure and that the assert statements involving 2 decimal places in the test itself are fine.

(2) It seems there are two functions with the name `test_kdom`: one in `python/tests/binary_grating.py` and the other in `python/tests/kdom.py`. The former needs to be renamed to something else (i.e., `test_binary_grating`). Two functions with the same name in the test suite name space may be causing unexpected issues during the `make check`.

stevengj:
Parallel failures should be fixed by #578.

PR #558, opened 2018-10-15
Fix memory leaks in Simulation.get_eigenmode and enable building without MPB

ChristopherHogan:
There were actually 3 leaks here. `Gk` and `kdom` are now returned directly as Python objects (`Gk` via `_get_eigenmode_Gk` and `kdom` is a numpy array passed as an output argument to `_get_eigenmode`), so Python gc will handle them properly. Also added a destructor that calls `destroy_eigenmode_data` for `meep::eigenmode_data`. 
@stevengj @oskooi

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -167,13 +167,8 @@ script:
     fi
   - >
     if [[ "${BUILD_WITHOUT_MPB}" == "1" ]]; then
-      ../configure
-        --enable-maintainer-mode
-        --prefix=$HOME/local
-        --with-libctl=$HOME/local/share/libctl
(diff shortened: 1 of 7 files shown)


stevengj:
We should make sure that Meep+Python compiles even if MPB is not installed.  This might involve adding `#ifdef HAVE_MPB` at appropriate places.

stevengj:
Would be nice to have a regression test for this.  Maybe just compile Meep with MPB disabled (`./configure ac_cv_header_mpb_h=no ....`) and check that it compiles.

PR #563, opened 2018-10-17
Added missing MPB (python) magnetic field density output:

centaure:
Added output of time-averaged magnetic field density (i.e. H-field)
as documented in [1].

[1] https://mpb.readthedocs.io/en/latest/Python_User_Interface/#bandoutput-functions


ChristopherHogan:
:+1:

PR #570, opened 2018-10-19
Add changes requested in issue 490

### LINKED ISSUE #490: Add field information to dft_flux from frequency domain solver
Is there already an easy way to add the field information calculated from the frequency domain solver to a `dft_flux` object (preferably in python)? That way, we can use the existing `get_eigenmode` routines by passing the modified `flux` object with the frequency domain solver.

ChristopherHogan:
Closes #490.
@stevengj @oskooi

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -1237,9 +1237,8 @@ class Simulation(object):
     def solve_cw(self, tol=1e-8, maxiters=10000, L=2):
         if self.fields is None:
             raise RuntimeError('Fields must be initialized before using solve_cw')
-        res = self.fields.solve_cw(tol, maxiters, L)
         self._evaluate_dft_objects()
-        return res
+        return self.fields.solve_cw(tol, maxiters, L)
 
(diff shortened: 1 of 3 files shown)


python/simulation.py
```diff
@@ -1237,7 +1237,9 @@ def modal_volume_in_box(self, box=None, center=None, size=None):
     def solve_cw(self, tol=1e-8, maxiters=10000, L=2):
         if self.fields is None:
             raise RuntimeError('Fields must be initialized before using solve_cw')
-        return self.fields.solve_cw(tol, maxiters, L)
+        res = self.fields.solve_cw(tol, maxiters, L)
+        self._evaluate_dft_objects()
```
stevengj: This line needs to go before `solve_cw`, I think, so that the `update_dfts()` call in `solve_cw` will take effect?

ChristopherHogan: Fixed. This is green.

PR #571, opened 2018-10-19
Add __rmul__ operator for scalar * Matrix and vec * Matrix

ChristopherHogan:
@stevengj @oskooi

### DIFF
--- a/python/geom.py
+++ b/python/geom.py
@@ -396,6 +396,14 @@ class Matrix(object):
         else:
             raise TypeError("No operation known for 'Matrix * {}'".format(type(m)))
 
+    def __rmul__(self, left_arg):
+        if type(left_arg) is Vector3:
+            return self.mv_mult(left_arg)
+        elif isinstance(left_arg, Number):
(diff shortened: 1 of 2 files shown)


python/geom.py
```diff
@@ -396,6 +396,14 @@ def __mul__(self, m):
         else:
             raise TypeError("No operation known for 'Matrix * {}'".format(type(m)))
 
+    def __rmul__(self, left_arg):
+        if type(left_arg) is Vector3:
```
stevengj: multiplying a vector on right by a matrix, i.e. `Vector3 * Matrix`, is not defined.    It seems like `__rmul__` should only work if `left_arg` is a Number or a Matrix.

ChristopherHogan: I removed the vector from the `__rmul__` case.  I think we only need to worry about  the Number case in `__rmul__` because it only gets called if `__mul__` returns `NotImplmentedError`, and `__mul__` should handle all cases where the left operand is a `Matrix`.

PR #577, opened 2018-10-22
Fix memory leak in get_dft_array

### LINKED ISSUE #575: Memory increasing steadily when calling get_dft_array from python
I am building an optimizer for designing a 3D material.  The simulation is run through Meep and I end up running many simulations in the whole process.  I found that at some point, after several iterations, the process would crash from running out of memory.  I was able to isolate the issue down to a small python program segment.  In the code segment below, the issue seems to be with calling sim.get_dft_array().  I have htop running in another window and whenever this program gets to that point, the memory usage jumps up a lot and stays there.  On the next iteration of the loop, the memory usage does not come down again, but instead jumps up by the same amount.  As I understand, there is some connection from python to the c++ backend through SWIG and I do not really understand how that all works.  I'm not sure what would be causing this behavior.

`
import math
import meep as mp
import time
import numpy as np
import sys

cell = mp.Vector3(5, 5, 5.9)
resolution = 40

geom = [mp.Block(size=mp.Vector3(2, 2, 2), center=mp.Vector3(0, 0, 0), material=mp.Medium(epsilon=1.53**3))]

                    boundary_layers=[],
                    geometry=geom,
                    sources=[],
                    resolution=resolution)

top_source_J = mp.Source(mp.GaussianSource(2, fwidth=2),
	component=mp.Ex, amplitude=1.0, center=mp.Vector3(0, 0, 2.5), size=mp.Vector3(5, 5, 0))

while(True):
	sim.reset_meep()
		                boundary_layers=[],
		                geometry=geom,
		                sources=[top_source_J],
		                resolution=resolution)

	dft_block_fields = sim.add_dft_fields([mp.Ex, mp.Ey, mp.Ez], 1, 3, 3,
		center=mp.Vector3(0, 0, 0), size=mp.Vector3(2, 2, 2))

	sim.run(until=1)
	for k in range(0, 100):
		sim.get_dft_array(dft_block_fields, mp.Ex, 0)

`

ChristopherHogan:
Fixes #575.
@stevengj @oskooi

### DIFF
--- a/python/meep.i
+++ b/python/meep.i
@@ -372,6 +372,7 @@ PyObject *_get_dft_array(meep::fields *f, dft_type dft, meep::component c, int n
     }
 
     PyObject *py_arr = PyArray_SimpleNewFromData(rank, arr_dims, NPY_CDOUBLE, dft_arr);
+    PyArray_ENABLEFLAGS((PyArrayObject*)py_arr, NPY_ARRAY_OWNDATA);
     delete[] arr_dims;
 
     return py_arr;
--- a/python/meep.i
+++ b/python/meep.i
@@ -375,7 +375,6 @@ PyObject *_get_dft_array(meep::fields *f, dft_type dft, meep::component c, int n
 
     PyObject *py_arr = PyArray_SimpleNew(rank, arr_dims, NPY_CDOUBLE);
     memcpy(PyArray_DATA((PyArrayObject*)py_arr), dft_arr, sizeof(std::complex<double>) * length);
-    PyArray_ENABLEFLAGS((PyArrayObject*)py_arr, NPY_ARRAY_OWNDATA);
     delete[] dft_arr;
     delete[] arr_dims;
 
--- a/python/meep.i
+++ b/python/meep.i
@@ -373,14 +373,10 @@ PyObject *_get_dft_array(meep::fields *f, dft_type dft, meep::component c, int n
         length *= dims[i];
     }
 
-    // Allocate our own version of dft_arr with malloc because
-    // numpy will eventually call free on it.
-    std::complex<double> *my_dft_arr = (std::complex<double> *)malloc(sizeof(std::complex<double>) * length);
-    memcpy(my_dft_arr, dft_arr, sizeof(std::complex<double>) * length);
-    delete[] dft_arr;
-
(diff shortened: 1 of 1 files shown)


python/meep.i
```diff
@@ -372,6 +372,7 @@ PyObject *_get_dft_array(meep::fields *f, dft_type dft, meep::component c, int n
     }
 
     PyObject *py_arr = PyArray_SimpleNewFromData(rank, arr_dims, NPY_CDOUBLE, dft_arr);
+    PyArray_ENABLEFLAGS((PyArrayObject*)py_arr, NPY_ARRAY_OWNDATA);
```
stevengj: This will eventually cause [NumPy to call `free`](https://github.com/numpy/numpy/blob/31f19645a13ee221f037ade57f13006127aa4c46/numpy/core/src/multiarray/alloc.c#L257-L274), so it is not safe to call with an array allocate with `new` like `get_dft_array.

ChristopherHogan: It would be nice to be able to get the array size and allocate a numpy array from Python like we do in `Simulation.get_array`, but I don't see a way to get the dft array size, so I'm copying the dft data to my own `malloc`ed array.

python/meep.i
```diff
@@ -366,12 +366,20 @@ PyObject *_get_dft_array(meep::fields *f, dft_type dft, meep::component c, int n
...
+    size_t length = 1;
     npy_intp *arr_dims = new npy_intp[rank];
     for (int i = 0; i < rank; ++i) {
         arr_dims[i] = dims[i];
+        length *= dims[i];
     }
 
-    PyObject *py_arr = PyArray_SimpleNewFromData(rank, arr_dims, NPY_CDOUBLE, dft_arr);
+    // Allocate our own version of dft_arr with malloc because
+    // numpy will eventually call free on it.
```
stevengj: Seems like it would be simpler to just do:
```c++
PyObject *py_arr = PyArray_SimpleNew(rank, arr_dims, NPY_CDOUBLE);
memcpy(PyArray_DATA(py_arr), dft_arr, sizeof(std::complex<double>) * length);
```
i.e. to let Python do the allocation.

ChristopherHogan: Fixed.

python/meep.i
stevengj: I don't think you need to set the `NPY_ARRAY_OWNDATA` flag if you are using `PyArray_SimpleNew`

PR #581, opened 2018-10-25
Expose more ldos data

ChristopherHogan:
A user on the mailing list wanted programmatic access to ldos data printed by the step function `mp.dft_ldos`. I added a public `DftLdos` function that returns a C++ `dft_ldos` that can then be

1. passed to `mp.dft_ldos`
2. passed to `mp.get_ldos_freqs`
3. queried for `dft_ldos::omega_min`

@stevengj @oskooi

### DIFF
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -1178,13 +1178,17 @@ As `load_force_data`, but negates the Fourier-transformed fields after they are 
 
 Meep can also calculate the LDOS (local density of states) spectrum, as described in [Tutorial/Local Density of States](Python_Tutorials/Local_Density_of_States.md). To do this, you simply pass the following step function to your `run` command:
 
-**`dft_ldos(fcen, df, nfreq)`**
+**`dft_ldos(fcen=None, df=None, nfreq=None, ldos=None)`**
 —
-Compute the power spectrum of the sources (usually a single point dipole source), normalized to correspond to the LDOS, in a frequency bandwidth `df` centered at `fcen`, at `nfreq` frequency points.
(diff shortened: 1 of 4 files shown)


doc/docs/Python_User_Interface.md
```diff
@@ -1178,13 +1178,17 @@ As `load_force_data`, but negates the Fourier-transformed fields after they are
 
 Meep can also calculate the LDOS (local density of states) spectrum, as described in [Tutorial/Local Density of States](Python_Tutorials/Local_Density_of_States.md). To do this, you simply pass the following step function to your `run` command:
 
-**`dft_ldos(fcen, df, nfreq)`**
+**`dft_ldos(fcen=None, df=None, nfreq=None, ldos=None)`**
 —
-Compute the power spectrum of the sources (usually a single point dipole source), normalized to correspond to the LDOS, in a frequency bandwidth `df` centered at `fcen`, at `nfreq` frequency points.
+Compute the power spectrum of the sources (usually a single point dipole source), normalized to correspond to the LDOS, in a frequency bandwidth `df` centered at `fcen`, at `nfreq` frequency points. One can also pass in an `ldos` created with `DftLdos` as `dft_ldos(ldos=my_ldos)`. This is useful for passing to `get_ldos_freqs` or accessing C++ properties like `ldos.omega_min`.
```
stevengj: Maybe it should be `ldos.freqs()` instead of `get_ldos_freqs(ldos)`?

stevengj:
Probably `omega` should be `freq` here and elsewhere in the API.  (Internally in C++, we call this "omega", but it is actually the frequency, not the angular frequency.)

ChristopherHogan:
For reference, here is the email that prompted this PR:
> Now I have the ldos_data but I can't read the associated frequencies. They get outputted in the console but when I try to get them from my ldos object, I get the Error:
> 
> AttributeError:
> 'Simulation' object has no attribute 'omega_min'
> or
> 'Source' object has no attribute 'omega_min'
> 
> What exactly is now my ldos object which the documentation is referring to?

doc/docs/Python_User_Interface.md
ChristopherHogan: Fixed.

PR #585, opened 2018-10-30
Rescale length units in GDSII files to comport with revisions to libGDSII

### LINKED ISSUE #555: Inconsistent length scale of imported coupler.gds in MEEP
In the [coupler.gds](https://github.com/stevengj/meep/blob/master/python/examples/coupler.gds) example, the height * width of the cell is 34400 um * 8000 um. But when running the MEEP [example python script](https://github.com/stevengj/meep/blob/master/python/examples/coupler.py), the size is taken as 34.4 um * 8 um (since we use a length scale of 1 um in the MEEP example). 

We should stick to the same length scale.

HomerReid:
Together with [the latest libGDSII master branch (commit c5cc55f)](https://github.com/HomerReid/libGDSII), fixes #555.


stevengj:
To do: the configure script should add some kind of version check for libGDSII so that people don't accidentally build the latest Meep with an old/incompatible version of libGDSII.

PR #591, opened 2018-11-05
Support PML and Absorbers in fragment_stats

ChristopherHogan:
This takes the list of `boundary_layers` and splits it into three lists of `meep::volume`s. One for boundary regions that don't overlap, one for where 2 boundary regions overlap, and one for where 3 overlap. These lists are passed to C++ to compute the number of PML pixels in each fragment, separated by degree of overlap. If the boundary layers are `Absorber`s, these pixel amounts are added to `num_nonzero_conductivity_pixels` instead (multiplied by degree of overlap).
@stevengj @oskooi


ChristopherHogan:
* Don't multiply Absorber pixels by overlap factor.
* Absorbers and PMLs are not mutually exclusive.

This is ready and green.

PR #593, opened 2018-11-06
Read epsilon input from numpy array

### LINKED ISSUE #584: Set geometry from numpy array
I think it would be convenient if one could pass an array to the simulation to set the geometry, similar to how it is possible to set the source distribution via `amp_data`. Perhaps `eps_data` as an argument to the simulation object. At the moment this is somewhat possible by creating a material array, saving this to HDF5 and loading the file via `epsilon_input_file`, but it would be great if this could be done directly.
This would also be a useful feature for the adjoint approach discussed before as it would allow for very fine control over the geometry.

ChristopherHogan:
Closes #584. Reuses the epsilon input file mechanism of `material_data` to allow setting the `default_material` to a numpy array. 
@stevengj @oskooi

### DIFF
--- a/python/typemap_utils.cpp
+++ b/python/typemap_utils.cpp
@@ -481,10 +481,11 @@ static int pymaterial_to_material(PyObject *po, material_type *mt) {
             PyErr_SetString(PyExc_TypeError, "Numpy array must be C-style contiguous.");
             return 0;
         }
-        md = (material_data *)malloc(sizeof(*md));
+        md = new material_data();
         md->which_subclass=material_data::MATERIAL_FILE;
         md->epsilon_dims[0] = md->epsilon_dims[1] = md->epsilon_dims[2] = 1;
-        md->epsilon_data = (double*)PyArray_DATA(pao);
+        md->epsilon_data = new realnum[PyArray_SIZE(pao)];
+        memcpy(md->epsilon_data, (realnum*)PyArray_DATA(pao), PyArray_SIZE(pao) * sizeof(realnum));
 
         for (int i = 0; i < PyArray_NDIM(pao); ++i) {
             md->epsilon_dims[i] = (size_t)PyArray_DIMS(pao)[i];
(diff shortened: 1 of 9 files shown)


python/typemap_utils.cpp
```diff
@@ -474,6 +474,28 @@ static int pymaterial_to_material(PyObject *po, material_type *mt) {
...
+        PyArrayObject *pao = (PyArrayObject*)po;
+
+        if (!PyArray_ISCARRAY(pao)) {
+            PyErr_SetString(PyExc_TypeError, "Numpy array must be C-style contiguous.");
+            return 0;
+        }
+        md = (material_data *)malloc(sizeof(*md));
+        md->which_subclass=material_data::MATERIAL_FILE;
+        md->epsilon_dims[0] = md->epsilon_dims[1] = md->epsilon_dims[2] = 1;
+        md->epsilon_data = (double*)PyArray_DATA(pao);
```
stevengj: Won’t this cause problems if the numpy array is garbage-collected before we are done with the material? Or can that never happen here?

ChristopherHogan: I think the array should be alive as long as the `Simulation` instance because they have to exist in the same scope. It could be garbage collected in a situation like this:

```py
def init():
    arr = np.array()  # eps data
    sim = mp.Simulation(default_material=arr)
    return sim

sim = init()
sim.run()
sim.reset_meep()
sim.default_material=mp.air  # arr is marked for collection
sim.run()
```
but in that case we're no longer using the `arr` material.

 I was worried that `libmeepgeom` may delete `epsilon_data` when numpy is supposed to own it, but it turns out that never happens. I think this PR is good to go, but I think there is a leak when `epsilon_data` contains the result of `h5file.read` in the `MATERIAL_FILE` case. I'll look into that separately.

stevengj:
Even if the typemap is safe in the particular case we are calling now, it creates a potential problem for any user code that calls a geometry function taking a material argument.

It seems like the safest route would be to:

1. Make a copy of the data in the typemap.

2. Modify `material_gc` to deallocate the data — it seems like we need to do this anyway, since we may have a memory leak for `make_file_material` ?

ChristopherHogan:
* Make a copy of the numpy array.
* Delete `epsilon_data` in `freearg` typemaps (`material_gc` is specifically for cleaning up susceptibilities allocated in user material functions).
* Fix memory leak in `make_file_material` in libmeepgeom and in `scheme/structure.cpp`.
* Consistently use `new` over `malloc` in typemaps.

Issue #595, opened 2018-11-06
more efficient eigenmode solver for parallel machines


stevengj:
Because of #594, we can now skip most of the eigenmode calculation on all processes ≠ 0.  Saves a bit of CPU energy and might be easier to diagnose if there is a problem with the eigenmode solver.

More seriously, we should also reconsider how we get the ε data from Meep to MPB, since it looks like it currently does a `broadcast` for every grid point that is evaluated, which is potentially quite slow.

Issue #568, opened 2018-10-18
fix binary_grating test


stevengj:
This is temporarily disabled (f1d972dfd39cf91e13ba92191fc8d866fe6bf899) because it appears to be subject to occasional random failures in MPI mode.  Not clear what is causing it.

oskooi:
After some investigation, it seems there is a likely bug somewhere in `fields::get_mode_flux_overlap` (and also `fields::get_mode_mode_overlap`) for the case involving MPI *and* a non-zero `k_point`. The mode coefficients are equivalent for the serial and parallel case when `k_point` is `0`. This is the first part of the test.

For the second part of the test where `k_point` is non-zero, `fields::get_eigenmode` produces identical results (`kdom` values as well as the field profiles) for the serial and parallel case. This demonstrates that the MPB-related functions are working correctly. The only difference in the results were the values of the mode coefficients.

Perhaps @HomerReid can provide further insights?

oskooi:
Unfortunately, it seems that #578 does not resolve this issue. As a further demonstration, the same erratic behavior is observed after replacing the random number generator `rand_r` in `src/mpb.cpp:342`:

```cpp
ASSIGN_SCALAR(H.data[i], rand_r(&seed) * 1.0/RAND_MAX, rand_r(&seed) * 1.0/RAND_MAX);
```
with just hard-coded random values:

```cpp
ASSIGN_SCALAR(H.data[i], 0.548724, 0.837532);
```
This would suggest that the MPB phase is in fact not the source of the bug. It doesn't seem that random numbers are used anywhere else in `src/mpb.cpp` or functions related to `get_eigenmode_coefficients`. 

Another possibility: what about the MPI calls to communicate MPB fields among different Meep chunks when MPB has been built serially?

stevengj:
In case there is still something that is causing the MPB solution to be nondeterministic, i.e. different on different processes (when the eigenmode region spans multiple processes), a simple way to synchronize them should be to add:
```c
broadcast(0, (double*) H.data,  2 * H.n * H.p);
```
in mpb.cpp before line 435 (before the call to `maxwell_compute_h_from_H`).

HomerReid:
On line 853 of `src/dft.cpp`, try adding this:

```
debug_printf("%e %e \n",real(overlap), imag(overlap));
```

[...]

PR #596, opened 2018-11-06
fix sync of eigenmode calc when no mode is found

stevengj:
This is an improved version of #594 that skips most of the eigenmode calculation on non-master processes (#595), which fixes a potential deadlock if the processes didn't agree in the case where no mode is found.

Also, it fixes a memory leak when no mode is found (returning `NULL`).   It also gets rid of the `rand_r` call from #568, which is no longer needed since `H` is only initialized on the master process.


ChristopherHogan:
`make check` uses 2 processes when Meep is built `--with-mpi`. It looks like `binary_grating.py` is the hanging test, so you could try running it manually with `mpirun -n 2 python path/to/python/tests/binary_grating.py`.

stevengj:
Looks like it is fixed now.  I'll merge once the latest CI tests are green.

Issue #560, opened 2018-10-16
material rotation


stevengj:
It would be good to add a couple of functions to rotate anisotropic materials, so that you can easily use such materials from the material library while preserving the freedom to orient them in any desired direction.  There are two cases:
```py
rotate_uniaxial_material(material, new_extraordinary_axis) # only need to supply one vector
rotate_material(material, rotation_matrix) # need a 3x3 rotation matrix or similar
```

stevengj:
ε, μ, and σ (for Lorentzian susceptibilies) are all described by a symmetric 3x3 matrix M in Meep (given in terms of the diag and offdiag components).   To rotate such a matrix (a "rank 2 tensor), if you are rotating all the coordinates x to Rx where R is a 3x3 rotation matrix, you need to rotate M to RMRᵀ where Rᵀ is the transpose.

PR #603, opened 2018-11-12
Add material rotation

ChristopherHogan:
Addresses half of #560. Still need a `rotate_uniaxial`. I'll add documentation once the interface is approved.
@stevengj @oskooi

### DIFF
--- a/python/tests/geom.py
+++ b/python/tests/geom.py
@@ -349,7 +349,7 @@ class TestMedium(unittest.TestCase):
         sim.add_flux(15, 6, 2, fregion)
         check_warnings(sim)
 
-    def test_rotate(self):
+    def test_transform(self):
 
         e_sus = [mp.LorentzianSusceptibility(sigma_diag=mp.Vector3(1, 2, 3),
                                              sigma_offdiag=mp.Vector3(12, 13, 14)),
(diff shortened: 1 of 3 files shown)


python/tests/geom.py
```diff
@@ -349,6 +349,29 @@ def check_warnings(sim, should_warn=True):
...
+
+        e_sus = [mp.LorentzianSusceptibility(sigma_diag=mp.Vector3(1, 2, 3),
+                                             sigma_offdiag=mp.Vector3(12, 13, 14)),
+                 mp.DrudeSusceptibility(sigma_diag=mp.Vector3(1, 2, 3),
+                                        sigma_offdiag=mp.Vector3(12, 13, 14))]
+
+        mat = mp.Medium(epsilon_diag=mp.Vector3(1, 2, 3), epsilon_offdiag=mp.Vector3(12, 13, 14),
+                        E_susceptibilities=e_sus)
+
+        rot_matrix = mp.Matrix(mp.Vector3(1, 0, 0), mp.Vector3(0, 0, 1), mp.Vector3(0, -1, 0))
```
oskooi: It would be better to use a [rotation matrix](https://en.wikipedia.org/wiki/Rotation_matrix#Basic_rotations) with off-diagonal components (e.g., clockwise rotation about the *z* axis):

```py
import math

rot_angle = math.radians(23.9)
rot_matrix = mp.Matrix(mp.Vector3(math.cos(rot_angle),math.sin(rot_angle),0), mp.Vector3(-math.sin(rot_angle),math.cos(rot_angle),0), mp.Vector3(0,0,1))
```

oskooi: Yours is fine. (It would have been good to use a rotation matrix with non-zero elements everywhere but this will suffice.)

ChristopherHogan: I see what you're saying. I'll add such a test.

stevengj:
It might be nice to generalize this to include [arbitrary coordinate transforms](http://math.mit.edu/~stevenj/18.369/coordinate-transform.pdf).  In particular, this means that you divide RMRᵀ by det(R).

stevengj:
Maybe call it `transform` instead of `rotate` in that case.

stevengj:
Thanks.  I updated the docs.

PR #607, opened 2018-11-13
Add verbose flag to get_eigenmode_coefficients to pass to get_eigenmode

ChristopherHogan:
@stevengj @oskooi


ChristopherHogan:
* Added default params to docs.
This is green now.

Issue #536, opened 2018-10-02
better load balancing


stevengj:
Given the fragment statistics and a fit (linear regression?) from these to wall-clock time (for 1-processor runs), it would be good to use this for better load balancing: try to divide up the computational cell to make the estimated cost roughly equal on each process.  Also note that you want to divide a cell along the "long" axis first, and in general get roughly cubical chunks, to minimize the surface area (communication) cost, all other things being equal.

The current routine that does this is `structure::choose_chunkdivision` in `structure.cpp`, which was written by Mihai Ibanescu many years ago IIRC.  It actually tries to do some kind of load balancing with the `split_by_effort` method, but was hampered by the fact that we didn't have a good way to estimate the computational effort.

Ideally, we could just plug a better "effort" estimate in and it would just work, but I suspect that the whole algorithm may need to be re-thought.

soamaven:
As a data point, this pull cut my simulations time in half. Thank you @ChristopherHogan!

PR #610, opened 2018-11-16
Make fragment_stats available in choose_chunkdivision

ChristopherHogan:
This is in preparation for better load balancing (#536).
@stevengj @oskooi


ChristopherHogan:
Closing since this will need more restructuring once we actually use `fragment_stats` in `structure.cpp`.

PR #625, opened 2018-11-28
Add high level phase_in_material and set_boundary

ChristopherHogan:
I added some basic documentation that probably needs fleshed out.
@stevengj @oskooi

### DIFF
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -899,9 +899,9 @@ Meep supports a large number of functions to perform computations on the fields.
 —
 Sets the condition of the boundary on the specified side in the specified direction. See the [Constants (Enumerated Types)](#Constants (Enumerated Types)) section for valid `side`, `direction`, and `boundary_condition` values.
 
-**`phase_in_material(structure, time)`**
+**`phase_in_material(newstructure, phasetime)`**
 —
-`structure` should be the `structure` field of another `Simulation` object.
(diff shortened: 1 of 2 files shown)


stevengj:
`phase_in_material` is an ancient and little-used function, so it would be good to verify that it still works.

A simple test would be to initialize the structure to a dielectric cylinder, and then tell it to phase into another structure with (e.g.) a shifted half-overlapping cylinder over some time `T`.  Then run the simulation, outputting epsilon at each time step.  You should see the first cylinder gradually disappear while the second cylinder appears (with ε constant in the regions where they overlap).

oskooi:
The `phase_in_material` routine still works as demonstrated by the following script and accompanying image.

```py
import meep as mp

cell_size = mp.Vector3(6,6,0)

geometry1 = [mp.Cylinder(center=mp.Vector3(),radius=1.0,material=mp.Medium(index=3.5))]

sim1 = mp.Simulation(cell_size=cell_size,
                    geometry=geometry1,
                    resolution=20)

sim1.init_sim()

geometry2 = [mp.Cylinder(center=mp.Vector3(1,1),radius=1.0,material=mp.Medium(index=3.5))]

sim2 = mp.Simulation(cell_size=cell_size,
                    geometry=geometry2,
                    resolution=20)

sim2.init_sim()

sim1.fields.phase_in_material(sim2.structure,10.0)

sim1.run(mp.at_beginning(mp.output_epsilon),
         mp.at_every(0.5,mp.output_epsilon), until=10)
```
phase_in_materials

PR #639, opened 2018-12-19
add phase map demonstration to binary grating tutorial example

oskooi:


doc/docs/Python_Tutorials/Mode_Decomposition.md
```diff
@@ -306,6 +306,138 @@ In the limit where the grating periodicity is much larger than the wavelength an
...
 
+We can also use the complex mode coefficients to compute the phase of the diffraction orders. This can be used to generate a phase map of the binary grating as a function of its geometric parameters. Phase maps are important for the design of subwavelength phase shifters such as those used in metalenses. In this demonstration, which is adapted from the previous example, we compute the transmittance spectra and phase map of the zeroth diffraction order (at 0°) for an E<sub>z</sub>-polarized planewave pulse spanning wavelengths of 0.4 to 0.6 μm which is normally incident on a binary grating with a periodicity of 0.35 μm and height of 0.6 μm. The duty cycle of the grating is varied from 0.1 to 0.9 in separate runs.
+
```
stevengj: paragraph to add:

Note that it is generally only the *relative* phase (the phase difference) between different structures.  The overall mode coefficient α is multiplied by a complex number given by the source amplitude, as well as an arbitrary (but deterministic) phase choice by our mode solver (MPB)—but as long as you keep the source fixed as you vary the parameters of the structure, the relative phases are meaningful.  More generally, for metasurface design it is preferable to use the overall complex amplitude α (both phase and amplitude), as we describe in [this 2018 paper](https://www.osapublishing.org/oe/abstract.cfm?uri=oe-26-26-33732): you can use α to optimize for a desired wavefront, as in equation (3) in section 3.1, or more generally for some arbitrary function of the far field as in section 3.2.

stevengj: added.

PR #640, opened 2018-12-19
Pass is_integrated flag to SourceTime constructor

ChristopherHogan:
@stevengj @oskooi


stevengj:
(I think this flag mainly matters for custom sources.  For the built-in source types like gaussian or cw, I think they are defined in such a way as to make the integrated and non-integrated options equivalent up to roundoff errors.   Would be good to check this.)

PR #642, opened 2018-12-19
Reuse fields if able

ChristopherHogan:
Pymeep was reallocating the fields on every iteration of `find_k`, which significantly affected performance compared to Scheme. This copies the MPB code to reuse the fields if able, and clean up properly otherwise.
@stevengj @oskooi


ChristopherHogan:
This needs more testing.

PR #644, opened 2018-12-21
Fix performance issues in ModeSolver.find_k

ChristopherHogan:
This fixes two issues that made python's `mpb.ModeSolver.find_k` about 7x slower than the corresponding Scheme `find-k`.

1. Scheme reuses fields between runs when it detects that a suitable `maxwell_data` is already available. This PR adds the same functionality for Python. ~~However, only the `default_material` and `geometry` can be modified between runs. Scheme uses global variables for modifiable parameters, but in Python we have to pass the information from the `ModeSolver` Python object to the `mode_solver` C++ instance.~~

2. For readability I had split an `if` clause into multiple lines in `python/geom.py:find_root_deriv`. However, in Scheme the final condition (which I called `cond`) was only evaluated if the first 2 `and` conditions were true, whereas in Python they were evaluated every time, leading to extra unnecessary work.

Python's find_k is now about 20% faster than Scheme's (although Python doesn't output epsilon by default like Scheme does).

I'll add a documentation PR to MPB explaining how to get the performance gains of reusing fields.

@stevengj @oskooi

### DIFF
--- a/libpympb/pympb.cpp
+++ b/libpympb/pympb.cpp
@@ -214,6 +214,7 @@ mode_solver::mode_solver(int num_bands,
   force_mu(force_mu),
   use_simple_preconditioner(use_simple_preconditioner),
   grid_size(grid_size),
+  nwork_alloc(0),
   eigensolver_nwork(eigensolver_nwork),
   eigensolver_block_size(eigensolver_block_size),
   last_parity(-2),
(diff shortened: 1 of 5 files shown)


ChristopherHogan:
Added property accessors that allow changes to the Python `ModeSolver` to propagate to the C++ `mode_solver`.

PR #646, opened 2018-12-31
Source slice

HomerReid:
The pure-text format of problem definitions in meep is very convenient for scripting, but without visual confirmation of the geometry you specified it can be easy to make inadvertent errors in e.g. the placement and size of geometrical entities. Visualization of the material geometry (e.g. `output_epsilon`) has been supported for a long time, but I wasn't aware of any similar functionality for *source* distributions. In addition to offering a useful sanity check on text-format input files, the ability to visualize the locations and amplitudes of sources may prove useful for analyzing the more complicated source distributions that arise in applications such as [adjoint-based sensitivity analysis](https://github.com/stevengj/meep/pull/600).

This PR extends the existing `array_slice` protocol, used thus far for field-component arrays, to the acquisition of *source*-component arrays. It adds a new routine named `get_source_slice` to the python and C++ interfaces, which is handled internally by adding a new branch to the existing `do_get_array_slice` master routine in `array_slice.cpp.` The C++ and python calling conventions are similar to those for `get_array_slice` / `get_array`: the user inputs a `Volume` and gets back a multidimensional array describing sources at the grid points it encompasses. The specific significance of the entries in the array depend on the input parameter `source_slice_type`, with 4 possibilities currently supported:

|`source_slice_type`| Significance of entries in `source_slice` array               |
|--------------------------|-------------------------------------------------------------------------|
| `Re Ex`                  | Real part of *x*-directed electric current amplitude       | 
| `Im Hz`                   | Imag part of *y*-directed magnetic current amplitude    | 
| `norm`                    | Sum of squared magnitudes of all source components | 
| `indicator`              | Binary indicator: 1 or 0 for points with/without sources |

As anticipated in the comments to #635, one use for `source_slice` functionality is to verify the implementation of sources defined by `geometric_objects.`. In `python/examples/object_source.py` I have put a somewhat contrived example code that illustrates both`geometric_object_sources` and `source_slices.` The code simply creates a geometry with three distinct source distributions--one conventional source defined by a `Volume` and two sources defined by `geometric_objects` (one `Cylinder` and one `Prism`)--and captures various types of `source_slice` to visualize aspects of the source distributions, which involve spatially-varying complex amplitudes in addition to the positional restrictions imposed by the defining `geometric_object`s.
source_slices


stevengj:
Looks good modulo the three comments above.

PR #649, opened 2019-01-02
add documentation for is_integrated to Source and other fixes

### LINKED ISSUE #645: hidden feature to initialize fields using Bessel functions
There is a hidden feature [`fields::initialize_field`](https://github.com/stevengj/meep/blob/master/src/initialize.cpp#L126-L139) for initializing fields using [Bessel functions](https://en.wikipedia.org/wiki/Bessel_function) via the [GNU Scientific Library](https://www.gnu.org/software/gsl/) (GSL). This would be useful for launching guided modes in [optical fibers](https://en.wikipedia.org/wiki/Optical_fiber). However, it is missing a Python SWIG wrapper as well as documentation in the manual (and possibly a tutorial example) even though it is available by default in the [C++ interface](https://meep.readthedocs.io/en/latest/Build_From_Source/#building-from-source).

oskooi:
~~Closes #645.~~


src/initialize.cpp
```diff
@@ -1,152 +0,0 @@
...
-    LOOP_OVER_VOL(gv, c, i) {
-      IVEC_LOOP_LOC(gv, here);
-      complex<double> val = func(here);
-      f[c][0][i] += real(val);
-      if (!is_real) f[c][1][i] += imag(val);
-    }
-  }
-}
-
-} // namespace meep
```
ChristopherHogan: You probably also need to remove `initialize.cpp` from `libmeep_la_SOURCES` in `src/Makefile.am`. Travis is working now so committing that change should run the tests.

ChristopherHogan:
Due to #650, Travis is flagging this as green even though the test suite failed with

```
../../../tests/cylindrical.cpp: In function ‘int test_pattern(double (*)(const meep::vec&), int, const char*)’:
../../../tests/cylindrical.cpp:290:8: error: ‘class meep::fields’ has no member named ‘initialize_field’
     f1.initialize_field(Hp, checkers);
        ^
../../../tests/cylindrical.cpp:291:7: error: ‘class meep::fields’ has no member named ‘initialize_field’
     f.initialize_field(Hp, checkers);
       ^
Makefile:944: recipe for target 'cylindrical.o' failed
```
I'll push a fix for Travis to fail builds when any exit code it nonzero.

PR #655, opened 2019-01-08
Array metadata replacement

ChristopherHogan:
This replaces and closes #552. I made the following changes:
* Rebased off latest master to fix several conflicts
* Allow `fields::get_array_slice_dimensions` to take a `direction[3]` instead of `int[3]` by writing a wrapper for SWIG called `_get_array_slice_dimensions`.
* Fixed failing `h5test` by passing `is`, `ie`, and `wherec` by reference to `compute_boundary_weights`.

The only thing I'm unsure of is why we subtract the `X` direction on line 391 in `array_slice.cpp`:
```c++
for(int r=0; r<rank; r++)
      dirs[r]=(meep::direction)(data->ds[r] - X);
```
Why is that necessary? Was it just for casting to an `int` when `dirs` was `int[3]`?

@stevengj @oskooi @HomerReid


stevengj:
Yes, it seems like `dirs[r]=data->ds[r]` should be fine.

HomerReid:
> Why is that necessary? Was it just for casting to an int when dirs was int[3]?

Yes, this was part of an attempt to avoid SWIG/python problems associated with arrays of `directions` that I wasn't able to solve, and now that you've solved them my workaround is obsolete. Thanks!

PR #656, opened 2019-01-08
Add high level initialize_field

### LINKED ISSUE #654: Python wrapper for initialize_field
It would be nice for the routine [`fields::initialize_field`](https://github.com/NanoComp/meep/blob/master/src/initialize.cpp#L126-L139), which is used by [python/tests/multilevel_atom.py](https://github.com/NanoComp/meep/blob/master/python/tests/multilevel_atom.py#L65), to have a high-level Python wrapper in order to hide the code internals.

We can then add documentation for this feature which is currently missing from the [manual](https://meep.readthedocs.io/en/latest/Python_User_Interface/).

ChristopherHogan:
Closes #654.
@stevengj @oskooi

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -1011,6 +1011,11 @@ class Simulation(object):
         warnings.warn('init_fields is deprecated. Please use init_sim instead', DeprecationWarning)
         self.init_sim()
 
+    def initialize_field(self, cmpnt, amp_func):
+        if self.fields is None:
+            self.init_sim()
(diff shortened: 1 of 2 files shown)


stevengj:
That was easy!  Also needs a documentation item.

ChristopherHogan:
I think @oskooi was going to add the documentation in a different PR.

stevengj:
Okay.

PR #657, opened 2019-01-08
Setup clang-format code beautification

stevengj:
The best tool I can find for automatic beautification (indenting, brace placement, etcetera) of C++ code is [`clang-format`](https://clang.llvm.org/docs/ClangFormat.html), which seems to be the most modern and well-maintained open-source tool for this right now

This PR doesn't actually run `clang-format` yet (which will cause conflicts on every other PR), but it prepares the way, via:

* A `.clang-format` file in the top-level directory sets the default format when you run `clang-format`. 
 The `clang-format` defaults (LLVM style, I think) seem pretty reasonable, with a couple of tweaks. 

* For it to format our `LOOP_IN_FOO` and `FOR_FOO` looping macros correctly, the loop bodies must be enclosed in braces.  That is the cause of most of the changes in this PR.

* We don't want to run `clang-format` on the SWIG-generated files (mainly because it is very slow), so I renamed them to `*.cxx` to easily distinguish them from hand-written code

You can just run `find . -name '*.*pp' | xargs clang-format -i` in the source tree to automatically fix all of the code spacing. 

Once we actually run `clang-format` and check into master, any existing PR will get lots of conflicts.  You might be able to do `git rebase -Xignore-all-space master` to mostly fix this.  Alternatively, you could run `clang-format` on your files and rebase with the reformatted files.  Could be a little bit of a pain for large PRs.

cc @ChristopherHogan

### DIFF
--- a/src/boundaries.cpp
+++ b/src/boundaries.cpp
@@ -268,6 +268,8 @@ void fields::find_metals() {
 		if (on_metal_boundary(here))
 		  chunks[i]->num_zeroes[ft]++;
 	      }
+      }
+        }
         typedef realnum *realnum_ptr;
         chunks[i]->zeroes[ft] = new realnum_ptr[chunks[i]->num_zeroes[ft]];
         size_t num = 0;
@@ -281,8 +283,6 @@ void fields::find_metals() {
     	      }
           }
         }
-      }
-    }
       }
     }
 }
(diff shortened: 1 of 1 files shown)


stevengj:
See also this article on rebasing branches after running `clang-format`:  https://engineering.mongodb.com/post/succeeding-with-clangformat-part-3-persisting-the-change

stevengj:
Looks like I misplaced a brace in boundaries.cpp.

PR #662, opened 2019-01-10
beautify code with clang-format

stevengj:
Following up on #657, this PR makes no changes except to run `clang-format -i` on all of the C++ code, giving it a uniform formatting.  (Previously, the code was a mishmash of styles due to different programmers and editor settings over the years, along with mangled indentation from tab/spaces conversions.)

Once this is merge, existing PR's and other branches will have zillions of conflicts, but if you follow a very careful procedure then this can be dealt with automatically:

1. First rebase on top of the *previous* commit to this one and fix any conflicts.

2. Then run `clang-format -i` on your code and commit it.

3. Then merge with this commit, ignoring conflicts with `git rebase -s`.

Summary:

## Instructions for rebasing other branches after this PR:
```sh
# update repo
git checkout master && git pull origin master # update master
rm -f python/*-python.cpp python/.deps/*-python.Plo # rm old swig files

# switch to your branch
git checkout MYBRANCH

# rebase onto PREVIOUS commit
git rebase 225014eb831dc710d50442463313c110562b9784^

# reformat and rebase to THIS PR's commit
find . -name '*.*pp' | xargs clang-format -i
git commit -a -m "reformat"
git rebase -s ours 225014eb831dc710d50442463313c110562b9784

# finish rebase
git rebase master
```

PR #663, opened 2019-01-10
Split chunks based on costs obtained from fragment_stats

### LINKED ISSUE #536: better load balancing

ChristopherHogan:
Closes #536. Adds `grid_volume::split_by_cost` which attempts to divide the cell into chunks that take a similar amount of work. The work estimate is produced by `fragment_stats::cost`, which was obtained from running linear regression on training data from random simulations. Scheme will fall back to the old `grid_volume::split_by_effort` method. Additionally, there is a `Simulation` constructor parameter for Python, `split_chunks_evenly`, that will force the use of `split_by_effort`.  Here are some benchmark results from the following simulation (a tall skinny cell with an expensive flux plane at the top):

```python
import argparse
import time
import meep as mp

def main(args):
    sx = 2
    sy = 30
    sz = 1
    fcen = 0.15
    df = 0.1

    src = mp.Source(mp.GaussianSource(fcen, fwidth=df), mp.Ey, mp.Vector3())

    sim = mp.Simulation(cell_size=mp.Vector3(sx, sy, sz),
                        sources=[src],
                        resolution=10,
                        split_chunks_evenly=args.split_evenly)

    flux = sim.add_flux(fcen, df, 500, mp.FluxRegion(center=mp.Vector3(0, 12), size=mp.Vector3(2, 1)))
    start = time.time()
    sim.run(until=200)
    print("bench:{},{},{}".format(args.split_evenly, mp.count_processors(), time.time() - start))

if __name__ == '__main__':
    parser.add_argument('-e', '--split-evenly', action='store_true', default=False,
                                        help='Use the old split_by_effort method to create chunks')
    args = parser.parse_args()
    main(args)
```

split_by_cost_results

The `meep::structure` is now created in C++ (`create_structure_and_set_materials`) instead of Python. Using the geometry in `structure::choose_chunkdivision` would require making copies, or passing it around, both of which would require more significant changes. `create_structure_and_set_materials` allows the Python geometry to be converted to C++ once, then deleted when it returns.

### Remaining issues
1. `split_by_cost` doesn't work with cylindrical coordinates yet so it falls back to `split_by_effort`
2. `split_by_cost` breaks the `structure::dump` and `structure::load` features. When you run the initial simulation and dump the structure file, the chunks are split optimally based on the geometry, but when you attempt to load the file, there is no geometry passed in so the chunks are split evenly, resulting in chunk size mismatches. My temporary solution to this is to pass `split_chunks_evenly=True` to a `Simulation` when you know you want to dump and reuse a structure file. I guess `structure::load` needs to recreate the chunks with the correct sizes.

### DIFF
--- a/libpympb/pympb.cpp
+++ b/libpympb/pympb.cpp
@@ -2287,7 +2287,7 @@ double mode_solver::compute_energy_in_objects(geometric_object_list objects) {
     return 0.0;
   }
 
-  geom_fix_objects0(objects);
+  geom_fix_object_list(objects);
 
   int n1 = mdata->nx;
   int n2 = mdata->ny;
--- a/python/meep.i
+++ b/python/meep.i
@@ -787,6 +787,9 @@ meep::volume_list *make_volume_list(const meep::volume &v, int c,
     delete $1;
 }
 
+// For some reason SWIG needs the namespaced version too
+%apply material_type { meep_geom::material_type };
+
 // Typemap suite for get_array_metadata
 
 %typecheck(SWIG_TYPECHECK_POINTER, fragment="NumPy_Fragments") double* xtics {
@@ -1110,6 +1113,9 @@ meep::volume_list *make_volume_list(const meep::volume &v, int c,
         delete[] $1.items;
     }
 }
+
+// For some reason SWIG needs the namespaced version too
+%apply material_type_list { meep_geom::material_type_list };
 
 // Typemap suite for custom_src_time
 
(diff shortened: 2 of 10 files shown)


stevengj:

stevengj:
Rebased.

PR #676, opened 2019-01-17
compute Poynting flux of far fields in dft_near2far

oskooi:
Replaces and closes #27.

A test of this new feature has been made a part of the revised `python/tests/antenna-radiation.py` which computes the Poynting flux of an E<sub>z</sub>-polarized dipole source from the fields in three different arrangements and verifies that they are equivalent.

We still need to add a high-level Python interface as well as documentation.

### DIFF
--- a/src/near2far.cpp
+++ b/src/near2far.cpp
@@ -342,6 +342,10 @@ double *dft_near2far::flux(direction df, const volume &where, double resolution)
     direction dirs[3] = {X,Y,Z};
     int rank = 0, N = 1;
     double vol = 1;
+
+    if (coordinate_mismatch(where.dim, df) || where.dim == Dcyl)
+      abort("cannot get flux for near2far: co-ordinate mismatch");
+
     LOOP_OVER_DIRECTIONS(where.dim, d) {
       dims[rank] = int(floor(where.in_direction(d) * resolution));
       if (dims[rank] <= 1) {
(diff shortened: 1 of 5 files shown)


src/near2far.cpp
```diff
@@ -334,6 +335,81 @@ void dft_near2far::save_farfields(const char *fname, const char *prefix, const v
...
+      x.set_direction(dirs[0], where.in_direction_min(dirs[0]) + i0*dx[0]);
+      for (int i1 = 0; i1 < dims[1]; ++i1) {
+	x.set_direction(dirs[1], where.in_direction_min(dirs[1]) + i1*dx[1]);
+	for (int i2 = 0; i2 < dims[2]; ++i2) {
+	  x.set_direction(dirs[2], where.in_direction_min(dirs[2]) + i2*dx[2]);
+	  farfield_lowlevel(EH1, x);
+	  int idx = (i0 * dims[1] + i1) * dims[2] + i2;
+	  for (int i = 0; i < Nfreq; ++i)
+	    for (int k = 0; k < 6; ++k)
+	      EH_[(k * N + idx) + (6 * N * i)] = EH1[i * 6 + k];
```
stevengj: Looks like you can get rid of the `EH_` array completely.  Instead, you can just merge this loop with the subsequent one, and accumulate `F` directly from the `EH1` array.

src/near2far.cpp
```diff
@@ -334,6 +335,81 @@ void dft_near2far::save_farfields(const char *fname, const char *prefix, const v
...
+    direction dirs[3] = {X,Y,Z};
+    int rank = 0, N = 1;
+    LOOP_OVER_DIRECTIONS(where.dim, d) {
+      dims[rank] = int(floor(where.in_direction(d) * resolution));
+      if (dims[rank] <= 1) {
+	dims[rank] = 1;
+	dx[rank] = 0;
+      }
+      else
+	dx[rank] = where.in_direction(d) / (dims[rank] - 1);
```
stevengj: You should accumulate the product of the nonzero `dx` elements and multiply this by the final flux at the end in order to get the integral.

src/near2far.cpp
```diff
@@ -334,6 +335,81 @@ void dft_near2far::save_farfields(const char *fname, const char *prefix, const v
...
+	  for (int i = 0; i < Nfreq; ++i)
+	    for (int k = 0; k < 6; ++k)
+	      EH_[(k * N + idx) + (6 * N * i)] = EH1[i * 6 + k];
+	}
+      }
+    }
+
+    delete[] EH1;
+    sum_to_master(EH_, EH, 6*N*Nfreq);
+    delete[] EH_;
```
stevengj: (This is all gone once you eliminate `EH` and `EH_`, since you will accumulate `F` directly from `EH1`.)

PR #681, opened 2019-01-23
Use fragment_stats to split chunks by cost

### LINKED ISSUE #536: better load balancing

ChristopherHogan:
Closes #536.  Attempts to divide the cell into chunks that take a similar amount of work. The work estimate is produced by `fragment_stats::cost`, which was obtained from running linear regression on training data from random simulations. 

The default is still to use `split_by_effort`, which splits the chunks evenly without taking cost into account, but setting the `Simulation` constructor parameter `split_chunks_evenly=False` will use `split_by_cost`.  Here are some benchmark results from the following simulation (a tall skinny cell with an expensive flux plane at the top):

```python
import argparse
import time
import meep as mp

def main(args):
    sx = 2
    sy = 30
    sz = 1
    fcen = 0.15
    df = 0.1

    src = mp.Source(mp.GaussianSource(fcen, fwidth=df), mp.Ey, mp.Vector3())

    sim = mp.Simulation(cell_size=mp.Vector3(sx, sy, sz),
                        sources=[src],
                        resolution=10,
                        split_chunks_evenly=args.split_evenly)

    flux = sim.add_flux(fcen, df, 500, mp.FluxRegion(center=mp.Vector3(0, 12), size=mp.Vector3(2, 1)))
    start = time.time()
    sim.run(until=200)
    print("bench:{},{},{}".format(args.split_evenly, mp.count_processors(), time.time() - start))

if __name__ == '__main__':
    parser.add_argument('-e', '--split-evenly', action='store_true', default=False,
                                        help='Use the old split_by_effort method to create chunks')
    args = parser.parse_args()
    main(args)
```

bench_results

The `meep::structure` is now created in C++ (`create_structure_and_set_materials`) instead of Python. Using the geometry in `structure::choose_chunkdivision` would require making copies, or passing it around, both of which would require more significant changes. `create_structure_and_set_materials` allows the Python geometry to be converted to C++ once, then deleted when it returns.

The number of processors is divided into prime factors and the cell is split into chunks starting from the largest prime factor towards the smallest in order to achieve a more even grid split for powers of 3. Here are some examples of the improvement

**Left**: `split_by_effort`                                                  **Right**: `split_by_cost`
## Cell of air with 9 processors
nine_procs
## Cell of air with 9 processors and PML
nine_procs_pml
## Tall skinny cell with expensive flux plane at the top, 3 processors
(`split_by_effort` puts all the work on 1 processor).
three_procs_flux_plane
## Tall skinny cell with expensive flux plane at the top, 9 processors.
nine_procs_flux_plane
This can be improved by splitting in the `X` direction, but the gain is not above the 30% threshold so `split_by_cost` splits in the `Y` direction. Using machine learning to get the actual communication cost should improve this.

### Remaining issues
1. `split_by_cost` doesn't work with cylindrical coordinates yet so it falls back to `split_by_effort`
2. `split_by_cost` breaks the `structure::dump` and `structure::load` features. When you run the initial simulation and dump the structure file, the chunks are split optimally based on the geometry, but when you attempt to load the file, there is no geometry passed in so the chunks are split evenly, resulting in chunk size mismatches. Need to pass a list of volumes to these functions.

### DIFF
--- a/src/structure.cpp
+++ b/src/structure.cpp
@@ -133,12 +133,11 @@ void structure::choose_chunkdivision(const grid_volume &thegv, int desired_num_c
 
     grid_volume vi;
     if (meep_geom::fragment_stats::resolution == 0 ||
-        meep_geom::fragment_stats::has_non_medium_material() ||
-        gv.dim == Dcyl || meep_geom::fragment_stats::split_chunks_evenly) {
+        meep_geom::fragment_stats::has_non_medium_material() || gv.dim == Dcyl ||
+        meep_geom::fragment_stats::split_chunks_evenly) {
       // Fall back to split_by_effort method
       vi = gv.split_by_effort(desired_num_chunks, i, num_effort_volumes, effort_volumes, effort);
-    }
-    else {
+    } else {
       vi = gv.split_by_cost(desired_num_chunks, i);
     }
 
--- a/src/structure.cpp
+++ b/src/structure.cpp
@@ -72,6 +72,150 @@ structure::structure(const grid_volume &thegv, double eps(const vec &), const bo
   }
 }
 
+static std::vector<int> get_prime_factors(int n) {
+  int initial_n = n;
+  std::vector<int> result;
+
+  while (n % 2 == 0) {
+    result.push_back(2);
+    n /= 2;
+  }
+
+  for (int i = 3; i <= sqrt(n); i += 2) {
+    while (n % i == 0) {
+      result.push_back(i);
+      n /= i;
+    }
+  }
+
+  if (n > 5) {
+    // If we end up with a prime number greater than 5, then start over with n -1 in order to get
+    // the largest number that is a multiple of 2, 3, or 5.
+    return get_prime_factors(initial_n - 1);
+  } else if (n >= 2 && n <= 5) {
+    result.push_back(n);
+  }
+
+  return result;
+}
+
+static double get_cost_of_volume(grid_volume gv) {
+  geom_box box = meep_geom::gv2box(gv.surroundings());
+  meep_geom::fragment_stats fstats(box);
+  fstats.compute();
+  return fstats.cost();
+}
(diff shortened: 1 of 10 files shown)


src/structure.cpp
```diff
@@ -123,22 +169,42 @@ void structure::choose_chunkdivision(const grid_volume &thegv, int desired_num_c
...
+  std::vector<grid_volume> chunk_volumes;
+
+  if (meep_geom::fragment_stats::resolution == 0 ||
+      meep_geom::fragment_stats::has_non_medium_material() || gv.dim == Dcyl ||
+      meep_geom::fragment_stats::split_chunks_evenly) {
+    for (int i = 0; i < adjusted_num_chunks; i++) {
+      grid_volume vi = gv.split_by_effort(adjusted_num_chunks, i, num_effort_volumes, effort_volumes,
+                                          effort);
+      chunk_volumes.push_back(vi);
+    }
```
stevengj, line 192: Seems like we should have continued to use `desired_num_chunks` for the `split_by_effort` code, rather than `adjusted_num_chunks` (used for the new `split_by_cost` code)?

PR #691, opened 2019-01-30
Allow any callable object for user material function

ChristopherHogan:
Requested by Ian Sage on the mailing list. Allows passing a class instance to the `material_function` parameter of the `Simulation` constructor, provided the class's `__call__` method is implemented.
@stevengj @oskooi


stevengj:
Since `PyCallable_Check` always returns true in Python 3 for any native Python class (it only [seems to return false](https://github.com/python/cpython/blob/ab67281e95de1a88c4379a75a547f19a8ba5ec30/Objects/object.c) for C-created Python classes that are missing the `tp_call` slot), this is problematic.

As Chris notes, we still don't support class methods because the current implementation needs to set the `eps` property.  The current workaround is just to wrap a callable object in a `lambda`.

PR #697, opened 2019-02-01
Support load/dump structure when using split_by_cost

ChristopherHogan:
* `structure::dump` saves the `nums` and `origin` fields of each chunk's `grid_volume` in the structure h5 file.
* `structure::load` reads that data and recreates the chunks with the correct `grid_volume`s. Since this required access to the `boundary_region`s, I removed `Simulation.load_structure` from the docs. The public interface should be the `load_structure` argument to the `Simulation` constructor because it's too complicated for users to create the `boundary_region` object.
* It turns out that the `save_flux` and `load_flux` methods already work with `split_by_cost`, I guess because they deal with `dft_chunks`, which `split_by_cost` doesn't touch.
@stevengj @oskooi


ChristopherHogan:
Added support for cylindrical coordinates and documented the `split_chunks_evenly` flag.

src/structure_dump.cpp
```diff
@@ -354,11 +384,66 @@ void structure::set_chiP_from_file(h5file *file, const char *dataset, field_type
...
+    int origins_idx = i * gv_dim;
+    int nums_idx = i * 3;
+    grid_volume new_gv = gv;
+    vec new_origin(new_gv.dim);
+    LOOP_OVER_DIRECTIONS(gv.dim, d) {
+      new_origin.set_direction(d, origins[origins_idx++]);
+      new_gv.set_num_direction(d, nums[nums_idx + ((int)d % 3)]);
+    }
+    new_gv.set_origin(new_origin);
+    delete chunks[i];
```
stevengj: Probably want `if (chunks[i]->refcount-- <= 1) delete chunks[i];`

stevengj: and don't call `changing_chunks` (because that also allocates new chunks, so you would end up doing that twice)

src/structure_dump.cpp
```diff
@@ -354,11 +384,66 @@ void structure::set_chiP_from_file(h5file *file, const char *dataset, field_type
...
 
-void structure::load(const char *filename) {
+void structure::load(const char *filename, boundary_region &br) {
   h5file file(filename, h5file::READONLY, true);
 
   if (!quiet) master_printf("reading epsilon from file \"%s\"...\n", filename);
 
+  // Load chunk grid_volumes and adjust chunks (necessary when split_by_cost is used)
+  int gv_dim = number_of_directions(gv.dim);
+  size_t origins_sz = num_chunks * gv_dim;
```
stevengj: Just use `* 3` here

src/structure_dump.cpp
```diff
@@ -354,11 +384,66 @@ void structure::set_chiP_from_file(h5file *file, const char *dataset, field_type
...
+  if (am_master()) {
+    size_t gv_nums_start = 0;
+    file.read_chunk(1, &gv_nums_start, &nums_dims, nums);
+  }
+  file.prevent_deadlock();
+  broadcast(0, nums, nums_sz);
+
+  for (int i = 0; i < num_chunks; ++i) {
+    int origins_idx = i * gv_dim;
+    int nums_idx = i * 3;
```
stevengj: Can just use `i * 3` below once you make the change above.

stevengj:
In the future we might want to refactor this into a separate routine that just loads the ε and μ values etcetera without changing the chunks, and conversely you might want something that just loads the chunks geometry without loading ε and μ.

stevengj:
That is, you would leave the old `Simulation.load_structure` as-is, and add a new `chunk_layout` argument to the `Simulation` constructor that takes the filename of a "chunk layout" file (or possibly another `Simulation` object).

src/structure_dump.cpp
```diff
@@ -354,6 +384,70 @@ void structure::set_chiP_from_file(h5file *file, const char *dataset, field_type
   }
 }
 
+void structure::load_chunk_layout(const char *filename, boundary_region &br,
+                                  std::vector<grid_volume> gvs) {
```
stevengj: It seems like this should be two methods:
```
void structure::load_chunk_layout(const char *filename, boundary_region &br);
void structure::load_chunk_layout(std::vector<grid_volume> gvs, boundary_region &br);
```
where the former calls the latter.

ChristopherHogan: Done.

PR #700, opened 2019-02-05
Add Python method to get farfields as array

### LINKED ISSUE #680: get_farfields routine to obtain far fields as NumPy array
Currently, there are only two routines available for obtaining the [far fields](https://meep.readthedocs.io/en/latest/Python_User_Interface/#near-to-far-field-spectra): 

1. `get_farfield`: returns the far field at a single point
2. `output_farfields`: writes the far fields in a given [`Volume`](https://meep.readthedocs.io/en/latest/Python_User_Interface/#volume) to an HDF5 file 

It would be useful to have an additional third routine which returns the far fields in a given `Volume` as a NumPy array. This would facilitate post processing directly from within a single Python script without having to write and read from disk.

As an extension of this feature, rather than restrict the far field output to a rectilinear `Volume`, the region could be generalized to a [`GeometricObject`](https://meep.readthedocs.io/en/latest/Python_User_Interface/#geometricobject) (e.g., `Sphere`, `Prism`, `Cylinder`, etc.)

ChristopherHogan:
Closes #680.
* Add `dft_near2far::get_farfields_array`
* `dft_near2far::save_farfields` calls `get_farfields_array`.
* Add wrapper `_get_farfields_array` so Python can call `dft_near2far::get_farfields_array`.
* Add `cavity_farfield.py` test to make sure the `get_farfields` array is the same as the h5 file from `output_farfields`.

### TODO
* ~~What is the ideal shape or return type for `get_farfields`? Currently it's a dict of numpy arrays with keys `Ex`, `Ey`, ... `Hz`, each with shape `(d1, d2, d3)` with dimensions of size 1 collapsed.~~
* ~~Test with `Nfreq` > 1.~~
@stevengj @oskooi


stevengj:
For `Nfreqs > 1`, let's just do the same thing as `output_farfields`: each array has an extra dimension corresponding to the frequencies.

python/simulation.py
```diff
@@ -1316,6 +1316,24 @@ def _add_near2far(self, fcen, df, nfreq, near2fars):
     def get_farfield(self, f, v):
         return mp._get_farfield(f.swigobj, py_v3_to_vec(self.dimensions, v, is_cylindrical=self.is_cylindrical))
 
+    def get_farfields(self, near2far, resolution, where=None, center=None, size=None):
+        vol = self._volume_from_kwargs(where, center, size)
+        result = mp._get_farfields_array(near2far.swigobj, vol, resolution)
+        res_ex = result[0] + result[1] * 1j
```
stevengj: Since we do this a lot, to avoid allocating additional temporary arrays, we might want a function like:
```py
def complexarray(re, im):
    z = im*1j
    z += re # add re to z in-place
    return z
```

ChristopherHogan: Done.

PR #701, opened 2019-02-06
add five FAQS and minor fixes to docs

oskooi:
- How do I compute the effective index of an eigenmode of a lossy waveguide?
- How do I compute the time average of the harmonic fields?
- How do I create a circularly-polarized planewave source in cylindrical coordinates?
- Why doesn't the continuous-wave (CW) source produce an exact single-frequency response?
- How do I create a focused beam with a Gaussian envelope?


doc/docs/FAQ.md
```diff
@@ -190,10 +192,26 @@ There are at least two possible reasons due to using: (1) a `material_function`
...
+
 ### How do I compute the group velocity of a mode?
 
 There are two possible approaches for computing the group velocity: (1) compute the [band diagram](https://en.wikipedia.org/wiki/Band_diagram) ω(**k**) using [`Harminv`](Python_User_Interface.md#harminv), fit it to a polynomial, and calculate its derivative using a [finite difference](https://en.wikipedia.org/wiki/Finite_difference), or (2) excite the mode using a narrowband pulse and compute the ratio of the flux to energy density.
 
+### How do I compute the time average of the harmonic fields?
+
+For a linear system, you can use a [ContinuousSource](Python_User_Interface.md#continuoussource) with `force_complex_fields=True` and time-step the fields until all transients have disappeared. The instantaneous steady-state field is equivalent to the time average.
+
```
stevengj: This isn't correct.   It's only quantities like |E|^2 and Re[E* x H] that are equivalent to the time average.

oskooi:
The time-average FAQ has been fixed. Separately, there was an error in the documentation of [`ContinuousSource`](https://meep.readthedocs.io/en/latest/Python_User_Interface/#continuoussource) involving a `cutoff` parameter which does not exist. This has been replaced with a description of the `slowness` parameter (which was missing). A figure has also been added to illustrate the effect of the `width` and `slowness` parameters in the turn-on function of the CW source.

doc/docs/FAQ.md
```diff
@@ -316,11 +316,15 @@ No. Meep does not support non-orthogonal grids with spatially varying resolution
...
 
-### Why does the continuous-wave (CW) source produce a finite bandwidth response?
+### Why doesn't the continuous-wave (CW) source produce a single-frequency response?
 
-The [ContinuousSource](Python_User_Interface.md#continuoussource) does not produce an exact single frequency response due to its finite turn-on time (which is described by a hyperbolic tangent function). If you Fourier transform the response, the finite turn-on time will produce other frequency components. After the turn-on, the continuous source is described by a sinusoid of the given frequency.
+The [ContinuousSource](Python_User_Interface.md#continuoussource) does not produce an exact single frequency response due to its finite turn-on time which is described by a hyperbolic tangent function. If you Fourier transform the response, the finite turn-on time will produce additional frequency components. After the turn-on, the CW source is described by a sinusoid of the given frequency.
```
stevengj: I would say that asymptotically the resulting fields *are* the single frequency response.  It's just that if you Fourier transform the whole simulation you will see a finite bandwidth due to the turn-on.

oskooi:
A description of the asymptotic limit of the CW source as well as the effect of its turn-on on the Fourier transform has been added for clarification.

jball1982:
> The time-average FAQ has been fixed. Separately, there was an error in the documentation of [`ContinuousSource`](https://meep.readthedocs.io/en/latest/Python_User_Interface/#continuoussource) involving a `cutoff` parameter which does not exist. This has been replaced with a description of the `slowness` parameter (which was missing). A figure has also been added to illustrate the effect of the `width` and `slowness` parameters in the turn-on function of the CW source.

Sorry to dredge up an old topic, but after years of sticking with the C++ interface to MEEP, I've been familiarizing myself with the Python interface, and I've found some of the differences to be confusing. I was searching the Github issues for this very question, because while "slowness" is described in the source code comments (I'm looking at code for MEEP 1.26.0 on Fossies), it seems that the slowness can't be set, but if you pass a cutoff value by keyword to the Python instruction, its __init__ will pass the "cutoff" value to the C++ constructor as the "slowness" value....do I have this correct? It's been a bit confusing trying to parse the differences between C++ and Python.

doc/docs/FAQ.md
```diff
@@ -190,9 +192,37 @@ There are at least two possible reasons due to using: (1) a `material_function`
...
+
+To obtain the loss in time, you make your computational cell a cross-section of your waveguide (i.e. 2d for a waveguide with constant cross-section), and set Bloch-periodic boundary conditions via the `k_point` input variable &mdash; this specifies your (real) β. You then treat it exactly the same as a [resonant-cavity problem](Python_Tutorials/Resonant_Modes_and_Transmission_in_a_Waveguide_Cavity.md#resonant-modes): you excite the system with a short pulse source, monitor the field at some point, and then analyze the result with [`Harminv`](Python_User_Interface.md#harminv); all of which is done if you call `run_kpoints`. This will give you the complex ω at the given β, where the imaginary part is the loss rate in time. Note: the loss in a uniform waveguide, with no absorption or disorder, is zero, even in the discretized system.
+
```
stevengj, line 214: Am I missing something, or does this FAQ never answer the question?  At the end, the effective index is usually defined as β/ω.

All the business with the group velocity is only if you need the *complex* effective index.  If you just want the real effective index (e.g. for a lossless waveguide), you can skip it.  Of course, in that case you could also use MPB.

oskooi: @jball1982: this was a bug. Should be fixed in #2499.

oskooi: @stevengj: this FAQ specifically addresses the case of *lossy* waveguides for which the answer provided would seem sufficient. Anyhow, for completeness, I have updated this FAQ in #2500 to also describe the case of lossless waveguides.

jball1982: @oskooi Thanks Ardavan, I appreciate your help. Despite the documentation on [readthedocs](https://meep.readthedocs.io/), I'm still having some trouble connecting the dots between C++ and Python in some cases. I am going to dig a little more and perhaps create a discussion on here soon with some questions.

Issue #604, opened 2018-11-12
get_eigenmode_coefficients for degenerate eigenmodes in uniform media


oskooi:
For a 2d cell in *x* and *y* with a non-zero k<sub>z</sub>, `get_eigenmode` seems to be computing an incorrect set of modes. This is demonstrated by the following script.

```py
import meep as mp
import math

resolution = 50        # pixels/μm                                                                                                                                    

dpml = 2               # PML thickness                                                                                                                                
dmat = 10              # length of material                                                                                                                           

sx = dpml+dmat+dpml
sy = 10

cell_size = mp.Vector3(sx,sy,0)
pml_layers = [mp.Absorber(thickness=dpml,direction=mp.X)]

wvl = 0.5              # center wavelength                                                                                                                            
fcen = 1/wvl           # center frequency                                                                                                                             

ng = 1.5
glass = mp.Medium(index=ng)

# out-of-plane (XZ) rotation angle of incident planewave; CCW about Y axis, 0 degrees along in-plane direction                                                        
theta_out = math.radians(20.3)

# in-plane (XY) rotation angle of incident planewave; CCW about Z axis, 0 degrees along +X axis                                                                       
theta_in = math.radians(10.7)

# k (in source medium) with correct length (plane of incidence: XY)                                                                                                   
k = mp.Vector3(math.cos(theta_in)*math.cos(theta_out),math.sin(theta_in)*math.cos(theta_out),math.sin(theta_out)).scale(fcen*ng)

sim = mp.Simulation(resolution=resolution,
                    cell_size=cell_size,
                    boundary_layers=pml_layers,
                    k_point=k,
                    default_material=glass)

sim.init_sim()

mon_pt = mp.Vector3(0.5*sx-dpml-0.2*dmat,0,0)

for num_band in range(1,7):
    sim.get_eigenmode(fcen, mp.X, mp.Volume(center=mon_pt, size=mp.Vector3(0,sy,0)), num_band, k, verbose=True)
```

The output shows that the dominant wavevectors of the six modes are *not* unique and instead come in pairs.

```
Dominant planewave for band 1: (2.813578,0.022404,1.040807)
Dominant planewave for band 2: (2.813578,0.022404,1.040807)
Dominant planewave for band 3: (2.812597,-0.077596,1.040807)
Dominant planewave for band 4: (2.812597,-0.077596,1.040807)
Dominant planewave for band 5: (2.811003,0.122404,1.040807)
[...]

PR #703, opened 2019-02-08
Reuse Maxwell data in get_eigenmode_coefficients

### LINKED ISSUE #647: more efficient mode solver for get_eigenmode_coefficients
When `get_eigenmode_coefficients` is running for many modes and many frequencies, it calls `get_eigenmode` over and over.  On *every* call it runs `create_maxwell_data` and `set_maxwell_dielectric`, even though these are the same for every mode.   It would be faster to share this between different bands/frequencies, maybe by a lower-level `get_eigenmode` routine that allows you to pass `mdata` as a parameter.

ChristopherHogan:
Closes #647. I tested this in `python/tests/mode_decomposition.py` with 1000 frequencies and saw a ~4.5x speedup over `master`.
@stevengj @oskooi

### DIFF
--- a/src/meep.hpp
+++ b/src/meep.hpp
@@ -1562,7 +1562,7 @@ public:
   void *get_eigenmode(double omega_src, direction d, const volume where, const volume eig_vol,
                       int band_num, const vec &kpoint, bool match_frequency, int parity,
                       double resolution, double eigensolver_tol, bool verbose = false,
-                      double *kdom = 0);
+                      double *kdom = 0, void **user_mdata = 0);
 
   void add_eigenmode_source(component c, const src_time &src, direction d, const volume &where,
                             const volume &eig_vol, int band_num, const vec &kpoint,
@@ -1865,7 +1865,7 @@ void green3d(std::complex<double> *EH, const vec &x, double freq, double eps, do
 
 // non-class methods for working with mpb eigenmode data
 //
-void destroy_eigenmode_data(void *vedata);
+void destroy_eigenmode_data(void *vedata, bool destroy_mdata = true);
 std::complex<double> eigenmode_amplitude(void *vedata, const vec &p, component c);
 double get_group_velocity(void *vedata);
 vec get_k(void *vedata);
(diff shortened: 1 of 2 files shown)


oskooi:
For reference, when running [python/examples/binary_grating_oblique.py](https://github.com/NanoComp/meep/blob/master/python/examples/binary_grating_oblique.py) which involves computing 96 eigenmodes at a single frequency (split into two separate batches of 58 and 38 modes), the total run time decreased from `151.8873 s` to `146.9405 s` which is just a 3% improvement using this PR. It seems that a large fraction of the runtime is spent by MPB in computing the eigenmodes rather loading the dielectric data. If this is indeed the case, there doesn't seem to be another approach to speed up these types of calculations.

FilipDominec:
@oskooi I agree that "[it] doesn't seem to be another approach to speed up" in the most general case of monitor planes being placed possibly _in the middle_ of a dielectric structure (e.g. a waveguide).

However, in many cases including the grating example, the monitor is located in a _homogeneous medium_. Therefore, MPB will return analytically known oblique plane waves, differing only by their transverse k-vectors and field polarisations. This means MPB is actually not needed for monitors in homogeneous media.

Without deeper knowledge of the python-meep's new API, I suggest there should be something like ```get_planewave_coefficients``` with the same function signature to ```get_eigenmode_coefficients```. For many people, this would be an ultimate speedup of simulations.

oskooi:
For homogeneous media, the eigenmodes are just planewaves (ω=ck/n) and can indeed be computed analytically which would thus avoid calling MPB entirely. It's not clear though whether this would offer much of a speedup since MPB is already quite fast for calculations involving planewaves in homogeneous media (i.e., converging in ~1-2 iterations, less than a second or two in walltime). Anyhow, if we were to support such a feature, it would probably be better to include it as part of `get_eigenmode_coefficients` (letting Meep decide whether to invoke MPB or to compute the planewaves analytically) rather than create a separate function `get_planewave_coefficients`. (A similar approach can be applied to the eigenmode source.) Feel free to create a separate issue for this feature. Note: we are still working on the API for 3d planewaves, see #604.

PR #717, opened 2019-02-15
Fix link issue on MacOS

### LINKED ISSUE #660: libmeepgeom link problem on MacOS
There seems to be some problem on MacOS (which is currently not tested on Travis) with how `libmeepgeom` was merged into `libmeep` (#630).     It builds, but when I try to run one of the test programs, for example, I get:
```sh
unixprompt$ tests/aniso_disp 
dyld: Symbol not found: _default_material
  Referenced from: /Users/stevenj/Code/meep/src/.libs/libmeep.12.dylib
  Expected in: flat namespace
 in /Users/stevenj/Code/meep/src/.libs/libmeep.12.dylib
Abort trap: 6
```
cc @ChristopherHogan

ChristopherHogan:
Fixes #660, hopefully. I wan't able to test it on a Mac though.
@stevengj @oskooi


ChristopherHogan:
I tested it through conda-forge's CI, but it doesn't seem to work. I can't ssh to the machine to debug though.

PR #734, opened 2019-02-20
Add -lctlgeom to LIBS

### LINKED ISSUE #660: libmeepgeom link problem on MacOS

ChristopherHogan:
Fixes #660 according to @underclocked, although this still won't work when using `-dead_strip_dylibs`.
@stevengj @oskooi

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -423,14 +423,8 @@ AC_SUBST(LIBCTL_CPPFLAGS)
 ##############################################################################
 
 # check for -lctlgeom:
-AC_CHECK_LIB(ctlgeom, vector3_dot, [have_libctlgeom=yes], [have_libctlgeom=no])
+AC_CHECK_LIB(ctlgeom, vector3_dot, [have_libctlgeom=yes; LIBS="-lctlgeom $LIBS"], [have_libctlgeom=no])
 AM_CONDITIONAL(WITH_LIBCTLGEOM, test x"$have_libctlgeom" = "xyes")
-
-if test "x$have_libctlgeom" = xyes; then
-   LIBCTLGEOM_LIBS="-lctlgeom"
-fi
-
-AC_SUBST(LIBCTLGEOM_LIBS)
 
 ##############################################################################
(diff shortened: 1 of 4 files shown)


configure.ac
```diff
@@ -423,15 +423,9 @@ AC_SUBST(LIBCTL_CPPFLAGS)
 ##############################################################################
 
 # check for -lctlgeom:
-AC_CHECK_LIB(ctlgeom, vector3_dot, [have_libctlgeom=yes], [have_libctlgeom=no])
+AC_CHECK_LIB(ctlgeom, vector3_dot, [have_libctlgeom=yes; LIBS="-lctlgeom $LIBS"], [have_libctlgeom=no])
```
stevengj: This doesn't seem right to me, since we don't want to link `-lctlgeom` from the `meep` executable that uses Scheme.   That's why I had the separate `LIBCTLGEOM_LIBS` variable.

stevengj: Oh, nevermind, this is correct.   Now that we have included libmeepgeom in libmeep, we need to always link libctlgeom.

I think there is no conflict with Scheme because, although Scheme compiles its own version of the libctlgeom routines, it does so in C++ within a `ctlio::` namespace?  I'm not sure how it gets the right version of functions like `object_of_point_in_tree`, that aren't namespaced in `geom.c`, however … maybe the linker does the right thing because `geom.o` is listed before `-lctlgeom` in the link command for `meep`?

PR #746, opened 2019-02-27
Streamline array slice interface

### LINKED ISSUE #661: collapse=False parameter is ignored in slicing routines
The recently added `get_dft_array_metadata` routine accepts four parameters: `dft_cell=None, vol=None, center=None, size=None`. However, it seems that the three geometric parameters &mdash; `vol`, `center`, and `size` &mdash; are never used. This is actually consistent with `get_dft_array` which also does not provide for the slicing of its `dft` object (unlike `get_array`). Thus, unless (or until) we provide this additional functionality for both `get_dft_array` and `get_dft_array_metadata`, we should remove the three unused parameters from `get_dft_array_metadata`.

As a demonstration, the following script extracts two different slices using `get_dft_array_metadata` by specifying unique values of `center` and `size`.

```py
import meep as mp
import numpy as np

resolution = 20

n = 3.4
w = 1
r = 1
pad = 4
dpml = 2

sxy = 2*(r+w+pad+dpml)
vol = mp.Volume(mp.Vector3(), size=mp.Vector3(sxy,sxy))

c2 = mp.Cylinder(radius=r)

fcen = 0.118
df = 0.08

                    geometry=[c1,c2],
                    sources=[src],
                    resolution=resolution,

(x1,y1,z1,w1) = sim.get_dft_array_metadata(dft_cell=dfts,center=mp.Vector3(0.5*sxy-dpml-2.3),size=mp.Vector3(0,sxy-2*dpml,0))
(x2,y2,z2,w2) = sim.get_dft_array_metadata(dft_cell=dfts,center=mp.Vector3(),size=mp.Vector3(2.3,2.3,0))
```
(note: due to #658, symmetries have been omitted)

The two sets of parameters (`x1`,`y1`,`z1`,`w1`) and (`x2`,`y2`,`z2`,`w2`) should be different but turn out to be in fact equal:
```
>>> w1.shape
(320, 320)
>>> w2.shape
(320, 320)
>>> np.where(w1!=w2)
(array([], dtype=int64), array([], dtype=int64))
```
Rather unexpectedly, if the `dft_cell` parameter is removed, the two slices turn out to be different.
```py
(x1,y1,z1,w1) = sim.get_dft_array_metadata(center=mp.Vector3(0.5*sxy-dpml-2.3),size=mp.Vector3(0,sxy-2*dpml,0))
(x2,y2,z2,w2) = sim.get_dft_array_metadata(center=mp.Vector3(),size=mp.Vector3(2.3,2.3,0))
```
```
>>> w1.shape
(242,)
>>> w2.shape
(48, 48)
```

### LINKED ISSUE #670: streamlining interface to slicing routines
Four suggestions for creating a more streamlined interface to the slicing routines:

1. change the interface of `get_array(vol=None, center=None, size=None, component=mp.Ez, ...)` to `get_array(component, vol=None, center=None, size=None, ...)` (i.e., `component` is the first argument) in order to be consistent with `get_source_slice`

2. rename `get_source_slice` to `get_source` to be consistent with `get_array`

3. remove `get_dft_array_metadata` and have just a single `get_array_metadata` which accepts a `dft_cell` keyword argument and set `collapse=True` by default for everything

4. `get_array` should have `collapse=True` by default (currently it is false)

ChristopherHogan:
Closes #670. Closes #661.
* `get_array` takes `component` as first argument (remains a keyword to keep as much backward compatibility as possible).
* `get_dft_array_metadata` is deprecated.
* Removed `collapse` parameter from `get_array_metadata` (defaults to `True`).
* Deprecated `get_source_slice` and added `get_source`. 

@HomerReid may want to review the changes.
@stevengj @oskooi

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -1606,7 +1606,12 @@ class Simulation(object):
         cmd = re.sub(r'\$EPS', self.last_eps_filename, opts)
         return convert_h5(rm_h5, cmd, *step_funcs)
 
-    def get_array(self, vol=None, center=None, size=None, component=mp.Ez, cmplx=None, arr=None):
+    def get_array(self, component=None, vol=None, center=None, size=None, cmplx=None, arr=None):
+        if component is None:
+            raise ValueError("component is required")
+        if isinstance(component, mp.Volume) or isinstance(component, mp.volume):
+            raise ValueError("The first argument must be the component")
+
         dim_sizes = np.zeros(3, dtype=np.uintp)
 
         if vol is None and center is None and size is None:
@@ -1659,7 +1664,7 @@ class Simulation(object):
         else:
             raise ValueError("Invalid type of dft object: {}".format(dft_swigobj))
 
-    def get_source_slice(self, component, vol=None, center=None, size=None):
+    def get_source(self, component, vol=None, center=None, size=None):
         if vol is None and center is None and size is None:
             v = self.fields.total_volume()
         else:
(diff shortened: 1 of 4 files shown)


python/simulation.py
```diff
@@ -1671,28 +1676,36 @@ def get_source_slice(self, component, vol=None, center=None, size=None):
...
+
+    def get_array_metadata(self, vol=None, center=None, size=None, dft=None):
+        if dft:
+            vol = dft_cell.where
+        if vol is None and center is None and size is None:
+            v = self.fields.total_volume()
+        else:
+            v = self._volume_from_kwargs(vol, center, size)
+        dims = np.zeros(3, dtype=np.uintp)
+        rank, dirs = mp._get_array_slice_dimensions(self.fields, v, dims, True)
```
stevengj: Should be `bool(dft)` rather than `True`, I think?

PR #747, opened 2019-02-27
Add dft_energy Python support

ChristopherHogan:
## TODO
1. ~~Tests~~
2. ~~I don't know what `DftEnergy.num_components` should be. It's the value that we multiply the number of dft pixels by in the `fragment_stats`.  For `dft_near2far` it's 4, for `dft_force` it's 6, for `dft_flux` it's 4, etc.~~
@stevengj @oskooi


stevengj:
`DftEnergy.num_components` should be 12 (in 3d): E, D, H, and B fields, 3 components each.

ChristopherHogan:
Added test and fixed remaining TODO items.

PR #753, opened 2019-03-04
add missing volume member to dft_energy class and Scheme example for computing group velocity

oskooi:
Fixes a bug in the `dft_energy` class from #744 where the `volume` of the `dft` object was not being stored as a class member and therefore not used to normalize the energy density values as part of the quadrature.

Also adds a Scheme example demonstrating the equivalence of computing the group velocity of a waveguide mode (from the [`EigenModeSource` tutorial](https://meep.readthedocs.io/en/latest/Python_Tutorials/Eigenmode_Source/)) using two different methods: (1) as the ratio of the Poynting flux to energy density and (2) directly via `get-eigenmode-coefficients` involving mode decomposition. This example can be ported to Python as part of #747. The test would verify that the relative error in the two values converges monotonically to 0 as the `resolution` is increased (e..g., 10, 20, and 40).

### DIFF
--- a/src/dft.cpp
+++ b/src/dft.cpp
@@ -446,18 +446,21 @@ dft_flux fields::add_dft_flux(const volume_list *where_, double freq_min, double
 
 dft_energy::dft_energy(dft_chunk *E_, dft_chunk *H_,
                        dft_chunk *D_, dft_chunk *B_,
-                       double fmin, double fmax, int Nf)
-{
+                       double fmin, double fmax, int Nf, const volume &where_)
+  : Nfreq(Nf), E(E_), H(H_), D(D_), B(B_), where(where_) {
   if (Nf <= 1) fmin = fmax = (fmin + fmax) * 0.5;
   freq_min = fmin;
-  Nfreq = Nf;
   dfreq = Nf <= 1 ? 0.0 : (fmax - fmin) / (Nf - 1);
-  E = E_; H = H_; D = D_; B = B_;
 }
 
-dft_energy::dft_energy(const dft_energy &f) {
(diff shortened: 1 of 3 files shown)


src/meep.hpp
```diff
@@ -1082,6 +1082,7 @@ class dft_energy {
   double freq_min, dfreq;
   int Nfreq;
   dft_chunk *E, *H, *D, *B;
+  volume where;
```
stevengj: This needed only if you add an `fields::output_dft` method for `dft_energy`.

src/dft.cpp
```diff
@@ -494,33 +497,35 @@ double *dft_energy::total() {
...
-      D = add_dft(direction_component(Dx, d), where->v, freq_min, freq_max, Nfreq,
-		  true, 1.0, D);
-      H = add_dft(direction_component(Hx, d), where->v, freq_min, freq_max, Nfreq,
-		  true, 1.0, H);
-      B = add_dft(direction_component(Bx, d), where->v, freq_min, freq_max, Nfreq,
-		  true, 1.0, B);
+      E = add_dft(direction_component(Ex, d), where->v, freq_min, freq_max, Nfreq, true, 1.0, E);
+      D = add_dft(direction_component(Dx, d), where->v, freq_min, freq_max, Nfreq, false, 1.0, D);
+      H = add_dft(direction_component(Hx, d), where->v, freq_min, freq_max, Nfreq, true, 1.0, H);
+      B = add_dft(direction_component(Bx, d), where->v, freq_min, freq_max, Nfreq, false, 1.0, B);
```
stevengj: The change to pass `false` for two of the fields is what is fixing your dV factor.   Adding the `where` field to `dft_energy` is unrelated (and is actually unused in this PR).

PR #769, opened 2019-03-14
WIP: periodic green's functions

### LINKED ISSUE #763: near to far field transformation with Bloch-periodic boundaries
The [`near to far field transformation`](https://meep.readthedocs.io/en/latest/Python_User_Interface/#near-to-far-field-spectra) feature only works with [non-periodic regions](https://www.mail-archive.com/meep-discuss@ab-initio.mit.edu/msg06261.html) in free space. It would be useful to also support Bloch-periodic boundaries via the `k_point` for e.g. computing the focal length and other imaging properties of a metasurface lens directly from its far-field intensity profile. This type of calculation is also possible using [mode decomposition](https://meep.readthedocs.io/en/latest/Python_User_Interface/#mode-decomposition) but requires additional post-processing for the user which may be cumbersome.

The challenge is implementing the periodic Green's functions. Naively, you just sum periodic copies of the Green's function but that converges very slowly. @HomerReid has implemented a much more sophisticated technique based on [Ewald summation](https://en.wikipedia.org/wiki/Ewald_summation) in [SCUFF-EM](http://homerreid.github.io/scuff-em-documentation/) which perhaps can be adapted for use in Meep. Homer's code is for 3d but we would also need the 2d case.

stevengj:
First attempt (currently untested) to hack in support for periodic Green's functions (closes #763).

That is, it detects whether any directions of the near-to-far surface coincide with a periodic direction of the computational cell.  If so, in that direction it just does a naive Bloch-periodic summation of the Green's function.  Currently, it just sums 10 copies.   For far-away surfaces we may need many more, and in principle should do something like Ewald summation, but for now this is hopefully good enough to experiment with.

*Update*: modulo the bugfix reported below, @oskooi reports that it seems to be working in a quick test.  To do:

- [x] Fix the `x0` typo noted below.
- [ ] A better test.   e.g. the binary grating example, but comparing the near2far fields with the explicitly computed fields (in a sufficiently large computational cell). Can be a separate PR as long as this is undocumented and opt-in.
- [x] The current hack of summing 10 copies is not acceptable in the long run.  Should we do something adaptive?  Or should the lattice sum be a user parameter?  The basic problem is that, the farther away you get, the more copies you need to sum.  The only way around this scaling is to do something fancy like Ewald summation, it seems? *Update: just made it a user parameter for now.*
- [x] Fix the handling of mirror symmetries etc. that chop up the near2far surface, which currently stop the code from working as in @oskooi's example below. 
- [x] In the current PR, the lattice sum turns on automatically if it detects that the near2far surface coincides with a periodic direction.   Since this is currently experimental, my inclination is that it shouldn't turn on by default.   Maybe the best thing is for the lattice sum to be a user parameter that defaults to 1.  In the short term, this will allow users to turn it on and make it sufficiently large for their needs as an option.

### DIFF
--- a/src/near2far.cpp
+++ b/src/near2far.cpp
@@ -265,7 +265,7 @@ void dft_near2far::farfield_lowlevel(std::complex<double> *EH, const vec &x) {
               xs.set_direction(periodic_d[1], x0.in_direction(periodic_d[1]) + i1*period[1]);
             double phase = phase0 + i1*periodic_k[1];
             std::complex<double> cphase = std::polar(1.0, phase);
-            green(EH6, x, freq, eps, mu, x0, c0, f->dft[Nfreq * idx_dft + i]);
+            green(EH6, x, freq, eps, mu, xs, c0, f->dft[Nfreq * idx_dft + i]);
             for (int j = 0; j < 6; ++j)
               EH[i * 6 + j] += EH6[j] * cphase;
           }
--- a/src/near2far.cpp
+++ b/src/near2far.cpp
@@ -444,7 +444,7 @@ double *dft_near2far::flux(direction df, const volume &where, double resolution)
 static double approxeq(double a, double b) { return fabs(a - b) < 0.5e-11 * (fabs(a) + fabs(b)); }
 
 dft_near2far fields::add_dft_near2far(const volume_list *where, double freq_min, double freq_max,
-                                      int Nfreq) {
+                                      int Nfreq, int Nperiods) {
   dft_chunk *F = 0; /* E and H chunks*/
   double eps = 0, mu = 0;
   volume everywhere = where->v;
@@ -497,10 +497,9 @@ dft_near2far fields::add_dft_near2far(const volume_list *where, double freq_min,
 
     for (int i = 0; i < 2; ++i) {
       if (has_direction(v.dim, fd[i]) &&
-          boundaries[High][fd[i]] == Periodic && boundaries[Low][fd[i]] == Periodic &&
-          float(w->v.in_direction(fd[i])) == float(v.in_direction(fd[i]))) {
+          boundaries[High][fd[i]] == Periodic && boundaries[Low][fd[i]] == Periodic) {
         periodic_d[i] = fd[i];
-        periodic_n[i] = 10; // just hard-code a 10x10 supercell for now
+        periodic_n[i] = Nperiods;
         period[i] = v.in_direction(fd[i]);
         periodic_k[i] = 2*pi*real(k[fd[i]]) * period[i];
       }
--- a/src/near2far.cpp
+++ b/src/near2far.cpp
@@ -30,16 +30,30 @@ using namespace std;
 namespace meep {
 
 dft_near2far::dft_near2far(dft_chunk *F_, double fmin, double fmax, int Nf, double eps_, double mu_,
-                           const volume &where_)
+                           const volume &where_, const direction periodic_d_[2],
+                           const int periodic_n_[2], const double periodic_k_[2], const double period_[2])
     : Nfreq(Nf), F(F_), eps(eps_), mu(mu_), where(where_) {
   if (Nf <= 1) fmin = fmax = (fmin + fmax) * 0.5;
   freq_min = fmin;
   dfreq = Nf <= 1 ? 0.0 : (fmax - fmin) / (Nf - 1);
+  for (int i = 0; i < 2; ++i) {
+    periodic_d[i] = periodic_d_[i];
+    periodic_n[i] = periodic_n_[i];
+    periodic_k[i] = periodic_k_[i];
+    period[i] = period_[i];
+  }
 }
 
 dft_near2far::dft_near2far(const dft_near2far &f)
     : freq_min(f.freq_min), dfreq(f.dfreq), Nfreq(f.Nfreq), F(f.F), eps(f.eps), mu(f.mu),
-      where(f.where) {}
+      where(f.where) {
+  for (int i = 0; i < 2; ++i) {
+    periodic_d[i] = f.periodic_d[i];
+    periodic_n[i] = f.periodic_n[i];
+    periodic_k[i] = f.periodic_k[i];
+    period[i] = f.period[i];
+  }
+}
 
 void dft_near2far::remove() {
   while (F) {
@@ -241,9 +255,21 @@ void dft_near2far::farfield_lowlevel(std::complex<double> *EH, const vec &x) {
       x0 = f->S.transform(x0, f->sn) + rshift;
       for (int i = 0; i < Nfreq; ++i) {
         double freq = freq_min + i * dfreq;
-        green(EH6, x, freq, eps, mu, x0, c0, f->dft[Nfreq * idx_dft + i]);
-        for (int j = 0; j < 6; ++j)
-          EH[i * 6 + j] += EH6[j];
+        vec xs(x0);
+        for (int i0 = -periodic_n[0]; i0 <= periodic_n[0]; ++i0) {
+          if (periodic_d[0] != NO_DIRECTION)
+            xs.set_direction(periodic_d[0], x0.in_direction(periodic_d[0]) + i0*period[0]);
+          double phase0 = i0*periodic_k[0];
+          for (int i1 = -periodic_n[1]; i1 <= periodic_n[1]; ++i1) {
+            if (periodic_d[1] != NO_DIRECTION)
+              xs.set_direction(periodic_d[1], x0.in_direction(periodic_d[1]) + i1*period[1]);
+            double phase = phase0 + i1*periodic_k[1];
+            std::complex<double> cphase = std::polar(1.0, phase);
+            green(EH6, x, freq, eps, mu, x0, c0, f->dft[Nfreq * idx_dft + i]);
+            for (int j = 0; j < 6; ++j)
+              EH[i * 6 + j] += EH6[j] * cphase;
+          }
+        }
       }
       idx_dft++;
     }
@@ -423,6 +449,10 @@ dft_near2far fields::add_dft_near2far(const volume_list *where, double freq_min,
   double eps = 0, mu = 0;
   volume everywhere = where->v;
 
+  direction periodic_d[2] = {NO_DIRECTION, NO_DIRECTION};
+  int periodic_n[2] = {0, 0};
+  double periodic_k[2] = {0, 0}, period[2] = {0, 0};
+
   for (const volume_list *w = where; w; w = w->next) {
     everywhere = everywhere | where->v;
     direction nd = component_direction(w->c);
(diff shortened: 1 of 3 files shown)


oskooi:
There is likely a simple bug in this PR because it does not appear to actually do anything. This is demonstrated using the script below which computes the far-field intensity profile relative to vacuum at λ=0.5 μm along a line for a [binary grating](https://meep.readthedocs.io/en/latest/Python_Tutorials/Mode_Decomposition/#transmittance-spectra-for-planewave-at-normal-incidence). The test involves verifying that the diffraction spectra contains only *odd* orders, m=±1 (±2.9°), ±3 (±8.6°), ±5 (±14.5°), etc., and that the ratio of the peaks for any two orders m<sub>1</sub> and m<sub>2</sub> matches the analytic result (m<sub>2</sub>)<sup>2</sup>/(m<sub>1</sub>)<sup>2</sup>.

```py
import meep as mp
import math
import numpy as np

resolution = 20        # pixels/μm                                                                                                                                         

sx = dpml+dsub+gh+dpad+dpml

sources = [mp.Source(mp.GaussianSource(fcen, fwidth=df), component=mp.Ez, center=src_pt)]

k_point = mp.Vector3(0,0,0)

glass = mp.Medium(index=1.5)

                    cell_size=mp.Vector3(sx),
                    boundary_layers=pml_layers,
                    k_point=k_point,
                    default_material=glass,
                    sources=sources)

mon_pt = mp.Vector3(0.5*sx-dpml-0.5*dpad)
n2f_mon = sim.add_near2far(fcen, 0, 1, mp.Near2FarRegion(center=mon_pt, size=mp.Vector3()))

sim.run(until_after_sources=mp.stop_when_fields_decayed(50, mp.Ez, mon_pt, 1e-9))

ff_x = 1e8     # location of line (10 m away from grating)                                                                                                                 
ff_angle = 20  # half angle subtended by the line                                                                                                                          
ff_y = 2*ff_x*math.tan(math.radians(ff_angle)) # length of line                                                                                                            
ff_npts = 800  # number of evenly-spaced points along line to compute far fields                                                                                           
ff_res = ff_npts/ff_y

ff_source = sim.get_farfields(n2f_mon, ff_res, center=mp.Vector3(ff_x), size=mp.Vector3(y=ff_y))

sim.reset_meep()

num_cells = 15  # number of unit cells in super cell                                                                                                                        
sy = num_cells*gp

geometry = [mp.Block(material=glass, size=mp.Vector3(dpml+dsub,mp.inf,mp.inf), center=mp.Vector3(-0.5*sx+0.5*(dpml+dsub)))]

for n in range(num_cells):
    geometry.append(mp.Block(material=glass, size=mp.Vector3(gh,gdc*gp,mp.inf), center=mp.Vector3(-0.5*sx+dpml+dsub+0.5*gh,-0.5*sy+(n+0.5)*gp)))

symmetries=[mp.Mirror(mp.Y)]

                    cell_size=mp.Vector3(sx,sy),
                    boundary_layers=pml_layers,
                    geometry=geometry,
                    k_point=k_point,
                    sources=sources,
                    symmetries=symmetries)

n2f_mon = sim.add_near2far(fcen, 0, 1, mp.Near2FarRegion(center=mon_pt, size=mp.Vector3(y=sy)))

sim.run(until_after_sources=mp.stop_when_fields_decayed(50, mp.Ez, mon_pt, 1e-9))

ff_grating = sim.get_farfields(n2f_mon, ff_res, center=mp.Vector3(ff_x), size=mp.Vector3(y=ff_y))

ff_ys = np.linspace(-0.5*ff_y,0.5*ff_y,ff_npts)
ang = np.degrees(np.arctan(ff_ys/ff_x))
ff_enh = abs(ff_grating['Ez'][0])**2/abs(ff_source['Ez'][0])**2

plt.figure()
plt.plot(ang,ff_enh,'bo-')
plt.xlabel('angle (degrees)')
plt.ylabel('far-field enhancement (relative to vacuum)');
plt.title('number of unit cells: {}'.format(num_cells))                                                                                                                                                                                                             
plt.show()
```
In the first part of the test, which does *not* involve this PR, the unit cell is periodically tiled to create a supercell. The results are shown below and demonstrate that as the number of unit cells is increased (`1`, `7`, and `15`), the far-field spectra converges to the expected result. In fact, the ratio of the peaks for m=+1 and m=+3 is converging to the analytic result (+3)<sup>2</sup>/(+1)<sup>2</sup>=9 since the transmittance is 4/(mπ)<sup>2</sup>. The same is true for peaks m=+3/+5 (2.8) and m=+5/+7 (2.0).

![binary_grating_farfield_enhancement](https://user-images.githubusercontent.com/7152530/54471004-e24a3700-476e-11e9-9eac-5b0d6b002dc8.png)

In the second test which is for this PR, only a *single* unit cell with periodic boundaries is used. The far-field spectra is shown below. It is identical to the left plot from the figure above. This indicates that this PR is having no effect.

![unitcell_farfield_ang20](https://user-images.githubusercontent.com/7152530/54471006-ec6c3580-476e-11e9-9373-7aac507dd22f.png)

The correct result would be for the far-field spectra computed for a single unit cell using this PR to *exactly* match that of a 20 unit-cell supercell without this PR.

src/near2far.cpp
```diff
@@ -465,6 +495,17 @@ dft_near2far fields::add_dft_near2far(const volume_list *where, double freq_min,
       default: abort("invalid normal direction in dft_near2far!");
     }
 
+    for (int i = 0; i < 2; ++i) {
+      if (has_direction(v.dim, fd[i]) &&
+          boundaries[High][fd[i]] == Periodic && boundaries[Low][fd[i]] == Periodic &&
+          float(w->v.in_direction(fd[i])) == float(v.in_direction(fd[i]))) {
```
oskooi: It seems the problem lies in this particular if statement, particularly the third and final test condition involving `in_direction`, which always evaluates to `false` and thus the subsequent initialization of the arrays `periodic_d`, `periodic_n`, `period`, and `periodic_k` to non-zero values is never performed.

src/near2far.cpp
```diff
@@ -241,9 +255,21 @@ void dft_near2far::farfield_lowlevel(std::complex<double> *EH, const vec &x) {
...
+        for (int i0 = -periodic_n[0]; i0 <= periodic_n[0]; ++i0) {
+          if (periodic_d[0] != NO_DIRECTION)
+            xs.set_direction(periodic_d[0], x0.in_direction(periodic_d[0]) + i0*period[0]);
+          double phase0 = i0*periodic_k[0];
+          for (int i1 = -periodic_n[1]; i1 <= periodic_n[1]; ++i1) {
+            if (periodic_d[1] != NO_DIRECTION)
+              xs.set_direction(periodic_d[1], x0.in_direction(periodic_d[1]) + i1*period[1]);
+            double phase = phase0 + i1*periodic_k[1];
+            std::complex<double> cphase = std::polar(1.0, phase);
+            green(EH6, x, freq, eps, mu, x0, c0, f->dft[Nfreq * idx_dft + i]);
```
stevengj: whoops: this should be `xs`, not `x0`!

oskooi:
With the one fix from `src/near2far.cpp:268`, the results are now as expected.

unitcell_farfield_ang20

supercell21_farfield_ang20

oskooi:
For reference, the following script was used to generate results for the unit-cell calculation. Note that `symmetries` has been removed.

```py
import meep as mp
import math
import numpy as np

resolution = 20        # pixels/μm                                                                                                                   

sx = dpml+dsub+gh+dpad+dpml
sy = gp

sources = [mp.Source(mp.GaussianSource(fcen, fwidth=df), component=mp.Ez, center=src_pt)]

k_point = mp.Vector3(0,0,0)

glass = mp.Medium(index=1.5)

                    cell_size=mp.Vector3(sx,0,0),
                    boundary_layers=pml_layers,
                    k_point=k_point,
                    default_material=glass,
                    sources=sources)

mon_pt = mp.Vector3(0.5*sx-dpml-0.5*dpad)
n2f_mon = sim.add_near2far(fcen, 0, 1, mp.Near2FarRegion(center=mon_pt, size=mp.Vector3()))

sim.run(until_after_sources=mp.stop_when_fields_decayed(50, mp.Ez, mon_pt, 1e-9))

ff_x = 1e8
ff_angle = 20
ff_y = 2*ff_x*math.tan(math.radians(ff_angle)) # first 9 modes                                                                                       
ff_npts = 800
ff_res = ff_npts/ff_y

ff_source = sim.get_farfields(n2f_mon, ff_res, center=mp.Vector3(ff_x), size=mp.Vector3(y=ff_y))

sim.reset_meep()

                    cell_size=mp.Vector3(sx,sy,0),
                    boundary_layers=pml_layers,
                    geometry=geometry,
                    k_point=k_point,
                    sources=sources)

n2f_mon = sim.add_near2far(fcen, 0, 1, mp.Near2FarRegion(center=mon_pt, size=mp.Vector3(y=sy)))

sim.run(until_after_sources=mp.stop_when_fields_decayed(50, mp.Ez, mon_pt, 1e-9))

ff_grating = sim.get_farfields(n2f_mon, ff_res, center=mp.Vector3(ff_x), size=mp.Vector3(y=ff_y))

ff_ys = np.linspace(-0.5*ff_y,0.5*ff_y,ff_npts)
ang = np.degrees(np.arctan(ff_ys/ff_x))

plt.figure()
plt.plot(ang,abs(ff_grating['Ez'][0])**2/abs(ff_source['Ez'][0])**2,'bo-')
plt.xlabel('angle (degrees)')
plt.ylabel('far-field enhancement (relative to source)');
plt.title('unit cell with periodic boundaries using PR #769')
plt.show()
```

stevengj:
To use it, pass `nperiods=10` to `add_near2far`.

oskooi:
`nperiods` is working as shown below. However, `symmetries` is still not working.

unitcell_farfield_ang20_nperiods7

supercell15_farfield_ang20

unitcell_farfield_sym_ang20

oskooi:
updated script for unit-cell calculation which includes `nperiods` in the call to `add_near2far` based on this PR:

```py
import meep as mp
import math
import numpy as np

resolution = 20        # pixels/μm                                                                                                             

sx = dpml+dsub+gh+dpad+dpml
sy = gp

sources = [mp.Source(mp.GaussianSource(fcen, fwidth=df), component=mp.Ez, center=src_pt)]

k_point = mp.Vector3(0,0,0)

glass = mp.Medium(index=1.5)

                    cell_size=mp.Vector3(sx),
                    boundary_layers=pml_layers,
                    k_point=k_point,
                    default_material=glass,
                    sources=sources)

mon_pt = mp.Vector3(0.5*sx-dpml-0.5*dpad)
n2f_mon = sim.add_near2far(fcen, 0, 1, mp.Near2FarRegion(center=mon_pt, size=mp.Vector3()))

sim.run(until_after_sources=mp.stop_when_fields_decayed(50, mp.Ez, mon_pt, 1e-9))

ff_x = 1e8
ff_angle = 21
ff_y = 2*ff_x*math.tan(math.radians(ff_angle)) # first 9 modes                                                                                 
ff_npts = 800
ff_res = ff_npts/ff_y

ff_source = sim.get_farfields(n2f_mon, ff_res, center=mp.Vector3(ff_x), size=mp.Vector3(y=ff_y))

sim.reset_meep()

# symmetries=[mp.Mirror(mp.Y)]                                                                                                                 
symmetries = []

                    cell_size=mp.Vector3(sx,sy),
                    boundary_layers=pml_layers,
                    geometry=geometry,
                    k_point=k_point,
                    sources=sources,
                    symmetries=symmetries)

nd = 10
n2f_mon = sim.add_near2far(fcen, 0, 1, mp.Near2FarRegion(center=mon_pt, size=mp.Vector3(y=sy)), nperiods=nd)

sim.run(until_after_sources=mp.stop_when_fields_decayed(50, mp.Ez, mon_pt, 1e-9))

ff_grating = sim.get_farfields(n2f_mon, ff_res, center=mp.Vector3(ff_x), size=mp.Vector3(y=ff_y))

ff_ys = np.linspace(-0.5*ff_y,0.5*ff_y,ff_npts)
ang = np.degrees(np.arctan(ff_ys/ff_x))

plt.figure()
plt.plot(ang,abs(ff_grating['Ez'][0])**2/abs(ff_source['Ez'][0])**2,'bo-')
plt.xlabel('angle (degrees)')
plt.ylabel('far-field enhancement (relative to source)');
plt.title('unit cell with periodic boundaries, nperiods = 7, using PR #769 \n without symmetries')
plt.show()
```

oskooi:
Working now: results are identical with and without `symmetries`.

unitcell_farfield_ang21_nperiods10

unitcell_farfield_ang21_nperiods10_nosym

PR #785, opened 2019-03-26
Send master_printf output to Python's stdout

### LINKED ISSUE #782: I/O callback for Python I/O integration
It would be nice to have the `master_printf` function (used for all of our C++ stdout printing) have the option of printing to the Python stdout stream, since this is the only way for output to be captured by a Jupyter notebook.   One way to do this would be to:

1. Add a global callback function, e.g. `void (*master_printf_callback)(const char *s, void *data) = NULL` and a global `void *master_printf_callback_data = NULL`, both of which default to `NULL`.

2. If `master_printf_callback != NULL`, have `master_printf` call [`vasprintf`](https://linux.die.net/man/3/vasprintf) to print to a string `s` instead of `vprintf` to print to `stdout`.   Then call `master_printf_callback(s, master_printf_callback_data)`.

3. In the Python front-end, have it set `master_printf_callback` to a function that calls `PySys_WriteStdout("%s", s)` to output to Python's `sys.stdout `

ChristopherHogan:
Closes #782. This causes all output to appear within a Jupyter notebook.  However, it's not perfect yet, as anything printed from MPB or libctl still goes to C stdout, and hence, the console.
@stevengj, @oskooi

### DIFF
--- a/python/meep.i
+++ b/python/meep.i
@@ -37,6 +37,7 @@
 namespace meep {
     size_t dft_chunks_Ntotal(dft_chunk *dft_chunks, size_t *my_start);
     typedef std::complex<double> (*amplitude_function)(const vec &);
+    void (*master_printf_callback)(const char *s);
 }
 
 #ifdef HAVE_MPB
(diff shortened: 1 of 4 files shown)


stevengj:
MPB has a similar function called `mpi_one_printf`, so in principle we could do the same thing there.

PR #787, opened 2019-03-27
include interpolation weights in DFT slice output

### LINKED ISSUE #784: discretization artifacts in get_dft_array
Following #769 which was recently merged, the [near to far field transformation](https://meep.readthedocs.io/en/latest/Python_User_Interface/#near-to-far-field-spectra) feature supports periodic boundaries. The following test (which will eventually be added to [`python/tests`](https://github.com/NanoComp/meep/tree/master/python/tests)), based on a [binary grating](https://meep.readthedocs.io/en/latest/Python_Tutorials/Mode_Decomposition/#diffraction-spectrum-of-a-binary-grating), attempts to verify this property by computing the E<sub>z</sub>  "far" fields (i.e. 40 wavelengths away) along a line spanning the entire length of the cell using two methods in a single simulation: (1) `add_near2far` and (2) `add_dft_fields`. In the limit of increasing `resolution` and `nperiods` (for `add_near2far`), the L<sub>2</sub>-norm of the difference of the magnitude of the two far-field datasets should monotonically converge to zero. However, this is not the case due to the `dft_fields` as demonstrated below. The `dft_near2far` fields are fine.

The following is a schematic of the simulation geometry and script.

n2f_dft_grating_geometry

```py
import meep as mp
import math
import numpy as np

resolution = 50        # pixels/μm                                                                                                                              

dpad = 20.0            # padding between grating and PML                                                                                                        

sx = dpml+dsub+gh+dpad+dpml
sy = gp

glass = mp.Medium(index=1.5)

k_point = mp.Vector3(0,0,0)

symmetries=[mp.Mirror(mp.Y)]

                    cell_size=mp.Vector3(sx,sy),
                    boundary_layers=pml_layers,
                    geometry=geometry,
                    k_point=k_point,
                    sources=sources,
                    symmetries=symmetries)

nd = 20
n2f_pt = mp.Vector3(-0.5*sx+dpml+dsub+gh+1.0)
n2f_obj = sim.add_near2far(fcen, 0, 1, mp.Near2FarRegion(center=n2f_pt, size=mp.Vector3(y=sy)), nperiods=nd)

dft_pt = mp.Vector3(0.5*sx-dpml)
dft_obj = sim.add_dft_fields([mp.Ez], fcen, fcen, 1, center=dft_pt, size=mp.Vector3(y=sy))

sim.run(until_after_sources=300)

n2f_Ez = sim.get_farfields(n2f_obj, resolution, center=dft_pt, size=mp.Vector3(y=sy))
dft_Ez = sim.get_dft_array(dft_obj, mp.Ez, 0)

norm_Ez = LA.norm(np.absolute(n2f_Ez['Ez'][0])-np.absolute(dft_Ez[1:-1]))

plt.figure()
plt.plot(np.absolute(n2f_Ez['Ez'][0]),'bo-',label='near2far')
plt.plot(np.absolute(dft_Ez),'rs-',label='dft_fields')
plt.xlabel('y-coordinate index')
plt.ylabel(r'|E$_z$|')
plt.legend(loc='lower right')
plt.title(r"resolution = {}, nperiods = {}, |error(|E$_z$|)|$_2$ = {:.5f}".format(resolution, nd, norm_Ez))
plt.show()
```

Note the use of `symmetries`. The results are identical to those with no `symmetries`.

As the `resolution` is gradually increased (`nperiods` is fixed at `20`), the magnitude of the fields computed by `add_dft_fields` (red curve) vary by nearly a factor of `2` and thus so does the norm. This is shown in the following figures for six values of the `resolution`: `30`, `40`, `50`, `60`, `70`, and `80`. The norm values are `4.02242`, `39.28917`, `2.47782`, `48.46916`, `1.63235`, and `56.27043`. The fields computed by `get_farfields` (blue curve) do not vary strongly with the `resolution`.

n2f_dft_res30_nd20

n2f_dft_res40_nd20

n2f_dft_res50_nd20

n2f_dft_res60_nd20

n2f_dft_res70_nd20

n2f_dft_res80_nd20

stevengj:
This ensures that empty dimensions are collapsed correctly.  I think this should fix #784 — what was happening is that at "unlucky" resolutions where the DFT slice fell between two pixels, the `collapse_empty_dims` was summing over two pixels in the empty direction, but without the interpolation weights this effectively doubled the field value, resulting in the factor-of-two oscillation observed in #784.

~~One downside of this PR is that, at the edges of the output volume, the edge pixels might be multiplied by an integration weight factor (between 0 and 1).   It might be nice to only include the interpolation weights for empty dimensions, but not for edge pixels, but that would require some additional work (e.g. a different `IVEC_LOOP_WEIGHT` function or a modified `loop_in_chunks`).~~ *Update:* Now the PR only includes the empty-dimension interpolation weights, and not the edge integration weights.

### DIFF
--- a/src/dft.cpp
+++ b/src/dft.cpp
@@ -731,7 +731,7 @@ cdouble dft_chunk::process_dft_component(int rank, direction *ds, ivec min_corne
      a copy of these with the weights for non-empty dimensions set to 1. */
   vec s0i(s0), s1i(s1), e0i(e0), e1i(e1);
   LOOP_OVER_DIRECTIONS(fc->gv.dim, d) {
-    if (empty_dim[d]) {
+    if (!empty_dim[d]) {
       s0i.set_direction(d, 1.0);
       s1i.set_direction(d, 1.0);
       e0i.set_direction(d, 1.0);
--- a/src/dft.cpp
+++ b/src/dft.cpp
@@ -39,6 +39,7 @@ struct dft_chunk_data { // for passing to field::loop_in_chunks as void*
   double dt_factor;
   bool include_dV_and_interp_weights;
   bool sqrt_dV_and_interp_weights;
+  bool empty_dim[5];
   dft_chunk *dft_chunks;
 };
 
@@ -93,6 +94,8 @@ dft_chunk::dft_chunk(fields_chunk *fc_, ivec is_, ivec ie_, vec s0_, vec s1_, ve
   dft = new complex<realnum>[N * Nomega];
   for (size_t i = 0; i < N * Nomega; ++i)
     dft[i] = 0.0;
+  for (int i = 0; i < 5; ++i)
+    empty_dim[i] = data->empty_dim[i];
 
   next_in_chunk = fc->dft_chunks;
   fc->dft_chunks = this;
@@ -170,6 +173,10 @@ dft_chunk *fields::add_dft(component c, const volume &where, double freq_min, do
   data.dt_factor = dt / sqrt(2.0 * pi);
   data.include_dV_and_interp_weights = include_dV_and_interp_weights;
   data.sqrt_dV_and_interp_weights = sqrt_dV_and_interp_weights;
+  data.empty_dim[0] = data.empty_dim[1] = data.empty_dim[2] = data.empty_dim[3] = data.empty_dim[4] = false;
+  LOOP_OVER_DIRECTIONS(where.dim, d) {
+    data.empty_dim[d] = where.in_direction(d) == 0;
+  }
   data.dft_chunks = chunk_next;
   loop_in_chunks(add_dft_chunkloop, (void *)&data, where, use_centered_grid ? Centered : c);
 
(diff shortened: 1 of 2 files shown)


oskooi:
Running this on the test in #784 for a `resolution` of `40` (which was producing incorrect results), the "interior" points returned by `get_dft_array` are now correct. It would be nice to also correct the endpoints.

n2f_dft_res40_nd20

oskooi:
For some reason, the latest commit seems to not have fixed the endpoints and the interior points are just as they were in #784 (i.e. off by an overall scale factor).

n2f_dft_res40_nd20

stevengj:
@oskooi, should be fixed now.

oskooi:
Indeed everything is working correctly now (i.e. `get_dft_array` is no longer resolution-dependent and the endpoints are fixed). The norm of the relative difference in the two fields is decreasing (mostly monotonically). Shown below are results for resolutions of `40`, `50`, `60`, and `70` with norms of `3.31022`, `2.47782`, `1.49521`, and `1.63235`.

n2f_dft_res40_nd20

n2f_dft_res50_nd20

n2f_dft_res60_nd20

n2f_dft_res70_nd20

PR #788, opened 2019-03-28
test for near2far with periodic boundaries

oskooi:
Adds a test to the `python/tests` suite based on the binary-grating example in #784: verifies that the norm of the difference in the far-fields computed using `add_near2far` and `add_dft_fields` converges to zero with increasing `resolution`. Runs on a single Kaby Lake 4.2Ghz processor in ~`40s`.

### DIFF
--- a/python/Makefile.am
+++ b/python/Makefile.am
@@ -92,6 +92,7 @@ TESTS =                                   \
     $(MODE_COEFFS_TEST)                   \
     $(MODE_DECOMPOSITION_TEST)            \
     $(TEST_DIR)/multilevel_atom.py        \
+    $(TEST_DIR)/n2f_periodic.py           \
     $(TEST_DIR)/oblique_source.py         \
     $(TEST_DIR)/physical.py               \
     $(TEST_DIR)/pw_source.py              \
--- a/python/tests/n2f_periodic.py
+++ b/python/tests/n2f_periodic.py
@@ -57,7 +57,7 @@ class TestNear2FarPeriodicBoundaries(unittest.TestCase):
             n2f_Ez = sim.get_farfields(n2f_obj, res[j], center=dft_pt, size=mp.Vector3(y=sy))
             dft_Ez = sim.get_dft_array(dft_obj, mp.Ez, 0)
 
-            norm[j] = LA.norm(np.absolute(n2f_Ez['Ez'][0])-np.absolute(dft_Ez[1:-1]))
+            norm[j] = LA.norm(n2f_Ez['Ez']-dft_Ez[1:-1])
             print("norm:, {}, {:.5f}".format(res[j],norm[j]))
             sim.reset_meep()        
 
--- a/python/tests/n2f_periodic.py
+++ b/python/tests/n2f_periodic.py
@@ -1,0 +1,68 @@
+import unittest
+import meep as mp
+import math
+import numpy as np
+from numpy import linalg as LA
+
+class TestNear2FarPeriodicBoundaries(unittest.TestCase):
+
+    def test_nea2far_periodic(self):
+        dpml = 1.0             # PML thickness
+        dsub = 3.0             # substrate thickness
+        dpad = 20.0            # padding between grating and PML
+        gp = 10.0              # grating period
+        gh = 0.5               # grating height
+        gdc = 0.5              # grating duty cycle
+
+        sx = dpml+dsub+gh+dpad+dpml
+        sy = gp
+
+        pml_layers = [mp.PML(thickness=dpml,direction=mp.X)]
+
+        wvl = 0.5              # center wavelength
+        fcen = 1/wvl           # center frequency
+        df = 0.05*fcen         # frequency width
+
+        src_pt = mp.Vector3(-0.5*sx+dpml+0.5*dsub)
+        sources = [mp.Source(mp.GaussianSource(fcen, fwidth=df), component=mp.Ez, center=src_pt, size=mp.Vector3(y=sy))]
+
+        glass = mp.Medium(index=1.5)
+        geometry = [mp.Block(material=glass, size=mp.Vector3(dpml+dsub,mp.inf,mp.inf), center=mp.Vector3(-0.5*sx+0.5*(dpml+dsub))),
+                    mp.Block(material=glass, size=mp.Vector3(gh,gdc*gp,mp.inf), center=mp.Vector3(-0.5*sx+dpml+dsub+0.5*gh))]
+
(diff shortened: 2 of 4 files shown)


stevengj:
I'm not entirely happy with the convergence rate you saw in #784 — it seems you are seeing only 1st-order convergence?

oskooi:
The convergence rate appears to be somewhere between first- and second-order as shown in the figure below for several values of `nperiods`.

n2f_norm

Strangely, this test seems to be failing randomly on Travis with the error message:

```
ERROR: test_nea2far_periodic (__main__.TestNear2FarPeriodicBoundaries)
  File "../../../python/tests/n2f_periodic.py", line 60, in test_nea2far_periodic
    norm[j] = LA.norm(np.absolute(n2f_Ez['Ez'][0])-np.absolute(dft_Ez[1:-1]))
IndexError: too many indices for array
```
This is because the size of the `dft_fields` array returned by `get_dft_array` is inconsistent and does not match the `near2far` array for certain resolutions (in this case `30`). However, everything runs fine on my local machine. Perhaps this is a rounding-error related bug in #787 which was merged yesterday?

ChristopherHogan:
Valgrind shows several errors when running `n2f_periodic.py`. I'm looking into it.

stevengj:
Looks first-order to me.  Would be good to check the near2far and dft-array fields separately — i.e. plot the errors of both as a function of resolution compared to a high-resolution reference.

oskooi:
The `dft_fields` are converging to *first* order and the `near2far` to second order as shown in the following figure. This might indicate that the interpolation in `get_dft_array` is still only first order.
n2f_dft_conv

For reference, the script that was used to generate the data as well as the figure is shown below.

```py
import meep as mp
import math
import numpy as np

dpad = 20.0            # padding between grating and PML                                                                                                                   

sx = dpml+dsub+gh+dpad+dpml
sy = gp

glass = mp.Medium(index=1.5)

k_point = mp.Vector3(0,0,0)

n2f_pt = mp.Vector3(-0.5*sx+dpml+dsub+gh+1.0)
ff_pt = mp.Vector3(0.5*sx-dpml)
ff_res = 10

def grating_ff(res):
                        cell_size=mp.Vector3(sx,sy),
                        boundary_layers=pml_layers,
                        geometry=geometry,
                        k_point=k_point,
                        sources=sources,
                        symmetries=symmetries)

    n2f_obj = sim.add_near2far(fcen, 0, 1, mp.Near2FarRegion(center=n2f_pt, size=mp.Vector3(y=sy)), nperiods=10)
    dft_obj = sim.add_dft_fields([mp.Ez], fcen, fcen, 1, center=ff_pt, size=mp.Vector3(y=sy))
    sim.run(until_after_sources=300)
    n2f_Ez = sim.get_farfields(n2f_obj, ff_res, center=ff_pt, size=mp.Vector3(y=sy))
    dft_Ez = sim.get_dft_array(dft_obj, mp.Ez, 0)
    sim.reset_meep()
    return n2f_Ez, dft_Ez

res_ref = 30
n2f_ref, dft_ref = grating_ff(res_ref)
y_ref = np.linspace(-0.5*sy,0.5*sy,res_ref*sy+2)

r = range(20,110,10)
norm_n2f = np.empty(len(r))
norm_dft = np.empty(len(r))
for j in range(len(r)):
    n2f_res, dft_res = grating_ff(r[j])
    y_res = np.linspace(-0.5*sy,0.5*sy,r[j]*sy+2)
    norm_n2f[j] = LA.norm(np.absolute(n2f_res['Ez'][0])-np.absolute(n2f_ref['Ez'][0]))
    norm_dft[j] = LA.norm(np.absolute(dft_res)-np.absolute(np.interp(y_res,y_ref,dft_ref)))
    print("norm:, {}, {:.6f}, {:.6f}".format(r[j],norm_n2f[j],norm_dft[j]))

plt.figure(dpi=150)
plt.loglog(r,norm_dft,'ro-',label='dft_fields')
plt.loglog(r,norm_n2f,'bo-',label='near2far')
plt.loglog(r,np.divide(2e2,r),'k--',label='linear reference')
plt.loglog(r,np.divide(5e2,r**2),'k-',label='quadratic reference')

plt.xlabel('resolution')
plt.ylabel('error in near2far or dft_fields')
plt.title('error in far fields of a binary-grating unit cell')
plt.legend(loc='lower left')
plt.show()
```

stevengj:
Why is `res_ref` only 30?   Shouldn't it be much higher than any of the tested `r` values?

Your `norm_dft[j]` value has a number of points that grows linearly with resolution, which cancels one order of convergence I think.   I think you want the norm corresponding to ∫|res-ref|²dy, which means you need to multiply by a factor of `dy = 1/resolution`.

(Also, the `np.absolute` calls shouldn't be there — you should just take the `norm` of the difference to make sure the phase is correct too.)

oskooi:
`res_ref=200` was used to generate the results which is double the maximum resolution of `100`. (`res_ref=30` was placed there by accident when transcribing.)

oskooi:
Following your suggestion of removing `np.absolute` and multiplying `norm_dft` by `1/resolution` produces the following results:

n2f_dft_conv2

The convergence rate for both methods is now second order (or actually just slightly greater). These results were produced using:

```py
res_ref = 200
n2f_ref, dft_ref = grating_ff(res_ref)
y_ref = np.linspace(-0.5*sy,0.5*sy,res_ref*sy+2)

r = range(20,110,10)
norm_n2f = np.empty(len(r))
norm_dft = np.empty(len(r))
for j in range(len(r)):
    n2f_res, dft_res = grating_ff(r[j])
    y_res = np.linspace(-0.5*sy,0.5*sy,r[j]*sy+2)
    norm_n2f[j] = LA.norm(n2f_res['Ez'][0]-n2f_ref['Ez'][0])
    norm_dft[j] = LA.norm(dft_res-np.interp(y_res,y_ref,dft_ref))/r[j]
    print("norm:, {}, {:.6f}, {:.6f}".format(r[j],norm_n2f[j],norm_dft[j]))
```

`n2f_res['Ez']` is an array of size `(1,100)` (i.e. it is *not* collapsed). This is why it is necessary to use `n2f_res['Ez'][0]` to obtain a true 1d array of size `(100,)`. The norm is therefore computed for all 100 points.

oskooi:
The convergence rate of the difference in the `near2far` and `dft_fields` (with the `np.absolute` removed) is between first and second order as shown below.

n2f_norm2

ChristopherHogan:
I think `dft.cpp:986` should be `reduced_stride[nd] = 0;` instead of `reduced_stride[nd - 1] = 0;`

PR #790, opened 2019-03-28
fixes for parallel get_field, get_eps, etcetera

### LINKED ISSUE #685: get_field_point does not support parallel use
The [current implementation](https://github.com/NanoComp/meep/blob/master/src/monitor.cpp#L198-L219) of [`get_field_point`](https://meep.readthedocs.io/en/latest/Python_User_Interface/#field-computations) does not include parallel support (i.e., MPI). A workaround involving `integrate_field_function` was previously proposed by @stevengj in the [mailing list](https://www.mail-archive.com/meep-discuss@ab-initio.mit.edu/msg04548.html) but has yet to be implemented.

stevengj:
Fixes #685.

Actually, there was some parallel code in the `fields::get_field` implementation — see the [`broadcast` call in `fields_chunk::get_field`](https://github.com/NanoComp/meep/blob/a5308c11167503fcfff5467af739ecef583bebbc/src/monitor.cpp#L187) which dates way back to e8d6a0458d98eee3e11e969822b68dd180167e25.  However, I think it is incorrect — it should be broadcasting from the process that owns the data, not from `n_proc()`.

This PR updates the code and hopefully is correct now, though it still needs tests.  In particular, in our usual style it now does all of the communication in the `fields::` methods and not in `fields_chunk`.  Also, the routines now take an optional `parallel=true` argument — you can pass `false` to omit the communication, which is useful in reducing the communications in the `vec` methods that do interpolation, since they can do the communication only *after* the interpolation rather than for each interpolated point.

I also removed some calls to a sorting routine before interpolation.  If I recall correctly, @droundy inserted these calls to make sure that interpolation gave *exactly* the same results (including rounding errors) regardless of the chunk divisions or parallelization.  We stopped guaranteeing this a while back because it was too much trouble for more complicated features.  It should still give the same results up to roundoff errors.

@oskooi, would be good to set up a test for this.  A simple thing to do is to call `get_field` and `get_eps` for a bunch of points (some in the interior and some right on the boundary between chunks) for both serial and parallel runs of the same structure and check that they match to machine precision.


oskooi:
The following test computes the difference in the values returned by `get_field_point` and `get_epsilon_point` at 19 equally-spaced points along a line through the 2d cell with sinusoidal ε with reference values from a serial run. If the difference is larger than 1e-10, it is shown in the output.

The following figure shows the permittivity and field profile. The green dotted line indicates the region over the ε or E<sub>z</sub> is computed.

sinusoid_eps_ez

```py
import meep as mp
import numpy as np
import math

sxy = 6                 # cell size                                                                                                                                                                 
dpml = 1                # thickness of PML                                                                                                                                                          

def sinusoid(p):
    r = (p.x**2+p.y**2)**0.5
    return mp.Medium(index=1.0+math.sin(2*math.pi*r)**2)

sinusoid.do_averaging = True

                     size=mp.Vector3(sxy,sxy),
                     material=sinusoid)]

src = [mp.Source(mp.GaussianSource(1.0, fwidth=0.1),
                 component=mp.Ez,
                 center=mp.Vector3())]

                    geometry=geometry,
                    subpixel_tol=1e-4,
                    subpixel_maxeval=1000,
                    sources=src,
                    k_point=mp.Vector3(),
                    resolution=20,
                    symmetries=[mp.Mirror(mp.X),mp.Mirror(mp.Y)],

x = np.linspace(0.865692,2.692867,19)

# reference values for Ez and epsilon from serial run                                                                                                                                               
ez_ref = [  -0.0001935467,
            0.0001230734,
            0.0013052441,
            0.0031842283,
            0.0036496679,
            0.0025336365,
            0.0014203228,
            0.0005547447,
            -0.0003031574,
            -0.0007824044,
            -0.0007147878,
            -0.0005227611,
            -0.0003401799,
            -0.0000800129,
            0.0001161430,
            0.0001209453,
            0.0000700215,
            0.0000280849,
            0.0000054367 ]

eps_ref = [  1.6541850413,
             2.1230242102,
             2.7331921116,
             3.3724323885,
             3.8204408186,
             3.8334517538,
             3.3209494246,
             2.4715212877,
             1.6593727598,
             1.1664608262,
             1.0642874690,
             1.3884526280,
             2.2104433509,
             3.2984497712,
             3.8905429818,
             3.3925394850,
             2.2233197108,
             1.3125774576,
             1.0529948731 ]

for j in range(x.size):
    ez_diff = np.real(sim.get_field_point(mp.Ez, mp.Vector3(x[j],-0.394862)))-ez_ref[j]
    if abs(ez_diff) > 1e-10:
        print("ez_diff:, {}, {:.10f}".format(j,ez_diff))

    eps_diff = sim.get_epsilon_point(mp.Vector3(x[j],2.967158))-eps_ref[j]
    if abs(eps_diff) > 1e-10:
	print("eps_diff:, {}, {:.10f}".format(j,eps_diff))
```

Prior to this PR, running a parallel simulation with 2 and 3 processors reveals inconsistencies for `get_epsilon_point` (but *not* `get_field_point`):

**2 processes**
```
-----------
Initializing structure...
Halving computational cell along direction x
Working in 2D dimensions.
Computational cell is 6 x 6 x 0 with resolution 20
     block, center = (0,0,0)
          size (6,6,0)
time for set_epsilon = 2.0467 s
-----------
run 0 finished at t = 200.0 (8000 timesteps)
eps_diff:, 9, 0.0006610281
```
**3 processes**
```
Using MPI version 3.1, 3 processes
-----------
Initializing structure...
Halving computational cell along direction x
Working in 2D dimensions.
Computational cell is 6 x 6 x 0 with resolution 20
     block, center = (0,0,0)
          size (6,6,0)
time for set_epsilon = 1.35169 s
-----------
run 0 finished at t = 200.0 (8000 timesteps)
eps_diff:, 13, -0.0002919241
```
This behavior is expected since currently `get_field_point`/`get_epsilon_point` does not support MPI.

However, what is unexpected is what happens when this PR is used for a *serial* run:

```
-----------
Initializing structure...
Halving computational cell along direction x
Working in 2D dimensions.
Computational cell is 6 x 6 x 0 with resolution 20
     block, center = (0,0,0)
          size (6,6,0)
time for set_epsilon = 2.09689 s
-----------
run 0 finished at t = 200.0 (8000 timesteps)
ez_diff:, 0, -0.0000003666
ez_diff:, 1, 0.0000002258
ez_diff:, 2, 0.0000025511
ez_diff:, 3, 0.0000062830
ez_diff:, 4, 0.0000072290
ez_diff:, 5, 0.0000050158
ez_diff:, 6, 0.0000027763
ez_diff:, 7, 0.0000009869
ez_diff:, 8, -0.0000007795
ez_diff:, 9, -0.0000016780
eps_diff:, 9, 0.0006610281
ez_diff:, 10, -0.0000014167
ez_diff:, 11, -0.0000009126
ez_diff:, 12, -0.0000004483
ez_diff:, 13, 0.0000000429
ez_diff:, 14, 0.0000002728
ez_diff:, 15, 0.0000002057
ez_diff:, 16, 0.0000001054
ez_diff:, 17, 0.0000000370
ez_diff:, 18, 0.0000000057
```
It seems that *all* the values returned by `get_field_point` (but *not* `get_epsilon_point`) differ from their expected values. This discrepancy persists for parallel runs as well.

stevengj:
Since you are using subpixel smoothing, you might be running into a problem due to the fact that some pixes for `do_averaging` were recently merged into master but were not in this PR?

I've rebased this PR against master, but I would also recommend testing without subpixel smoothing, as the `do_averaging` feature is still experimental and may be buggy.

oskooi:
The following modified test, which does not involve subpixel smoothing, passes for the master branch as well as this PR.

```py
from __future__ import division

import unittest
import meep as mp
import numpy as np
import math

class TestGetPoint(unittest.TestCase):

    def test_get_point(self):
        sxy = 6                 # cell size                                                                                                                                                           
        dpml = 1                # thickness of PML                                                                                                                                                    

        def sinusoid(p):
            r = (p.x**2+p.y**2)**0.5
            return mp.Medium(index=1.0+math.sin(2*math.pi*r)**2)

                             size=mp.Vector3(sxy,sxy),
                             material=sinusoid)]

        src = [mp.Source(mp.GaussianSource(1.0, fwidth=0.1),
                         component=mp.Ez,
                         center=mp.Vector3())]

                            geometry=geometry,
                            sources=src,
                            k_point=mp.Vector3(),
                            resolution=20,
                            symmetries=[mp.Mirror(mp.X),mp.Mirror(mp.Y)],

        ## reference values for Ez and epsilon from serial run                                                                                                                                        
        ez_ref = [ -0.0002065983,
                   -0.0001954795,
                   -0.0000453570,
                   0.0000311267,
                   -0.0000121473,
                   -0.0000410032,
                   -0.0000341301,
                   -0.0000275021,
                   -0.0000397990,
                   -0.0000351730,
                   0.0000079602,
                   0.0000227437,
                   -0.0001092821,
                   -0.0002202751,
                   -0.0001408186,
                   0.0006325076,
                   0.0024890489,
                   0.0027476069,
                   0.0014815873,
                   0.0004714913,
                   -0.0004332029,
                   -0.0007101315,
                   -0.0003818581,
                   -0.0000748507,
                   0.0001408819,
                   0.0001119776,
                   0.0000395008,
                   0.0000078844,
                   -0.0000010431 ]

        eps_ref = [ 1.6458346134,
                    1.2752837068,
                    1.0974010956,
                    1.0398089537,
                    1.0465784716,
                    1.0779924737,
                    1.1059439286,
                    1.1135579291,
                    1.0971979186,
                    1.0653178566,
                    1.0391657283,
                    1.0513779677,
                    1.1466009312,
                    1.3882154483,
                    1.8496939317,
                    2.5617731415,
                    3.3788212533,
                    3.9019494270,
                    3.6743431894,
                    2.7285622651,
                    1.6635165033,
                    1.0891237010,
                    1.1485969863,
                    1.9498398061,
                    3.3100416367,
                    3.9038800599,
                    2.8471862395,
                    1.4742605488,
                    1.0370162714 ]

        x = np.linspace(-0.865692,2.692867,29)
        for j in range(x.size):
            self.assertAlmostEqual(np.real(sim.get_field_point(mp.Ez, mp.Vector3(x[j],-0.394862))),ez_ref[j],places=10)
            self.assertAlmostEqual(sim.get_epsilon_point(mp.Vector3(x[j],2.967158)),eps_ref[j],places=10)

if __name__ == '__main__':
    unittest.main()
```

stevengj:
Actually, now that I look at it, the old code called `broadcast(n_proc(), ...)`, which is correct as `n_proc()` is the process that owns the chunk.  The current PR does a lot less communication, however, since it only does the `sum_to_all` *after* interpolation.

PR #600, opened 2018-11-09
Adjoint based optimization

HomerReid:
WIP: preliminary interface specification, documentation, python code for adjoint-based optimization

[AdjointDocumentation.pdf](https://github.com/stevengj/meep/files/2567194/AdjointDocumentation.pdf)

PR #795, opened 2019-03-29
Adjoint solver

HomerReid:
Adjoint solver module.

The [Official documentation](https://homerreid.github.io/meep-adjoint-documentation/AdjointSolver/Overview/) aspires to be thorough and comprehensive, though currently has some incomplete sections. I will create a separate PR for the source of this documentation, but it uses a mkdocs theme that is not immediately compatible with the existing MEEP doc tree.  

A sketchier documentation page is the quickstart, included in this PR as a single file within the existing 
MEEP doc tree: [AdjointQuickStart](https://github.com/HomerReid/meep/blob/adjoint_solver/doc/docs/Python_Tutorials/AdjointQuickStart.md).

Closes #600.


python/Makefile.am
```diff
@@ -166,23 +168,40 @@ endif
...
+                  $(srcdir)/adjoint/Objective.py        	\
+                  $(srcdir)/adjoint/OptimizationProblem.py	\
+                  $(srcdir)/adjoint/ParallelDesignTester.py	\
+                  $(srcdir)/adjoint/Visualization.py
+
+######################################################################
+# finally, specification of what gets installed in the meep python
+# module directory of the python site-packages installation
+# Q: Why is this not redundant since e.g. the HL_IFACE files should
+#    already be installed by virtue of being in pkgpython_PYTHON
```
ChristopherHogan: This section assembles the `meep` package hierarchy in the build directory so that the normal 
```bash
make
make check
make install
```
workflow tests the built package before installing anything. If there were any adjoint solver tests, `make check` would currently be unable to import the `meep.adjoint` module.

python/adjoint/Objective.py
```diff
@@ -0,0 +1,661 @@
...
+    # mp.get_dft_array(), but 'zero-padded:' when the low-level DFT object
+    # does not have data for the requested component (perhaps because it vanishes
+    # identically by symmetry), this routine returns an array of the expected
+    # dimensions with all zero entries, instead of a rank-0 array that prints
+    # out as a single preposterously large or small floating-point number,
+    # which is the not-very-user-friendly behavior of mp.get_dft_array().
+    ######################################################################
+    def get_EH_slice(self, c, nf=0):
+        EH = self.sim.get_dft_array(self.dft_obj, c, nf)
+        return EH if np.ndim(EH)>0 else 0.0j*np.zeros(self.slice_dims)
```
stevengj: It seems like this is a bugfix that should be incorporated into `get_dft_array` itself?

python/adjoint/Objective.py
```diff
@@ -0,0 +1,661 @@
...
+    #########################################################
+    def eval_fq(self, nf=0):
+        return self.obj_func.get_fq(self.dft_cells,nf=nf)
+
+    def eval_gradf(self, nf=0):
+        cell=self.dft_cells[-1] # design cell
+        EH_forward=cell.get_EH_slices(nf,label='forward')
+        EH_adjoint=cell.get_EH_slices(nf) # no label->current simulation
+        self.dfdEps=np.sum( [EH_forward[nc]*EH_adjoint[nc]
+                               for nc,c in enumerate(cell.components) if c in Exyz], 0 )
```
stevengj: Doesn't a comprehension like this produce a Python `list` and not a numpy array?  On a Python list, my recollection is that you might as well do `sum` … `np.sum` offers no advantage and may actually be slower.  Not that it matters here since the dominant cost will be building the list in the first place.

python/adjoint/Basis.py
```diff
@@ -0,0 +1,440 @@
...
+
+    def __call__(self, p=[0.0,0.0]):
+        b=np.zeros(self.dim)
+        for idx,val in self.contributors(p):
+           b[idx]=val
+        return b
+
+    def eval(self, p=[0.0,0.0], beta_vector=None):
+        bv = beta_vector if beta_vector else self.beta_vector
+        return sum( [ bv[idx]*val for idx,val in self.contributors(p) ] )
```
stevengj: Eventually we will want a native C++ implementation of this sort of finite-element basis (or just bilinear interpolation on a grid, rather than triangles), since for optimization you will eventually want to use something like this `eval` for a material function, but a pure-Python implementation like this will be pretty slow when setting up ε on a lot of grid points.   (Especially in practical cases one will want even more complicated basis functions, like the SIMP method with a smoothing filter, in order to do optimization with a "binarization" pass and a resolution constraint.)

stevengj:
I want to be cautious about incorporating finite-element modules for interpolation of the degrees of freedom from an arbitrary mesh (as opposed to simple bilinear interpolation from a Cartesian grid), for three reasons:

1. This adds a flexibility that one is unlikely to need: in practice, topology optimization in photonics is oriented towards fabrication by lithography or 3d printing, and both of these have uniform spatial resolution and have degrees of freedom lying within box-like domains.

2. Practical topology optimization requires one to use additional nonlinear filtering operations in order to constrain the minimum feature size, "binary-ize" the structure, etcetera (e.g. impose connectivity constraints).   This seems much easier to do an a uniform Cartesian grid.

smartalecH:
Regarding the choice of a basis mesh/method:

I've run various adjoint experiments with the newer adjoint module that uses fenics to mesh and interpolate, trying to verify that the adjoint gradient approaches the finite difference gradient. It seems that the gradient w.r.t. the design variables (p) is very sensitive to the projection operation needed to go from ∂J / ∂ϵ to ∂J / ∂p (where J is the objective function).

Is this behavior expected (I would assume so since the projection is a least-squares estimation problem)? Would certain bases more closely approximate the finite-difference method better than others?

@stevengj @oskooi

PR #796, opened 2019-03-29
Adjoint documentation

HomerReid:
Markdown source for the [official meep adjoint module documentation](https://homerreid.github.io/meep-adjoint-documentation/AdjointSolver/Overview/), organized here as part of the MEEP doc tree, but not compatible with the existing MEEP documentation build due a different markdown theme.


stevengj:
How hard would it be to just use the same stylesheet at the rest of the docs?  Since most of this PR is in markdown, isn't it stylesheet-agnostic?

stevengj:
Obsoleted by #1167.

Issue #759, opened 2019-03-08
array-slice interpolation


stevengj:
Currently `array_slice.cpp` sets `snap_empty_dims=true` when calling `loop_in_chunks`.  This means that slice locations are "rounded" to the nearest pixel rather than being interpolated.  (See also the discussion in #661.)

It would be nice to do proper interpolation here:

1. In `array_slice.cpp`, pass `snap_empty_dims=false` when calling `loop_in_chunks`

2. Include the interpolation weights in the fields stored in the slice array.

3. Call `collapse_empty_dimensions` from `dft.cpp` to collapse empty dimensions to the interpolated result.

stevengj:
Once this is resolved we can fix the get_array_metadata function, which now has a hard time because DFT slices and array slices are different so the same metadata can't necessarily be correct for both.

PR #811, opened 2019-04-10
Revised array metadata

### LINKED ISSUE #658: get_dft_array_metadata does not support symmetry objects
There is a bug in the recently added `get_dft_array_metadata` routine when used with symmetry objects. This is demonstrated in the following example which is based on the [frequency-domain solver tutorial example](https://meep.readthedocs.io/en/latest/Scheme_Tutorials/Frequency_Domain_Solver/) involving a ring resonator. The modal volume is computed in two separate runs using different methods: (1) via the `modal_volume_in_box` routine from the frequency-domain fields computed via `solve_cw` and (2) via the Fourier-transformed fields from a pulsed source using `get_dft_array` and cubature weights from `get_dft_array_metadata` (following a similar approach in [python/tests/modal-volume-metadata.py](https://github.com/NanoComp/meep/blob/master/python/tests/modal_volume_metadata.py#L22-L36)). 

```py
import meep as mp
import numpy as np

resolution = 30

n = 3.4
w = 1
r = 1
pad = 4
dpml = 2

sxy = 2*(r+w+pad+dpml)
vol = mp.Volume(mp.Vector3(), size=mp.Vector3(sxy-2*dpml,sxy-2*dpml))

c2 = mp.Cylinder(radius=r)

fcen = 0.118
df = 0.08

# CW source                                                                                                                                                                     
src = mp.Source(mp.ContinuousSource(fcen,fwidth=df), mp.Ez, mp.Vector3(r+0.1))

                    geometry=[c1,c2],
                    sources=[src],
                    resolution=resolution,
                    force_complex_fields=True,
                    symmetries=[mp.Mirror(mp.Y)],

sim.init_sim()                                                                                                                                                                
sim.solve_cw(1e-8, 10000, 10)                                                                                                                                                 
cw_modal_volume = sim.modal_volume_in_box(box=vol)                                                                                                                            

sim.reset_meep()

# pulsed source                                                                                                                                                                 

                    geometry=[c1,c2],
                    sources=[src],
                    resolution=resolution,
                    symmetries=[mp.Mirror(mp.Y)],  ## results are different if symmetries are left out                                                                                                                   

(Ex,Ey,Ez) = [sim.get_dft_array(dfts, c, 0) for c in [mp.Ex, mp.Ey, mp.Ez]]
(X,Y,Z,W)  = sim.get_dft_array_metadata(dft_cell=dfts)
# slightly annoying: we need an epsilon array with empty dimensions collapsed,                                                                                                  
# something not currently provided by any C++ function; for now just                                                                                                            
# create it via brute-force python loop, although this could get slow.                                                                                                          
Eps=np.zeros(0)
for x in X:
    for y in Y:
        for z in Z:
            Eps=np.append(Eps,sim.get_epsilon_point(mp.Vector3(x,y,z)))
Eps=np.reshape(Eps,np.shape(W))
EpsE2 = np.real(Eps*(np.conj(Ex)*Ex + np.conj(Ey)*Ey + np.conj(Ez)*Ez))
pulse_modal_volume = np.sum(W*EpsE2)/np.max(EpsE2)                                                                                                                            

print("modal volume:, {:.6f}, {:.6f}".format(cw_modal_volume,pulse_modal_volume))
```
When `symmetries=[mp.Mirror(mp.Y)]` is removed in the second run involving time-stepping a pulsed source, the results are nearly identical which is expected:
```
modal volume:, 4.033377, 4.014716
```
However, when `symmetries=[mp.Mirror(mp.Y)]` is included, the result based on `get_dft_array_metadata` is different:
```
modal volume:, 4.033377, 11.627477
```

### LINKED ISSUE #667: get_array_metadata not working when k_point is set
The [`get_array`](https://meep.readthedocs.io/en/latest/Python_User_Interface/#array-slices) routine does not seem to be working properly when `k_point` is used. This is demonstrated in the following example of a 2d cell consisting of a 1d lattice of slanted blocks. A side-by-side comparison of the dielectric profile obtained using `get_epsilon`  is generated for two cases: with a `k_point` of `mp.Vector3(0,0,0)` (periodic boundaries) and without.

```py
import meep as mp
import numpy as np
import math

resolution = 50

cell_size = mp.Vector3(5,2,0)

k_point = mp.Vector3(0,0,0)

rot_angle = 30

geometry = []
for j in range(5):
    geometry.append(mp.Block(material=mp.Medium(index=1.5),
                             center=mp.Vector3(-2+j,0,0),
                             size=mp.Vector3(0.5,3,0),
                             e1 = mp.Vector3(1,0,0).rotate(mp.Vector3(0,0,1), math.radians(rot_angle)),
                             e2 = mp.Vector3(0,1,0).rotate(mp.Vector3(0,0,1), math.radians(rot_angle))))

                    cell_size=cell_size,
                    k_point=k_point,
                    geometry=geometry,
                    ensure_periodicity=False)
sim.init_sim()
eps = sim.get_epsilon()
(x,y,z,w) = sim.get_array_metadata()

plt.figure(dpi=150)

plt.subplot(1,2,1)
plt.pcolormesh(x,y,np.transpose(eps), shading='gouraud')
plt.axis([-2.5, 2.5, -1, 1])
plt.xlabel("x")
plt.xticks([t for t in np.arange(-2.5,3.5,1.0)])
plt.ylabel("y")
plt.yticks([t for t in np.arange(-1.0,1.5,0.5)])
plt.title("with k_point")

sim.reset_meep()

                    cell_size=cell_size,
                    geometry=geometry,
                    ensure_periodicity=False)
sim.init_sim()
eps = sim.get_epsilon()
(x,y,z,w) = sim.get_array_metadata()

plt.subplot(1,2,2)
plt.pcolormesh(x,y,np.transpose(eps), shading='gouraud')
plt.axis([-2.5, 2.5, -1, 1])
plt.xlabel("x")
plt.xticks([t for t in np.arange(-2.5,3.5,1.0)])
plt.ylabel("y")
plt.yticks([t for t in np.arange(-1.0,1.5,0.5)])
plt.title("without k_point")

plt.tight_layout()
plt.show()
```
The dielectric profile is expected to be identical but is not.

eps

### LINKED ISSUE #758: pass snap_empty_dims to get_array_metadata
Currently `get_array_metadata` sets `snap_empty_dims=true` unconditionally.  This should really be a parameter, I think — it is `true` for array slices, but not for DFT slices (which are interpolated and collapsed).   By setting `snap_empty_dims=true` when getting DFT slice metadata we are potentially off by half a pixel in one of the coordinates, I think?

(`snap_empty_dims` should likewise be a parameter to `get_array_slice_dimensions`.)

HomerReid:
Fixes #658, fixes #667, fixes #758. Also prepares for #759 by extending the array collapsing routine to handle double-valued arrays.

In the new approach, the `weights` array returned by `get_array_metadata` may alternatively be fetched by calling the usual `get_array` or `get_dft_array` routines with `component` set to `NO_COMPONENT.`

If the environment variable `MEEP_ARRAY_XYZDIR` is set to 'X', 'Y', or 'Z', then the array slice produced for `NO_COMPONENT` instead stores the corresponding component of the grid-point coordinate. This is not intended for regular use but may be useful for debugging complicated cases.


stevengj:
Still failing:
```
FAIL: tests/cavity_arrayslice.py
FAIL: tests/dft_fields.py
FAIL: tests/simulation.py
```

ChristopherHogan:
> FAIL: tests/simulation.py

`python/tests/simulation.py:602` should be  `slice = sim.get_source(mp.Ez)`

> FAIL: tests/cavity_arrayslice.py

Passing `True` as the collapse (last) parameter to `mp._get_array_slice_dimensions` in `python/simulation.py:1709` fixes this. Is `collapse=True` supposed to be the default now for `get_array`?

> FAIL: tests/dft_fields.py

The results of `output_dft` and `get_dft_array` don't agree. The array shapes are the same, but the data is different.

stevengj:
> Is collapse=True supposed to be the default now for get_array?

Until #759 is fixed, I think it should be passing collapse=false, snap=true

stevengj:
> The results of output_dft and get_dft_array don't agree

Maybe one is now doing interpolation and the other isn't?

python/simulation.py
```diff
@@ -1706,7 +1706,7 @@ def get_array(self, component=None, vol=None, center=None, size=None, cmplx=None
         else:
             v = self._volume_from_kwargs(vol, center, size)
 
-        _, dirs = mp._get_array_slice_dimensions(self.fields, v, dim_sizes, False)
+        _, dirs = mp._get_array_slice_dimensions(self.fields, v, dim_sizes, False, True)
```
ChristopherHogan: This won't work. `mp._get_array_slice_dimensions` in `python/meep.i:498` needs to be updated to take the `snap_empty_dims` parameter and pass it through to `fields::get_array_slice_dimensions`.

ChristopherHogan:
`dft.cpp:1070` looks wrong. I think `size_t array_size = dims[0] * (rank==1 ? dims[1] : 1);` should be `size_t array_size = dims[0] * (rank==2 ? dims[1] : 1);`. This change makes the `dft_fields.py` test pass.

python/simulation.py
stevengj: Okay, I think it's fixed now?

stevengj:
It seems like it should handle `rank == 3` as well; I pushed a fix.

stevengj:
Is this okay to merge now?

ChristopherHogan:
Yes, I think so.

PR #817, opened 2019-04-11
GDSII layers

### LINKED ISSUE #514: GDSII file introspection
Would be good to have some introspection functions for GDSII files, e.g. a function to return a list of layer numbers present in a file.

HomerReid:
Fixes #514. Requires most recent repository version of `libGDSII.`

````python
>>> mp.GDSII_layers('python/examples/coupler.gds')              
Out[2]: [0, 1, 2, 3, 4, 5, 31, 32]
```


stevengj:
Would be nice to have the configure script check for a version of libGDSII so that it can give a more helpful error message than a compilation error.   I'm a bit worried that people will be confused if they download the new Meep and it fails to compile because they have an older version of libGDSII installed.

@oskooi commented that it would also be nice to get string labels in the file as well.

stevengj:
I updated it so that it still compiles with older versions of libgdsii

PR #823, opened 2019-04-15
Estimate memory usage of a simulation

ChristopherHogan:
I probably need some help to get this right. It will be useful for choosing the correct AWS machines for data collection.
@stevengj @oskooi

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -1146,12 +1146,12 @@ class Simulation(object):
         E_realnums = self.fragment_stats.num_pixels_in_box * (2 if is_complex else 1) * realnums_per_grid_point
         H_realnums = self.fragment_stats.num_pixels_in_box * (2 if is_complex else 1) * realnums_per_grid_point
         D_realnums = self.fragment_stats.num_pixels_in_box * (2 if is_complex else 1) * realnums_per_grid_point
+        chi1inv_realnums = self.fragment_stats.num_pixels_in_box * 9
 
         Mu_realnums = 0
         if self.has_mu():
-            Mu_realnums = 9 * self.fragment_stats.num_pixels_in_box * (2 if is_complex else 1) + H_realnums
+            Mu_realnums = chi1inv_realnums + H_realnums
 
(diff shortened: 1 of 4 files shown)


python/simulation.py
```diff
@@ -1115,6 +1116,34 @@ def require_dimensions(self):
...
 
+    def get_estimated_memory_usage(self):
+        if self.fields is None:
+            self.collect_stats = True
+            self.init_sim()
+
+        if self.fragment_stats is None:
+            self.fragment_stats = self._compute_fragment_stats(self.structure.user_volume)
+
+        is_periodic = self.k_point and self.k_point != mp.Vector3(0, 0, 0)
```
stevengj: probably call this `is_complex`, and also check if `force_complex` has been set

python/simulation.py
```diff
@@ -1115,6 +1116,34 @@ def require_dimensions(self):
...
+        E_realnums = self.fragment_stats.num_pixels_in_box * (2 if is_periodic else 1)
+        M_realnums = self.fragment_stats.num_pixels_in_box * (2 if is_periodic else 1)
```
stevengj: What is M?

python/simulation.py
```diff
@@ -1115,6 +1116,34 @@ def require_dimensions(self):
...
+        D_realnums = self.fragment_stats.num_pixels_in_box * (2 if is_periodic else 1)
```
stevengj: Shouldn't these be multiplied by 3 (at least in 3d(, since they are 3-component vector fields?

python/simulation.py
ChristopherHogan: I guess it was my short hand for "magnetic." My notes from when we talked about this say "E, M, and D" are the minimum three things that are always present.

python/simulation.py
ChristopherHogan: Done

python/simulation.py
ChristopherHogan: Done. I'm using `sim.dimensions` so it's correct for 1d and 2d too.

ChristopherHogan:
This should be ready.

PR #830, opened 2019-04-18
test for get_array_metadata

oskooi:
Adds a Python test for `get_array_metadata` involving `symmetries` and `k_point` following #811 which was recently merged.

### DIFF
--- a/python/tests/array_metadata.py
+++ b/python/tests/array_metadata.py
@@ -1,0 +1,65 @@
+import unittest
+import meep as mp
+import numpy as np
+
+class TestArrayMetadata(unittest.TestCase):
+
+    def test_array_metadata(self):
+        resolution = 25
+
+        n = 3.4
+        w = 1
+        r = 1
+        pad = 4
+        dpml = 2
+
+        sxy = 2*(r+w+pad+dpml)
+        nonpml_vol = mp.Volume(mp.Vector3(), size=mp.Vector3(sxy-2*dpml,sxy-2*dpml))
+
+        c1 = mp.Cylinder(radius=r+w, material=mp.Medium(index=n))
+        c2 = mp.Cylinder(radius=r)
+
+        fcen = 0.118
+        df = 0.08
+
+        # CW source
+        src = mp.Source(mp.ContinuousSource(fcen,fwidth=df), mp.Ez, mp.Vector3(r+0.1))
+
+        sim = mp.Simulation(cell_size=mp.Vector3(sxy,sxy),
+                            geometry=[c1,c2],
+                            sources=[src],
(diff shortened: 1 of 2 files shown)


python/tests/array_metadata.py
```diff
@@ -0,0 +1,65 @@
...
+                            boundary_layers=[mp.PML(dpml)])
+
+        dft_obj = sim.add_dft_fields([mp.Ez], fcen, fcen, 1, where=nonpml_vol)
+        sim.run(until_after_sources=100)
+
+        Ez = sim.get_dft_array(dft_obj,mp.Ez,0)
+        (X,Y,Z,W) = sim.get_array_metadata(dft_cell=dft_obj)
+        Eps = sim.get_array(vol=nonpml_vol,component=mp.Dielectric)
+        EpsE2 = np.real(Eps*np.conj(Ez)*Ez)
+        pulse_modal_volume = np.sum(W*EpsE2)/np.max(EpsE2)
```
stevengj: You might want to instead use some test integrand that involves (x,y,z) as well, e.g. the same integral but put a factor of (x^2 + 2y^2 + 3z^2) in the numerator.

For comparison, compute the same integral using `integrate_field_function` in the `ContinuousSource` simulation.

oskooi:
The test has been updated. It seems that there is a similar test [`modal_volume_metadata.py`](https://github.com/NanoComp/meep/blob/master/python/tests/modal_volume_metadata.py) that was previously added by @HomerReid but which is never actually used since it is not included in [`python/Makefile.am`](https://github.com/NanoComp/meep/blob/master/python/Makefile.am). Perhaps this file was used for internal development and thus should be removed?

python/tests/array_metadata.py
```diff
@@ -14,49 +14,64 @@ def test_array_metadata(self):
...
-                            boundary_layers=[mp.PML(dpml)])
+                            symmetries=symmetries,
+                            boundary_layers=pml_layers)
 
         sim.init_sim()
         sim.solve_cw(1e-6, 1000, 10)
-        cw_modal_volume = sim.modal_volume_in_box(box=nonpml_vol)
+
+        def electric_energy(r, ex, ey, ez, eps):
+            return np.real(eps * (np.conj(ex)*ex + np.conj(ey)*ey + np.conj(ez)*ez))
```
stevengj: To make sure the `X`, `Y`, `Z` arrays are correct, see my suggestion below about multiplying by a function of `r` too.

(We don't need to test this on a physically meaningful integrand.)

oskooi:
Following #831, the test has been updated to include an arbitrary position-dependent function in the integrand.

PR #843, opened 2019-04-26
Fix shape for get_farfields and output_farfields

### LINKED ISSUE #842: get_farfields returns arrays with incorrect shape
From @oskooi:

> The routine get_farfields returns a dictionary of NumPy arrays but the 
shape of these arrays for a 2d Near2FarRegion is 3d when it should be 
2d. This is demonstrated in the script below which produces the 
following output:
```
dict_keys(['Hx', 'Ex', 'Hy', 'Hz', 'Ey', 'Ez'])
(1, 1000, 21)
(1000, 21)
```

> Note that the shape for the array variable input_ff is (1, 1000, 21) 
whereas it should just be (1000, 21). To convert the 3d array to 2d 
requires specifying the first dimension via input_ff['Ez'][0] which 
should be unnecessary.

```python
import meep as mp
import math
import numpy as np

resolution = 30        # pixels/μm

sx = dpml+dsub+gh+dpad+dpml
cell_size = mp.Vector3(sx)

wvl_min = 0.4           # min wavelength
wvl_max = 0.6           # max wavelength
fmin = 1/wvl_max        # min frequency
fmax = 1/wvl_min        # max frequency
fcen = 0.5*(fmin+fmax)  # center frequency
df = fmax-fmin          # frequency width

sources = [mp.Source(mp.GaussianSource(fcen, fwidth=df), component=mp.Ez, center=src_pt, size=mp.Vector3())]

k_point = mp.Vector3()

glass = mp.Medium(index=1.5)

                    cell_size=cell_size,
                    boundary_layers=pml_layers,
                    k_point=k_point,
                    default_material=glass,
                    sources=sources)

nfreq = 21
n2f_pt = mp.Vector3(0.5*sx-dpml-0.5*dpad)
n2f_obj = sim.add_near2far(fcen, df, nfreq, mp.Near2FarRegion(center=n2f_pt, size=mp.Vector3()))

sim.run(until_after_sources=mp.stop_when_fields_decayed(50, mp.Ez, n2f_pt, 1e-9))

ff_distance = 1e8     # far-field distance from near fields (arbitrary)
ff_angle = 50         # far-field cone angle
ff_length = 2*ff_distance*math.tan(math.radians(ff_angle))
ff_npts = 1000
ff_res = ff_npts/ff_length

input_ff = sim.get_farfields(n2f_obj, ff_res, center=mp.Vector3(-0.5*sx+dpml+dsub+gh+ff_distance), size=mp.Vector3(y=ff_length))

print(input_ff.keys())
print(input_ff['Ez'].shape)
print(input_ff['Ez'][0].shape)
```

ChristopherHogan:
Fixes #842. This issue also existed for `output_farfields` (which calls `dft_near2far::save_farfields`). The documentation states that "dimensions that = 1 are omitted," but only trailing empty dimensions were being removed. Ideally the removal of empty dimensions could be done in `dft_near2far::get_farfields_array` so the collapsing code wouldn't need to be duplicated. Is that doable?
@stevengj @oskooi

### DIFF
--- a/src/near2far.cpp
+++ b/src/near2far.cpp
@@ -22,6 +22,7 @@
 
 #include <meep.hpp>
 #include <assert.h>
+#include <cstring>
 #include "config.h"
 #include <math.h>
 
@@ -331,6 +332,19 @@ realnum *dft_near2far::get_farfields_array(const volume &where, int &rank, size_
     }
   }
   sum_to_all(EH_, EH, 6 * 2 * N * Nfreq);
+
+  /* collapse singleton dimensions */
+  const int ndims = 4;
+  size_t collapsed_dims[ndims] = {};
+  int collapsed_rank = 0;
+  for (int i = 0; i < rank; ++i) {
+    if (dims[i] > 1) {
+      collapsed_dims[collapsed_rank++] = dims[i];
+    }
+  }
+  rank = collapsed_rank;
+  std::memcpy(dims, collapsed_dims, ndims * sizeof(int));
+
   delete[] EH_;
   delete[] EH1;
   return EH;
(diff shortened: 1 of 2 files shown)


stevengj:
Putting this in `get_farfields_array` seems sensible to me.  Note that the rank-collapsing must be done *after* the loops in that routine — just tack a little postprocessing of `dims` and `rank` on to the end.

src/near2far.cpp
```diff
@@ -331,6 +332,19 @@ realnum *dft_near2far::get_farfields_array(const volume &where, int &rank, size_
...
+  const int ndims = 4;
+  size_t collapsed_dims[ndims] = {};
+  int collapsed_rank = 0;
+  for (int i = 0; i < rank; ++i) {
+    if (dims[i] > 1) {
+      collapsed_dims[collapsed_rank++] = dims[i];
+    }
+  }
+  rank = collapsed_rank;
+  std::memcpy(dims, collapsed_dims, ndims * sizeof(size_t));
```
stevengj: Seems like it would be cleaner to do:
```cxx
/* collapse singleton dimensions */
int ireduced = 0;
for (int i = 0; i < rank; ++i)
    if (dims[i] > 1) dims[ireduced++] = dims[i];
rank = ireduced;
```

ChristopherHogan: Done.

PR #844, opened 2019-04-27
fix minor typos of docs

Maruoka842:
In my case,
```
print(harminv_instance.modes)
```
return 
```
[Mode(freq=0.11680245319740999, decay=-0.0006834209964856166, Q=85.45424694152504, amp=(0.022364090007081165-0.05966479702405166j), err=(1.943059072730648e-08+0j))]
```
So, the component of freq.real, freq.decay and freq.decay don't exist.
They should be freq and decay.


ChristopherHogan:
:+1:

PR #856, opened 2019-05-03
Add progress output for get_farfields_array

### LINKED ISSUE #849: displaying the progress status and total time of near2far calculations in output
In certain situations involving a large number of far-field points, the [`near2far`](https://meep.readthedocs.io/en/latest/Python_User_Interface/#near-to-far-field-spectra) routines `get_farfields` and `output_farfields` can comprise the majority of the wall-clock time (e.g., the tutorial example in #845). Currently, the progress status of the `near2far` calculation is *not* included in the output. Unlike the progress status for the time-stepping (i.e., `on time step 4467 (time=55.8375), 0.0549067 s/step`) which provides an approximate insight into its expected completion, no such information is available for `near2far` calculations. In fact, there is *no* mention at all in the output that the far-fields are being computed. It would be useful to include some indication of the `near2far` progress in the output. This information could also facilitate with debugging to improve performance.

ChristopherHogan:
Closes #849.
@stevengj @oskooi

### DIFF
--- a/src/meep.hpp
+++ b/src/meep.hpp
@@ -42,7 +42,7 @@ typedef float realnum;
 typedef double realnum;
 #endif
 
-#define MIN_OUTPUT_TIME 4.0 // output no more often than this many seconds
+#define MEEP_MIN_OUTPUT_TIME 4.0 // output no more often than this many seconds
 
 extern bool quiet; // if true, suppress all non-error messages from Meep
 
--- a/src/meep.hpp
+++ b/src/meep.hpp
@@ -42,6 +42,8 @@ typedef float realnum;
 typedef double realnum;
 #endif
 
+#define MIN_OUTPUT_TIME 4.0 // output no more often than this many seconds
+
 extern bool quiet; // if true, suppress all non-error messages from Meep
 
 const double pi = 3.141592653589793238462643383276;
--- a/src/meep.hpp
+++ b/src/meep.hpp
@@ -1369,6 +1369,7 @@ enum time_sink {
   FieldOutput,
   FourierTransforming,
   MPBTime,
+  GetFarfieldsTime,
   Other
 };
 
@@ -1799,9 +1800,6 @@ private:
   double times_spent[Other + 1];
   // fields.cpp
   void figure_out_step_plan();
-  // time.cpp
-  void am_now_working_on(time_sink);
-  void finished_working();
   // boundaries.cpp
   bool chunk_connections_valid;
   void find_metals();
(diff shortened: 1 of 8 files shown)


src/meep.hpp
```diff
@@ -42,6 +42,8 @@ typedef float realnum;
 typedef double realnum;
 #endif
 
+#define MIN_OUTPUT_TIME 4.0 // output no more often than this many seconds
```
stevengj: Should be renamed to `MEEP_MIN_OUTPUT_TIME` if it is in the global header.

ChristopherHogan: Fixed.

ChristopherHogan:
I can't use the `time_sink` infrastructure to time `farfield_lowlevel` because `am_now_working_on` and `finished_working` are methods of the `fields` class, while `farfield_lowlevel` is a `dft_near2far` method.

oskooi:
The output from running this PR on a test case involving [python/examples/binary_grating_n2f.py](https://github.com/NanoComp/meep/blob/master/python/examples/binary_grating_n2f.py) is:
```
...
get_farfields_array working on point 470 of 500 (94% done)
get_farfields_array working on point 477 of 500 (95% done)
get_farfields_array working on point 484 of 500 (96% done)
get_farfields_array working on point 491 of 500 (98% done)
get_farfields_array working on point 498 of 500 (99% done)

Elapsed run time = 476.4813 s

Field time usage:
        connecting chunks: 0.113281 s
            time stepping: 120.414 s
            communicating: 4.5503 s
     Fourier transforming: 25.4002 s
        getting farfields: 322.603 s
          everything else: 1.09053 s
```

It would be useful to include the metric `s/point` (meaningful for assessing performance) as part of the `get_farfields_array working on point ...` output similar to `s/step` for time stepping: `on time step 21406 (time=428.12), 0.00672292 s/step`. Also, note that the sum of all the sub entries under `Field time usage` is 474.171311 s. This means that 2.30998 s is still unaccounted for; should this difference just be included in `everything else:` so that the sum is consistent?

oskooi:
There is no interval progress output for `get_farfields` (i.e., `get_farfields_array working on point 434 of 500 (86% done)`) when using MPI; the only output is the total time at the end of the run:
```
...
field decay(t = 800.32): 3.1235531072986875e-11 / 0.09814324428153094 = 3.182647089124699e-10
run 0 finished at t = 800.32 (40016 timesteps)

Field time usage:
        connecting chunks: 0.00317784 s
            time stepping: 1.66425 s
            communicating: 0.361284 s
     Fourier transforming: 0.00176502 s
        getting farfields: 323.816 s
          everything else: 1.31586 s
```

PR #860, opened 2019-05-07
Add stop_after_walltime and stop_on_interrupt

### LINKED ISSUE #850: Feature request: simulation timeout run function
It would be nice to incorporate a `run` function that stops the simulation after a specified time interval (actual time, not MEEP time) has passed. 

This would be especially useful for jobs that are submitted to clusters with designated walltimes. The user could pass the walltime to the MEEP solver so that any progress could be saved before the job scheduler kills the job.

I'm thinking of something similar to the `opt.set_maxtime(maxtime)` within [nlopt's](https://meep.readthedocs.io/en/latest/Python_User_Interface/#run-functions) API.

ChristopherHogan:
Closes #850.
Usage:
```py
sim.run(until=mp.stop_after_walltime(60*60))
sim.run(until=200, stop_on_interrupt=True)
```
@stevengj @oskooi

### DIFF
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -1532,7 +1532,7 @@ A common point of confusion is described in [The Run Function Is Not A Loop](The
 —
 Run the simulation until a certain time or condition, calling the given step functions (if any) at each timestep. The keyword argument `until` is *either* a number, in which case it is an additional time (in Meep units) to run for, *or* it is a function (of no arguments) which returns `True` when the simulation should stop. If `stop_on_interrupt` is set to `True`, instead of terminating when receiving a SIGINT or SIGTERM signal from the system, the simulation will abort time stepping and continue executing any code that follows the `run` function (e.g., outputting fields).
 
-**`run(step_functions..., until_after_sources=condition/time, stop_on_interrupt=False`)**
+**`run(step_functions..., until_after_sources=condition/time, stop_on_interrupt=False)`**
 —
 Run the simulation until all sources have turned off, calling the given step functions (if any) at each timestep. The keyword argument `until_after_sources` is either a number, in which case it is an *additional* time (in Meep units) to run for after the sources are off, *or* it is a function (of no arguments). In the latter case, the simulation runs until the sources are off *and* `condition` returns `True`.
 
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -1528,9 +1528,9 @@ A common point of confusion is described in [The Run Function Is Not A Loop](The
 
 ### Run Functions
 
-**`run(step_functions..., until=condition/time, stop_on_interrupt=False)`**
+**`run(step_functions..., until=condition/time)`**
 —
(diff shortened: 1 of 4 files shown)


stevengj:
I think it would be better to have a way to combine multiple stopping conditions:
```py
sim.run(until=[200, mp.stop_after_walltime(3600), stop_on_interrupt()])
```
where `until` is now a list of stopping conditions, and the simulation stops if any of them returns true.

doc/docs/Python_User_Interface.md
```diff
@@ -1530,9 +1530,9 @@ A common point of confusion is described in [The Run Function Is Not A Loop](The
 
 **`run(step_functions..., until=condition/time)`**
 —
-Run the simulation until a certain time or condition, calling the given step functions (if any) at each timestep. The keyword argument `until` is *either* a number, in which case it is an additional time (in Meep units) to run for, *or* it is a function (of no arguments) which returns `True` when the simulation should stop.
+Run the simulation until a certain time or condition, calling the given step functions (if any) at each timestep. The keyword argument `until` is *either* a number, in which case it is an additional time (in Meep units) to run for, *or* it is a function (of no arguments) which returns `True` when the simulation should stop. `until` can also be a list of stopping conditions which may include a number and additional functions.
 
-**`run(step_functions..., until_after_sources=condition/time)`**
+**`run(step_functions..., until_after_sources=condition/time, stop_on_interrupt=False)`**
```
stevengj: the `stop_on_interrupt` keyword has been removed, no?

python/simulation.py
```diff
@@ -1266,16 +1273,20 @@ def _run_sources_until(self, cond, step_funcs):
...
-            new_cond = f
+        ts = self.fields.last_source_time()
+        new_conds = []
+        for i in range(len(cond)):
+            if isinstance(cond[i], numbers.Number):
+                new_conds.append((ts - self.round_time()) + cond[i])
+            else:
+                def f(sim):
+                    return cond[i](sim) and sim.round_time() >= ts
+                new_conds.append(f)
```
stevengj: It doesn't seem like we need this logic anymore.  Can't we just do
```
cond.append(ts)
```
to append `ts` to the other conditions?

ChristopherHogan: I don't think that works. Wouldn't It ignore other stopping conditions? E.g., if we have

```py
sim.run(until_after_sources=mp.stop_when_fields_decayed(...))
```

it would stop after time `ts` whether or not the fields have decayed.

doc/docs/Python_User_Interface.md
ChristopherHogan: Correct. Fixed.

python/simulation.py
stevengj: Why can't we just add it to the list of stopping conditions, rather than modifying the existing conditions, since the new code stops if any of the conditions are true?

ChristopherHogan: What you're saying is true for arguments to `until`, but for arguments to `until_after_sources` a stopping condition implies that 2 things must be true:

1. The stopping condition function returns `True`
2. The time is after the "last source time".

So, in a case like this

```py
sim.run(until=mp.stop_after_walltime(10))
```
the simulation should stop after 10 seconds, no matter what. But I would expect this

```py
sim.run(until_after_sources=mp.stop_after_walltime(10))
```

to stop after 10 seconds only if the last source time has been hit. If it takes 20 seconds to hit the last source time, then it should not stop after only 10 seconds, but continue the full 20 seconds.

Am I misunderstanding what `until_after_sources` is supposed to do?

stevengj: Ah, good point.  Another way to put it is that `until_after_sources` is *and* whereas our list of conditions is joined with *or*.

PR #862, opened 2019-05-07
Python implementation for medium evaluations

smartalecH:
The python portion of #475 that is used to evaluate material dispersion profiles at a given frequency/wavelength. Useful for making sure your profile really is what you expect. Also great for doing several MPB simulations across a band.

Here is a simple test script that pulls the profiles of Ag, Au, Si, and Cu and compares them to their original sources (stored in the attached h5 file):

```python
import numpy as np
import meep as mp
from meep.materials import Ag, Au, Cu, Si
import h5py


dataNames     = ['Ag','Au','Cu','Si']
testMaterials = [Ag,Au,Cu,Si]
hf = h5py.File('materialData.hdf5', 'r')

plt.figure(figsize=(4,8))

for iter in range(len(testMaterials)):
    # Pull Meep data
    M = testMaterials[iter]
    f0 = M.valid_freq_range[0]
    f1 = M.valid_freq_range[1]
    wavelength = np.linspace(1/f0,1/f1,1000)
    freq = 1/wavelength
    eps  = M.get_eps(freq)
    n = [np.real(np.sqrt(num.x)) for num in eps]
    k = [np.imag(np.sqrt(num.x)) for num in eps]

    # Pull refractiveindex.info data
    ref = hf.get(dataNames[iter])
    wavelength_ref = ref[:,0]
    n_ref = ref[:,1]
    if M !=Si:
        k_ref = ref[:,2]

    linewidth = 2
    # Set up plot
    plt.subplot(421 + 2*iter)
    plt.plot(wavelength_ref,n_ref,color='Red',linewidth=linewidth,label='ref')
    plt.plot(wavelength,n,'--',color='Blue',linewidth=linewidth,label='meep')
    plt.title("{} (real)".format(dataNames[iter]))
    plt.legend()
    plt.xlabel('Wavelength ($\mu$m)')
    plt.grid(True)
    plt.ylabel('n')

    if M !=Si:
        plt.subplot(421 + 2*iter+1)
        plt.plot(wavelength_ref,k_ref,color='Red',linewidth=linewidth,label='ref')
        plt.plot(wavelength,k,'--',color='Blue',linewidth=linewidth,label='meep')
        plt.title("{} (imag)".format(dataNames[iter]))
        plt.legend()
        plt.grid(True)
        plt.xlabel('Wavelength ($\mu$m)')
        plt.ylabel('k')
    

plt.tight_layout()
plt.savefig('materialEval.png')
plt.show()
```
materialEval

[materialData.hdf5.zip](https://github.com/NanoComp/meep/files/3154110/materialData.hdf5.zip)

### DIFF
--- a/python/geom.py
+++ b/python/geom.py
@@ -252,22 +252,18 @@ class Medium(object):
         else:
             freqs = np.squeeze(freq)
 
-        numFreqs = freqs.size
+        freqs = freqs[:,np.newaxis,np.newaxis]
 
         # Initialize with instantaneous dielectric tensor
-        epsmu = np.tile(np.array(Matrix(mp.Vector3(diag.x, offdiag.x, offdiag.y),
-                     mp.Vector3(offdiag.x, diag.y, offdiag.z),
-                     mp.Vector3(offdiag.y, offdiag.z, diag.z))),(numFreqs,1,1))
+        epsmu = np.expand_dims(Matrix(diag=diag,offdiag=offdiag),axis=0)
 
         # Iterate through susceptibilities
         for i_sus in range(len(susceptibilities)):
             epsmu = epsmu + susceptibilities[i_sus].eval_susceptibility(freqs)
         
         # Account for conductivity term
-        conductivity = np.tile(np.array(Matrix(mp.Vector3(conductivity_diag.x, conductivity_offdiag.x, conductivity_offdiag.y),
-                    mp.Vector3(conductivity_offdiag.x, conductivity_diag.y, conductivity_offdiag.z),
-                    mp.Vector3(conductivity_offdiag.y, conductivity_offdiag.z, conductivity_diag.z))),(numFreqs,1,1))
-        epsmu = (1 + 1j/np.tile(freqs,(3,3,1)).T * conductivity) * epsmu
+        conductivity = np.expand_dims(Matrix(diag=conductivity_diag,offdiag=conductivity_offdiag),axis=0)
+        epsmu = (1 + 1j/freqs * conductivity) * epsmu
 
         # Convert list matrix to 3D numpy array size [freqs,3,3]
         return np.array(epsmu)
@@ -280,9 +276,7 @@ class Susceptibility(object):
         self.sigma_offdiag = sigma_offdiag
 
     def transform(self, m):
-        sigma = Matrix(mp.Vector3(self.sigma_diag.x, self.sigma_offdiag.x, self.sigma_offdiag.y),
-                       mp.Vector3(self.sigma_offdiag.x, self.sigma_diag.y, self.sigma_offdiag.z),
-                       mp.Vector3(self.sigma_offdiag.y, self.sigma_offdiag.z, self.sigma_diag.z))
+        sigma = Matrix(diag=self.sigma_diag,offdiag=self.sigma_offdiag)
         new_sigma = (m * sigma * m.transpose()) / abs(m.determinant())
         self.sigma_diag = mp.Vector3(new_sigma.c1.x, new_sigma.c2.y, new_sigma.c3.z)
         self.sigma_offdiag = mp.Vector3(new_sigma.c2.x, new_sigma.c3.x, new_sigma.c3.y)
@@ -296,12 +290,8 @@ class LorentzianSusceptibility(Susceptibility):
         self.gamma = gamma
     
     def eval_susceptibility(self,freq):
-
-        sigma = np.tile(np.array(Matrix(mp.Vector3(self.sigma_diag.x, self.sigma_offdiag.x, self.sigma_offdiag.y),
-                       mp.Vector3(self.sigma_offdiag.x, self.sigma_diag.y, self.sigma_offdiag.z),
-                       mp.Vector3(self.sigma_offdiag.y, self.sigma_offdiag.z, self.sigma_diag.z))),(freq.size,1,1))
-
-        return np.tile(self.frequency*self.frequency / (self.frequency*self.frequency - freq*freq - 1j*self.gamma*freq),(3,3,1)).T * sigma
+        sigma = np.expand_dims(Matrix(diag=self.sigma_diag,offdiag=self.sigma_offdiag),axis=0)
+        return self.frequency*self.frequency / (self.frequency*self.frequency - freq*freq - 1j*self.gamma*freq) * sigma
 
 
 class DrudeSusceptibility(Susceptibility):
@@ -312,12 +302,8 @@ class DrudeSusceptibility(Susceptibility):
         self.gamma = gamma
     
     def eval_susceptibility(self,freq):
-
-        sigma = np.tile(np.array(Matrix(mp.Vector3(self.sigma_diag.x, self.sigma_offdiag.x, self.sigma_offdiag.y),
-                       mp.Vector3(self.sigma_offdiag.x, self.sigma_diag.y, self.sigma_offdiag.z),
-                       mp.Vector3(self.sigma_offdiag.y, self.sigma_offdiag.z, self.sigma_diag.z))),(freq.size,1,1))
-
-        return np.tile(-self.frequency*self.frequency / (freq*(freq + 1j*self.gamma)),(3,3,1)).T * sigma
+        sigma = np.expand_dims(Matrix(diag=self.sigma_diag,offdiag=self.sigma_offdiag),axis=0)
+        return -self.frequency*self.frequency / (freq*(freq + 1j*self.gamma)) * sigma
 
 
 class NoisyLorentzianSusceptibility(LorentzianSusceptibility):
@@ -489,10 +475,14 @@ class Prism(GeometricObject):
 
 class Matrix(object):
 
-    def __init__(self, c1=Vector3(), c2=Vector3(), c3=Vector3()):
+    def __init__(self, c1=Vector3(), c2=Vector3(), c3=Vector3(), diag=Vector3(), offdiag=Vector3()):
         self.c1 = c1
         self.c2 = c2
         self.c3 = c3
+        if c1 == c2 == c3 == Vector3():
+            self.c1 = Vector3(diag.x,offdiag.x,offdiag.y)
+            self.c2 = Vector3(offdiag.x,diag.y,offdiag.z)
+            self.c3 = Vector3(offdiag.y,offdiag.z,diag.z)
 
     def __getitem__(self, i):
         return self.row(i)
--- a/python/geom.py
+++ b/python/geom.py
@@ -211,6 +211,7 @@ class Medium(object):
         self.H_chi2_diag = H_chi2_diag
         self.H_chi3_diag = H_chi3_diag
         self.D_conductivity_diag = D_conductivity_diag
+        self.D_conductivity_offdiag = Vector3()
         self.B_conductivity_diag = B_conductivity_diag
         self.valid_freq_range = valid_freq_range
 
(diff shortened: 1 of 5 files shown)


python/geom.py
```diff
@@ -234,6 +234,27 @@ def transform(self, m):
 
         for s in self.H_susceptibilities:
             s.transform(m)
+    def get_eps(self,freq):
```
stevengj: Since we'll probably also want a `get_mu`, maybe this should be refactored into a `_get_epsmu(tensor, susceptibilities, conductivity)` function that is called by `get_eps` and `get_mu`

And maybe the functions should just be called `epsilon(self, freq)` and `mu(self, freq)`?

python/geom.py
```diff
@@ -234,6 +234,27 @@ def transform(self, m):
...
+        # Clean the input
+        freq = np.squeeze([freq])
+
+        # Compensate for scalar inputs
+        if len(freq.shape) == 0:
+            freq = np.array([freq])
+
+        # Initialize with instantaneous dielectric tensor, use numpy arrays for
+        # convenience and speed.
+        eps = np.array([self.epsilon_diag.x,self.epsilon_diag.y,self.epsilon_diag.z],dtype='complex128')
```
stevengj: What about the off-diagonal parts?

smartalecH:
Thanks for the suggestions! The function `epsilon()` now outputs a list of `matrix` objects corresponding to the dielectric tensor at that particular frequency.

Here is a test case showing both the ordinary and extraordinary axes for lithium niobate and barium borate:

```python

hf = h5py.File('materialData_ani.hdf5', 'r')

dataNames = ["BaB2","LiNO"]
testMaterials = [BaB2O4,LiNbO3]
names = ["BaB2O4","LiNbO3"]
plt.figure(figsize=(4,4))
for iter in range(len(testMaterials)):
    # Pull Meep data
    M = testMaterials[iter]
    f0 = M.valid_freq_range[0]
    f1 = M.valid_freq_range[1]
    wavelength = np.linspace(1/f0,1/f1,1000)
    freq = 1/wavelength
    eps  = M.epsilon(freq.tolist())
    ne = [np.real(np.sqrt(num.c3.z)) for num in eps]
    no = [np.real(np.sqrt(num.c1.x)) for num in eps]

    # Pull refractiveindex.info data
    ref = hf.get(dataNames[iter] + '_e')
    wavelength_ref = ref[:,0]
    ne_ref = ref[:,1]
    ref = hf.get(dataNames[iter] + '_o')
    wavelength_ref = ref[:,0]
    no_ref = ref[:,1]

    linewidth = 2
    # Set up plot
    plt.subplot(221 + 2*iter)
    plt.plot(wavelength_ref,ne_ref,color='Red',linewidth=linewidth,label='ref')
    plt.plot(wavelength,ne,'--',color='Blue',linewidth=linewidth,label='meep')
    plt.title("{}".format(dataNames[iter]))
    plt.legend()
    plt.xlabel('Wavelength ($\mu$m)')
    plt.grid(True)
    plt.ylabel('n$_e$')


    plt.subplot(221 + 2*iter+1)
    plt.plot(wavelength_ref,no_ref,color='Red',linewidth=linewidth,label='ref')
    plt.plot(wavelength,no,'--',color='Blue',linewidth=linewidth,label='meep')
    plt.title("{}".format(dataNames[iter]))
    plt.legend()
    plt.grid(True)
    plt.xlabel('Wavelength ($\mu$m)')
    plt.ylabel('n$_o$')
    

plt.tight_layout()
plt.savefig('materialEval_ani.png')

plt.show()
```
materialEval_ani

The previous simulations also still work.

I still haven't refactored the function into `_get_epsmu(tensor, susceptibilities, conductivity)` quite yet. Where would you like that function to reside?

Also, the `Matrix` class seems to have trouble accessing its elements with the bracket notation, even with the `__getitem__` overload. That's why I resorted to the odd, `num.c3.z` notation. The class is also missing copy overloads. We may want to look into that.

python/geom.py
smartalecH: Just fixed this by returning the entire dielectric tensor as a list. 

Do you know of any good materials to test this with? The uniaxial materials below obviously only test different values along the diagonal, but not any off-diagonal elements. I'm having trouble finding dispersion data for the off-diagonal terms of triclinic/monoclinic materials.

python/geom.py
smartalecH: I think this is a good idea.

Where would you prefer that `-get_epsmu()` reside? Would it return both `eps` and `mu`?

stevengj: That is, you would have
```
_get_epsmu(diag, offdiag, susceptibilities, conductivity_diag, conductivity_offdiag, freqs)
```
and from `epsilon` you would call `_get_epsmu(self.epsilon_diag, self.epsilon_offdiag, self.E_susceptibilities, self.D_conductivity_diag, self.D_conductivity_offdiag, freqs)` whereas for `mu` you would pass the corresponding magnetic quantities.

Also, I think it would be convenient to return a number rather than a list if `freqs` is a number.

stevengj:
Here is some anisotropic data on liquid crystal, for example: https://www.tandfonline.com/doi/full/10.1080/15421400600655816?scroll=top&needAccess=true&

stevengj:
For a test, just comparing the output to two hard-coded values should be sufficient.

stevengj:
To handle scalars and numpy arrays, you could do something like:
```jl
if hasattr(freq, "__iter__"):
    freqs = freq
else:
    freqs = [freq]
```
then, at the end, you can do something like
```py
if hasattr(freq, "__iter__"):
    return result
else:
    return result[1]
```

smartalecH:
I just pushed an update that refactored the method to use a common `_get_epsmu()` function. I vectorized the computation using `numpy` arrays (I'm not sure there's actually any gains here). 

The function can take either a scalar, list, or `numpy` array. It computes everything and outputs a `numpy` array with size (N,3,3) where N is the number of frequency points passed.

I added a test script that tests Si, LiNO3, and Au. I'm still looking for a decent material to test nondiagonal components.

I also added one more Lorentzian to the Au and Ag entries to match the cited publication.

python/geom.py
```diff
@@ -237,31 +240,37 @@ def transform(self, m):
...
         
-        return eps
+        # Account for conductivity term
+        conductivity = np.tile(np.array(Matrix(mp.Vector3(conductivity_diag.x, conductivity_offdiag.x, conductivity_offdiag.y),
+                    mp.Vector3(conductivity_offdiag.x, conductivity_diag.y, conductivity_offdiag.z),
+                    mp.Vector3(conductivity_offdiag.y, conductivity_offdiag.z, conductivity_diag.z))),(numFreqs,1,1))
+        epsmu = (1 + 1j/np.tile(freqs,(3,3,1)).T * conductivity) * epsmu
+
+        # Convert list matrix to 3D numpy array size [freqs,3,3]
+        return np.array(epsmu)
```
stevengj: probably we want a 3x3 (2d) array if `freq` was a number.

python/geom.py
```diff
@@ -237,31 +240,37 @@ def transform(self, m):
...
-            D_conductivity = Matrix(mp.Vector3(self.D_conductivity_diag.x, self.D_conductivity_offdiag.x, self.D_conductivity_offdiag.y),
-                       mp.Vector3(self.D_conductivity_offdiag.x, self.D_conductivity_diag.y, self.D_conductivity_offdiag.z),
-                       mp.Vector3(self.D_conductivity_offdiag.y, self.D_conductivity_offdiag.z, self.D_conductivity_diag.z))
-            ones_matrix = Matrix(mp.Vector3(1, 1, 1),
-                       mp.Vector3(1, 1, 1),
-                       mp.Vector3(1, 1, 1))
-            eps[i_freq] = (ones_matrix + 1j/freq[i_freq] * D_conductivity) * eps[i_freq]
+        epsmu = np.tile(np.array(Matrix(mp.Vector3(diag.x, offdiag.x, offdiag.y),
+                     mp.Vector3(offdiag.x, diag.y, offdiag.z),
+                     mp.Vector3(offdiag.y, offdiag.z, diag.z))),(numFreqs,1,1))
```
stevengj: Shouldn't this be a `(numFreqs,3,3)` array of 3x3 matrices, rather than an array of `Matrix`?   (Probably we need t a utility function to make a 3x3 hermitian numpy matrix from `diag` and `offdiag`.)

Also, I think you can just make it of size `(1,3,3)` and not use `np.tile` here, since when you multiply it by a `(numfreqs, 3,3)` array the numpy broadcasting will implicitly do the tiling for you.

(You should hardly ever need to call `np.tile` because of broadcasting, in fact.)

python/geom.py
```diff
@@ -237,31 +240,37 @@ def transform(self, m):
...
+        # Iterate through susceptibilities
+        for i_sus in range(len(susceptibilities)):
+            epsmu = epsmu + susceptibilities[i_sus].eval_susceptibility(freqs)
         
-        return eps
+        # Account for conductivity term
+        conductivity = np.tile(np.array(Matrix(mp.Vector3(conductivity_diag.x, conductivity_offdiag.x, conductivity_offdiag.y),
+                    mp.Vector3(conductivity_offdiag.x, conductivity_diag.y, conductivity_offdiag.z),
+                    mp.Vector3(conductivity_offdiag.y, conductivity_offdiag.z, conductivity_diag.z))),(numFreqs,1,1))
+        epsmu = (1 + 1j/np.tile(freqs,(3,3,1)).T * conductivity) * epsmu
```
stevengj: `conductivity` and `epsmu` have size `(1,3,3)`, then `epsmu = (1 + 1j/np.reshape(freqs,(numFreqs,1,1)) * conductivity) * epsmu` should I think have the desired result due to broadcasting (no need for `np.tile`).

python/geom.py
smartalecH: Actually, the `Matrix` class has an `__array__` overload function that automatically converts it into a numpy array when wrapped by the `np.array()` function. I just use the `Matrix` class for readability.

I agree, however, that we could include a utility function for these types of Matrix initializations. 

I'll also leverage the broadcasting better. I'll have to add an additional dimension to all of the `Matrix` objects to convert it from a 3x3 `numpy` array to a 1x3x3. You're right, there's no reason to use `tile` if I set up all the arrays right.

python/geom.py
smartalecH: `conductivity` and `epsmu` are initially size (3,3), which is why I was using tile. But it makes more sense to add singleton dimensions and let `numpy` handle the broadcasting. I'll add this in the next commit as well.

smartalecH:
My latest commit leverages broadcasting to streamline the computation.

I've also modified the Matrix constructor to allow the user to either specify three column vectors (as in the current implementation) or to specify a diagonal vector and an off-diagonal vector. If we decide to keep this, we should modify the docs too.

python/geom.py
smartalecH: I now squeeze the output array to remove the singleton dim for a scalar `freq`

oskooi:
Note that you need to add the test to [`python/Makefile.am`](https://github.com/NanoComp/meep/blob/master/python/Makefile.am); otherwise it is not part of the test suite and will not run during `make check`.

python/materials.py
```diff
@@ -258,12 +258,17 @@
 Ag_frq4 = 9.083*eV_um_scale      # 0.137 um
 Ag_gam4 = 0.916*eV_um_scale
 Ag_sig4 = Ag_f4*Ag_plasma_frq**2/Ag_frq4**2
+Ag_f5 = 5.646
+Ag_frq5 = 20.29*eV_um_scale
```
oskooi: This additional Lorentzian term corresponds to a resonance at a wavelength of 0.061 μm (or 61 nm) that is so far outside the optical range that it should not affect results but it will increase the computational expense due to the additional storage and time stepping of auxiliary fields. (This is why it was originally left out.) Is it really necessary?

python/materials.py
```diff
@@ -291,12 +296,17 @@
 Au_frq4 = 4.304*eV_um_scale      # 0.288 um
 Au_gam4 = 2.494*eV_um_scale
 Au_sig4 = Au_f4*Au_plasma_frq**2/Au_frq4**2
+Au_f5 = 4.384
+Au_frq5 = 13.32*eV_um_scale
```
oskooi: This Lorentzian resonance is at 0.093 μm (or 93 nm). Similar to the additional Ag term; is it really necessary?

python/materials.py
smartalecH: Right, but it seems to be a pretty broad resonance because even though it's so far away, it does alter the shape of the resonances between 100 nm and 300 nm. After that, however, it's negligible.

python/materials.py
smartalecH: Same as before. Let me know if you want me to remove it.

python/geom.py
stevengj: If `freq` was a list or array of length 1, not sure we want to do this… probably only squeeze if `freq` was a scalar.

python/geom.py
```diff
@@ -436,10 +475,14 @@ def __init__(self, vertices, height, axis=Vector3(z=1), center=None, **kwargs):
...
 
-    def __init__(self, c1=Vector3(), c2=Vector3(), c3=Vector3()):
+    def __init__(self, c1=Vector3(), c2=Vector3(), c3=Vector3(), diag=Vector3(), offdiag=Vector3()):
         self.c1 = c1
         self.c2 = c2
         self.c3 = c3
+        if c1 == c2 == c3 == Vector3():
+            self.c1 = Vector3(diag.x,offdiag.x,offdiag.y)
+            self.c2 = Vector3(offdiag.x,diag.y,offdiag.z)
+            self.c3 = Vector3(offdiag.y,offdiag.z,diag.z)
```
stevengj: The `offdiag` parts represent a complex Hermitian matrix in general, so you should take the conjugate of the lower-triangular part of the matrix with `np.conj` (which works for both real and complex values).

smartalecH: good catch -- fixed with latest commit.

PR #863, opened 2019-05-08
Implementation of gyrotropic susceptibilities

### LINKED ISSUE #60: Feature request: gyrotropic media
I would like to re-open the topic of gyrotropic media in MEEP. (Related terms are also Faraday effect, nonreciprocal behaviour, magnetooptics etc. Mathematically, all these are represented by nonzero imaginary part of the off-diagonal components in the permittivity tensor.)

While I found no access to this functionality in its current version, there are hints it has been implemented years ago: 
 * "Complex dielectric tensors are also useful for gyrotropic (magneto-optic) media. We have some internal modified versions of Meep that support this. However, merging these modifications into an official Meep hasn't happened yet, partially because I have another student who has been rewriting the PML and timestepping code"  (January 2008, https://www.mail-archive.com/meep-discuss@ab-initio.mit.edu/msg01383.html)
 * "We are also grateful to (...) Y. Chong for early support in Meep for gyrotropic media." (http://ab-initio.mit.edu/~oskooi/papers/Oskooi10.pdf)

This would be a very interesting feature for many researchers. (I know it from personal collaboration with other people, and it has also been asked for repeatedly in the mailing lists.) 

I believe the transfer of MEEP development to Github can finally get together people with the know-how with people having time and motivation to make it work: **Could we merge the obviously existing nonreciprocal MEEP code, perhaps as a mere prototype or experimental feature?**

seewhydee:
Closes #60 ...


stevengj:
Overall, this looks great!

Have you tried it out on any example problems yet?  Once there are some minimal tests that indicate that this is working, I would be happy to merge as an experimental feature since it doesn't really affect the rest of the code.   If you have an example reproducing something from a paper to use a tutorial that would be good too.

My biggest concern (modulo bugs) is whether this will introduce instabilities.

seewhydee:
Haven't done any rigorous testing yet, will work on that soon. Off the top of my head, there are a couple of straightforward tests to be do in 2D:

* In a uniform 2D waveguide with out-of-plane gyrotropy vector, the dispersion relation can be calculated analytically and compared to simulation results for varying gyrotropy.
* In a 2D disk of gyrotropic medium with out-of-plane gyrotropy vector, I think there should be an analytic solution for the Bessel-type normal modes, which should experience a Zeeman-like splitting. This can be compared to the resonance frequencies found in simulation.

For 3D, I'm not sure; it may be necessary to compare to a Comsol calculation.

seewhydee:
Sorry for the long silence. Here's an update.

I wasn't too happy with the equation of motion that just adds a precession term to the Lorentzian damped harmonic oscillator, since it does not reproduce the right frequency dependence for materials like ferrites. So I spent some time implementing another equation of motion, the [Landau Lifshitz Gilbert equation](https://en.wikipedia.org/wiki/Landau%E2%80%93Lifshitz%E2%80%93Gilbert_equation), which does give the right frequency dependence (this is on the "landau-lifshitz-gilbert" branch). [Implementation notes here](https://github.com/NanoComp/meep/files/3222559/meep_gyrotropy_note_2.pdf).

However, there's a stability problem that I haven't been able to fix. For relatively weak gyrotropy and short runtime, the simulation behaves as expected. The [attached figure](https://user-images.githubusercontent.com/4715025/58406475-42792580-809c-11e9-94e0-fc179e2896a3.png) plots Ex and Ey for a wave propagates along a gyrotropy axis z; indeed, the Ex component precesses into Ey. But with increasing t, a spatial instability sets in and creates an oscillation on the length scale of the grid.

It doesn't seem to be a problem with the Euler integration of the Landau Lifshitz Gilbert equation by itself -- for an isolated point, the integration doesn't seem to exhibit any instability. There seems to be some kind of bad interaction with the EM fields on the grid.

Kinda stumped right now, so suggestions welcome.

stevengj:
Have you tried increasing the "courant" parameter (i.e. shrinking Δt)?

seewhydee:
After setting Courant to 0.25, the instability still occurs, but appears at a later time.

I'll play around with some alternative discretization schemes for the LLG equation and see if it makes any difference.

If it still doesn't work, I guess we can fall back on the Lorentzian + precession implementation of gyrotropy, which doesn't seem to have such a problem.

seewhydee:
Hi Steven,

Focusing on the Lorentzian + precession model for now; Alex (@acerjan) did some additional testing on this and found that a remaining stability problem when there are different materials in a simulation. To nail this down, I need to ask a few questions about the Yee grid implementation.

First question: in the previous Lorentzian susceptibility model, at each grid point belonging to Ex, Ey, or Ez, only the polarization vector P in the field direction is calculated, correct? For example, on a grid point belonging to Ex, the LOOP_OVER_VOL_OWNED macro ensures that only Px is calculated. If the medium is anisotropic, the values of Ey and Ez are available for the computation at the Ex grid point, and there's memory allocated for Py and Pz. But the Py and Pz components are never updated.

Second: my code replaces LOOP_OVER_VOL_OWNED with LOOP_OVER_VOL. This is so that (e.g.) at each Ex grid point, not only Px but also Py and Pz are calculated, with the appropriate gyrotropy contribution. From these three P components, only one of them (e.g. Px) subsequently participates in the FDTD updating for that grid point. This means doing 3x the usual work, but I don't see an easy way around that. Does this approach sound correct to you?

Third: what exactly is the role of needs_W_notowned? I didn't know what this did so I previously didn't turn it on for gyrotropic media. Regardless, the other field components seemed to be available anyway (e.g. the values of Ey and Ez were available at Ex grid points). Do I need to ensure that needs_W_notowned is always true for all directions (as though the medium is anisotropic)? This might be the source of the problem Alex was seeing with non-uniform structures, but I'm not sure...

acerjan:
I've put together some relatively simple code that reproduces my stability issue. There are two files attached here, the first is a minimal scheme ctl file which simulates an Hz source in a punctured gyrotropic medium which encounters stability issues towards the end of its runtime. The second file is a shell script to run the scheme code. Currently, the .sh and .ctl files are set up to output the 2D Hz fields and create a .gif, in which you will see the stability issue, shown here:

gyro_stab_boundary_nLayers_01

If you change the variable epsTopoOutMO to be 0 from its current value of 0.5, the code will turn off the gyrotropic material, and no instability will be seen.

Please let me know if there is something else I can do to help with the development of this feature. I will note that once this stability issue is ironed out, I do have some simple code that can hopefully be used as an example of finding surface states at the boundary between a topological insulator and trivial medium.

[gyroStability.txt](https://github.com/NanoComp/meep/files/3234104/gyroStability.txt)
[run_stability.txt](https://github.com/NanoComp/meep/files/3234105/run_stability.txt)

seewhydee:
Whoops, it looks like I was confused; LOOP_OVER_VOL doesn't mean what I seemed to think it meant.

I've updated the susceptibility code to use LOOP_OVER_VOL_OWNED, but updating the 3 polarization components explicitly. The basic approach is still the same: the P vector at each grid point is driven by a single field component, but itself has 3 components.

Alex's test case seems to work OK now without destabilizing. Alex, can you check again?

seewhydee:
Actually, hang on --- there's still a problem with the current scheme. Gonna have to think about it more...

seewhydee:
Hey @acerjan... I think I got my confusions worked out, and after the latest fixes the test case doesn't seem to be blowing up. Could you try again? ¯\_(ツ)_/¯

acerjan:
The test code I posted works, but I'm able to generate an instability using a slightly different geometry. Let me try to strip this down to a slightly simpler system. This time, the error might be coming from the interface between two gyrotropic media, but hard to say.

acerjan:
Ok, it looks like the error is still coming from the interface between gyrotropic and non-gyrotropic media, but it only seems to appear with non-zero k. Shown here is a gif of the Hz field generated by the attached ctl file in which you can clearly see that there is still a stability issue.

gyro_stab_boundary_nLayers_01

[gyroStability_v2.txt](https://github.com/NanoComp/meep/files/3238432/gyroStability_v2.txt)

oskooi:
Three things to try to determine if the instability is related to the discretization:
- turn off subpixel averaging since it is used by default
- increase the `resolution`
- decrease the `Courant` number

seewhydee:
Subpixel averaging, resolution, and courant number don't seem to have much effect on the instability. It seems to kick in whenever the bias (precession frequency) or the magnitude of the k-point becomes too large. Maybe it's something to do with boundary conditions are handled? I still don't have a good understanding of that part of the code.

Alex, is it possible to check the quantitative match between the Meep implementation and MPB for small values of k and bias (and/or dielectric contrast), where the instability doesn't show up? It would help narrow down the range of possible problems. I've gone through the damped harmonic oscillator implementation multiple times and am fairly sure it's correct, but it would be nice to get some confirmation.

seewhydee:
Aha, found the interaction with the boundary conditions: gyrotropic_susceptibility needed custom num_cinternal_notowned_needed and num_cinternal_notowned_ptr methods so that complex boundaries are handled right.

Should be fixed now. Try again?

seewhydee:
Here's a quantitative comparison showing good agreement with the theoretical solution for a Faraday-rotating plane wave:

Faraday-rotation-comparison

The Python file has been added to python/examples/, and is discussed in the new file doc/docs/Python_Tutorials/Gyrotropic_Media.md in the repository.

acerjan:
Awesome! This new version works! Shown here is a set of simulations at the interface between a topological insulator, based on the simple model of [Haldane and Raghu](https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.100.013904), and a trivial insulator comprised of a triangular lattice of air holes in a high dielectric background. The source is placed near the interface, while absorbing boundary conditions are added far away from the boundary. As you can see, this setup properly allows for the harmonic inversion routine to find the chiral edge state at the boundary as the one good remaining crystal momentum is swept.

TI_boundary_meep

My goal is to put the code used to generate this plot together as an additional example for the gyrotropic media feature, but I have a couple remaining questions as the bandgap in the above plot isn't as large as I expected it to be. (Bandgap should be from 0.42 to 0.50, as I found using mpb.)

1) To simulate something as close to a simple, constant complex-Hermitian dielectric tensor, I chose to work with a lossless gyrotropic Drude susceptibility, loosely documented in Yidong's (@seewhydee) notes in #60, meep_gyrotropy_note.pdf. Am I correct that for this purpose, working with the Drude form of the susceptibility is simpler than the Lorentzian form?

2) I rederived Eq. 4 in meep_gyrotropy_note.pdf, and I want to confirm that I either screwed up my algebra somewhere or that there are a pair of typos in the notes -- that the term in front of Eq. 4b should be omega_0^2 sigma b / omega^3 (as it is currently written is has the wrong units I think), and that the denominator in both Eqs. 4a and 4b is (1 + i gamma / omega)^2 - (b/omega)^2. Or have I missed something?

seewhydee:
Hi Alex, I found a typo in the susceptibility update equation. Please try again when you have the time.

Regarding the frequency domain dielectric function, please see if you agree [with my derivation here](https://github.com/NanoComp/meep/files/3250339/meep_gyrotropy_note_2.pdf). I currently don't know how gyrotropic media would interact with PML, sorry.

stevengj:
> Third: what exactly is the role of needs_W_notowned? I didn't know what this did so I previously didn't turn it on for gyrotropic media. Regardless, the other field components seemed to be available anyway (e.g. the values of Ey and Ez were available at Ex grid points). Do I need to ensure that needs_W_notowned is always true for all directions (as though the medium is anisotropic)? This might be the source of the problem Alex was seeing with non-uniform structures, but I'm not sure...

When you are updating the polarization component Pₖ at an "owned" point (which is all each chunk is required to update), for an isotropic material it only needs the associated electromagnetic field (E or H) component Wₖ at the same "owned" point, so no communication is required.

However, for an anisotropic material we also need other field components Wⱼ for j≠k, and thanks to the Yee grid we need to get these by averaging from neighboring grid points (to get Wⱼ at the location of Pₖ to second-order accuracy).  But these neighboring grid points might not be "owned" if you are at the boundary of your chunk, so they must be found by explicit communication with other processes/chunks.

If the susceptibility returns `needs_W_notowned == true` for a given component, therefore, the code sets up some additional communications, as described here: https://github.com/NanoComp/meep/blob/6762b82852ff3c594822e08801ed0b86d10bb309/src/boundaries.cpp#L319-L334

By default, `needs_W_notowned` is determined from whether the material is isotropic.  In your Lorentz + precession, you need not-owned P points for your cross-product term, but your model only requires diagonal elements of W (the E field typically) for a diagonal σ, so I think you are fine.  On the other hand, for the Landau-Lifshitz-Gilbert equation, you *do* require other components of E, so I think you need to return `true` for `needs_W_notowned` in that case.

stevengj:
Regarding PML, for our formulation (http://math.mit.edu/~stevenj/papers/OskooiJo11.pdf) I don't immediately see any problem with gyrotropic media.   You are changing the P equation (the computation of "U" from "W" in the notation of the linked paper), but since the dP/dt equation is completely local in space it is not affected by coordinate stretching.

(It would be a problem for PML if your gyrotropic material introduced backwards-wave modes, where the group velocity dω/dk has a sign opposite of k, but I don't think you are doing that?)

acerjan:
I also found a bug in my code (forgot to change the point source to be a line source when I added a finite thickness), but even with the newly updated version, my bandgap is still a bit smaller than I expected. I've also been through your new set of notes and I agree with your derivation.

It may be that the smaller bandgap I'm seeing is entirely correct, and is due to changes in the effective dielectric tensor for frequencies further away from the central frequency, and that the thing to do is run a series of very narrow sources to build up the band structure that way. For reference, the bandgap I'm expecting is ~20% of the central frequency. This change will take a bit of time to implement, so it'll have to wait until I return from holiday.

seewhydee:
@acerjan: the topological band gap you're testing on is really wide, and dispersion is highly likely to be a significant issue. For the quantitative verification, I suggest simply testing against a ~2-3% bandgap.

@stevengj: I think the code is converging, so could you take a look at the proposed interface and comment on whether it's acceptable? (The Scheme interface will track the Python one, once finalized).

* [Theoretical formulation in Materials.md](https://github.com/seewhydee/meep/blob/landau-lifshitz-gilbert/doc/docs/Materials.md#gyrotropic-media)

* [Python interface with 3 new susceptibility classes](https://github.com/seewhydee/meep/blob/landau-lifshitz-gilbert/doc/docs/Python_User_Interface.md#gyrotropiclorentziansusceptibility-or-gyrotropicdrudesusceptibility)

stevengj:
If the problem is dispersion, simply double the resolution and see if the agreement improves significantly.   I don't recall ever having a big problem with dispersion over a 20% bandwidth, which gets you nowhere near the Nyquist frequency.

stevengj:
If we can have a sanity test of this feature, I think we can merge it?

seewhydee:
Fine by me. As of now, Faraday rotation of a plane wave quantitatively matches theory for all three types of gyrotropic media. For inhomogenous structures containing both gyrotropic and non-gyrotropic media, the simulations run without field instabilities, but we have not quantitatively checked the results yet. As a clearly-labelled experimental feature, it's OK to merge whenever you like.

Alex will compare the topological photonic crystal band gap size when he gets back, and there are a bunch of other tests I'm interested in doing (e.g. looking at the dispersion relation of magnetic surface plasmons) before removing the "experimental" label.

stevengj:
> As of now, Faraday rotation of a plane wave quantitatively matches theory for all three types of gyrotropic media.

Great!  Can you add an automated test to `make check` for this?    A Python test in the `python/tests` directory would be fine.

seewhydee:
Sure, done: https://github.com/seewhydee/meep/blob/master/python/tests/faraday_rotation.py

stevengj:
Looks good to merge once tests are green.

ChristopherHogan:
The `faraday_rotation.py` test needs to be added to the `TESTS` variable in `python/Makefile.am`. Otherwise the test won't run.

seewhydee:
Oh, right. Added it.

acerjan:
After returning from a couple of different trips, I've finally had a moment to run simulations which account for the dispersion in MEEP's model of gyrotropic material relative to MPB's. Although Steven's suggestion of doing self-consistent frequency calculations in mpb would probably have been less computer runtime, just running a bunch of meep simulations with different choices of the gyrotropic media was faster to set up, and computer resources are bountiful.

As a reminder, I'm simulating the boundary between a topological and a trivial photonic crystal, so we expect to see a chiral edge state. Of course, mpb finds both edge states, but meep should only find one, as the other border is buried in the absorbing region. As can be seen in this plot comparing mpb (red lines) with meep (blue +s), the results are similar, though some slight differences are seen -- the meep frequencies seem to be reduced by ~2-3% compared with mpb, but the overall shape, and bandgap are nearly the same. Is this a typical amount of discrepancy between the two simulation techniques? Or is this indicative of an error?

TI_boundary_meep_mpb

If you think the error is small enough to not be concerned, I'm happy to prepare some of this as an example for meep / mpb, just let me know what you'd like to have added.

seewhydee:
Could this be a due to discretization? Does changing the resolution alter the frequency shift?

It certainly would be great to have a photonic TI as a Meep tutorial, but doing repeated simulations over a frequency sweep seems like an unnecessary complication for that purpose. Could you try picking parameters where the bandgap is about 1/10 the present size (say a 3% bandgap rather than a ~20% bandgap)? Then dispersion effects ought to be much less pronounced, and it should be possible to compare Meep results to MPB results (using mid-gap values) without a frequency sweep.

acerjan:
Yeah, i'll move to a 3% bandgap or something smaller so i can check the discretization more easily. It still might require 2 or 3 runs, but not the ~26 used in the above plot.

PR #865, opened 2019-05-08
Output seconds/point in get_farfields_array progress

ChristopherHogan:
@stevengj @oskooi

### DIFF
--- a/src/near2far.cpp
+++ b/src/near2far.cpp
@@ -324,12 +324,13 @@ realnum *dft_near2far::get_farfields_array(const volume &where, int &rank, size_
       x.set_direction(dirs[1], where.in_direction_min(dirs[1]) + i1 * dx[1]);
       for (size_t i2 = 0; i2 < dims[2]; ++i2) {
         x.set_direction(dirs[2], where.in_direction_min(dirs[2]) + i2 * dx[2]);
-        if (!quiet && wall_time() > start + MEEP_MIN_OUTPUT_TIME) {
+        double t;
+        if (!quiet && (t = wall_time()) > start + MEEP_MIN_OUTPUT_TIME) {
           size_t this_point = (dims[1]*dims[2]*i0) + (dims[2]*i1) + i2 + 1;
           master_printf("get_farfields_array working on point %zu of %zu (%d%% done), %g s/point\n",
                         this_point, total_points, (int)((double)this_point/total_points*100),
-                        (wall_time() - start) / (std::max(1, (int)(this_point - last_point))));
-          start = wall_time();
+                        (t - start) / (std::max(1, (int)(this_point - last_point))));
+          start = t;
           last_point = this_point;
         }
         farfield_lowlevel(EH1, x);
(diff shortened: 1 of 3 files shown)


src/near2far.cpp
```diff
@@ -325,9 +326,11 @@ realnum *dft_near2far::get_farfields_array(const volume &where, int &rank, size_
         x.set_direction(dirs[2], where.in_direction_min(dirs[2]) + i2 * dx[2]);
         if (!quiet && wall_time() > start + MEEP_MIN_OUTPUT_TIME) {
           size_t this_point = (dims[1]*dims[2]*i0) + (dims[2]*i1) + i2 + 1;
-          master_printf("get_farfields_array working on point %zu of %zu (%d%% done)\n",
-                        this_point, total_points, (int)((double)this_point/total_points*100));
+          master_printf("get_farfields_array working on point %zu of %zu (%d%% done), %g s/point\n",
+                        this_point, total_points, (int)((double)this_point/total_points*100),
+                        (wall_time() - start) / (std::max(1, (int)(this_point - last_point))));
           start = wall_time();
```
stevengj: Probably you should eval `t = wall_time()` once and assign it to a variable rather than 3 times, more to use a consistent time than for performance, not that it matters too much.

stevengj: i.e.
```c
double t;
if (!quiet && (t = wall_time()) > start + MEEP_MIN_OUTPUT_TIME) {
    ...
}
```

ChristopherHogan: Done.

src/near2far.cpp
```diff
@@ -323,11 +324,14 @@ realnum *dft_near2far::get_farfields_array(const volume &where, int &rank, size_
       x.set_direction(dirs[1], where.in_direction_min(dirs[1]) + i1 * dx[1]);
       for (size_t i2 = 0; i2 < dims[2]; ++i2) {
         x.set_direction(dirs[2], where.in_direction_min(dirs[2]) + i2 * dx[2]);
-        if (!quiet && wall_time() > start + MEEP_MIN_OUTPUT_TIME) {
+        double t;
+        if (!quiet && (t = wall_time()) > start + MEEP_MIN_OUTPUT_TIME) {
```
stevengj: If we want this to work in the parallel case then we may need to add:
```
if (!quiet) all_wait();
```
to make the timing accurate on the master process.

The question is, what is the performance impact of putting synchronization in the loop like this?  I'm not sure

ChristopherHogan:
The synchronization doesn't seem to affect performance at all.  Also calling `stdout.flush` from C instead of python so it flushes on every call to `master_printf` instead of just during time stepping.

stevengj:
You tried it with a big parallel calculation?

ChristopherHogan:
I tried it with 4 processors on the [binary_grating_n2f example](python/examples/binary_grating_n2f.py)  which spent around 350 seconds in `get_farfields_array`, both with and without the `all_wait()`.

stevengj:
I was more worried about the scaling for a large number of processors (~100).

ChristopherHogan:
Can I just crank the resolution in binary_grating_n2f.py, or does it need to be modified in some other way to make a realistic example for 100 procs?

oskooi:
To test for a large number of processors, you can increase three parameters simultaneously: `resolution`, `nfreq`, and `nperiods`. (Also, you'll probably just want to run the last of the three `run`s involving the supercell.)

Issue #861, opened 2019-05-07
speeding up near2far


stevengj:
Two options:

- [x] use `#pragma omp parallel for` in `farfield_lowlevel`, maybe by moving `for (int i = 0; i < Nfreq; ++i)` to be the outermost loop.  This only works if you have cores to spare and we compile with OpenMP. #868

- [x] in 2d, we can switch to a faster Bessel-function calculation.   In a quick test, ~~the [Fortran routines by Amos](https://people.sc.fsu.edu/~jburkardt/f77_src/toms644/toms644.html)~~ libc `jn` and `yn` routines are about 4-5x faster than the ones in GSL.  Precompiled binaries of these are available in https://github.com/JuliaMath/openspecfun at https://github.com/JuliaMath/OpenspecfunBuilder/releases, for example, if we don't want to require a Fortran compiler. #869

For the first option, would be nice to try it out.  Note that you will need to compile with `-fopenmp`

stevengj:
I added a branch `omp_near2far` with the changed loop (871b4ca06d22d54dd7e9fe7717c37e7fe6687fd7).  You will need to manually add the `-fopenmp` flag by configuring
```
./configure CXX="g++ -fopenmp"
```
or similar

oskooi:
This seems to be working and giving rise to linear speedup as demonstrated in the figure below. The test involves timing the `get_farfields` calculations in [`python/examples/binary_grating_n2f.py`](https://github.com/NanoComp/meep/blob/master/python/examples/binary_grating_n2f.py) involving 21 frequencies and 500 points for 5 values of `OMP_NUM_THREADS` (1-5) for two cases: (1) a unit cell with `add_near2far` involving `nperiods=20` and (2) a super cell with 41 unit cells and `nperiods=1` (i.e., default: no tiling). Note: since the total time for `get_farfields` is not yet directly available, pending #856, the values were computed by subtracting from the "Elapsed run time" all the components in "Field time usage" (e.g., "time stepping, "communicating", etc).

![n2f_openmp_semilogy](https://user-images.githubusercontent.com/7152530/57340770-16493500-70ec-11e9-8360-da92244e2509.png)

The only unusual thing was that it worked for the serial version (i.e., `CXX="g++ -fopenmp"`) but *not* with the OpenMPI compiler (i.e., `CXX="mpic++ -fopenmp"`).

stevengj:
[...]

Issue #228, opened 2018-03-09
OpenMP parallelism


stevengj:
Might be nice to support this in addition to MPI.  The advantage of OpenMP is that the runtime system can automatically load balance, because no data movement is required.  However, it only works with a single shared-memory multi-core compute node, so you'd still need MPI to distribute large simulations.

(See also #130.)

stevengj:
See #565 for a rough draft implementation.

PR #868, opened 2019-05-09
openmp for near2far calculation

stevengj:
This adds an experimental `--with-openmp` flag to the `configure` script that turns on OpenMP compilation, and uses it for the `near2far` calculation for part 1 of #861 (only parallelizing if you are computing for multiple frequencies, however).

To use it, `./configure --with-openmp` and set the `OMP_NUM_THREADS` environment variable before running Meep (or Python).

In the long run, it would be nice to use more comprehensive OpenMP parallelism (#228).

Although this parallelizes over the different frequencies because that was simplest, we could alternatively parallelize at over spatial points in the near or far field.

### DIFF
--- a/src/near2far.cpp
+++ b/src/near2far.cpp
@@ -250,7 +250,9 @@ void dft_near2far::farfield_lowlevel(std::complex<double> *EH, const vec &x) {
     component c0 = component(f->vc); /* equivalent source component */
 
     vec rshift(f->shift * (0.5 * f->fc->gv.inva));
-#pragma omp parallel for
+#ifdef HAVE_OPENMP
+#  pragma omp parallel for
+#endif
     for (int i = 0; i < Nfreq; ++i) {
       double freq = freq_min + i * dfreq;
       size_t idx_dft = 0;
--- a/src/near2far.cpp
+++ b/src/near2far.cpp
@@ -250,12 +250,13 @@ void dft_near2far::farfield_lowlevel(std::complex<double> *EH, const vec &x) {
     component c0 = component(f->vc); /* equivalent source component */
 
     vec rshift(f->shift * (0.5 * f->fc->gv.inva));
-    size_t idx_dft = 0;
-    LOOP_OVER_IVECS(f->fc->gv, f->is, f->ie, idx) {
-      IVEC_LOOP_LOC(f->fc->gv, x0);
-      x0 = f->S.transform(x0, f->sn) + rshift;
-      for (int i = 0; i < Nfreq; ++i) {
-        double freq = freq_min + i * dfreq;
+#pragma omp parallel for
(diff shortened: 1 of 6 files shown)


src/near2far.cpp
```diff
@@ -250,12 +250,15 @@ void dft_near2far::farfield_lowlevel(std::complex<double> *EH, const vec &x) {
...
     vec rshift(f->shift * (0.5 * f->fc->gv.inva));
-    size_t idx_dft = 0;
-    LOOP_OVER_IVECS(f->fc->gv, f->is, f->ie, idx) {
-      IVEC_LOOP_LOC(f->fc->gv, x0);
-      x0 = f->S.transform(x0, f->sn) + rshift;
-      for (int i = 0; i < Nfreq; ++i) {
-        double freq = freq_min + i * dfreq;
+#ifdef HAVE_OPENMP
+#  pragma omp parallel for
+#endif
```
oskooi: For some reason, I am unable to compile with OpenMP via `--with-openmp` on my local machine using this `#HAVE_OPENMP` macro (i.e., the build is successful but the `examples/binary_grating_n2f.py` test uses just a single thread/process for the `get_farfields` calculation regardless of the value for `OMP_NUM_THREADS`). To fix this problem, I had to revert back to using just `#pragma omp parallel for` (i.e., removing the `#ifdef` and `#endif` lines) and `CXX="g++ -fopenmp"`. Perhaps this is related to the `openmp` -related changes in `configure.ac` in this PR?

stevengj: Whoop, I forgot to add `OPENMP_CXXFLAGS` to `CXXFLAGS` … try it again.

oskooi:
This is working and giving linear speed up as demonstrated previously in #861 and again as shown below for both serial and MPI (`np=2`) on a single, shared-memory, 8-core (Kaby Lake 4.2GHz) machine. The time for the `get_farfields` calculations was obtained directly from the `getting farfields` category of the `Field time usage` statistics following #856.

n2f_openmp_semilogy_2

n2f_openmp_mpi_np2_semilogy

As part of this PR, it would be good to also provide documentation in two places: (1) the `--with-openmp` configure flag in [Build from Source/Optional Dependencies](https://meep.readthedocs.io/en/latest/Build_From_Source/#optional-dependencies) and (2) for `get_farfields` and `output_farfields` in [Python User Interface/Near to Far Field Spectra](https://meep.readthedocs.io/en/latest/Python_User_Interface/#near-to-far-field-spectra). 

The related tutorial example [Near to Far Field Spectra/Diffraction Spectrum of a Finite Binary Grating](https://meep.readthedocs.io/en/latest/Python_Tutorials/Near_to_Far_Field_Spectra/#diffraction-spectrum-of-a-finite-binary-grating) will be updated in a separate PR with these results.

PR #870, opened 2019-05-11
fix minor typo

Maruoka842:
According to /meep/python/simulation.py, the function, run_k_points is defined as
```
def run_k_points(self, t, k_points):
```
So, the T in the document should be lower case, t.


stevengj:
Thanks!

PR #876, opened 2019-05-14
Add Python quiet function

### LINKED ISSUE #874: expose/document `quiet`
Right now you can access this with `mp.cvar.quiet`, but it would be nicer to have a function
```py
mp.quiet(quietval=True)
```
So that you can just do `mp.quiet()` to turn on quiet mode.

ChristopherHogan:
Closes #874.

### DIFF
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -938,7 +938,7 @@ Miscellaneous Functions
 
 **`meep.quiet(quietval=True)`**
 —
-Pass `True` (the default) to silence all Meep output. Output can be enabled again by passing `False`. This sets a global variable so the value will persist across runs within the same script.
+Meep ordinarily prints various diagnostic and progress information to standard output. This output can be suppressed by calling this function with `True` (the default). The output can be enabled again by passing `False`. This sets a global variable, so the value will persist across runs within the same script.
 
 ### Output File Names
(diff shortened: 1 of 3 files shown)


doc/docs/Python_User_Interface.md
```diff
@@ -936,6 +936,10 @@ Utility function that takes a `meep::volume` `vol` and returns the center and si
 Miscellaneous Functions
 -----------------------
 
+**`meep.quiet(quietval=True)`**
+—
+Pass `True` (the default) to silence all Meep output. Output can be enabled again by passing `False`. This sets a global variable so the value will persist across runs within the same script.
```
stevengj: Maybe explain what output is affected. For example:

> Meep ordinarily prints various diagnostic and progress information to standard output. This output can be suppressed by...

klein-j:
I like the idea of this patch, but I don't think it is going far enough. The amount of lines printed by meep where only reduced by about 20% in my program. The selection is also pretty unintuitive. "Field time usage" is omitted, but "Meep: using complex fields." or "Working in 1D dimensions" is still shown. Those are very particular and fine grained information that I am usually aware of rather than some sort of warnings. I don't see why those would still be displayed in silent mode.

In short I think it would be nice to silence all output (and even better if multiple levels of log details would be configurable).

PR #911, opened 2019-06-11
add ffmpeg package to build from source instructions and docs update

oskooi:
Since the Debian/Ubuntu [`ffmpeg`](https://tracker.debian.org/pkg/ffmpeg) package is ~100 MB and used only by [`Animate2D`](https://meep.readthedocs.io/en/latest/Python_User_Interface/#animate2d), we may want to consider adding a e.g. `--without-animation` configure flag to make this an optional dependency. For now, #909 provides a workaround.


stevengj:
`ffmpeg` is not used for `configure` or building, and now that #909 is merged the tests will pass without it.

PR #918, opened 2019-06-13
Explicit public API for fragment_stats

ChristopherHogan:
@stevengj @oskooi

### DIFF
--- a/src/meepgeom.hpp
+++ b/src/meepgeom.hpp
@@ -99,6 +99,12 @@ struct fragment_stats {
 
   fragment_stats() {}
   fragment_stats(geom_box &bx);
+
+  void compute();
+  double cost() const;
+  void print_stats() const;
+
+private:
   void update_stats_from_material(material_type mat, size_t pixels);
   void compute_stats();
   void count_anisotropic_pixels(medium_struct *med, size_t pixels);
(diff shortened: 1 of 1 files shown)


src/meepgeom.hpp
```diff
@@ -99,6 +99,12 @@ struct fragment_stats {
 
   fragment_stats() {}
   fragment_stats(geom_box &bx);
+
+  void compute();
+  double cost() const;
+  void print_stats() const;
```
stevengj: If we are going to make this public, it should probably take a `FILE*` argument defaulting to `stdout`.

ChristopherHogan: To clarify

1. These were always public. I'm just marking helper methods as `private`.

2. I don't mean "public" in the general C++ sense here. I don't expect users to ever use these functions from C++, since they only work if the simulation is run from Python (because Python currently passes things like dft volumes and pml volumes). 

3. Passing `split_chunks_evenly=False` implicitly uses `fragment_stats`, but the user doesn't have to know anything about it. The explicit use case for `fragment_stats` from Python (the entire API) is 

```python
sim = mp.Simulation(..., collect_stats=True)
sim.init_sim()
# get the cost of the whole cell
sim.fragment_stats.cost()
# print the number of each pixel type in the cell
sim.fragment_stats.print_stats()
```

Adding a `FILE*` argument to `print_stats` is problematic because (from the SWIG docs):
```
There is a list of known-to-be-broken features in Python 3:
  - No more support for FILE* typemaps, because PyFile_AsFile has been dropped in Python 3.
```
This is true for the new SWIG 4.0 as well.

PR #919, opened 2019-06-14
Enable Eigenmode Features with Dispersive Materials

### LINKED ISSUE #415: Enhancement: Medium Evaluation

smartalecH:
The finished version of #475.

Modifies the `get_eigenmode` routines such that dispersive materials are now supported. Specifically, I modified the `get_chi1inv()` function, which now evaluates the (real) inverse permittivity by summing all of the susceptibilities at a given point in a given direction for a specific field component at a specific frequency. It should be backward-compatible, as the default frequency is 0 (i.e. only the DC frequency is evaluated in this case).

As mentioned, it only evaluates the real permittivity, since MPB's iterative solver can (currently) only solve for real eigenvalues.

That being said, this method still enables a lot of important features, one of them being __S parameter extraction for dispersive materials__ (provided the materials are mostly dielectric).

Closes #415.

I've implemented a test that compares the actual permittivity, the return value of `chi1()`, and the "effective index" of a uniform medium (i.e. just the refractive index). All three should be the same.

I check this for Silicon, SIlver, and Lithium Niobate so that we get a range of susceptibilities.


smartalecH:
It looks like the serial implementation fails because the new routine doesn't like materials that have been rotated. It's able to pull the epsilon at those values just fine (using `sim.fields.get_chi1inv()`). But when MPB checks the dielectrics, it throws the "invalid dielectric function" error.

stevengj:
If MPB is giving that error it must be that your ε matrix is not positive-definite.

smartalecH:
This latest iteration fixes the original bug, which means that the routine can now use the `get_eigenmode()` routines with dispersive materials.

In addition, I modified the `array_slice` and the `h5fields` routines so that the user can optionally specify a frequency and they will evaluate the permittivity/permeability at that frequency. This is essential for the visualization module and useful for other post-processing applications.

The `array_slice` routines seem to be working fine (I've implemented additional tests within `dispersive_eigenmode.py`). The `h5` routines fail when executed in parallel. For some reason, there's a communication error with the `get_chi1inv_at_pt()` function in `monitor.cpp`. This is the same function used for the `get_eigenmode` routines and the `array_slice` routines (with no issues there).

I'm a little stumped, so if anyone else with more experience with parallel h5 generation could look at it, that would be great.

src/monitor.cpp
```diff
@@ -160,108 +160,139 @@ complex<double> fields_chunk::get_field(component c, const ivec &iloc) const {
...
+       if (conductivity[c][d]) {
+         double conductivityCur = conductivity[c][d][idx];
+         eps = std::complex<double>(1.0, (conductivityCur/omega)) * eps;
+       }
+      // Return chi1 inverse, take the real part since no support for loss in mpb yet
+      // TODO: Add support for metals
+      if (eps.imag() == 0.0){
+        res = 1.0 / eps.real();
+      }else{
+        res = 1.0 / (std::sqrt(eps).real() * std::sqrt(eps).real());
```
stevengj: What's going on here?   Why not just `res = real(1.0 / eps)`?

smartalecH:
Output:

```
Initializing structure...
Working in 2D dimensions.
Computational cell is 16 x 8 x 0 with resolution 10
     block, center = (0,0,0)
          size (1e+20,1,1)
-----------
Initializing structure...
Working in 2D dimensions.
Computational cell is 16 x 8 x 0 with resolution 10
     block, center = (0,0,0)
          size (1e+20,1,1)
time for set_epsilon = 0.0751069 s
lorentzian susceptibility: frequency=0.000905797, gamma=0
lorentzian susceptibility: frequency=0.88125, gamma=0
time for set_epsilon = 0.0731909 s
lorentzian susceptibility: frequency=0.000905797, gamma=0
lorentzian susceptibility: frequency=3.31657, gamma=0
lorentzian susceptibility: frequency=0.88125, gamma=0
lorentzian susceptibility: frequency=3.31657, gamma=0
-----------
-----------
creating output file "./test_animation-eps-000000.00.h5"...
creating output file "./test_animation-eps-000000.00.h5"...
HDF5-DIAG: Error detected in HDF5 (1.10.4) thread 0:
  #000: H5F.c line 444 in H5Fcreate(): unable to create file
    major: File accessibilty
    minor: Unable to open file
  #001: H5Fint.c line 1364 in H5F__create(): unable to open file
    major: File accessibilty
    minor: Unable to open file
  #002: H5Fint.c line 1615 in H5F_open(): unable to lock the file
    major: File accessibilty
    minor: Unable to open file
  #003: H5FD.c line 1640 in H5FD_lock(): driver lock request failed
    major: Virtual File Layer
    minor: Can't update object
  #004: H5FDsec2.c line 941 in H5FD_sec2_lock(): unable to lock file, errno = 35, error message = 'Resource temporarily unavailable'
    major: File accessibilty
    minor: Bad file ID accessed
meep: error on line 455 of ../../src/h5file.cpp: error opening HDF5 output file
Primary job  terminated normally, but 1 process returned
a non-zero exit code. Per user-direction, the job has been aborted.
mpirun detected that one or more processes exited with non-zero status, thus causing
the job to be terminated. The first process to do so was:

  Process name: [[39506,1],0]
  Exit code:    1
```

src/monitor.cpp
```diff
@@ -160,108 +160,139 @@ complex<double> fields_chunk::get_field(component c, const ivec &iloc) const {
...
+// Useful if you already know the exact location of the point you are interested in
+// (i.e. you know what idx should be). This is used with the get_array() routines.
+double structure_chunk::get_chi1inv_at_pt(component c, direction d, int idx, double omega) const {
+  double res = 0.0;
+  if (is_mine()){
+    res =
+        chi1inv[c][d] ? chi1inv[c][d][idx] : (d == component_direction(c) ? 1.0 : 0);
+    if (res != 0){
+      // Get instaneous dielectric (epsilon)
+      std::complex<double> eps(1.0 / res,0.0);
```
stevengj: This seems problematic for off-diagonal components, or in general for anisotropic materials, since you need a matrix inverse in that case.

src/monitor.cpp
smartalecH: What we really want is `real(1/n) ^2`, which is not the same as `real(1.0 / eps)` if `eps` is complex.

src/monitor.cpp
smartalecH: Addressed in 6eac066

python/geom.py
```diff
@@ -239,6 +239,29 @@ def transform(self, m):
         for s in self.H_susceptibilities:
             s.transform(m)
 
+    def rotate(self, rotations):
+        for rot in rotations:
+            if np.count_nonzero(rot) != 1:
+                raise ValueError("Each rotation vector should only have 1 coordinate.")
```
stevengj: would be more general to give a rotation vector (direction = axis, length = angle).  Formula: https://en.wikipedia.org/wiki/Rotation_matrix#Rotation_matrix_from_axis_and_angle

python/tests/dispersive_eigenmode.py
```diff
@@ -0,0 +1,147 @@
...
+import meep as mp
+import numpy as np
+from meep import mpb
+import h5py
+class TestDispersiveEigenmode(unittest.TestCase):
+    # ----------------------------------------- #
+    # ----------- Helper Functions ------------ #
+    # ----------------------------------------- #
+    # Directly cals the C++ chi1 routine
+    def call_chi1(self,material,omega):
```
stevengj: would be nice to have a documented API for this.    The natural thing would be an `epsilon(omega)` and `mu(omega)` method of material, but are these names already taken by fields?

stevengj: Oh, nevermind, we already have these.

src/monitor.cpp
```diff
@@ -160,108 +160,208 @@ complex<double> fields_chunk::get_field(component c, const ivec &iloc) const {
...
       }
   return 0.0;
 }
 
-double structure_chunk::get_chi1inv(component c, direction d, const ivec &iloc) const {
-  if (is_mine())
-    return chi1inv[c][d] ? chi1inv[c][d][gv.index(c, iloc)]
-                         : (d == component_direction(c) ? 1.0 : 0);
-  return 0.0;
+/* Set Vinv = inverse of V, where both V and Vinv are real-symmetric matrices.*/
```
stevengj: not symmetric

stevengj: also not real in general

src/monitor.cpp
```diff
@@ -160,108 +160,208 @@ complex<double> fields_chunk::get_field(component c, const ivec &iloc) const {
...
+  Vinv[1 + 3*0] = 1/det * (V[1 + 3*2]*V[2 + 3*0] - V[1 + 3*0]*V[2 + 3*2]);
+  Vinv[1 + 3*1] = 1/det * (V[0 + 3*0]*V[2 + 3*2] - V[0 + 3*2]*V[2 + 3*0]);
+  Vinv[1 + 3*2] = 1/det * (V[0 + 3*2]*V[1 + 3*0] - V[0 + 3*0]*V[1 + 3*2]);
+  Vinv[2 + 3*0] = 1/det * (V[1 + 3*0]*V[2 + 3*1] - V[1 + 3*1]*V[2 + 3*0]);
+  Vinv[2 + 3*1] = 1/det * (V[0 + 3*1]*V[2 + 3*0] - V[0 + 3*0]*V[2 + 3*1]);
+  Vinv[2 + 3*2] = 1/det * (V[0 + 3*0]*V[1 + 3*1] - V[0 + 3*1]*V[1 + 3*0]);
 }
 
-double structure::get_chi1inv(component c, direction d, const vec &loc, bool parallel) const {
+double structure_chunk::get_chi1inv_at_pt(component c, direction d, int idx, double omega) const {
```
stevengj: should have a fast branch for omega=0

src/monitor.cpp
```diff
@@ -160,108 +160,208 @@ complex<double> fields_chunk::get_field(component c, const ivec &iloc) const {
...
+    } else if (is_D(c)) {
+      comp_list[0] = Dx; comp_list[1] = Dy; comp_list[2] = Dz;
+      my_stuff = D_stuff;
+    } else if (is_B(c)) {
+      comp_list[0] = Bx; comp_list[1] = By; comp_list[2] = Bz;
+      my_stuff = B_stuff;
+    }
+
+    std::vector<double> chi1_inv_tensor(9,0);
+    std::vector<double> chi1_tensor(9,0);
```
stevengj: aren't these heap allocated, aren't they?  Maybe better to just have a `double chi1_tensor[9]`.

src/monitor.cpp
```diff
@@ -160,108 +160,208 @@ complex<double> fields_chunk::get_field(component c, const ivec &iloc) const {
...
+    for (int com_it=0; com_it<3;com_it++){
+      for (int dir_int=0;dir_int<3;dir_int++){
+        if (chi1inv[comp_list[com_it]][dir_int] )
+          chi1_inv_tensor[com_it + 3*dir_int] = chi1inv[comp_list[com_it]][dir_int][idx];
+        else if(dir_int == component_direction(comp_list[com_it]))
+          chi1_inv_tensor[com_it + 3*dir_int] = 1;
+        else
+          chi1_inv_tensor[com_it + 3*dir_int] = 0;
+      }
+    }
```
stevengj: couldn't this be a static initializer?  i.e. `double chi1_tensor[9] = {1,0,0, 0,1,0, 0,0,1};` ?

src/monitor.cpp
```diff
@@ -160,108 +160,208 @@ complex<double> fields_chunk::get_field(component c, const ivec &iloc) const {
...
+      for (int dir_int=0;dir_int<3;dir_int++){
+        std::complex<double> eps(chi1_tensor[com_it  + 3*dir_int],0.0);
+        component cc = comp_list[com_it];
+        direction dd = (direction)dir_int;
+        // Loop through and add up susceptibility contributions
+        // locate correct susceptibility list
+        susceptibility *my_sus = chiP[my_stuff];
+        while (my_sus) {
+          if (my_sus->sigma[cc][dd]) {
+            double sigma = my_sus->sigma[cc][dd][idx];
```
stevengj: This is a little worrisome — doesn't this correspond to different points in the Yee grid for the same `idx`?   Seems like you have to do some averaging with neighboring values like in h5 output and array slices.

stevengj:
Note that this is not really doing the interpolation from the Yee grid correctly, so it is making a first-order error in principle.  In practice, that error only shows up at the boundaries between materials, where it is hard to be accurate anyway, so let's just accept the 1st-order error for now and leave it to another PR to do the interpolation correctly.

src/monitor.cpp
smartalecH: addressed in e28c961

python/geom.py
smartalecH: addressed in e28c961

src/monitor.cpp
smartalecH: adressed in e28c961

src/monitor.cpp
smartalecH: addressed in e28c961

src/monitor.cpp
smartalecH: addressed in e28c961

smartalecH:
The PR should be ready for merging. Here's a summary of the features:
* Enable `get_eigenmode` routines with dispersive, anisotropic materials (excluding gyromagnetic materials, for now).
* `get_array` and `output_h5` routines now accept an `omega` parameter and can evaluate the permittivity/permeability at that frequency. Defaults to 0, which performs the previous behavior.
* Visualization module can accept an `omega` parameter to plot the dielectric at that frequency. Defaults to the center frequency of the first source in the `sources` list.
* Small bug fixes in the visualization module (with updated tests)
* Added a `rotate` function to the `Medium` class.

As discussed, a better interpolation scheme is needed to improve the accuracy around the material boundaries.

src/monitor.cpp
```diff
@@ -225,29 +225,31 @@ double structure::get_chi1inv(component c, direction d, const ivec &origloc, dou
...
-  Vinv[0 + 3*0] = 1/det * (V[1 + 3*1]*V[2 + 3*2] - V[1 + 3*2]*V[2 + 3*1]);
-  Vinv[0 + 3*1] = 1/det * (V[0 + 3*2]*V[2 + 3*1] - V[0 + 3*1]*V[2 + 3*2]);
-  Vinv[0 + 3*2] = 1/det * (V[0 + 3*1]*V[1 + 3*2] - V[0 + 3*2]*V[1 + 3*1]);
-  Vinv[1 + 3*0] = 1/det * (V[1 + 3*2]*V[2 + 3*0] - V[1 + 3*0]*V[2 + 3*2]);
-  Vinv[1 + 3*1] = 1/det * (V[0 + 3*0]*V[2 + 3*2] - V[0 + 3*2]*V[2 + 3*0]);
-  Vinv[1 + 3*2] = 1/det * (V[0 + 3*2]*V[1 + 3*0] - V[0 + 3*0]*V[1 + 3*2]);
-  Vinv[2 + 3*0] = 1/det * (V[1 + 3*0]*V[2 + 3*1] - V[1 + 3*1]*V[2 + 3*0]);
-  Vinv[2 + 3*1] = 1/det * (V[0 + 3*1]*V[2 + 3*0] - V[0 + 3*0]*V[2 + 3*1]);
-  Vinv[2 + 3*2] = 1/det * (V[0 + 3*0]*V[1 + 3*1] - V[0 + 3*1]*V[1 + 3*0]);
+  if (det.real() == 0 && det.imag() == 0) abort("meep: Matrix is singular, aborting.\n");
```
stevengj: doesn't `det == 0.0` work?

smartalecH: I was just being safe -- do you want it changed?

stevengj: `det == 0.0` seems clearer and is safe.

(You can't do `det == 0`, on the other hand, because C++ doesn't let you mix `complex<double>` with scalars of other types.)

smartalecH: Just fixed with 8590db1

src/monitor.cpp
```diff
@@ -160,108 +160,211 @@ complex<double> fields_chunk::get_field(component c, const ivec &iloc) const {
...
+            double sigma = my_sus->sigma[cc][dd][idx];
+            eps += my_sus->chi1(omega,sigma);
+          } 
+          my_sus = my_sus->next;
+        }
+
+        // Account for conductivity term
+        if (conductivity[cc][dd]) {
+          double conductivityCur = conductivity[cc][dd][idx];
+          eps = std::complex<double>(1.0, (conductivityCur/omega)) * eps;
```
stevengj, line 316: This line looks off to me — it seems like you are doing an element-wise product rather than a matrix multiplication?   

Since σ can only be a diagonal matrix currently, a matrix multiplication would multiply by `conductivity[cc][cc]` (the diagonal element from row `cc`), no?

stevengj:
What are the `step_func_args` that were added to some functions, like `output_epsilon`?  They don't seem to be used.

PR #927, opened 2019-06-19
print more timing statistics

### LINKED ISSUE #924: print more timing info
Right now, we collect timing info across all processes, but only print out the statistics from the master process.

It might be more informative to modify [`print_times`](https://github.com/NanoComp/meep/blob/4051db54fc26e816af1a0babda77019231427924/src/time.cpp#L65-L69) to print the average and variance (across processes) instead.  The variance (as a % of the average) would give some information about load-balancing efficiency.

stevengj:
Closes #924.

### DIFF
--- a/src/time.cpp
+++ b/src/time.cpp
@@ -58,14 +58,32 @@ static const char *ts2n(time_sink s) {
   return "everything else";
 }
 
-static void pt(double ts[], time_sink s) {
-  if (ts[s]) master_printf("    %21s: %g s\n", ts2n(s), ts[s]);
+static void pt(double mean[], double stddev[], time_sink s) {
+  if (mean[s] != 0) {
+    if (stddev[s] != 0)
+      master_printf("    %21s: %g s +/- %g s\n", ts2n(s), mean[s], stddev[s]);
+    else
+      master_printf("    %21s: %g s\n", ts2n(s), mean[s]);
+  }
 }
 
 void fields::print_times() {
+  double mean[Other + 1], square_times[Other + 1], stddev[Other + 1];
+  int n = count_processors();
+
+  for (int i = 0; i <= Other; ++i)
+    square_times[i] = times_spent[i] * times_spent[i];
+  sum_to_master(times_spent, mean, Other+1);
+  sum_to_master(square_times, stddev, Other+1);
+  for (int i = 0; i <= Other; ++i) {
+    mean[i] /= n;
+    stddev[i] -= n*mean[i]*mean[i]
+    stddev[i] = n == 1 || stddev[i] <= 0 ? 0.0 : sqrt(stddev[i] / (n-1));
+  }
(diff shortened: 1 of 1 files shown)


stevengj:
Would be nice to verify that the standard deviation of the timing is related to load balancing.

For example, you could try running the benchmark in #681, where we know that `split_by_cost` improved the load balancing compared to `split_by_effort`, and verify that the timing variance goes down as well for `split_by_cost`.

oskooi:
This seems to be working *but* the results for `split_by_cost` and `split_by_effort` are *swapped* which might indicate that there is a bug in the `split_chunks_evenly` parameter of the `Simulation` object that is mixing up the boolean values (i.e., `split_chunks_evenly=True` actually corresponds to `split_by_cost` rather than `split_by_effort`).

Running the unmodified example in #681 for the two cases and plotting the results (just the time-stepping portion of the simulation) is shown below.

**split_by_effort**
```
$ for n in `seq 4`; do mpirun -n ${n} python tall_skinny_cell_3d.py -e |grep "time stepping"; done
            time stepping: 1.68271 s
            time stepping: 0.966419 s +/- 0.00839127 s
            time stepping: 0.675613 s +/- 0.0180802 s
            time stepping: 0.553308 s +/- 0.0140741 s
```
**split_by_cost**
```
$ for n in `seq 4`; do mpirun -n ${n} python tall_skinny_cell_3d.py |grep "time stepping"; done
            time stepping: 1.68216 s
            time stepping: 1.57454 s +/- 0.0353577 s
            time stepping: 1.6826 s +/- 0.0899637 s
            time stepping: 1.73653 s +/- 0.0841767 s
```
time_split_cost_even

Note that the magnitude of the error bars for the "`split_by_effort`" case is smaller than those for "`split_by_cost`" which is expected if the results are swapped.

stevengj:
Anyway, it seems like the variance is doing what it is supposed to: it is larger in the case with poorer load-balancing.

src/time.cpp
```diff
@@ -58,14 +58,32 @@ static const char *ts2n(time_sink s) {
...
+  int n = count_processors();
+
+  for (int i = 0; i <= Other; ++i)
+    square_times[i] = times_spent[i] * times_spent[i];
+  sum_to_master(times_spent, mean, Other+1);
+  sum_to_master(square_times, stddev, Other+1);
+  for (int i = 0; i <= Other; ++i) {
+    mean[i] /= n;
+    stddev[i] -= n*mean[i]*mean[i];
+    stddev[i] = n == 1 || stddev[i] <= 0 ? 0.0 : sqrt(stddev[i] / (n-1));
```
stevengj: note that this is computing `stddev = ` ∑(xᵢ-x̄)² / (n-1) = (∑xᵢ² - nx̄²) / (n-1)

PR #931, opened 2019-06-21
more timing stats

stevengj:
Continuation of #927.   If `fields::verbose()` is set, then it prints out timing values from every process.

### DIFF
--- a/src/time.cpp
+++ b/src/time.cpp
@@ -85,6 +85,26 @@ void fields::print_times() {
   for (int i = 0; i <= Other; i++)
     pt(mean, stddev, (time_sink)i);
   master_printf("\n");
+
+  if (verbosity > 0) {
+    master_printf("\nField time usage for all processes:\n");
+    double *alltimes_tmp = new double[n * (Other+1)];
+    double *alltimes = new double[n * (Other+1)];
+    for (int i = 0; i <= Other; ++i) {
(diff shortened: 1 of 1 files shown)


oskooi:
```
$ mpirun -np 3 python ~/meep/python/examples/refl-angular.py
...
...
Field time usage:
        connecting chunks: 0.000521342 s +/- 1.58173e-05 s
            time stepping: 2.46249 s +/- 0.013184 s
            communicating: 0.91695 s +/- 0.0118404 s
     Fourier transforming: 0.248497 s +/- 0.338284 s
          everything else: 2.76524 s +/- 0.339444 s


Field time usage for all processes:
        connecting chunks: 0.000524998, 0.000504017, 0.000535011
            time stepping: 2.44917, 2.47553, 2.46278
          copying borders: 0, 0, 0
            communicating: 0.930335, 0.907844, 0.91267
        outputting fields: 0, 0, 0
     Fourier transforming: 0.639113, 0.0527964, 0.0535817
                      MPB: 0, 0, 0
        getting farfields: 0, 0, 0
          everything else: 2.37331, 2.9575, 2.96492
```

stevengj:
That looks good — note that the data is only printed to six significant digits, if you take this data and compute `std` in numpy or whatever, it will only reproduce the computed std dev approximately (the last digit will generally not match).

PR #937, opened 2019-06-27
Updated some Jupyter tutorials

smartalecH:


oskooi:
In each of the notebooks, why is it necessary to always specify the axis object via the figure handle as:
```
f = plt.figure(dpi=120)
sim.plot2D(ax=f.gca())
```
instead of just:
```
plt.figure(dpi=120)
sim.plot2D()
```
As mentioned in the [docs](https://meep.readthedocs.io/en/latest/Python_User_Interface/#data-visualization): "If no `ax` is supplied, then the routine will create a new figure and grab its axis."

smartalecH:
Right now, if the user doesn't specify an axis, then a new figure is _created_. This means that if you specify a figure before (let's call it figure 1), and don't pass it's axis, then everything will be plotted on a new figure (let's call it figure 2).

We could probably tweak this, but it gets very hairy with MPI (`matplotlib` just doesn't play well with MPI in general).

I know `sim.plot2D(ax=f.gca())` doesn't look as pretty as the alternative, but it actually protects the user in several cases (and enables some great functionality).

PR #944, opened 2019-07-02
no communications for isfinite check

stevengj:
Eliminates the communications noted in https://github.com/Simpetus/predict_walltime/issues/15#issuecomment-507769620, which is probably not expensive but confuses the timing statistics.

### DIFF
--- a/src/meep.hpp
+++ b/src/meep.hpp
@@ -1803,9 +1803,9 @@ public:
   void get_point(monitor_point *p, const vec &) const;
   monitor_point *get_new_point(const vec &, monitor_point *p = NULL) const;
 
-  std::complex<double> get_field(int c, const vec &loc) const;
+  std::complex<double> get_field(int c, const vec &loc, bool parallel = true) const;
   std::complex<double> get_field(component c, const vec &loc, bool parallel = true) const;
-  double get_field(derived_component c, const vec &loc) const;
+  double get_field(derived_component c, const vec &loc, bool parallel = true) const;
 
   // energy_and_flux.cpp
   void synchronize_magnetic_fields();
--- a/src/monitor.cpp
+++ b/src/monitor.cpp
@@ -63,11 +63,11 @@ void fields::get_point(monitor_point *pt, const vec &loc) const {
   }
 }
 
-complex<double> fields::get_field(int c, const vec &loc) const {
-  return (is_derived(c) ? get_field(derived_component(c), loc) : get_field(component(c), loc));
+complex<double> fields::get_field(int c, const vec &loc, bool parallel) const {
+  return (is_derived(c) ? get_field(derived_component(c), loc, parallel) : get_field(component(c), loc, parallel));
 }
 
-double fields::get_field(derived_component c, const vec &loc) const {
+double fields::get_field(derived_component c, const vec &loc, bool parallel) const {
   component c1 = Ex, c2 = Ex;
   double sum = 0;
   switch (c) {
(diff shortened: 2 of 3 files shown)


oskooi:
This PR seems to have done something strange to the timing statistics of the `everything else` bin even after changes in #946 are applied on top. This is demonstrated using an example of a tall, skinny, 2d cell with an expensive `FluxRegion` at one end run with 3 processors/chunks.

```py
import meep as mp

sx = 30
sy = 5
fcen = 1.0
df = 0.1
nfreq = 500

src = mp.Source(mp.GaussianSource(fcen,fwidth=df),
                component=mp.Ez,
                center=mp.Vector3())

                    sources=[src],
                    resolution=20,
                    verbose=True,
                    k_point=mp.Vector3(0.53,0.14,0),
                    split_chunks_evenly=True)

flux = sim.add_flux(fcen,
                    df,
                    nfreq,
                    mp.FluxRegion(center=mp.Vector3(14),size=mp.Vector3(y=sy)))                                                                                                                                          

sim.run(until=200)
```

**prior to this PR** (82ed0da)
```
Elapsed run time = 5.6079 s

Field time usage:
        connecting chunks: 0.00624348 s +/- 6.72307e-06 s
            time stepping: 0.791404 s +/- 0.0130607 s
            communicating: 0.10892 s +/- 0.00288693 s
     Fourier transforming: 1.5083 s +/- 2.61192 s
          everything else: 3.14133 s +/- 2.59801 s


Field time usage for all processes:
        connecting chunks: 0.00625037, 0.00624314, 0.00623694
            time stepping: 0.796284, 0.801322, 0.776606
          copying borders: 0, 0, 0
            communicating: 0.111922, 0.106163, 0.108674
        outputting fields: 0, 0, 0
     Fourier transforming: 0.000287041, 0.000334134, 4.52428
                      MPB: 0, 0, 0
        getting farfields: 0, 0, 0
          everything else: 4.64155, 4.64104, 0.141413
```


**this PR**
```
Elapsed run time = 5.5963 s

Field time usage:
        connecting chunks: 0.00596506 s +/- 0.000914024 s
            time stepping: 3.82357 s +/- 2.57174 s
            communicating: 0.214381 s +/- 0.0167141 s
     Fourier transforming: 1.4917 s +/- 2.58318 s
          everything else: 0.00347832 s +/- 0.00215778 s


Field time usage for all processes:
        connecting chunks: 0.00547808, 0.00539764, 0.00701946
            time stepping: 5.29906, 5.31765, 0.853997
          copying borders: 0, 0, 0
            communicating: 0.231732, 0.213024, 0.198387
        outputting fields: 0, 0, 0
     Fourier transforming: 0.000282357, 0.000314605, 4.4745
                      MPB: 0, 0, 0
        getting farfields: 0, 0, 0
          everything else: 0.0027478, 0.00178063, 0.00590654
```

The third chunk which contains the `FluxRegion` has a `Fourier transforming:` time which is much larger than that for the other two chunks. However, the time spent waiting for the one DFT chunk to finish `Fourier transforming` is now included as part of `time stepping` (**this PR**) rather than `everything else` (**prior to this PR**). This PR is incorrect.

What should be happening is that in `fields::step()`, [`update_dfts();`](https://github.com/NanoComp/meep/blob/master/src/step.hpp#L88) which is followed by `finished_working()`, should flip the final `working_on` timing state to the default/original `Other`. Instead it is being flipped to `Stepping`.

It looks like the timing stack [`was_working_on`](https://github.com/NanoComp/meep/blob/master/src/meep.hpp#1854) which has all elements [initialized to `Other`](https://github.com/NanoComp/meep/blob/master/src/fields.hpp#L51-52) is being overwritten to `Stepping` sometime during `am_now_working_on`. It's not clear what in this PR is causing this to occur.

stevengj:
> However, the time spent waiting for the one DFT chunk to finish Fourier transforming is now included as part of time stepping (this PR) rather than everything else (prior to this PR). This PR is incorrect.

It doesn't follow that this PR is incorrect.  It just means that there is some communication (like a `sum_to_all` call) elsewhere in the timestepping code that is forcing a barrier, and hence is causing the wait time to be included in the timestepping.  The solution is to wrap `am_working(MpiTime)` around those other communications (once they are found).

That is, this PR removed a synchronization point.   Hence synchronization got moved somewhere else, and unfortunately that other point isn't currently in `am_working(MpiTime)` .  That's not a bug in this PR.

PR #960, opened 2019-07-14
Allow iterables in place of Vector3 args

### LINKED ISSUE #928: accept tuples instead of `meep.Vector3` arguments
For arguments that take a `Vector3`, e.g. `size=...`, it would be nice to take a tuple instead.

Indeed, for `size=foo` and similar, maybe we can just call `meep.Vector3(*foo)` to splat the arguments.  That way, you could use `size=[3,4,5]`, `size=(3,4,5)`, `size=meep.Vector3(3,4,5)`, and so on.

ChristopherHogan:
Closes #928 
@stevengj @oskooi

### DIFF
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -1001,6 +1001,20 @@ Return the current simulation time in simulation time units (e.g. during a run f
 
 Occasionally, e.g. for termination conditions of the form *time* &lt; *T*?, it is desirable to round the time to single precision in order to avoid small differences in roundoff error from making your results different by one timestep from machine to machine (a difference much bigger than roundoff error); in this case you can call `Simulation.round_time()` instead, which returns the time rounded to single precision.
 
+**`Simulation.print_times()`**
+—
+Call after running a simulation to print the times spent on various types of work. Example output:
+
+```
+Field time usage:
+        connecting chunks: 0.0819176 s +/- 0.000428381 s
+            time stepping: 0.198949 s +/- 0.0225551 s
+            communicating: 0.410577 s +/- 0.278853 s
+        outputting fields: 0.512352 s +/- 0.0238399 s
+     Fourier transforming: 0.0738274 s +/- 0.0967926 s
+          everything else: 0.324933 s +/- 0.377573 s
+```
+
 ### Field Computations
 
 Meep supports a large number of functions to perform computations on the fields. Most of them are accessed via the lower-level C++/SWIG interface. Some of them are based on the following simpler, higher-level versions. They are accessible as methods of a `Simulation` instance.
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -438,7 +438,7 @@ Properties:
 
 **`x`, `y`, `z` [`float` or `complex`]**
 —
-The `x`, `y`, and `z` components of the vector.
+The `x`, `y`, and `z` components of the vector. Generally, functions that take a `Vector3` as an argument will accept an iterable (e.g., a tuple or list) and automatically convert to a `Vector3`.
 
 **`Vector3(x=0.0, y=0.0, z=0.0)`**
 —
--- a/python/geom.py
+++ b/python/geom.py
@@ -202,20 +202,20 @@ class Medium(object):
         if H_chi3:
             H_chi3_diag = Vector3(H_chi3, H_chi3, H_chi3)
 
-        self.epsilon_diag = epsilon_diag
-        self.epsilon_offdiag = epsilon_offdiag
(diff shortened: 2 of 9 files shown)


oskooi:
2 of the 8 tests on Travis (both for Python 3.7) are failing due to a timeout. Restarting them does not seem to help. Otherwise, this should be good to go.

stevengj:
Ardavan spotted a few other places where we might want to splat the arguments, in particular for `kpoint_func` and `eig_kpoint`.

stevengj:
The travis failure is a bit of a concern to me because it doesn't seem like it is happening on other PRs, though I don't see why it would be caused by this PR.

smartalecH:
I ran this on my local machine (python3.7) with mpi and mpb and it passed. Not sure what's up.

ChristopherHogan:
#963 exhibits similar behavior (hang on MPI Py3). I found a reliable local reproducer (`mpirun -n 2 python tests/visualization.py` in this PR WITHOUT the "Fix MPI hang in Py3 test" commit, which Alec fixed in #963) and was able to see an MPI deadlock where `rank 0` was waiting at `time.cpp:83 sum_to_master()`, while `rank 1` was waiting at `fields.cpp:508 or_to_all()`. I don't have time to investigate further tonight.

ChristopherHogan:
Valgrind didn't report any memory issues in the test.

smartalecH:
> (mpirun -n 2 python tests/visualization.py in this PR WITHOUT the "Fix MPI hang in Py3 test" commit, which Alec fixed in #963)

Good find. #963 revamps a lot of deadlocking within some of the `Animate2D` stepping routines, in addition to a revised test `visualization.py` (similar to your commit). I wonder if something else is going on too, however.

It's odd that the ranks are stopping where they are, don't you think?

ChristopherHogan:
Right, I don't think it's necessarily a problem with the `visualization` test. I was just saying that's the only reliable reproducer of the deadlock that I've seen.

ChristopherHogan:
I think I see what's going on. It comes down to the fact that Python is under no obligation to run garbage collection at any specific time, which means we can't control when `~fields` is called. In the case of the reproducer here we have the following flow:

1. Ranks 0 and 1 run `test_animation_output`
2. Rank 1 finishes first and moves on to `test_plot2D`, where it creates a new `sim` and waits for synchronization at the `or_to_all` in `fields::require_component`. For whatever reason, Python didn't do any garbage collection, so `~fields`, and hence, `print_times` hasn't been called on rank 1.
3. Rank 0 finishes `test_animation_output`. Here Python decides to do garbage collection, so the local `sim` object is reclaimed, which triggers `~fields`, and hence, `print_stats`, which waits at `sum_to_all`, but since rank 1 has skipped this path, deadlock results.

I verified this theory by forcing garbage collection at the end of `test_animation_output`.
```py
del sim
import gc; gc.collect()
```
In that case both `~fields` calls trigger at the correct time and everything is good.

This seems to imply that we can't rely on MPI synchronization in destructors. Thoughts? I'll dig into the SWIG docs.

stevengj:
> This seems to imply that we can't rely on MPI synchronization in destructors.

That seems to be a good policy to follow.

We should just get rid of the `print_times` call in `~fields` and call it explicitly whenever we need timing statistics.

ChristopherHogan:
Removed `print_times` from `~fields`. It's now required to call it explicitly via `sim.print_times()`.

> Ardavan spotted a few other places where we might want to splat the arguments, in particular for kpoint_func and eig_kpoint.

Done for `eig_kpoint`. `kpoint_func` takes a function, not a `Vector3`. Maybe you were thinking of `Simulation.k_point`? I didn't change that yet because that has special handling for values of `False`, and empty tuples and lists evaluate to `False`. I'd need to spend some more time on that to maintain backwards compatibility.

oskooi:
This PR does not permit the following:

replacing:

https://github.com/NanoComp/meep/blob/0d4f4afdcc3f97b845b07ce75d97ceb5138c8ba6/python/examples/binary_grating_oblique.py#L38-L39

with:

```
k = (fcen*ng).rotate((z=1), theta_in)
```

Is there no way to set this up using the iterable construct? Without the explicit `Vector3` in front of the parentheses, it does become a little difficult to read but nevertheless simplifies the expression.

ChristopherHogan:
> Is there no way to set this up using the iterable construct?

I don't believe there is a way. `rotate` is a method on `Vector3`, but you want to call it on a `tuple`. For that to work we would have to subclass the Python `tuple` builtin and add a `rotate` method, but even then, the literal `()` would still refer to the builtin, not the subclass, so you would have to write something like `rotatable_tuple().rotate()`, which isn't any different from writing `Vector3().rotate()`.

PR #962, opened 2019-07-16
print actual number of chunks used in choose_chunkdivision rather than desired number

oskooi:
Fixes a small bug in `structure::choose_chunkdivision` to print the actual number of chunks rather than the desired number (which may not be equivalent if the desired number is a prime number larger than `5`).


src/vec.cpp
```diff
@@ -1068,11 +1068,10 @@ grid_volume grid_volume::split_by_cost(int desired_chunks, int proc_num) const {
   grid_volume split_gv =
       split_at_fraction(!split_low, split_point, best_split_direction, num_in_split_dir);
 
-  if (split_low) {
+  if (split_low)
     return split_gv.split_by_cost(num_low, proc_num);
-  } else {
+  else
     return split_gv.split_by_cost(desired_chunks - num_low, proc_num - num_low);
-  }
```
stevengj: These changes are harmless, but will be reverted by the next time we run `clang-format`

stevengj: I'm fine with changing the style in this way (omit braces for single-line blocks, and maybe put `else` on its own line rather than grouping it with the closing brace), but we should dig through the `clang-format` docs and find out how to change this in the `meep/.clang-format` file so that the change will be applied consistently to the entire codebase.

PR #966, opened 2019-07-16
bisection search for faster splitting

stevengj:
This updates the `split_by_cost` function to do a bisection search along each axis to find a balanced splitting, rather than searching every possible splitting, which should speed things up for large simulations.

(In most cases the left and right costs should be monotonic functions of the splitting point, in which cases a bisection search should be equivalent to checking every possible point.)

`split_into_three` should have similar improvements, but that isn't touched by this PR.

### DIFF
--- a/src/meep/vec.hpp
+++ b/src/meep/vec.hpp
@@ -1030,6 +1030,9 @@ public:
 
 private:
   std::complex<double> get_split_costs(direction d, int split_point) const;
+  grid_volume split_by_cost(int desired_chunks, int proc_num,
+                            int best_split_point, direction best_split_direction, double left_effort_fraction) const;
+  void find_best_split(int desired_chunks, int &best_split_point, direction &best_split_direction, double &left_effort_fraction) const;
   grid_volume(ndim d, double ta, int na, int nb, int nc);
   ivec io;    // integer origin ... always change via set_origin etc.!
   vec origin; // cache of operator[](io), for performance
--- a/src/meep/vec.hpp
+++ b/src/meep/vec.hpp
@@ -1029,6 +1029,7 @@ public:
   double origin_z() const { return origin.z(); }
 
 private:
+  std::complex<double> get_split_costs(direction d, int split_point) const;
   grid_volume(ndim d, double ta, int na, int nb, int nc);
   ivec io;    // integer origin ... always change via set_origin etc.!
   vec origin; // cache of operator[](io), for performance
--- a/src/vec.cpp
+++ b/src/vec.cpp
@@ -1002,6 +1002,28 @@ double grid_volume::get_cost() const {
   return fstats.cost();
 }
 
+// return complex(left cost, right cost).  Should really be a tuple, but we don't want to require C++11? yet?
+std::complex<double> grid_volume::get_split_costs(direction d, int split_point) const {
+  double left_cost = 0, right_cost = 0;
+  if (split_point > 0) {
+    grid_volume v_left = *this;
+    v_left.set_num_direction(d, split_point);
+    left_cost = v_left.get_cost();
+  }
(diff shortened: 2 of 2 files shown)


oskooi:
This is failing `python/tests/ring-cyl.py` because it contains [`split_chunks_evenly=False`](https://github.com/NanoComp/meep/blob/master/python/tests/ring_cyl.py#L50). Since the fragment statistics (#681) does not support cylindrical coordinates, the fix is simple, e.g.:

`vec.cpp:1032`

change
```
   if (desired_chunks == 1) return *this;
```
to
```
   if (desired_chunks == 1 || dim == Dcyl) return *this;
```

oskooi:
`python/tests/chunks.py` (the only other test involving `split_chunks_evenly=False`) seems to hang indefinitely when run with `4`, `6`, or `8` processors whereas with `2` processors finishes in about a second:
```
$ mpirun -n 2 python/tests/chunks.py
Using MPI version 3.0, 2 processes
-----------
Initializing structure...
Splitting into 2 chunks by cost
time for choose_chunkdivision = 0.000123024 s
```

stevengj:
The `ring-cyl.py` test was working before, so I don't see why it should have suddenly stopped working?

I noticed a problem where my bisection search might not converge and hopefully fixed it.

oskooi:
Output from `sim.visualize_chunk()` using 8 processors/chunks:

chunk_layout_8

oskooi:
Output from master for 8 chunks/processors:

chunk_layout_master_8

stevengj:
Good, so it is coming up with the same (or almost the same…could differ by ±1 pixel) chunking.

stevengj:
Independent of this PR, it would be nice to understand why some of the corner chunks are a different process (color) than the surrounding chunks — to minimize communication costs, it seems like we would want to assign each process to a contiguous set of chunks where possible.

oskooi:
The following are the chunk layouts for a tall, skinny, 2d cell with an expensive flux region at one end for the `split_by_cost` algorithm from master and this PR for a range of processors/chunks: `4`, `6`, `8`, `10`, and `12`. The results are similar but not identical. In particular, note that for the case of `4` and `6` processors/chunks, the bisection search of this PR produces `4`/`6` equally-sized chunks (which is correct) but that the linear search of master produces unequal-sized chunks.

```py
import meep as mp

sx = 30
sy = 5
fcen = 1.0
df = 0.1
nfreq = 500

src = mp.Source(mp.GaussianSource(fcen,fwidth=df),
                component=mp.Ez,
                center=mp.Vector3())

                    sources=[src],
                    resolution=20,
                    verbose=True,
                    k_point=mp.Vector3(0.53,0.14,0),
                    split_chunks_evenly=False)

flux = sim.add_flux(fcen,
                    df,
                    nfreq,
                    mp.FluxRegion(center=mp.Vector3(14),size=mp.Vector3(y=sy)))

sim.init_sim()
sim.visualize_chunks()
if mp.am_master():
    plt.savefig('imbalanced_2d_chunk_layout.png')
```

**master**

`4` processors/chunks

imbalanced_2d_chunk_layout_master_4

`6` processors/chunks

imbalanced_2d_chunk_layout_master_6

`8` processors/chunks

imbalanced_2d_chunk_layout_master_8

`10` processors/chunks

imbalanced_2d_chunk_layout_master_10

`12` processors/chunks

imbalanced_2d_chunk_layout_master_12

**this PR**

`4` processors/chunks

imbalanced_2d_chunk_layout_4

`6` processors/chunks

imbalanced_2d_chunk_layout_6

`8` processors/chunks

imbalanced_2d_chunk_layout_8

`10` processors/chunks

imbalanced_2d_chunk_layout_10

`12` processors/chunks

imbalanced_2d_chunk_layout_12

PR #968, opened 2019-07-17
Fix find_k band_func ordering

### LINKED ISSUE #851: find_k only stores single eigenmode field info
It seems that the `find_k` function only records the first eigenmode's field into the ModeSolver object, even when it solves for multiple bands. Consequently, functions like `ModeSolver.get_efield()` can't access the different bands (or modes).

When trying to access multiple modes after a simulation, it throws the warning:
`Must have 1 <= band index <= num_bands (1)`

Here's a MWE:
```python
import meep as mp
from meep import mpb
import numpy as np


Si   = mp.Medium(index=3.45)
SiO2 = mp.Medium(index=1.45)
wavelength = 1.55
omega      = 1/wavelength
w = 1    # Si width (um)
h = 0.5  # Si height (um)
geometry = [mp.Block(size=mp.Vector3(mp.inf, w, h), material=Si)]
resolution = 32  # pixels/um
sc_y = 2  # supercell width (um)
sc_z = 2  # supercell height (um)
geometry_lattice = mp.Lattice(size=mp.Vector3(0, sc_y, sc_z))

num_modes = 4
# Setup a simulation object
ms = mpb.ModeSolver(
    geometry_lattice = geometry_lattice,
    geometry         = geometry,
    resolution       = resolution,
    num_bands        = num_modes
)

# ------------------------------------------------- #
# Solve for all the modes at once, then plot
# ------------------------------------------------- #

k = ms.find_k(
    p              = mp.EVEN_Z,
    omega          = omega,
    band_min       = 1,
    band_max       = num_modes,
    korig_and_kdir = mp.Vector3(1,0,0),
    tol            = 1e-4,
    kmag_guess     = omega * 3.45,
    kmag_min       = omega * 0.1,
    kmag_max       = omega * 4
)
print(wavelength * np.array(k))
eps = ms.get_epsilon()
plt.figure(figsize=(8,4))
# Plot the E fields
for mode in range(num_modes):
    print('Current band: {}'.format(mode+1))
    E   = ms.get_efield(which_band=mode+1,bloch_phase=False)
    plt.subplot(1,num_modes,mode+1)
    plt.imshow(eps.transpose(), interpolation='spline36', cmap='binary')
    plt.imshow(np.abs(np.squeeze(E[:,:,0,1]).transpose())**2, cmap='RdBu', alpha=0.9)
    plt.axis('off')
    plt.title('$E_{{y{}}}$'.format(mode))
plt.tight_layout()
plt.savefig('errorFields.png')

# ------------------------------------------------- #
# Solve for all the modes one by one
# ------------------------------------------------- #

plt.figure(figsize=(8,4))
for mode in range(num_modes):
    k = ms.find_k(
        p              = mp.EVEN_Z,
        omega          = omega,
        band_min       = mode+1,
        band_max       = mode+1,
        korig_and_kdir = mp.Vector3(1,0,0),
        tol            = 1e-4,
        kmag_guess     = omega * 3.45,
        kmag_min       = omega * 0.1,
        kmag_max       = omega * 4
    )
    E   = ms.get_efield(which_band=mode+1,bloch_phase=False)
    plt.subplot(1,num_modes,mode+1)
    plt.imshow(eps.transpose(), interpolation='spline36', cmap='binary')
    plt.imshow(np.abs(np.squeeze(E[:,:,0,1]).transpose())**2, cmap='RdBu', alpha=0.9)
    plt.axis('off')
    plt.title('$E_{{y{}}}$'.format(mode))
plt.tight_layout()
plt.savefig('correctFields.png')

plt.show()
```

The problematic routine simply routines the fundamental mode after each call:

errorFields

I can get the correct field profile if I iterate through the solver mode by mode (and only solve for that particular band):

correctFields

Hopefully this is just a small tweak in the python interface. I can look into it.

smartalecH:
Fixes #851.

### DIFF
--- a/python/solver.py
+++ b/python/solver.py
@@ -1040,7 +1040,7 @@ class ModeSolver(object):
             ks.append(mp.find_root_deriv(rootfun(b), tol, kmag_min, kmag_max, k0s[b - band_min]))
 
         if band_funcs:
-            for b, k in zip(range(band_max, band_max - nb, -1), reversed(ks)):
+            for b, k in zip(range(1, band_max +1), reversed(ks)):
                 self.num_bands = b
                 self.k_points = [korig + kdir1.scale(k)]
 


python/solver.py
```diff
@@ -1040,7 +1040,7 @@ def _rootfun(k):
             ks.append(mp.find_root_deriv(rootfun(b), tol, kmag_min, kmag_max, k0s[b - band_min]))
 
         if band_funcs:
-            for b, k in zip(range(band_max, band_max - nb, -1), reversed(ks)):
+            for b, k in zip(range(1, band_max +1), reversed(ks)):
```
stevengj: ks shouldn't be reversed now?

smartalecH: No — I think that was a bug before. The Ks are generated in reverse order to begin with.

PR #969, opened 2019-07-18
contrib: Build bash script for end users

d-a-v:
Automate full installation to any directory (not only `/usr/local` - *edit4*: now selectable as argument)
    
This script is very useful for
    
- users with electromagnetics skills in the need for the latest MEEP developments but lacking time to get sufficient system knowledge to reach their grail,
    
- system administrators constantly being asked for the latest MEEP versions by their impatient electromagnetics specialist users (sometimes for years)
    
This script has its place right in the source repository, so the (large) audience can improve and adapt it over time.
    
Based on https://github.com/NanoComp/meep/blob/9f68f390f68ecb8a18d98f51c8166c808a1e6f98/doc/docs/Build_From_Source.md#ubuntu-1604-and-1804
Build for ubuntu16.04, 18.04 and CentOS 7

*edit:* documentation update is [visible there](https://github.com/d-a-v/meep/blob/build/doc/docs/Build_From_Source.md#building-from-source)
*edit2:* Centos7 was not finishing the full build, fix included
*edit3:* works with the 3 OSes on a raw docker machine

### DIFF
--- a/contrib/build-meep.sh
+++ b/contrib/build-meep.sh
@@ -54,14 +54,6 @@ SUDO=
 mkdir -p ${SRCDIR}
 cd ${SRCDIR}
 
-RPATH_FLAGS="-Wl,-rpath,${DESTDIR}/lib:/usr/lib/x86_64-linux-gnu/hdf5/openmpi"
-LDFLAGS="-L${DESTDIR}/lib -L/usr/lib/x86_64-linux-gnu/hdf5/openmpi ${RPATH_FLAGS}"
-CFLAGS="-I${DESTDIR}/include -I/usr/include/hdf5/openmpi"
-CPPFLAGS=${CFLAGS}
-PKG_CONFIG_PATH=${DESDTIR}/pkgconfig
-export PKG_CONFIG_PATH
-export PATH=${DESTDIR}/bin:${PATH}
-
 gitclone ()
 {
     repo=${1##*/}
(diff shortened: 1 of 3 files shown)


contrib/build-meep.sh
```diff
@@ -0,0 +1,227 @@
+#!/bin/bash
+
+# latest version of this script can be found at:
+#   https://github.com/NanoComp/meep/blob/master/contrib/build-meep.sh
+
+# detect if script is in a directory called src/
+DESTDIR=$(pwd)
```
stevengj: Might be nice to get this from an argument to the script rather than using `pwd`.

contrib/build-meep.sh
```diff
@@ -0,0 +1,227 @@
...
+    else
+        git clone --depth=1 $1
+    fi
+}
+
+autogensh ()
+{
+    sh autogen.sh PKG_CONFIG_PATH="${PKG_CONFIG_PATH}" RPATH_FLAGS="${RPATH_FLAGS}" LDFLAGS="${LDFLAGS}" CFLAGS="${CFLAGS}" CPPFLAGS="${CPPFLAGS}" \
+        --disable-static --enable-shared --prefix="${DESTDIR}" \
+        --with-libctl=${DESTDIR}/share/libctl \
```
stevengj: Maybe also need `GEN_CTL_IO=${DESTDIR}/bin/gen-ctl-io`

d-a-v: Thanks for reviewing!
It is not necessary thanks to `export PATH=${DESTDIR}/bin:${PATH}`

contrib/build-meep.sh
d-a-v: That's true. It is now available.

d-a-v:
FWIW:
```
myusername@Windows10Host:~/meep$ cat /proc/version
Linux version 4.4.0-18362-Microsoft (Microsoft@Microsoft.com) (gcc version 5.4.0 (GCC) ) #1-Microsoft Mon Mar 18 12:02:00 PST 2019
myusername@Windows10Host:~/meep$ export PYTHONPATH=/home/myusername/meep/lib/python3.6/site-packages
myusername@Windows10Host:~/meep$ python3
Python 3.6.8 (default, Jan 14 2019, 11:02:34)
[GCC 8.0.1 20180414 (experimental) [trunk revision 259383]] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import meep
>>>
```

amlaas:
I just used this script under both Linux Ubuntu 18.04 64bits and  WSL on Windows10.
In both cases, it worked perfectly and compiled/installed everything needed to use meep/mpb with either the scheme or the python interface.
Thanks!

Antoine

PR #975, opened 2019-07-28
Treat fragments with object interfaces as anisotropic

### LINKED ISSUE #959: Account for anisotropic pixels on interfaces in fragment stats
> Moreover, right now our fragment statistics don't take into account the fact that anisotropic pixels are created everywhere you have interfaces. To handle this, if any object interface intercepts a fragment box, the entire box should be treated as anisotropic. -Steven

@stevengj, can you elaborate on this?
1. What do you mean by "object interface?"
2. To "treat a box as anisotropic", I assume I would do something like:
```
if (object interface intercepts)
  num_anisotropic_eps_pixels += num_pixels_in_box
  if (has_mu)
    num_anisotropic_mu_pixels += num_pixels_in_box
```

ChristopherHogan:
Closes #959. 
@stevengj @oskooi

### DIFF
--- a/src/meepgeom.cpp
+++ b/src/meepgeom.cpp
@@ -1647,6 +1647,7 @@ std::vector<meep::volume> fragment_stats::pml_2d_vols;
 std::vector<meep::volume> fragment_stats::pml_3d_vols;
 std::vector<meep::volume> fragment_stats::absorber_vols;
 bool fragment_stats::split_chunks_evenly = false;
+bool fragment_stats::eps_averaging = false;
 
 static geom_box make_box_from_cell(vector3 cell_size) {
   double edgex = cell_size.x / 2;
@@ -1693,7 +1694,7 @@ fragment_stats compute_fragment_stats(
     material_type default_mat, std::vector<dft_data> dft_data_list_,
     std::vector<meep::volume> pml_1d_vols_, std::vector<meep::volume> pml_2d_vols_,
     std::vector<meep::volume> pml_3d_vols_, std::vector<meep::volume> absorber_vols_, double tol,
-    int maxeval, bool ensure_per) {
+    int maxeval, bool ensure_per, bool eps_averaging) {
 
   fragment_stats::geom = geom_;
   fragment_stats::dft_data_list = dft_data_list_;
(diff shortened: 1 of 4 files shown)


src/meepgeom.cpp
```diff
@@ -1765,17 +1768,30 @@ void fragment_stats::compute_stats() {
...
+    // an object interface intercepts the box, which means we treat
+    // the entire box as anisotropic.
+    bool anisotropic_pixels_already_added = false;
+    if (overlap != 1.0) {
+      anisotropic_pixels_already_added = true;
+      num_anisotropic_eps_pixels += num_pixels_in_box;
+      if (mu_not_1(go->material)) {
+        num_anisotropic_mu_pixels += num_pixels_in_box;
+      }
+    }
```
stevengj: This only creates anisotropy if eps_averaging = True

src/meepgeom.cpp
```diff
@@ -1765,17 +1768,30 @@ void fragment_stats::compute_stats() {
     geometric_object *go = &geom.items[i];
     double overlap = box_overlap_with_object(box, *go, tol, maxeval);
 
+    // If the object doesn't overlap the entire box, that implies that
+    // an object interface intercepts the box, which means we treat
+    // the entire box as anisotropic.
```
stevengj: Maybe add a comment that this could give some false positives if there is another object with the same material behind the current object, but in practice it is probably reasonable to assume that there is a material interface somewhere in the box so we won't worry about fancier edge-detection methods for now.

src/meepgeom.cpp
```diff
@@ -1768,15 +1770,21 @@ void fragment_stats::compute_stats() {
...
+      // an object interface intercepts the box, which means we treat
+      // the entire box as anisotropic. This method could give some false
+      // positives if there is another object with the same material behind
+      // the current object, but in practice it is probably reasonable to
+      // assume that there is a material interface somwhere in the box so
+      // we won't worry about fancier edge-detection methods for now.
+      if (overlap != 1.0) {
+        anisotropic_pixels_already_added = true;
+        num_anisotropic_eps_pixels += num_pixels_in_box;
+        if (mu_not_1(go->material)) {
```
stevengj: This could give false negatives, since `go` could be something like vacuum, and it could be the material behind with mu.

But mu is not that common anyway so I'm inclined not to worry about it for now.

PR #976, opened 2019-07-29
Bisection search for split_into_three

oskooi:
Extends the bisection search approach of #966 to `grid_volume::split_into_three` which was missing.

### DIFF
--- a/src/vec.cpp
+++ b/src/vec.cpp
@@ -1127,77 +1127,29 @@ grid_volume grid_volume::split_by_cost(int desired_chunks, int proc_num, int bes
 }
 
 void grid_volume::split_into_three(std::vector<grid_volume> &result) const {
+  int best_low_split_point;
+  direction best_low_split_direction;
+  double left_effort_fraction;
+  find_best_split(3, best_low_split_point, best_low_split_direction, left_effort_fraction);
 
-  grid_volume best_low, best_mid, best_high;
-  double best_overall_split_measure = 1e20;
-  int best_low_split_point = 0;
-  int best_high_split_point = 0;
+  grid_volume low_gv = split_at_fraction(false, best_low_split_point, best_low_split_direction,
+                                         num_direction(best_low_split_direction));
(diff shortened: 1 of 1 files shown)


oskooi:
As a demonstration of the performance enhancement, the time for `choose_chunkdivision` with `split_chunks_evenly=False` for an [OLED](http://www.simpetus.com/projects.html#meep_oled) test case for `6` processes/chunks is `153.414 s` using this PR and `746.713 s` on master. This is a nearly 5X speedup.

src/vec.cpp
```diff
@@ -1127,77 +1127,29 @@ grid_volume grid_volume::split_by_cost(int desired_chunks, int proc_num, int bes
...
-      best_high = high_gv;
-    }
-  }
-  result.push_back(best_low);
-  result.push_back(best_mid);
-  result.push_back(best_high);
+  int best_low_split_point;
+  direction best_low_split_direction;
+  double left_effort_fraction;
+  find_best_split(3, best_low_split_point, best_low_split_direction, left_effort_fraction);
```
stevengj: I don't think the `find_best_split` method is really correct right now for `desired_chunks != 2`.   That's why @ChristopherHogan had a separate `split_into_three` function in the first place?

stevengj: At first glance, I think we should change two places:

1. `cost_diff` should return `right_cost / (1 - 1.0/desired_chunks) - left_cost * desired_chunks`;

2. `find_best_split` should set `split_measure` to `max(left_cost * desired_chunks, right_cost / (1 - 1.0/desired_chunks))`

stevengj: For example, if `desired_chunks == 3` then you would ideally like the `left_cost` to be 1/3 of the total and `right_cost` to be `2/3` of the total, so basically you want `right_cost == (desired_chunks-1) * left_cost

stevengj: For the specific cases of `desired_chunks == 1` or `== 2`, my proposal is equivalent to what it is doing now.  But I think it would be better to change it to my formula, which seems more general.

PR #984, opened 2019-08-09
Fine-grained timing statistics for structure class

oskooi:
The `structure` class contains four routines &mdash; `set_epsilon`, `set_conductivity`, `choose_chunkdivision`, and `add_susceptibility` &mdash; which in certain situations require non-negligible runtimes. This PR times these four routines separately and stores the results in a public variable which can be used for e.g. improving performance, debugging, etc.

### DIFF
--- a/src/meep.hpp
+++ b/src/meep.hpp
@@ -714,6 +714,7 @@ public:
   grid_volume *effort_volumes;
   double *effort;
   int num_effort_volumes;
+  double t_setepsilon, t_setconductivity, t_chunkdivision, t_addsusceptibility;
 
   ~structure();
   structure();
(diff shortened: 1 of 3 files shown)


src/meep.hpp
```diff
@@ -714,6 +714,7 @@ class structure {
   grid_volume *effort_volumes;
   double *effort;
   int num_effort_volumes;
+  double t_setepsilon, t_setconductivity, t_chunkdivision, t_addsusceptibility;
```
stevengj: This should be done by adding new `time_sink` types rather than explicit fields.

stevengj:
In a typical simulation, when you call `sim.run`, it first calls `sim.init_sim`, which initializes the structure (`_init_structure`) *before* creating the fields object (which starts the fields timer).  So I'm not sure why the structure-initialization time would be included in the fields timing for a typical run?

stevengj:
If you are doing your own timing, then just call `sim.init_sim()` explicitly before starting your timing.

stevengj:
Maybe in the future we should just make the timing stack global rather than putting it in the `fields` object.

PR #1007, opened 2019-09-03
FAQ on 2d structures with out-of-plane wavevector

oskooi:
- How do I model a 2d structure involving an out-of-plane wavevector?

PR #1009, opened 2019-09-04
test and documentation for special_kz

oskooi:
Adds a test to the Python `make check` suite based on the example described in #1007:[comment](https://github.com/NanoComp/meep/pull/1007#issuecomment-527747292). Documentation is provided  for both the Python and Scheme interface.

Also, `special_kz` can now be defined directly in the `Simulation` constructor (default: `False`). As @stevengj [proposed](https://github.com/NanoComp/meep/pull/1007#issuecomment-527858908), we may want to turn this feature on by default.

### DIFF
--- a/doc/docs/FAQ.md
+++ b/doc/docs/FAQ.md
@@ -481,9 +481,9 @@ Yes. You can use the [`set_boundary`](Python_User_Interface.md#field-computation
 
 ### How do I model a 2d structure involving an out-of-plane wavevector?
 
-To model e.g., fiber waveguides with 2d claddings, diffractive binary gratings with arbitrary incident planewaves in 3d, etc. in Cartesian coordinates, you would create a 2d cell in the $xy$ plane and specify a `k_point` with *non-zero* component in $z$. The resulting 3d simulation involves all electromagnetic fields (rather than a 2d simulation which involves a subset of the fields determined by the polarization of the current source).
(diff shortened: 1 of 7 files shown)


doc/docs/FAQ.md
```diff
@@ -481,9 +481,9 @@ Yes. You can use the [`set_boundary`](Python_User_Interface.md#field-computation
...
+To model e.g., fiber waveguides with 2d claddings, diffractive binary gratings with arbitrary incident planewaves in 3d, etc. in Cartesian coordinates, you would create a 2d cell in the $xy$ plane and specify a `k_point` with *non-zero* component in $z$. The resulting 3d simulation involves all electromagnetic fields (rather than a 2d simulation which involves a subset of the fields determined by the polarization of the current source). Performance can be improved by specifying `special_kz=True` in the `Simulation` constructor; this results in a 2d simulation with real rather than complex fields (as long as the $x$ and $y$ components of `k_point` are zero).
 
-Note: [mode decomposition](Python_Tutorials/Mode_Decomposition.md) is *not* currently supported for this use case (Issues [#291](https://github.com/NanoComp/meep/issues/291), [#604](https://github.com/NanoComp/meep/issues/604)).
+Note: [mode decomposition](Python_Tutorials/Mode_Decomposition.md) is *not* yet currently supported for this use case (Issues [#291](https://github.com/NanoComp/meep/issues/291), [#604](https://github.com/NanoComp/meep/issues/604)).
```
stevengj: Separate from handling the degenerate case directly (otherwise, kz works with mode decomposition), the eigenmode calculation also needs to handle special_kz.  In particular, for 2d simulations it should set kz to `fields.beta`

PR #1014, opened 2019-09-09
demonstrate breakdown of scalar theory in binary-grating tutorial

### LINKED ISSUE #1011: updated mode-decomposition tutorial
As discussed with @oskooi, it would be nice to do an updated version of https://meep.readthedocs.io/en/latest/Python_Tutorials/Mode_Decomposition/#transmittance-spectra-for-planewave-at-normal-incidence that shows the results as a function of period, in order to see the deviations from scalar-diffraction theory.

In particular, you could plot the transmitted power in the 0-th diffraction order as a function of period, showing:

1. For period >> λ, it matches scalar diffraction theory (is ≈ 0 for this structure).
2. For period << λ, it goes to the effective medium limit of transmission through a 3 layer structure (a layer of the mean ε (weighted by the duty cycle) sandwiched between the high and low ε).
3. In between, there is a (continuous) transition region where a full-wave calculation is required.

oskooi:
Closes #1011.


oskooi:
0th_order_vs_gp_3 5_res200_Ez_Hz

oskooi:
At an index contrast of `3.5`, the transmittance for E<sub>z</sub> and H<sub>z</sub> should be: `0.7256882155473792` and `0.8041965627867106`.

stevengj:
For n=3.5 and a λ=0.5µm, the wavelength is 0.14µm in the material.  So, you probably need to go to even smaller periods if you really want to see it asymptoting.

oskooi:
For the high-index case, going down to smaller periods (`0.005` μm), increasing the `resolution` (`1000` pixels/μm), and using more intermediate periodicity values (Δ=`0.005` μm) does indeed clearly show the asymptotic limit of the transmittance to the correct values set by effective-medium theory.

0th_order_vs_gp_3 5_res1000_Ez_Hz

Note that this is equivalent to just zooming into the left-most region of the figure above.

PR #1020, opened 2019-09-13
fix phase in get_field_point for special_kz calculations

stevengj:
Fixes #1018.

PR #1022, opened 2019-09-14
fix phase in get_field_point for special_kz calculations

### LINKED ISSUE #1018: get_field_point for 2d cell with non-zero kz and special_kz=True
(re: @stevengj's question in #1010:[comment](https://github.com/NanoComp/meep/pull/1010#issuecomment-528522428)).

When calling `get_field_point` in the case of a 2d cell with non-zero k<sub>z</sub> and `special_kz=True`, the ratio of the fields at any two points *outside* the cell along a line in the *z* direction should be defined by their relative phase difference exp(ik<sub>z</sub>*d*) where *d* is the separation distance.

However, as demonstrated in the following example, the value returned by `get_field_point` for two *different* points separated in *z* is the *same*. This is unexpected.

```py
import meep as mp
import math

resolution = 30 # pixels/um                                                                          

cell_size = mp.Vector3(14,14)

pml_layers = [mp.PML(thickness=2)]

                     size=mp.Vector3(mp.inf,1,mp.inf),

fsrc = 0.3  # frequency of eigenmode source                                                          
bnum = 1    # band number of eigenmode                                                               
kz = 0.2    # fixed out-of-plane wavevector component                                                

sources = [mp.EigenModeSource(src=mp.ContinuousSource(fsrc),
                              center=mp.Vector3(),
                              size=mp.Vector3(y=14),
                              eig_band=bnum,
                              eig_parity=mp.EVEN_Y,
                              eig_match_freq=True)]

                    resolution=resolution,
                    boundary_layers=pml_layers,
                    sources=sources,
                    geometry=geometry,
                    symmetries=[mp.Mirror(mp.Y)],
                    k_point=mp.Vector3(z=kz),
                    special_kz=True)

sim.run(until=50)

d = 3.5
ez1 = sim.get_field_point(mp.Ez, mp.Vector3(2.3,-5.7,4.8))
ez2 = sim.get_field_point(mp.Ez, mp.Vector3(2.3,-5.7,4.8+d))

print("field:, {}, {}".format(ez1.real,ez2.real))
print("phase:, {}, {}".format(ez2.real/ez1.real,math.cos(2*math.pi*kz*d)))
```

```
-----------
Initializing structure...
time for choose_chunkdivision = 0.000809389 s
Working in 2D dimensions.
Computational cell is 14 x 14 x 0 with resolution 30
     block, center = (0,0,0)
          size (1e+20,1,1e+20)
time for set_epsilon = 0.114979 s
-----------
MPB solved for omega_1(1.03923,0,0) = 0.320023 after 17 iters
MPB solved for omega_1(0.967463,0,0) = 0.30002 after 8 iters
MPB solved for omega_1(0.967391,0,0) = 0.3 after 4 iters
MPB solved for omega_1(0.967391,0,0) = 0.3 after 1 iters
run 0 finished at t = 50.0 (3000 timesteps)
field:, -0.00029456639310736254, -0.00029456639310736254
phase:, 1.0, -0.30901699437494756
```

In fact, `get_field_point` *always* returns `-0.00029456` regardless of its *z*-component. This would indicate that the *z*-component of `k_point` is just being ignored.

**Addendum**: results for `special_kz=False` (i.e., 3d cell) are exactly as expected.

stevengj:
Second try to fix #1018.

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -68,7 +68,9 @@ def py_v3_to_vec(dims, iterable, is_cylindrical=False):
         if is_cylindrical:
             return mp.veccyl(v3.x, v3.z)
         else:
-            return mp.vec(v3.x, v3.y)
+            v = mp.vec(v3.x, v3.y)
+            v.set_direction(mp.Z, v3.z) # for special_kz handling
+            return v
     elif dims == 3:
         return mp.vec(v3.x, v3.y, v3.z)
     else:
--- a/scheme/structure.cpp
+++ b/scheme/structure.cpp
@@ -138,7 +138,11 @@ vector3 vec_to_vector3(const meep::vec &pt) {
 meep::vec vector3_to_vec(const vector3 v3) {
   switch (dim) {
     case meep::D1: return meep::vec(v3.z);
-    case meep::D2: return meep::vec(v3.x, v3.y);
+    case meep::D2: {
+      meep::vec v(v3.x, v3.y);
+      v.set_direction(meep::Z, v3.z); // for special_kz handling
+      return v;
+    }
     case meep::D3: return meep::vec(v3.x, v3.y, v3.z);
     case meep::Dcyl: return meep::veccyl(v3.x, v3.z);
     default: meep::abort("unknown dimensionality in vector3_to_vec");
--- a/src/boundaries.cpp
+++ b/src/boundaries.cpp
@@ -154,7 +154,8 @@ inline bool fields::on_metal_boundary(const ivec &here) {
 
 bool fields::locate_point_in_user_volume(ivec *there, complex<double> *phase) const {
   // Check if a translational symmetry is needed to bring the point in...
-  if (!user_volume.owns(*there)) FOR_DIRECTIONS(d) {
+  if (!user_volume.owns(*there)) {
+    LOOP_OVER_DIRECTIONS(gv.dim, d) {
       if (boundaries[High][d] == Periodic &&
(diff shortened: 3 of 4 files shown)


oskooi:
Nope, still not working: results identical to #1020:[comment](https://github.com/NanoComp/meep/pull/1020#issuecomment-531425249).

stevengj:
Oh, you pass a `Vector3`, which gets converted to a 2d (`D2`) `meep::vec` with a nonzero z component, which gets converted to a set of `meep::ivec` vectors on the grid for interpolation — but the `ivec` conversion drops the `z` component.

So, I need to multiply by the exp(iβz) phase at a different point in the code.

oskooi:
Working now:

```
field:, 0.00031908659947579566, -9.860318191533325e-05
phase:, -0.30901699437494806, -0.30901699437494756
```

Note that the sign of the phase is correct.

oskooi:
The results above showed only the real part of the fields. The full complex fields are `0.00031908659947579566-1.5630735307953867e-19j` and `-9.860318191533325e-05-0.00030346938969391717j`. The ratio of the latter to the former is exactly exp(i2πk<sub>z</sub>d) where k<sub>z</sub> is 0.2 and d is 3.5.

stevengj:
I wonder where the tiny imaginary part at `z=0` is coming from?

oskooi:
E<sub>z</sub> is computed at `(2.3,-5.7,10)` and `(2.3,-5.7,10+d)`. Perhaps it's just a coincidence that  the imaginary part is 0 at `z=10`?

PR #1028, opened 2019-09-20
Python tutorial example for finite grating

### LINKED ISSUE #932: tutorial example for finite grating
As discussed with @oskooi, it might be nice to show an example of computing the diffracted components of a finite periodic surface grating.  Rough procedure:

1. Do two simulations: one with the grating and one with the plain surface.  Compute the Fourier-transformed fields at a given frequency ω (e.g. just Ez for a 2d simulation with the Ez polarization) and subtract them to get the scattered fields s(x) at ω, for a plane just above the surface.

2. Check: The scattered fields should be nonzero above the grating and decay to zero away from the grating.

3. Compute the Fourier transform: integrate a(k) = ∫ s(x) exp(ikx) dx.   |a(k)|² is then the amplituide of the corresponding Fourier component.  (You should see peaks corresponding to the diffracted orders of 2π/Λ of the grating, becoming sharper and sharper as the number of grating periods increases.)

In 3d, in principle it is the same procedure, except that (a) now you have a 2d Fourier transform and (b) you need to transform multiple transverse (Ex,Ey) field components separately, and then at each (kx,ky) decompose the fourier amplitudes a(kx,ky) into two polarizations (s and p).

oskooi:
Closes #932.  

Scheme version will be added as a separate PR.


leo-usa:
I run the finite grating example in the tutorial, but the result is different from the document. I just need someone to run it and confirm that the result in the document is correct. It's the code on this page:

https://meep.readthedocs.io/en/latest/Python_Tutorials/Near_to_Far_Field_Spectra/

My result is in the attachment, you can see it's different from the one shown on this page. I have run quite a few other MEEP examples including the binary grating and metasurface lens, the results are all consistent with ones shown in the document, so I suspect this finite grating result in the document is not correct.

Thanks!

Leo
Screen Shot 2021-01-26 at 8 01 55 PM

PR #1033, opened 2019-10-06
No more libctl dimensions global

ChristopherHogan:
As of NanoComp/libctl#43, it is no longer necessary for meepgeom to keep track of the libctl `dimensions` global variable. Fixes failing `refl_angular.py` test on master branch.
@stevengj @oskooi


stevengj:
Can we make it so that it still works with older libctl versions, or do we need to require a new release?

PR #1035, opened 2019-10-09
Add kz_2d option

### LINKED ISSUE #1026: account for implicit "i" factor in special_kz calculations
As documented here:

https://github.com/NanoComp/meep/blob/4e1802b653b1949a9cf33d82ce67583091e17e11/src/step_db.cpp#L121-L133

we implicitly store `i*(TM fields)` when doing special_kz calculations with real field variables.  This doesn't affect things like flux, but it should be accounted for in other calculations, e.g. when retrieving/outputting the fields or their Fourier transforms, handling source terms, etcetera.

If we want to make special_kz the default, we should probably force complex fields by default in that case, since otherwise it will be a bit confusing that the fields are real.

ChristopherHogan:
Fixes #1026.

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -649,6 +649,9 @@ class Simulation(object):
         else:
             raise ValueError("Invalid kz_2d option: {} not in [complex, real/imag, 3d]".format(kz_2d))
 
+        if self.special_kz and self.k_point is False:
+            self.k_point = Vector3()
+
     # To prevent the user from having to specify `dims` and `is_cylindrical`
     # to Volumes they create, the library will adjust them appropriately based
     # on the settings in the Simulation instance. This method must be called on
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -638,19 +638,18 @@ class Simulation(object):
         self.fragment_stats = None
         self._output_stats = os.environ.get('MEEP_STATS', None)
 
-        if kz_2d == "complex":
-            self.special_kz = True
-            self.force_complex_fields = True
-        elif kz_2d == "real/imag":
-            self.special_kz = True
-            self.force_complex_fields = False
-        elif kz_2d == "3d":
-            self.special_kz = False
-        else:
-            raise ValueError("Invalid kz_2d option: {} not in [complex, real/imag, 3d]".format(kz_2d))
-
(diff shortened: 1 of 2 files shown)


ChristopherHogan:
Using "complex" as the new default (as described in #1026) causes about half of the tests to fail (assuming it's correct to set the default `k_point` to `Vector3(0, 0, 0)` when using the "complex" option). Should the failing tests use the "3d" option?

python/simulation.py
```diff
@@ -639,6 +638,20 @@ def __init__(self,
...
+            self.special_kz = True
+            self.force_complex_fields = False
+        elif kz_2d == "3d":
+            self.special_kz = False
+        else:
+            raise ValueError("Invalid kz_2d option: {} not in [complex, real/imag, 3d]".format(kz_2d))
+
+        if self.special_kz and self.k_point is False:
+            self.k_point = Vector3()
+
```
oskooi: Lines 641-653 should be replaced with:

```
        self.special_kz = False
        if self.cell_size.z == 0 and self.k_point.z != 0:
            if kz_2d == "complex":
                self.special_kz = True
                self.force_complex_fields = True
            elif kz_2d == "real/imag":
                self.special_kz = True
                self.force_complex_fields = False
            elif kz_2d == "3d":
                self.special_kz = False
            else:
                raise ValueError("Invalid kz_2d option: {} not in [complex, real/imag, 3d]".format(kz_2d))
```

Setting `kpoint=0` when `special_kz=True` is incorrect since its non-zero z component is what is used in the calculations.

python/tests/special_kz.py
```diff
@@ -80,20 +80,28 @@ def test_special_kz(self):
...
+        t_complex = time() - start
 
         start = time()
-        Rmeep_kz = self.refl_planar(theta, True)
-        t_kz = time() - start
+        Rmeep_real_imag = self.refl_planar(theta, 'real/imag')
+        t_real_imag = time() - start
+
+        start = time()
+        Rmeep_3d = self.refl_planar(theta, '3d')
```
oskooi: The 3d simulation can probably be removed (in order to save time) since this particular test is only for the `special_kz=True` feature.

PR #1039, opened 2019-10-17
Add %newobject to create_structure_and_set_materials

### LINKED ISSUE #1038: Memory not freed up by Simulation
Hi, I'm running many Meep simulations in a binary, serially, and am seeing memory leakage. A small example:
```
import os
import psutil
import meep as mp

def evaluate():
  # Code from examples/straight-waveguide.py, with resolution = 30.
  cell = mp.Vector3(16,8,0)
  geometry = [mp.Block(mp.Vector3(mp.inf,1,mp.inf),
                       center=mp.Vector3(),
  sources = [mp.Source(mp.ContinuousSource(frequency=0.15),
                       component=mp.Ez,
                       center=mp.Vector3(-7,0))]
  pml_layers = [mp.PML(1.0)]
  resolution = 30
                      boundary_layers=pml_layers,
                      geometry=geometry,
                      sources=sources,
                      resolution=resolution)
  sim.run(until=1)

mp.quiet(True)
process = psutil.Process(os.getpid())
while True:
  evaluate()
  print(process.memory_info())
```

(Sample) Output:
```
pmem(rss=55906304, vms=2243817472, shared=17506304, text=47247360, lib=0, data=54673408, dirty=0)
pmem(rss=58003456, vms=2243817472, shared=17506304, text=47247360, lib=0, data=56770560, dirty=0)
pmem(rss=60100608, vms=2243817472, shared=17506304, text=47247360, lib=0, data=58867712, dirty=0)
pmem(rss=62197760, vms=2243817472, shared=17506304, text=47247360, lib=0, data=60964864, dirty=0)
pmem(rss=65789952, vms=2243817472, shared=18935808, text=47247360, lib=0, data=63062016, dirty=0)
pmem(rss=67887104, vms=2243817472, shared=18935808, text=47247360, lib=0, data=65159168, dirty=0)
pmem(rss=69984256, vms=2243817472, shared=18935808, text=47247360, lib=0, data=67256320, dirty=0)
...
```
I'm on Python 3.6, Meep 1.11.0, MPB 1.9.0, all compiled from source.

I ran the following C++ code, adapted from tests/bench.cpp:
```
#include <stdio.h>
#include <stdlib.h>
#include <meep.hpp>
using namespace meep;

double one(const vec &) { return 1.0; }
void evaluate(const double xmax, const double ymax, const double zmax, double eps(const vec &)) {
  const double a = 10.0;
  grid_volume gv = vol3d(xmax, ymax, zmax, a);
  structure s(gv, one);
  fields f(&s);
  f.add_point_source(Ez, 0.8, 0.6, 0.0, 4.0, vec(xmax * .5, ymax * .5, zmax * .5));
  f.step();
}

  initialize mpi(argc, argv);
  quiet = true;
  while(true) {
    evaluate(10.0, 10.0, 10.0, one);
    master_printf(".");
  }
  return 0;
}
```
Which does not seem to leak memory (as monitored by `top`).

I do not have any SWIG experience, and don't have an understanding of how the SWIG wrapped objects are destroyed once the reference counter goes to 0. Is there some cleanup method I could call to ensure that all fields etc are deallocated, for example?

Many thanks. Would be helpful even just to know if this is reproducible for y'all.

jeinarsson:
Without this decorator the memory allocated in meep::structure
is never recovered, because ~structure is never run.

Fixes #1038.

### DIFF
--- a/python/meep.i
+++ b/python/meep.i
@@ -1577,7 +1577,7 @@ PyObject *_get_array_slice_dimensions(meep::fields *f, const meep::volume &where
     import atexit
     atexit.register(report_elapsed_time)
 %}
-
+%newobject create_structure_and_set_materials;
 %inline %{
 
 size_t get_realnum_size() {


ChristopherHogan:
Thank you! This looks correct. The test failures are expected.

ChristopherHogan:
Replaced by #1042.

PR #1047, opened 2019-11-05
documentation for kz_2d for Python interface

### LINKED ISSUE #1025: document special_kz algorithm
Would be good to have a document that writes down exactly what equations we are evolving in this case.

oskooi:
Closes #1025.

Would be good to duplicate this feature in the Scheme interface as well.


stevengj:
This is probably not enough detail on how to use the `"real/imag"` option effectively; we should probably have a separate page on this (#1025) that we can refer to.

One option would be to not document the real/imag option for now, and just document the other two.

PR #1054, opened 2019-11-15
documentation page for 2d Cell with Out-of-Plane Wavevector

oskooi:
This was requested by @stevengj in #1047:[comment](https://github.com/NanoComp/meep/pull/1047#issuecomment-550121983).


stevengj:
Would be nice to add a simple example, e.g. out-of-plane incidence on the binary grating example, just computing transmission/reflection (skipping mode decomposition).

PR #1057, opened 2019-11-19
Python tutorial for differential/radar cross section

### LINKED ISSUE #1055: differential/radar cross section tutorial
Would be good to have a tutorial example showing the computation of the differential cross section (DCS), e.g. just updating the Mie tutorial.

To compute the DCS in a given direction, the simplest method is probably to use the near2far feature to get the fields, and hence the Poynting flux `F = r̂⋅Re[E*×H]/2` (= intensity in the outward direction), at a single point a large distance R >> (object diameter).     Then, if I understand the definition correctly, the DCS σ in that direction is `σ = R²F / (incident intensity)`.  (You would calculate the incident planewave intensity from the normalization run as in the current Mie example.)

As a check, you should also find that if you integrate the DCS over all spherical angles, the result should equal the scattering cross section:

image

(In a spherical Mie calculation, the DCS would depend on polar angle θ but not azimuthal angle φ, if we define the spherical coordinates relative to the incident planewave direction.)

To get the radar cross section (RCS) as mentioned in https://github.com/NanoComp/meep/pull/1045#issuecomment-547702985, you would simply compute the DCS in the "backwards" direction (i.e. backscattering) and multiply it by 4π.

(I got the 4π factor just by inspecting the definitions, but it's reassuring that other people have come to the same conclusion: https://physics.stackexchange.com/questions/148211/is-radar-cross-section-the-same-as-scattering-cross-section)

oskooi:
Closes #1055.

Scheme version and Jupyter notebook will be added in a separate PR.


oskooi:
As requested by @stevengj in #1054:[comment](https://github.com/NanoComp/meep/pull/1054#issuecomment-554513777), an example has been added to [2d Cell with Out-of-Plane Wavevector](https://meep.readthedocs.io/en/latest/2d_Cell_Special_kz/).

stevengj:
Sounds good.  We should eventually plot the DCS vs θ, though it would be good to have the analytical result from the pymie package first.

PR #1075, opened 2019-12-13
split by cost along longest direction

stevengj:
Always splitting along the longest direction is a more conservative choice since our split-by-cost model does not include a communication cost, so we want to avoid creating too many long/skinny chunks.

The plan is to merge this PR and then run some benchmark tests.  If it doesn't improve things we can always revert it.


stevengj:
Reverted since it didn't seem to improve things, and maybe made them worse.

PR #1078, opened 2019-12-17
Add progress bar for notebooks

ChristopherHogan:
Unfortunately the idea of using `_repr_markdown_` and `__repr__` to adapt the display to notebooks and the terminal won't work because IPython will always call [`__repr__` in addition to `_repr_markdown_`](https://github.com/ipython/ipython/issues/9771). This PR adds a progress bar for notebooks, but still prints the same text it was printing before. It looks like this:

Screenshot from 2019-12-16 18-36-29

Including the text kind of defeats the purpose of the progress bar, but I'm not sure we can do much better until there is a reliable way to detect if we're running in a notebook.
@stevengj @oskooi

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -2607,7 +2607,7 @@ def display_progress(t0, t, dt):
 
     def _disp(sim):
         t1 = mp.wall_time()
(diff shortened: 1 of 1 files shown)


stevengj:
Would be good to make this work in quiet mode, I think?

PR #1091, opened 2020-01-05
Get chunk communciation areas

### LINKED ISSUE #1069: function to output total chunk area
Should be easy to write.   Just add (1) a `size_t grid_volume::surfacearea()` function that returns the surface area in pixels, and (2) a method to `structure` that outputs the total surface area of all the chunks.

Could also be done in Python since the geom_box is part of the fragment stats.

ChristopherHogan:
Closes #1069. Adds `get_max_chunk_communication_area` and `get_avg_chunk_communication_area`. Unions all chunks on the same processor to a single volume. Counts all inner boundaries, but only counts outer boundaries that are periodic.

Usage:

```py
sim = mp.Simulation(...)
max_comm = sim.get_max_chunk_communication_area()
avg_comm = sim.get_avg_chunk_communication_area()
```
@stevengj @oskooi

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -1108,6 +1108,97 @@ class Simulation(object):
         if self.load_structure_file:
             self.load_structure(self.load_structure_file)
 
+    def _is_outer_boundary(self, vol, direction, side):
+
+        if direction == mp.X:
+            cell_size_in_dir = self.cell_size.x
+        elif direction == mp.Y:
+            cell_size_in_dir = self.cell_size.y
+        else:
+            cell_size_in_dir = self.cell_size.z
+
+        half_cell_size = cell_size_in_dir / 2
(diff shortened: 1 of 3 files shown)


python/simulation.py
```diff
@@ -1108,6 +1108,97 @@ def _init_structure(self, k=False):
...
+
+        # Union the chunk volumes that are on the same processor
+        idx = 0
+        result = []
+        for i in range(mp.count_processors()):
+            unioned_vol = vols[idx].surroundings()
+            idx += 1
+            while idx < len(owners) and owners[idx] == i:
+                unioned_vol = unioned_vol | vols[idx].surroundings()
+                idx += 1
```
stevengj: This code assumes that all of the chunks for a given process are stored consecutively… I guess this is true?

PR #1095, opened 2020-01-08
use_centered_grid

### LINKED ISSUE #1092: option to get DFT fields at Yee grid points
Might be useful (e.g. for adjoint calculations) to have an option `yee_grid` to `add_dft_fields` that allows you to pass `use_centered_grid=true` to the underlying `add_dft` function called in `fields::add_dft_fields`.

smartalecH:
Closes #1092 

Special thanks to @ChristopherHogan for help with the typemaps.

Added a test that just checks for proper execution -- it does not check if the underlying routine actually computes the correct DFT components at each yee grid point.


stevengj:
It might be clearer to call this flag `yee_grid=True` in the Python API?

smartalecH:
Addressed with latest commit. Also modified the docs.

PR #1100, opened 2020-01-10
Python tutorial and test for near2far for cylindrical coordinates

oskooi:
Should be merged after #1090. Also includes a test.

Scheme version and Jupyter notebook will be added in a separate PR.


stevengj:
As we discussed, you could compute the far fields along a line very far way, and then compute the magnitude of the Poynting vector at each point to get the power flowing in that direction.

Alternatively, you could do the Fourier transform of the near fields analogous to what you did for the 2d diffraction grating, following my instructions for how to do it in 3d.  That is, you are still doing the 2d Fourier transform over the near fields, exactly as if the structure was 3d.   Of course, when you perform the Fourier integrals, it will be more efficient to do the integrals in cylindrical coordinates.  The angular integrals can probably be done analytically since it looks like they are Bessel integrals (https://en.wikipedia.org/wiki/Bessel_function#Bessel's_integrals).   In principle, this is more direct, but it is easier to use the near2far code since that is already written.

PR #1110, opened 2020-01-28
improve mpi4py error message and update centos7 installer

d-a-v:
The change in `python/meep.i` provides a detailed error message that ***allows*** to find out causes of installation issues:
```
**
** failed to load python MPI module (mpi4py)
** /usr/local/lib64/python3.6/site-packages/mpi4py/MPI.cpython-36m-x86_64-linux-gnu.so: undefined symbol: ompi_mpi_logical8 
**
```
as compared to previously:
```
**
** failed to load python MPI module (mpi4py)
**
```

The updated installer builds a working version of meep inside a bare `centos:7` docker image.
The above error is solved by using `LD_PRELOAD` system environment variable.

### DIFF
--- a/contrib/build-meep.sh
+++ b/contrib/build-meep.sh
@@ -18,6 +18,7 @@ EOF
 
 [ -z "$1" ] && echo "(use -h for help)"
 
+test=false
 installdeps=true
 
 while [ ! -z "$1" ]; do
@@ -35,6 +36,13 @@ while [ ! -z "$1" ]; do
         -n)         # do not check for distribution dependencies
             installdeps=false
             ;;
+        -S)         # source directory (default: <installdir>/src)
+            SRCDIR="$2"
+            shift
+            ;;
+        -t)         # basic test
+            test=true
+            ;;
         *)
             echo "'$1' ?"
             help "$0"
@@ -45,16 +53,17 @@ done
 
 
 # detect wether DESTDIR is ending with src/
-[ -z ${DESTDIR} ] && DESTDIR=$(pwd)
-[ ${DESTDIR##*/} = src ] && DESTDIR=$(cd $(pwd)/..; pwd)
-SRCDIR=${DESTDIR}/src
+[ -z "${DESTDIR}" ] && DESTDIR=$(pwd)
+[ "${DESTDIR##*/}" = src ] && DESTDIR=$(cd $(pwd)/..; pwd)
+[ -z "$SRCDIR" ] && SRCDIR=${DESTDIR}/src
 
 cat << EOF
 
 This sript will download or update sources, compile and install MEEP.
-Please ensure the following final paths fit your needs:
+Please ensure the following final paths fit your needs
+Or use help '$0 -h' to change them:
     '${DESTDIR}/bin/meep'
-    '${DESTDIR}/lib/...'
+    '${DESTDIR}/lib(64)/...'
     '${DESTDIR}/share/...'
     '${DESTDIR}/...'
     '${SRCDIR}/<sources>'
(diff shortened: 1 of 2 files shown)


contrib/build-meep.sh
```diff
@@ -63,6 +72,28 @@ Press return to continue
...
+    export CXX=/usr/lib64/openmpi/bin/mpicxx
+    export PATH=${PATH}:/usr/lib64/openmpi/bin
+    RPATH_FLAGS="-Wl,-rpath,${DESTDIR}/lib64:/usr/lib64/openmpi/lib"
+    LDFLAGS="-L${DESTDIR}/lib64 -L/usr/lib64/openmpi/lib ${RPATH_FLAGS}"
+    CFLAGS="-I${DESTDIR}/include -I/usr/include/openmpi-x86_64/"
+    export PYTHONPATH=${DESTDIR}/lib/python3.6/site-packages
+    export PYTHONPATH=${PYTHONPATH}:/usr/local/lib64/python3.6/site-packages
+    export PYTHONPATH=${PYTHONPATH}:/usr/local/lib64/python3.6/site-packages/mpi4py
+    export LD_LIBRARY_PATH=${DESTDIR}/lib64:/usr/lib64/openmpi/lib:/usr/local/lib64/python3.6/site-packages/mpi4py/lib-pmpi:/usr/local/lib64/python3.6/site-packages/mpi4py
+    export LD_PRELOAD=/usr/lib64/openmpi/lib/libmpi.so
```
stevengj: Why is preloading needed?

contrib/build-meep.sh
```diff
@@ -63,6 +72,28 @@ Press return to continue
...
+    export CC=/usr/lib64/openmpi/bin/mpicc
+    export CXX=/usr/lib64/openmpi/bin/mpicxx
+    export PATH=${PATH}:/usr/lib64/openmpi/bin
+    RPATH_FLAGS="-Wl,-rpath,${DESTDIR}/lib64:/usr/lib64/openmpi/lib"
+    LDFLAGS="-L${DESTDIR}/lib64 -L/usr/lib64/openmpi/lib ${RPATH_FLAGS}"
+    CFLAGS="-I${DESTDIR}/include -I/usr/include/openmpi-x86_64/"
+    export PYTHONPATH=${DESTDIR}/lib/python3.6/site-packages
+    export PYTHONPATH=${PYTHONPATH}:/usr/local/lib64/python3.6/site-packages
+    export PYTHONPATH=${PYTHONPATH}:/usr/local/lib64/python3.6/site-packages/mpi4py
+    export LD_LIBRARY_PATH=${DESTDIR}/lib64:/usr/lib64/openmpi/lib:/usr/local/lib64/python3.6/site-packages/mpi4py/lib-pmpi:/usr/local/lib64/python3.6/site-packages/mpi4py
```
stevengj: Doesn't `mpicc` take care of the library path for us?

contrib/build-meep.sh
```diff
@@ -63,6 +72,28 @@ Press return to continue
...
+if $ubuntu; then
+    RPATH_FLAGS="-Wl,-rpath,${DESTDIR}/lib:/usr/lib/x86_64-linux-gnu/hdf5/openmpi"
+    LDFLAGS="-L${DESTDIR}/lib -L/usr/lib/x86_64-linux-gnu/hdf5/openmpi ${RPATH_FLAGS}"
+    CFLAGS="-I${DESTDIR}/include -I/usr/include/hdf5/openmpi"
+fi
+
+if $centos; then
+    export CC=/usr/lib64/openmpi/bin/mpicc
+    export CXX=/usr/lib64/openmpi/bin/mpicxx
+    export PATH=${PATH}:/usr/lib64/openmpi/bin
```
stevengj: why do we need to set PATH if we are giving the full path in CC and CXX?

python/meep.i
```diff
@@ -1563,8 +1563,8 @@ PyObject *_get_array_slice_dimensions(meep::fields *f, const meep::volume &where
     if with_mpi():
         try:
             from mpi4py import MPI
-        except ImportError:
-            print('\n**\n** failed to load python MPI module (mpi4py)\n**\n')
+        except ImportError as e:
+            print('\n**\n** failed to load python MPI module (mpi4py)\n**', e, '\n**\n')
```
stevengj: Would be good to split this change into a separate PR so that it can be merged independently.

d-a-v: done with #1140

contrib/build-meep.sh
d-a-v: This `PATH` will be removed in next commit

contrib/build-meep.sh
d-a-v: Yes, removed in next commit.

contrib/build-meep.sh
d-a-v: > Why is preloading needed?

I don't know the reason TBH but I can show the effect of not using it with centos 7.

If I'm missing something I don't know yet where. This is not needed for ubuntu.

```
[root@771c413e32e4 /]# python3 /tmp/test-meep.py 

**
** failed to load python MPI module (mpi4py)
** /usr/local/lib64/python3.6/site-packages/mpi4py/MPI.cpython-36m-x86_64-linux-gnu.so: undefined symbol: ompi_mpi_logical8 
**

Vector3<16.0, 8.0, 0.0>

Elapsed run time = 0.0002 s
[root@771c413e32e4 /]# export LD_PRELOAD=/usr/lib64/openmpi/lib/libmpi.so
[root@771c413e32e4 /]# python3 /tmp/test-meep.py 
Using MPI version 3.0, 1 processes
Vector3<16.0, 8.0, 0.0>

Elapsed run time = 0.0002 s
[root@771c413e32e4 /]# 
```
(It is a docker build <= me trying to explain 'root' :)

d-a-v:
My conclusion on this script is that it is easier to install everything in a docker environment (same OS and same destination directory as with native host), then copy the installed directories to the final host.
That was the only way I found really working without threat.
Otherwise libraries / binaries can be linked with several equivalent libraries at the same time. For example I saw openblas ***and*** atlas linked in `libmpb.so` even though `--with-blas=openblas` was passed to `configure`. It cannot happen when using the optional docker option for building.

I am opened to comments. This script was initially proposed to help meep users as it helped me to install meep for my users. Maintainers may have a different vision on how to build meep. I am OK if this appears too difficult to maintain in this repository, and removed, or if it should be improved in some way.

Note that I still have no answer about the bizarre need for LD_PRELOAD in the complex python interaction with all the dependencies on CentOS. I would be happy to get rid of it.

PR #1112, opened 2020-01-30
WIP: complex ε and μ output

### LINKED ISSUE #1108: complex ε and μ output
PR #919 by @smartalecH added support to output Re[ε(ω)] and Re[μ(ω)] at a given frequency ω.  It would be nice to support getting the full complex ε and μ as well.   The code [already computes this](https://github.com/NanoComp/meep/blob/db8d382e1e9078b3a52ee0c1420f00af399c5402/src/monitor.cpp#L257-L356) internally, but it currently takes the real part for output.

In particular, whenever a ω≠0 is specified and conductivity or susceptibilities are present, probably `output_epsilon`, `get_epsilon`, etcetera should probably output/return a complex array.

oskooi:
Fixes #1108 

There is just one thing that needs to be resolved: `output_epsilon` and `get_epsilon` output/return a complex array only when ω≠0 whereas conductivity or susceptibility needs to be present as well. I wasn't sure how to do this cleanly.

The following two lines need to be modified:

https://github.com/oskooi/meep/blob/a11f1982f8f487676b1f2be8c927a9b4bb53ce1d/python/simulation.py#L2013

https://github.com/oskooi/meep/blob/a11f1982f8f487676b1f2be8c927a9b4bb53ce1d/src/h5fields.cpp#L391


ChristopherHogan:
Hi Ardavan,

I've attached the signed document.

Chris

stevengj:
These parameters should be renamed from `omega` to `frequency`.

PR #1121, opened 2020-02-11
write output files from Python tests to temporary system directory rather than source subdirectory

oskooi:
Several of the Python tests create output files due to e.g. `output_epsilon`, `save_flux`, `Animate2D`, etc. which are placed in the `python/tests` subdirectory during `make check`. As a security measure, some operating systems prevent the test suite from writing any kind of file to disk in non-temporary system directories. This PR modifies the output directory of any test which creates a file to a temporary directory determined by the OS at runtime via [`tempfile.mkdtemp`](https://docs.python.org/3.6/library/tempfile.html#tempfile.mkdtemp). 

Additionally, all instances of `os.path.realpath` are replaced with `os.path.abspath` since on certain systems expanding `".."` using `realpath` is *not* allowed.

TODOs: 

1. ~~since [MPB](https://mpb.readthedocs.io/en/latest/Python_User_Interface/#the-modesolver-class) does not contain the equivalent of Meep's `use_output_directory` parameter, the only test that is left out is [python/tests/mpb.cpp](https://github.com/NanoComp/meep/blob/master/python/tests/mpb.py).~~

2. modify the C++ tests in `tests` which also generate output files to write them to a temporary directory.

~~**note**: the tests on Travis are failing due to the following error:~~

```
5280  File "../../../python/tests/holey_wvg_cavity.py", line 147, in <module>
5281    temp_dir = mp.comm.bcast(temp_dir, root=0)
5282AttributeError: module 'meep' has no attribute 'comm'
```

### DIFF
--- a/doc/docs/Build_From_Source.md
+++ b/doc/docs/Build_From_Source.md
@@ -191,7 +191,7 @@ Assuming you've set your `LDFLAGS` etcetera, the configure script should find al
 make check
 ```
 
-Note: several of the unit tests generate output files which are written to disk. The C++ test suite in `meep/tests` outputs its files in the same subdirectory. The Python test suite in `meep/python/tests` outputs its files to a temporary system directory.
+Note: several of the unit tests generate output files which are written to disk. The C++ test suite in `meep/tests` outputs its files in the same subdirectory. The Python test suite in `meep/python/tests` outputs its files to a temporary system directory (i.e., /tmp, etc.).
 
 The configure script accepts several flags to modify its behavior.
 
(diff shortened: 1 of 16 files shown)


stevengj:
Would be nice to make this easier to access from C++ and reduce amount of repeated code.

Maybe just update the `make_output_directory` function in C++ to add a no-argument version that creates a temporary directory?

stevengj:
Note that the analogous function from C is [`mkdtemp`](http://man7.org/linux/man-pages/man3/mkdtemp.3.html)

stevengj:
Does `mp.make_output_directory('foo', 'bar')` work now from Python?  Just wondering if the typemaps work here.

oskooi:
Following #1122, Python's `tempfile.mkdtemp` has been replaced with `meep.make_output_directory`. The C++ tests will be updated in a separate PR.

PR #1122, opened 2020-02-11
tempdir support

stevengj:
`mp.make_output_directory()` now makes a temporary directory with `mkdtemp`.

So, you can do `sim.use_output_directory(mp.make_output_directory())` to work in a temporary directory.

It would be nice to add a parameter to `use_output_directory` that automatically deletes the directory when the simulation destructor is called.

### DIFF
--- a/src/output_directory.cpp
+++ b/src/output_directory.cpp
@@ -150,7 +150,7 @@ got_tmpdir:
   strcat(strcpy(outdirname, tmpdir), meeptemplate);
 
   if (am_master() && !mkdtemp(outdirname)) {
-      abort("failed to create temporary output directory");
+      abort("failed to create temporary output directory \"%s\"", outdirname);
   }
   broadcast(0, outdirname, len);
   return outdirname;
--- a/src/output_directory.cpp
+++ b/src/output_directory.cpp
@@ -137,6 +137,17 @@ const char *make_output_directory(const char *exename, const char *jobname) {
   return outdirname;
 }
 
+/* similar to above, but creates a temporary directory */
+const char *make_output_directory() {
+  static char outdirname[] = "meepXXXXXX";
+  if (am_master()) {
+    if (!mkdtemp(outdirname))
+      abort("failed to create temporary output directory");
+  }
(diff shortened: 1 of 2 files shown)


oskooi:
This does not seem to be working as expected: `mp.make_output_directory()` creates a directory in the *current* directory rather than in one of the system directories (i.e. `/tmp`) in the way that Python's `tempfile.mkdtemp` does.

(As an aside, the two failing Travis tests `python/tests/bend_flux.py` and `python/tests/mode_coeffs.py` are both due to NanoComp/libctl#53 which was recently merged. The hard-coded values in the two tests likely need to be updated. cc @danielwboyce.)

src/output_directory.cpp
```diff
@@ -137,6 +128,17 @@ const char *make_output_directory(const char *exename, const char *jobname) {
   return outdirname;
 }
 
+/* similar to above, but creates a temporary directory */
+const char *make_output_directory() {
+  static char outdirname[] = "meepXXXXXX";
```
oskooi: The name of the output directory should probably include `/tmp` and therefore be changed to `"/tmp/meepXXXXXX"`. Otherwise, the directory is created in the current directory.

stevengj: Oh right.

Getting the tempdir properly is a little annoying.  Here is how [libuv does it](https://github.com/libuv/libuv/blob/7b28d36056d947e74e2c3347ee9bb04a1f5388c6/src/unix/core.c#L1069-L1117).  (libuv is MIT/expat licensed so we can grab its implementation.)

stevengj: Done.

stevengj:
It now puts it in the temporary directory (`/tmp` or similar).

stevengj:
I also noticed that, contrary to its name, `trash_output_directory` does not actually "trash" anything.  It just calls `mkdir` to create the directory, but this does not delete any existing directory.

src/output_directory.cpp
stevengj: Done.

PR #1126, opened 2020-02-15
write output from C tests to temporary directory

oskooi:
Following #1121 and #1122, the output from the C test suite is written to a temporary directory rather than a subdirectory. Of the 19 tests only one ([tests/h5test.cpp](https://github.com/NanoComp/meep/blob/master/tests/h5test.cpp)) actually produces HDF5 output files. For some reason, six of the other tests create an output directory even though no files were being written. This has been corrected.

Finally, following the discussion earlier today, [Tutorials/Eigenmode Source/Index-Guided Modes in a Ridge Waveguide](https://meep.readthedocs.io/en/latest/Python_Tutorials/Eigenmode_Source/#index-guided-modes-in-a-ridge-waveguide) has been updated to explicitly mention that the eigenmode source does not need to span the entire length of the cell.

(All tests pass when NanoComp/libctl#53 is excluded.)


tests/h5test.cpp
```diff
@@ -438,10 +445,13 @@ int main(int argc, char **argv) {
...
             master_printf("Checking %s...\n", name);
             if (!check_3d(funky_eps_3d, a, splitting, Sf3[iS], Ez, c3d[ic], gv_3d[igv], use_real,
-                          gv_3d_rank[igv], name))
+                          gv_3d_rank[igv], name, temp_dir))
               return 1;
           }
       }
 #endif /* HAVE_HDF5 */
+
+  if (am_master()) rmdir(temp_dir);
```
stevengj: This won't work.  `rmdir` only works for empty directories.

See also [this stackoverflow post](https://stackoverflow.com/questions/2256945/removing-a-non-empty-directory-programmatically-in-c-or-c).     It seems like we could implement the POSIX `ntfw` solution pretty easily to add a `meep::delete_directory` function.

stevengj: i.e. add
```cxx
#include <ftw.h>
static int rmpath(const char *path, const struct stat *s, int type, struct FTW *ftw) {
    (void) s; (void) ftw;
    return remove(path) < 0 ? -1 : 0;
}

// equivalent to rm -rf path
void delete_directory(const char *path) {
    nftw(argv[1], rmFiles,10, FTW_DEPTH|FTW_MOUNT|FTW_PHYS);
}
```

oskooi: Done.

src/output_directory.cpp
```diff
@@ -160,4 +161,14 @@ void trash_output_directory(const char *dirname) {
...
 }
 
+static int rmpath(const char *path, const struct stat *s, int t, struct FTW *ftw) {
+  (void) s; (void) t; (void) ftw; // unused
+  return remove(path);
+}
+
+// equivalent to rm -rf path
+void delete_directory(const char *path) {
+  nftw(path, rmpath, 10, FTW_DEPTH|FTW_MOUNT|FTW_PHYS);
```
stevengj: should be `if (am_master())`

PR #1127, opened 2020-02-16
update failing prism tests

oskooi:
This PR updates the two failing Python tests due to NanoComp/libctl#53 in order to restore Travis CI. One of the tests (`mode_coeffs.py`) requires just a minor update to three of its hard-coded values. ~~The other test (`bend_flux.py`) is failing due to something more significant that is broken in `libGDSII`; this is commented out until a fix can be found.~~

### DIFF
--- a/python/tests/mode_coeffs.py
+++ b/python/tests/mode_coeffs.py
@@ -81,7 +81,7 @@ class TestModeCoeffs(unittest.TestCase):
         self.assertTrue(res.kdom[0].close(mp.Vector3(0.604301, 0, 0)))
         self.assertTrue(res.kdom[1].close(mp.Vector3(0.494353, 0, 0), tol=1e-2))
         self.assertAlmostEqual(res.cscale[0],0.50000977,places=5)
-        self.assertAlmostEqual(res.cscale[1],0.50181621,places=5)
+        self.assertAlmostEqual(res.cscale[1],0.50096888,places=5)
         mode_power = mp.get_fluxes(mode_flux)[0]
 
         TestPassed = True
@@ -117,8 +117,8 @@ class TestModeCoeffs(unittest.TestCase):
         eval_point = mp.Vector3(0.7, -0.2, 0.3)
         ex_at_eval_point = emdata.amplitude(eval_point, mp.Ex)
         hz_at_eval_point = emdata.amplitude(eval_point, mp.Hz)
-        self.assertAlmostEqual(ex_at_eval_point, 0.45358518109307083+0.5335421986481814j)
-        self.assertAlmostEqual(hz_at_eval_point, 3.717865162096829-3.1592989829386298j)
+        self.assertAlmostEqual(ex_at_eval_point, 0.4887779638178009+0.48424014532428294j)
+        self.assertAlmostEqual(hz_at_eval_point, 3.4249236584603495-3.455974863884166j)
 
     def test_kpoint_func(self):
 
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -608,7 +608,7 @@ The vertices that make up the prism. They must lie in a plane that's perpendicul
 
 **`height` [`number`]**
 —
-The prism thickness, extruded in the direction of `axis`. `mp.inf` can be used for infinite height.
+The prism thickness, extruded in the direction of `axis`. `mp.inf` can be used for infinite height. No default value.
 
 **`axis` [`Vector3`]**
 —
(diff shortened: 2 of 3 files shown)


oskooi:
The GDSII prism test in `bend_flux.py` just required the height to be set to a non-zero value since it has a [default value of zero](https://github.com/NanoComp/meep/blob/master/src/meepgeom.hpp#L204-L205). The height parameter for `get_GDSII_prisms`, which was missing from the [documentation](https://meep.readthedocs.io/en/latest/Python_User_Interface/#gdsii-support), has also been added. Thanks to @danielwboyce for helping to resolve this issue.

python/tests/mode_coeffs.py
```diff
@@ -117,8 +117,8 @@ def test_modes(self):
         eval_point = mp.Vector3(0.7, -0.2, 0.3)
         ex_at_eval_point = emdata.amplitude(eval_point, mp.Ex)
         hz_at_eval_point = emdata.amplitude(eval_point, mp.Hz)
-        self.assertAlmostEqual(ex_at_eval_point, 0.45358518109307083+0.5335421986481814j)
-        self.assertAlmostEqual(hz_at_eval_point, 3.717865162096829-3.1592989829386298j)
+        self.assertAlmostEqual(ex_at_eval_point, 0.4887779638178009+0.48424014532428294j)
+        self.assertAlmostEqual(hz_at_eval_point, 3.4249236584603495-3.455974863884166j)
```
stevengj: Is this discretization error?

That is, do the old code and the new code converge to one another (at a reasonable rate) as you increase the resolution?

oskooi: Yes, this is indeed a discretization error: the values from the two versions of the code are converging to each other with increasing resolution.

stevengj: If you output epsilon in both cases, how do they differ?

stevengj: It would also be nice to plot the error vs resolution for both old and new codes (compared to a high-resolution run, I guess).

oskooi:
The following is a plot of the relative error in the flux of a 2d ridge waveguide (E<sub>z</sub> polarization; fundamental mode) created using a `Prism` object (identical to the structure in [python/tests/mode_coeffs.py](https://github.com/NanoComp/meep/blob/bc8c0cb2864f83bd11ba08cff90dc50cebb73445/python/tests/mode_coeffs.py#L10-L55)) vs. resolution for the "old" and "new" codes (i.e., before and after NanoComp/libctl#53). The results converge to the same value with increasing resolution. (Note: the "exact" value of the Poynting flux is computed at a resolution of 200 which is twice the maximum resolution of 100 in the plot.)

prism_power_rel_error

As additional verification, the epsilon profile is also identical for the two cases.

straight_waveguide_prism

The only significant change in `python/tests/mode_coeffs.py` is the hard-coded values of the [eigenmode calculation](https://github.com/NanoComp/meep/blob/bc8c0cb2864f83bd11ba08cff90dc50cebb73445/python/tests/mode_coeffs.py#L110-L121) (H<sub>z</sub> polarization). To investigate this further, a comparison of the real and imaginary parts as well as the magnitude of E<sub>x</sub> and H<sub>z</sub> over a 1d cross section of the waveguide (rather than just a single arbitrary point used in the test) for the two cases is shown below.

**pre libctl 53**

straight_waveguide_band_prelibctl53

**post libctl 53**

straight_waveguide_band_postlibctl53

The only difference is the phase in the real/imaginary parts of E<sub>x</sub>/H<sub>z</sub> as the magnitude is identical in both cases.

Thus, it seems that the changes introduced by NanoComp/libctl#53 are just minor discretization artifacts.

python/tests/mode_coeffs.py
oskooi: When these values are compared as the absolute-value squared, the differences are found to be minor.

**pre libctl 53**
abs(0.45358518109307083+0.5335421986481814j)**2 --> 0.4904067942455693
abs(3.717865162096829-3.1592989829386298j)**2 --> 23.803691427130342

**post libctl 53**
abs(0.4887779638178009+0.48424014532428294j)**2 --> 0.47339241625755824
abs(3.4249236584603495-3.455974863884166j)**2 --> 23.673864326080604

The relative differences in these two sets of values are 3.6% and 0.5%.

oskooi:
For additional reference, the following is a plot of error vs. resolution for the H<sub>z</sub> polarization. Note that the results for NanoComp/libctl#53 (in red) contain less oscillations than the old code. This is a noticeable improvement in the subpixel smoothing for Prism objects.

prism_power_rel_error_Hz

PR #1129, opened 2020-02-18
Python interface for slanted prisms

stevengj:
Adds `sidewall_angle` parameter to Python prism objects (defaults to zero), allowing them to interface the new slanted-prism feature in libctl 4.5 (NanoComp/libctl#53).

### DIFF
--- a/python/geom.py
+++ b/python/geom.py
@@ -493,7 +493,7 @@ class Ellipsoid(Block):
 
 class Prism(GeometricObject):
 
-    def __init__(self, vertices, height, axis=Vector3(z=1), center=None, **kwargs):
+    def __init__(self, vertices, height, axis=Vector3(z=1), center=None, sidewall_angle=0, **kwargs):
         centroid = sum(vertices, Vector3(0)) * (1.0 / len(vertices)) # centroid of floor polygon
         original_center = centroid + (0.5*height)*axis               # center as computed from vertices, height, axis
         if center is not None and len(vertices):
@@ -506,6 +506,7 @@ class Prism(GeometricObject):
         self.vertices = vertices
         self.height = height
         self.axis = axis
+        self.sidewall_angle = sidewall_angle
 
         super(Prism, self).__init__(center=center, **kwargs)
 
--- a/python/typemap_utils.cpp
+++ b/python/typemap_utils.cpp
@@ -764,10 +764,11 @@ static int pyellipsoid_to_ellipsoid(PyObject *py_ell, geometric_object *e) {
 
 static int pyprism_to_prism(PyObject *py_prism, geometric_object *p) {
   material_type material;
-  double height;
+  double height, sidewall_angle;
   vector3 axis, center;
 
   if (!get_attr_material(py_prism, &material) || !get_attr_dbl(py_prism, &height, "height") ||
+      !get_attr_dbl(py_prism, &sidewall_angle, "sidewall_angle") ||
       !get_attr_v3(py_prism, &center, "center") || !get_attr_v3(py_prism, &axis, "axis")) {
 
     return 0;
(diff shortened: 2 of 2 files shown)


stevengj:
Needs documentation as well.  Note that we decided to define the `center` (centroid) parameter with respect to a sidewall angle of zero, as otherwise changing the sidewall angle would change the vertical angle (because it shifts the centroid with respect to the base).

oskooi:
Some strange artifacts are found in the epsilon profile of a hexagonal prism in the *xy* plane of a 2d cell whenever the `sidewall_angle` is not zero as shown below.

```py
import meep as mp
import numpy as np
import math
import matplotlib
matplotlib.use('agg')

cell_size = mp.Vector3(2.5,2.5)

resolution = 50

vertices = [mp.Vector3(-1,0),
            mp.Vector3(-0.5,math.sqrt(3)/2),
            mp.Vector3(0.5,math.sqrt(3)/2),
            mp.Vector3(1,0),
            mp.Vector3(0.5,-math.sqrt(3)/2),
            mp.Vector3(-0.5,-math.sqrt(3)/2)]

geometry = [mp.Prism(vertices,                     
                     height=1.5,
                     center=mp.Vector3(),
                     sidewall_angle=np.radians(15),

                    cell_size=cell_size,
                    geometry=geometry)

sim.init_sim()
plt.figure()
sim.plot2D()
plt.savefig('prism_sidewall.png')
```

**sidewall_angle=0**

prism_sidewall_0

**sidewall_angle=np.radians(15)**

prism_sidewall_15

**sidewall_angle=np.radians(27)**

prism_sidewall_27

Note that the size of the cross section is shrinking as the `sidewall_angle` increases. This is correct given the intersection of the prism with the *z*=0 plane.

oskooi:
It seems there might be a small bug in the subpixel smoothing for non-zero sidewall angles. This is because when subpixel smoothing is disabled (`eps_averaging=False`), the epsilon profiles for the cases above show no artifacts:

**sidewall_angle=np.radians(15)**

prism_sidewall_15_noepsavg

**sidewall_angle=np.radians(27)**

prism_sidewall_27_noepsavg

stevengj:
The "fuzz" around the edges looks a little suspicious, but I'm fine with merging this as an undocumented "experimental" feature.

oskooi:
Artifacts along the prism boundaries also appear in 3d.

```py
import meep as mp
import numpy as np
import math
import matplotlib
matplotlib.use('agg')

cell_size = mp.Vector3(3,3,3)

resolution = 50

vertices = [mp.Vector3(-1,0),
            mp.Vector3(-0.5,math.sqrt(3)/2),
            mp.Vector3(0.5,math.sqrt(3)/2),
            mp.Vector3(1,0),
            mp.Vector3(0.5,-math.sqrt(3)/2),
            mp.Vector3(-0.5,-math.sqrt(3)/2)]

geometry = [mp.Prism(vertices,
                     axis=mp.Vector3(0,0,1),
                     height=1.5,
                     center=mp.Vector3(),
                     sidewall_angle=np.radians(10),

                    cell_size=cell_size,
                    geometry=geometry)

sim.run(mp.at_beginning(mp.output_epsilon),until=0)
```

```
$ h5topng -o eps-xy.png -z 75 prism_sidewall-eps-000000.00.h5
$ h5topng -o eps-xz.png -y 75 prism_sidewall-eps-000000.00.h5
$ h5topng -o eps-yz.png -x 75 prism_sidewall-eps-000000.00.h5
```

**eps-xy.png**

eps-xy

**eps-xz.png**

eps-xz

**eps-yz.png**

eps-yz

danielwboyce:
From a discussion with @stevengj, a good idea for debugging this may be to check `normal_to_prism` and `intersect_line_segment_with_prism` near one of the pixels that looks weird.

oskooi:
When the sidewall angle is 0, no artifacts appear along the boundaries.

**eps-xy.png**

eps-xy

**eps-xz.png**

eps-xz

**eps-yz.png**

eps-yz

PR #1132, opened 2020-02-20
create and delete temporary directory using unittest methods

oskooi:
#1121 involved creating and deleting a temporary directory directly within the `__main__` function. However, certain interpreters have script scope enforcement and do not allow the `temp_dir` directory name to be used outside of the `__main__` function leading to an error. Also, on certain systems, [`os.removedirs`](https://docs.python.org/3/library/os.html#os.removedirs) cannot delete non-empty directories and aborts with an error.

This PR moves the creation of the temporary directory to within the [`setUpClass`](https://docs.python.org/3.6/library/unittest.html#unittest.TestCase.setUpClass) method of the `unittest` mechanism and stores the directory name as a member of the `cls` object. The directory deletion is moved to the [`tearDownClass`](https://docs.python.org/3.6/library/unittest.html#unittest.TestCase.tearDownClass) method and `os.removedirs` is replaced with `meep.delete_output_directory` which can force the deletion of a non-empty directory and has a built-in `all_wait()` to prevent race conditions.

All the tests from #1121 have been updated with these changes.

### DIFF
--- a/python/tests/chunks.py
+++ b/python/tests/chunks.py
@@ -3,11 +3,13 @@ import meep as mp
 
 class TestChunks(unittest.TestCase):
 
-    def setUpClass(self):
-        self.temp_dir = mp.make_output_directory()
+    @classmethod
+    def setUpClass(cls):
+        cls.temp_dir = mp.make_output_directory()
 
-    def tearDownClass(self):
-        mp.delete_directory(self.temp_dir)
+    @classmethod
+    def tearDownClass(cls):
+        mp.delete_directory(cls.temp_dir)
 
     def test_chunks(self):
         sxy = 10
--- a/python/tests/chunks.py
+++ b/python/tests/chunks.py
@@ -1,8 +1,15 @@
 import unittest
 import meep as mp
-import os
+import shutil
 
 class TestChunks(unittest.TestCase):
+
+    def setUp(self):
+      self.temp_dir = mp.make_output_directory()
+
+    def tearDown(self):
+        if mp.am_master():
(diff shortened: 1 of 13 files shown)


python/tests/chunks.py
```diff
@@ -1,9 +1,17 @@
...
 
 class TestChunks(unittest.TestCase):
 
+    def setUp(self):
+      self.temp_dir = mp.make_output_directory()
+
+    def tearDown(self):
+        mp.all_wait()
+        if mp.am_master():
+            shutil.rmtree(self.temp_dir,ignore_errors=True)
```
stevengj: Can just do `mp.delete_directory(self.temp_dir)` now

stevengj: (need a rebase first)

python/tests/simulation.py
```diff
@@ -19,6 +18,10 @@ class TestSimulation(unittest.TestCase):
 
     def setUp(self):
         print("Running {}".format(self._testMethodName))
+        self.temp_dir = mp.make_output_directory()
+
+    def tearDown(self):
+        mp.delete_directory(self.temp_dir)
```
stevengj: This creates a temporary method for every test.   Better to do it in `setUpClass` and `tearDownClass`

oskooi: Done.

oskooi:
This is ready to be merged. Afterwards, it would be good to do a 1.13.1 tagged release.

stevengj:
Merged and tagged.

PR #1133, opened 2020-02-24
documentation for glancing-angle mode launcher

### LINKED ISSUE #1130: tutorial example on breakdown of glancing-angle mode launcher?
Although we support eigenmode sources (and mode coefficients) for oblique waveguides (see #698 and #940), of course problems will arise as the waveguide approaches glancing incidence to the source plane:

1. The intersection of the waveguide with the source plane becomes larger and larger, necessitating a bigger computational cell.

2. Modal dispersion becomes a problem: the sensitivity of the field pattern (in the source plane) to frequency becomes larger and larger, so an eigenmode source will become narrower bandwidth.  Probably discretization errors become worse as well.

It might be nice to have a tutorial example illustrating this, to give a rough rule of thumb of the maximum angle where this is practical.

(Of course, for angles greater than 45° you will always be better off rotating the source plane, but in some circumstances this might be inconvenient, e.g. if you are continuously optimizing the waveguide angle.)

oskooi:
Closes #1130.


doc/docs/Python_Tutorials/Eigenmode_Source.md
```diff
@@ -106,7 +108,13 @@ There are numerical dispersion artifacts due to the FDTD spatial and temporal di
...
+
+Finally, we demonstrate that as long as the line source intersects the waveguide *and* `eig_kpoint` is not nearly parallel to the direction of the line source, the mode can be properly launched. As shown in the field profiles below for the single-mode waveguide, there does not seem to be any noticeable distortion in the launched mode as the waveguide approaches glancing incidence to the source plane up to 80°, where the total power in the forward-propagating mode is 97%. Note that the line source spans the entire length of the cell extending into the PML region (not shown). In this example, the maximum rotation angle is ~84°, where the total power drops to 59% and the backward-propagating fields are clearly visible. Due to discretization errors, increasing the size of the cell will *not* improve results beyond 84°, unless you also increase resolution.  As the waveguide mode approaches glancing incidence, sensitivity to discretization errors increases (because the mode varies rapidly with frequency on a glancing-angle cross-section).  For waveguide angles much beyond 45° you probably want to simply change the orientation of the line source by 90°.
+
```
stevengj: Would be good to actually try increasing the cell size for the 83° and 84° cases, to see if the resolution is limiting.

stevengj: Since you tried it and increasing the cell size helps, it would be good to show the larger cell size (for which there are still resolution effects)

oskooi: Done.

Issue #1060, opened 2019-11-22
Bad ε-averaging from Prism


jeinarsson:
I'm getting strange ε output from a Prism, where the vertices are generated by vectorizing a bitmap by marching squares (see [skimage.measure.find_contours](https://scikit-image.org/docs/0.16.x/api/skimage.measure.html#find-contours)). Here's a small example:

```python
import meep as mp
import numpy as np
import matplotlib.pyplot as plt
from skimage import measure

cell = mp.Vector3(3,3,0)

# The size of the array representing the level set.
# Larger L give a higher fidelity circle with more and smaller edges.
L = 75 

xs = np.linspace(-1,1,L)
ys = np.linspace(-1,1,L)
X, Y = np.meshgrid(xs,ys)
design = X**2 + Y**2 < 1.0**2

# Extract contour and transform to design space.
contours = measure.find_contours(design, level=0.5)
contour = contours[0]  # we know there's only one contour.
contour = contour[:-1,:]  # exclude last point == first point.
contour /= L
contour -= .5
contour *= 2*R
vertices = [mp.Vector3(
    x=p[0], y=p[1], z=0.0
) for p in contour]

geometry = [mp.Prism(vertices, height=1, material=mp.Medium(epsilon=10))]
resolution = 25
sim = mp.Simulation(cell_size=cell,
                    geometry=geometry,
                    resolution=resolution)

fig, ax = plt.subplots(figsize=(10,10))
mp.quiet(True) # otherwise it outputs all prism vertices.
sim.plot2D(ax=ax, eps_parameters={'interpolation': 'none', 'cmap': 'Greys'})
mp.quiet(False)

for go in sim.geometry:
  if isinstance(go, mp.Prism):
    xs = [v.x+0.5/resolution for v in go.vertices] # the translation corrects for the fact that imshow plots pixel values half pixel
    ys = [v.y+0.5/resolution for v in go.vertices] # offset compared to skimage.measure.find_contour.
    ax.plot(xs, ys,'.-')
```
Which outputs the following figure (blue curve is vertices, raster is epsilon data):
![Output 1](https://user-images.githubusercontent.com/610827/69467070-55af3a00-0d3b-11ea-97f3-fb576f14d437.png)

Without knowing more, I suspect there is some edge case (no pun intended) issue here, because all edges that come out of marching cubes are horizontal, vertical or exactly 45 degrees.

Indeed, I can do the same experiment where I generate the vertices on a "perfect circle":
```python
[...]

PR #1135, opened 2020-02-26
Python test for subpixel smoothing of prisms

oskooi:
Based on the example in #1060 where a circular shape formed using a `Prism` with a large number of vertices was demonstrated not to be working properly (which has been fixed by NanoComp/libctl#46 and NanoComp/libctl#49), a test is added to the Python suite which verifies that the integral of ε over the entire cell for a `Prism` (ε=12 in air) where the number of vertices defined on the contour of a circle is repeatedly doubled converges to the same integral for a `Cylinder`. For additional verification, the test is performed twice with circles of different radii.

This test is mainly designed to verify that subpixel smoothing is working properly.

ToDo: once the sidewall angle feature is working, would be good to add a similar test in 3d.


doc/docs/Installation.md
```diff
@@ -123,6 +123,10 @@ print(mp.__version__)
 
 This will show something like `1.11.0-1-g415bc8eb` where the first three digits (`1.11.0`) refer to a stable tarball release, the following digit is the number of commits after this stable release, and the eight characters following the `g` in the final string refer to the commit hash.
 
+### Non-Networked Systems
+
+To install the PyMeep Conda package on a [non-networked system](https://docs.anaconda.com/anaconda/user-guide/tasks/install-packages/#installing-packages-on-a-non-networked-air-gapped-computer), you will need to obtain the bz2 tarball of the [official release](https://anaconda.org/conda-forge/pymeep/files) or [nightly build](https://anaconda.org/simpetus/pymeep/files).
+
```
woodenToaster: This won't work without the dependencies. I thought Erik solved this problem over a year ago. It looks like there's a package, [conda pack](https://github.com/conda/conda-pack), that allows one to relocate an environment to another machine. I've never tried it though.

oskooi: The docs has been updated to mention conda pack.

oskooi:
The second test from #1060 has been added involving a prism object with vertices pre-computed using [`skimage.measure.find_contours`](https://scikit-image.org/docs/0.8.0/api/skimage.measure.find_contours.html). This test was failing prior to NanoComp/libctl#46 and NanoComp/libctl#49.

PR #1139, opened 2020-02-29
WIP: x parity constraints for eigenmodes

### LINKED ISSUE #1109: x parity constraints for eigenmodes
This would be nice, as discussed in NanoComp/mpb#72.   However, implementing it in MPB is rather subtle as discussed in NanoComp/mpb#74  because of the interaction with MPB's transverse basis.

Seems like it might be more straightforward to just do a cyclic rotation of the coordinates in Meep's `src/mpb.cpp` so that the x direction is always the propagation direction.

danielwboyce:
Closes #1109. 

In process of making changes to cyclically rotate Meep's coordinate systems when passed into MPB so that wave propagation is always in the x direction. So of the math involved and its implementation (notation) is a bit opaque, so at the moment the largest part of the problem is understanding what exactly is happening in the `src/mpb.cpp` so I don't miss any necessary transformations, either from Meep to MPB or from MPB to Meep.

### DIFF
--- a/src/mpb.cpp
+++ b/src/mpb.cpp
@@ -75,11 +75,10 @@ static void meep_mpb_eps_coordcycle(symmetric_matrix *eps, symmetric_matrix *eps
       o[1] = o_copy[2];
       o[2] = o_copy[0];
       p = vec(p.y(), p.z(), p.x());
-      //operation on f ?
+      //TODO: operation on f ?
     }
   }
 
-  // need to find out what eps_inv is doing exactly, but I believe all the necessary changes are made above this
   eps_inv->m00 = real(f->get_chi1inv(Ex, X, p, omega));
   eps_inv->m11 = real(f->get_chi1inv(Ey, Y, p, omega));
   eps_inv->m22 = real(f->get_chi1inv(Ez, Z, p, omega));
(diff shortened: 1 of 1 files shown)


danielwboyce:
The `direction` object `d` hasn't been rotated correctly; I need to update it with the correct way for `direction` objects.

danielwboyce:
At the moment when I run `python/examples/mode-decomposition.py` with the compiled test version I'm getting an error that seems related to SWIG:

```
  File "/home/tests/mode-decomposition.no_rotation.testing_version.py", line 61, in <module>
    sim.run(until_after_sources=mp.stop_when_fields_decayed(50,mp.Ez,mon_pt,1e-9))
  File "/usr/local/lib/python3.6/site-packages/meep/simulation.py", line 2222, in run
    self.init_sim()
  File "/usr/local/lib/python3.6/site-packages/meep/simulation.py", line 1321, in init_sim
    self.add_source(s)
  File "/usr/local/lib/python3.6/site-packages/meep/simulation.py", line 1603, in add_source
    add_eig_src()
  File "/usr/local/lib/python3.6/site-packages/meep/__init__.py", line 3947, in add_eigenmode_source
    return _meep.fields_add_eigenmode_source(self, *args)
RuntimeError: meep: invalid dielectric function for MPB
```

danielwboyce:
Actually judging the error to be caused by SWIG seems a bit hasty; I'm just not quite sure what's caused it yet.

PR #1140, opened 2020-03-02
Add python details on Import error

d-a-v:
This is related with [#1110](https://github.com/NanoComp/meep/pull/1110#pullrequestreview-359140321) @stevengj


stevengj:
Thanks!

PR #1141, opened 2020-03-02
WIP: Allow unequal DFT frequencies

### LINKED ISSUE #1070: allow unequal DFT frequencies
Right now all of the DFT objects use equally spaced frequency points.   It would be nice to be able to pass in an arbitrary array of frequencies instead.

For example, one could use Chebyshev nodes for the frequencies and then do high-order Chebyshev interpolation using the [barycentric formula](https://www.chebfun.org/docs/guide/guide04.html).   (Indeed, this is potentially so useful that we might want to build Chebyshev interpolation into Meep — it fits well with the "illusion of continuity" philosophy.  We could give the illusion of computing the spectrum continuously in a given range to a given order of interpolation accuracy.)

As another example, I recently worked out ["optimal" solar-weighted Gaussian quadrature rules](https://nbviewer.jupyter.org/urls/math.mit.edu/~stevenj/Solar-Quadrature.ipynb?flush_cache=true) for integrating against the solar irradiance spectrum.    This is useful for solar-cell simulations, and the optimal frequency evaluation points are unequally spaced.

danielwboyce:
~~Fixes #1070.~~ In progress of changing the relevant DFT classes' properties so that they have an array of frequencies, not `freq_min` or `freq_max`. The existing constructors will be updated so that DFT objects can be defined with `freq_min`, `freq_max`, and `Nfreq`, but new constructors will also be added that use only `freqs` and `Nfreq`.


smartalecH:
Some swig advice:

When wrapping overloaded C constructors, Swig needs all of the parameters to be independently type-checked ([see docs](http://www.swig.org/Doc3.0/SWIGPlus.html#SWIGPlus_overloaded_methods)). This is good practice anyway... If you want to check a numpy array called `freqs`, you can do something like this:

```c
%typecheck(SWIG_TYPECHECK_POINTER, fragment="NumPy_Fragments") double* freqs {
    $1 = is_array($input);
}
```

(all within `meep.i` of course)

You'll then have to define a custom typemap. You can do something like this:

```
%typemap(in, fragment="NumPy_Macros") double* freqs {
    $1 = (double *)array_data($input);
}
```

Make sure you do this for all the `fields."flux"` methods (i.e. `add_dft_flux`, etc.)

You'll have to go through `simulation.py` and update all of the flux, force, far_field, eigenmode, etc methods.

danielwboyce:
As of now, the source code seems to be properly compiling (none of the constructors seem to have been broken). Now I'll need to update `python/meep.i` and `scheme/meep.i` to have support in the scripting languages, which should also make testing easier.

danielwboyce:
As of now the package compiles all the way through, but I'm sure I'm getting SWIG wrong. The version of `python/meep.i` that's in place has the minimal changes in place required to build the package with the changes to the C code and `simulation.py`. I notice that for some reason the overloaded constructors don't need me to define new typemaps or typechecks to compile; is it likely that the automatically-generated wrappers are correct?

While attempting to run the `python/examples/mie_scattering.py` tutorial example I'm getting the following error

```
  File "/home/tests/mie_scattering.testing_v.pass_frq_min_frq_max.py", line 51, in <module>
    sim.run(until_after_sources=10)
  File "/usr/local/lib/python3.6/site-packages/meep/simulation.py", line 2259, in run
    self._evaluate_dft_objects()
  File "/usr/local/lib/python3.6/site-packages/meep/simulation.py", line 1627, in _evaluate_dft_objects
    dft.swigobj = dft.func(*dft.args)
  File "/usr/local/lib/python3.6/site-packages/meep/simulation.py", line 1851, in _add_flux
    return self._add_fluxish_stuff(self.fields.add_dft_flux, freqs, nfreq, fluxes)
  File "/usr/local/lib/python3.6/site-packages/meep/simulation.py", line 1987, in _add_fluxish_stuff
    stuff = add_dft_stuff(vol_list, freqs, nfreq, *args)
  File "/usr/local/lib/python3.6/site-packages/meep/__init__.py", line 3959, in add_dft_flux
    return _meep.fields_add_dft_flux(self, *args)
NotImplementedError: Wrong number or type of arguments for overloaded function 'fields_add_dft_flux'.
  Possible C/C++ prototypes are:
    meep::fields::add_dft_flux(meep::volume_list const *,double *,int,bool)
    meep::fields::add_dft_flux(meep::volume_list const *,double *,int)
    meep::fields::add_dft_flux(meep::volume_list const *,double,double,int,bool)
    meep::fields::add_dft_flux(meep::volume_list const *,double,double,int)
    meep::fields::add_dft_flux(meep::direction,meep::volume const &,double *,int,bool)
    meep::fields::add_dft_flux(meep::direction,meep::volume const &,double *,int)
    meep::fields::add_dft_flux(meep::direction,meep::volume const &,double,double,int,bool)
    meep::fields::add_dft_flux(meep::direction,meep::volume const &,double,double,int)
```

Which seems to tell me that the `add_dft_flux` function is not playing nice with SWIG. Thing is there's no `add_dft_flux` function mentioned in `python/meep.i`! Was the wrapper getting generated automatically before? Will I need to make a typemap suite for it now? Which parameters will I need to check? _All_ of them, or just the ones that differ between the overloaded versions of the function (`double freq_min`, `double freq_max`, `double *freqs`)?

smartalecH:
The python/swig UI regarding DFT stuff is a little convoluted...largely because of the load balancing requirements. But luckily I think you only need some minor changes to get your code working.

First, let's remember that all the DFT functions using fluxes will call `add_dft_flux` through `_add_fluxish_stuff`:

https://github.com/NanoComp/meep/blob/6c10e745ab2e2e26627f82695c289fe5d79b6845/python/simulation.py#L1820-L1823

You can see that `_add_fluxish_stuff` assumes one particular constructor form for the corresponding `add_dft_stuff` function (`add_dft_flux` in this case):

https://github.com/NanoComp/meep/blob/6c10e745ab2e2e26627f82695c289fe5d79b6845/python/simulation.py#L1939-L1955

This is problematic for two reasons:

1. First, you need to generalize the python code so we can _either_ use the original constructor (i.e. specify center, width, and the number of freqs) or pass in an array (like you've already implemented on your branch).
2. You need to explicitly typecheck and typemap every argument of all the possible constructors. Even if it compiles, that doesn't mean it works! The typechecking mechanism is how swig determines which constructor to use. It assigns a weight to each paramter based on it's type and maps the cumulative weight to a table. You can see more about that in the docs I posted above.

Feel free to check out PR #1095 for an example. I "overloaded" `add_dft_fields` (in reality I just added a default argument, but SWIG treats this as an overloaded constructor). I had to typecheck the remaining parameters. I was getting the same error as you until I realized this small nuance.

Looking forward to this feature!

ChristopherHogan:
I think what you want is to apply the predefined numpy typemaps. If you have `double *freqs, int Nfreqs` pairs of arguments for the new functions and they're all named consistently, all you need in `python/meep.i`

```
%apply (double *IN_ARRAY1, int DIM1) {(double *freqs, int Nfreqs)};
```

ChristopherHogan:
https://docs.scipy.org/doc/numpy-1.13.0/reference/swig.interface-file.html

python/simulation.py
```diff
@@ -1812,22 +1834,34 @@ def load_minus_force_data(self, force, fdata):
...
+    # TODO: make add_flux_uneven, wrap add_flux around it, change parameters of _add_flux
+    def add_flux_uneven(self, freqs, nfreq, *fluxes):
+        freqs.sort()
+        flux = DftFlux(self._add_flux, [freqs, nfreq, fluxes])
         self.dft_objects.append(flux)
         return flux
 
-    def _add_flux(self, fcen, df, nfreq, fluxes):
+    def add_flux(self, fcen, df, nfreq, *fluxes):
+        freqs = np.linspace(fcen - df / 2, fcen + df / 2, nfreq).tolist()
```
ChristopherHogan: You can remove the `tolist()` call since the typemaps are expecting a numpy array.

python/meep.i
```diff
@@ -966,6 +966,25 @@ meep::volume_list *make_volume_list(const meep::volume &v, int c,
 // typemaps needed for add_dft_fields
 //--------------------------------------------------
 
+%typecheck(SWIG_TYPECHECK_POINTER, fragment="NumPy_Fragments") double* freqs {
+    $1 = is_array($input);
+}
+
+%typemap(in, fragment="NumPy_Macros") double* freqs {
```
ChristopherHogan: This typemap isn't needed. It's covered by the `%apply` below.

python/meep.i
```diff
@@ -966,6 +966,25 @@ meep::volume_list *make_volume_list(const meep::volume &v, int c,
...
+    $1 = (double *)array_data($input);
+}
+
+%typecheck(SWIG_TYPECHECK_INTEGER) int Nfreqs {
+    $1 = PyInteger_Check($input);
+}
+
+%typemap(in) int Nfreqs {
```
ChristopherHogan: This typemap isn't needed. It's covered by the `%apply` below.

danielwboyce:
Removing `tolist` where extraneous and the extra typechecks doesn't seem to have fixed the problem.

src/dft_ldos.cpp
```diff
@@ -22,16 +22,31 @@ using namespace std;
...
-    domega = 0;
-    Nomega = 1;
+// TODO: overload dft_ldos constructor
+dft_ldos::dft_ldos(double *freqs, int Nfreq) {
+  if (Nfreq < 1) { abort("Nfreq must be at least 1"); }
+  else {
+    Nomega = Nfreq;
+    double om[Nomega];
+    for (int i = 0; i < Nfreq; i++) { om[i] = freqs[i] * 2 * pi; }
+    omegas = om;
```
stevengj: This won't work.  `om` will no longer exist once this scope exits.  You need to allocate on the heap with e.g. `new`

src/dft_ldos.cpp
```diff
@@ -22,16 +22,31 @@ using namespace std;
 
 namespace meep {
 
-dft_ldos::dft_ldos(double freq_min, double freq_max, int Nfreq) {
-  if (Nfreq <= 1) {
-    omega_min = (freq_min + freq_max) * pi;
-    domega = 0;
-    Nomega = 1;
+// TODO: overload dft_ldos constructor
+dft_ldos::dft_ldos(double *freqs, int Nfreq) {
```
stevengj: If the object makes a copy of `freqs`, then this should be declared as `const double *freqs`.

src/dft.cpp
```diff
@@ -182,6 +189,36 @@ dft_chunk *fields::add_dft(component c, const volume &where, double freq_min, do
...
 }
 
+dft_chunk *fields::add_dft(component c, const volume &where, double freq_min, double freq_max,
+                           int Nfreq, bool include_dV_and_interp_weights,
+                           complex<double> stored_weight, dft_chunk *chunk_next,
+                           bool sqrt_dV_and_interp_weights, complex<double> extra_weight,
+                           bool use_centered_grid, int vc) {
+  if (Nfreq < 1) abort("Nfreq must be at least 1");
+  double dfreq = (freq_max - freq_min) / (Nfreq - 1);
+  double freqs[Nfreq];
```
stevengj: Same problem: this does not persist.

src/meep.hpp
```diff
@@ -1009,8 +1009,9 @@ class dft_chunk {
   void operator-=(const dft_chunk &chunk);
 
   // the frequencies to loop_in_chunks
-  double omega_min, domega;
+  // TODO: change dft_chunk properties
   int Nomega;
+  double *omegas;
```
stevengj: might want to use `std::vector<double>` here.

stevengj: also below

stevengj:
Before you worry about getting the Python interface to work, I would to get the C++ tests to work, e.g.  `tests/bend-flux-ll.cpp` and `tests/dft-fields.cpp`.

stevengj:
In fact, let's separate the Python stuff into a separate PR.  Just try to get the C++ plumbing in place in this PR.

danielwboyce:
Let's also add a C++ test for the new plumbing in this PR.

PR #1142, opened 2020-03-03
Python test for nonconvex prisms

oskooi:
#1135 added a test for a circular-shaped `Prism`. This PR extends that test to include three *non-convex* shapes where the curvature changes sign (while avoiding any sharp "kinks" or "corners" which can degrade the accuracy of subpixel smoothing). These shapes (or "blobs") are randomly generated using the parameterized function a<sub>1</sub>sin²(a<sub>2</sub>θ)/r<sup>a<sub>3</sub></sup>+b<sub>1</sub>cos²(b<sub>2</sub>θ)/r<sup>b<sub>3</sub></sup>. The vertices of the prism for each blob is pre-computed using [`skimage.measure.find_contours`](https://scikit-image.org/docs/0.8.0/api/skimage.measure.find_contours.html). The test involves verifying that the the integral of ε over the entire cell converges to the "exact" result with increasing number of vertices. The "exact" result is simply a prism with at least twice the number of vertices of the two test cases. This is shown in the figures below: (a) and (b) insets are the test cases and (c) is the "exact" shape.

**generating vertices of a non-convex blob using the marching squares method**
```py
import numpy as np
import matplotlib
matplotlib.use('agg')
from skimage import measure
import meep as mp

def blob(x,y):
    th = np.arctan(y/x)
    r = np.sqrt(x**2+y**2)
    if 0.5*np.sin(3.5*th)**2/r**1.19850984 + 0.6*np.cos(2.5*th)**2/r**2.098614 < 0.10:
        return 1.0
    else:
        return 2.0

nx = 320
xs = np.linspace(-5,5,nx)
XS, YS = np.meshgrid(xs,xs)

eps = np.zeros((nx,nx))
for i in range(nx):
    for j in range(nx):
        eps[i,j] = blob(XS[i,j],YS[i,j])
        
contours = measure.find_contours(eps, level=1.5)
contour = contours[0]
contour = contour[:-1,:]
contour /= nx
contour -= 0.5
contour *= 10

vertices = [mp.Vector3(x=p[0], y=p[1], z=0.0) for p in contour]
geometry = [mp.Prism(vertices, height=1.0, material=mp.Medium(epsilon=12))]

for v in vertices:
    print("prism:, {}, {}".format(v.x,v.y))

cell_size = mp.Vector3(10,10)
resolution = 25
                    geometry=geometry,
                    resolution=resolution)


fig, ax = plt.subplots(figsize=(10,10))
sim.plot2D(eps_parameters={'interpolation': 'none', 'cmap': 'Greys'})

for go in sim.geometry:
    if isinstance(go, mp.Prism):
        xs = [v.x+0.5/resolution for v in go.vertices] # the translation corrects for the fact that imshow plots pixel values half pixel
        ys = [v.y+0.5/resolution for v in go.vertices] # offset compared to skimage.measure.find_contour.
        ax.plot(xs, ys,'.-')
        
fig.savefig("blob_n{}".format(nx))
    
## blob1: 0.5*np.sin(3.5*th)**2/r**1.19850984 + 0.6*np.cos(2.5*th)**2/r**2.098614 < 0.10: 

## blob2: 0.8*np.sin(1.4*th)**2/r**1.19850984 + 1.2*np.cos(1.7*th)**2/r**1.398614 < 0.20: 

## blob3: 1.15*np.sin(1.8*th)**2/r**1.268509 + 0.94*np.cos(1.4*th)**2/r**1.498614 < 0.16: 
```

**blob 1**
blob1_prism

**blob 2**
blob2_prism

**blob 3**
blob3_prism


stevengj:
It might be nice to hard-code the "exact" result computed at a high resolution.

Otherwise you are testing that it converges to *something*, but if it converges to the wrong result the test will no longer pass.

stevengj:
Would be good to test that you get the same result with symmetry.

oskooi:
The computed values of the ε integral for the "exact" blob have been replaced with hard-coded values which speeds up the calculation significantly.

It turns out that the vertices computed by `skimage.measure.find_contours` are slightly offset which breaks the symmetry of the underlying blob (this can be seen in the figure above in the blue contour lines around each blob which is displaced slightly from the origin and also verified by large changes in the ε integral itself). The test involving the circle where the vertices are computed analytically (originally added in #1135) does preserve symmetry and thus has been expanded to verify that its result is the same with and without symmetry.

PR #1148, opened 2020-03-07
Fix operator delete[] in typemap for dft_force

jeinarsson:
The memory in `result` is allocated with `new[]` in `dft_force::force`,
and must therefore be deallocated with `delete[]`.

See eg
https://en.cppreference.com/w/cpp/memory/new/operator_delete

Some compilers/runtimes are more relaxed, but some throw runtime errors on detecting this mismatch.


stevengj:
LGTM, thank!

Issue #1051, opened 2019-11-13
Simulation result depends on # of processes, using geometry_center


fferen:
In Parallel Meep, the simulation result depends on the number of processes in some cases when using the `geometry_center` argument in `Simulation`. Minimal example here: https://pastebin.com/YShcMhHy

When running with `mpirun -np 28 python testbug.py`, I get ...`Energy: 0.9868491846495558`... for the total field energy after the first iteration, while with `mpirun -np 1 python testbug.py` I get `Energy: 1.0441432121065009`. The error in the energy is sometimes bigger or smaller but seems to get worse over time.

ChristopherHogan:
FYI, I'm seeing the same behavior in the scheme interface.

```scheme
(define-param res 15)
(define-param subH 0.8)
(define-param airH 1)
(define-param devW 4)
(define-param pmlW 1)

(define-param totW (+ devW (* pmlW 2)))
(define-param totH (+ airH subH (* pmlW 2)))

(set! geometry (list (make block
                       (size totW totW (+ subH pmlW))
                       (center 0 0 (* -1 (/ (+ subH pmlW) 2)))
                       (material (make dielectric (epsilon 2.4))))))

(set! pml-layers (list (make pml (thickness pmlW))))

(set! sources (list (make source
                      (src (make continuous-src
                             (frequency 1)
                             (width 0)))
                      (component Ez)
                      (center 0 0 0)
                      (size devW devW 0)
                      (amplitude 1))))

(set-param! resolution res)
(set! geometry-lattice (make lattice (size totW totW totH)))
(set! force-complex-fields? true)
(set! geometry-center (vector3 0 0 (/ (- airH subH) 2)))

(run-until 5)

(define-param cellVol (volume (center 0 0 (/ (- airH subH) 2)) (size totW totW totH)))
(print (field-energy-in-box cellVol))

```

stevengj:
It would be helpful to visualize the resulting ε functions and the Ez field at the end of the simulation, to see whether the materials or the source location have shifted depending on the number of CPUs.

danielwboyce:
There is a major error in how the ε function changes with resolution when `geometry_center` is defined by the user; on this geometry the ε function goes to zero everywhere between resolutions 13 and 17, no matter if Meep was compiled with MPI or not, and doesn't seem to be changed by how many cores it's run with when compiled with MPI.

[...]

PR #1149, opened 2020-03-10
Python test for GDSII prisms

oskooi:
Adds another test to [`python/tests/prism.py`](https://github.com/NanoComp/meep/blob/master/python/tests/prism.py) involving a non-convex "spiral" loaded from a GDSII file (which was generated using [gdspy](https://gdspy.readthedocs.io/en/stable/gettingstarted.html#polygonal-only-paths)) and verifies that the integral of ε over the cell matches a hard-coded reference value.

spiral_gds_res25


oskooi:
The parallel tests are failing because the result depends on the number of processes. This is due to the use of `geometry_center` and the bug reported in #1051.

**serial: 1 process**
```
epsilon-sum:, 423.2359686656507 (prism-gds)
```

**parallel: 2 processes**
```
epsilon-sum:, 423.8031458380916 (prism-gds)
```

**parallel: 3 processes**
```
epsilon-sum:, 424.2720743018822 (prism-gds)
```

Until #1051 is resolved, the only workaround is to remove `geometry_center` which involves modifying the GDS file.

oskooi:
For reference, here is the script that was used to generate the spiral pattern for the GDSII file [which is copied from the [gdspy manual](https://gdspy.readthedocs.io/en/stable/gettingstarted.html#polygonal-only-paths) and adjusted slightly to center the shape near (0,0)].

```py
import numpy as np
import gdspy

path2 = gdspy.Path(0.5, (0, 0))

# Start the path with a smooth Bezier S-curve                                                                                                                               
path2.bezier([(0, 5), (5, 5), (5, 10)])

def spiral(u):
    r = 4 - 3 * u
    theta = 5 * u * np.pi
    x = r * np.cos(theta) - 4
    y = r * np.sin(theta)
    return (x, y)

def dspiral_dt(u):
    theta = 5 * u * np.pi
    dx_dt = -np.sin(theta)
    dy_dt = np.cos(theta)
    return (dx_dt, dy_dt)

path2.parametric(spiral, dspiral_dt)

## center shape near (0,0)
path2.translate(-1.5,-7.3)

lib = gdspy.GdsLibrary()
cell = lib.new_cell('spiral')
cell.add(path2)
lib.write_gds('spiral.gds')
```

Issue #965, opened 2019-07-16
__swig_getmethods__ no longer exists in SWIG 4.0.0?


stevengj:
I am getting
```
>>> import meep
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/stevenj/Local/lib/python2.7/site-packages/meep/__init__.py", line 2631, in <module>
    class dft_ldos(object):
  File "/Users/stevenj/Local/lib/python2.7/site-packages/meep/__init__.py", line 2670, in dft_ldos
    __swig_getmethods__["freq_min"] = get_omega_min
NameError: name '__swig_getmethods__' is not defined
```
This is due to a call to the internal function `__swig_getmethods__` in `python/meep.i` introduced by #581.   It looks like this no longer exists?

stevengj:
If I comment out the `__swig_getmethods__` calls, I also get an error on `_newclass`:
```
>>> import meep
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/stevenj/Local/lib/python2.7/site-packages/meep/__init__.py", line 2631, in <module>
    class dft_ldos(object):
  File "/Users/stevenj/Local/lib/python2.7/site-packages/meep/__init__.py", line 2673, in dft_ldos
    if _newclass: freq_min = property(get_omega_min)
NameError: name '_newclass' is not defined
```

sh-zam:
Yes, apparently. It also broke lldb scripting.

stevengj:
Right now as a stopgap I'm just manually commenting out the `__swig_getmethods__` and `_newclass` calls, which only affect the `dft_ldos` feature.

oskooi:
While #1154 removes the `__swig_getmethods__` calls introduced in #581, building the master branch with SWIG 4.0.1 (by disabling [these lines](https://github.com/NanoComp/meep/blob/master/configure.ac#L523-L527) in `configure.ac`) on Ubuntu produces the following error in `python/tests/simulation.py`:

```
======================================================================
ERROR: test_vec_constructor (__main__.TestSimulation)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ubuntu/install/meep/python/meep/simulation.py", line 57, in vec
    return mp._vec(*args)
  File "/home/ubuntu/install/meep/python/meep/__init__.py", line 1050, in _vec
    val = _meep.new__vec(*args)
TypeError: Wrong number or type of arguments for overloaded function 'new__vec'.
[...]

PR #1154, opened 2020-03-19
arbitrary frequency grid for DFT functions in C interface

oskooi:
Progress towards #1070 (still needs Python interface). Closes #1141.

Adds support for arbitrary frequency grids for DFT functions. There are now two interfaces to all DFT functions: (1) the original equally-spaced grid specified by (`freq_min`, `freq_max`,`Nfreq`) and (2) the new unequally-spaced grid specified by `std::vector<double>`.

ToDo: Python interface and tutorial example to be added in a separate PR.


oskooi:
SWIG typemaps for the `freq` array are added. However, 2 of the 40 tests in the Python suite are failing (`dft_energy.py` and `geom.py`). Not sure why calling `sim.add_energy` in `dft_energy.py` is causing a segmentation fault.

stevengj:
As I commented when we discussed #1141, we should probably be using `std::vector<double> freqs` for every argument where you are passing `double *freqs, size_t Nfreqs`.

Then if you do `flux_object.freqs` in Python it will return an array of frequencies.

python/simulation.py
```diff
@@ -2967,7 +2975,7 @@ def scale_flux_fields(s, flux):
 
 
 def get_flux_freqs(f):
-    return np.linspace(f.freq_min, f.freq_min + f.dfreq * f.Nfreq, num=f.Nfreq, endpoint=False).tolist()
+    return np.linspace(f.freq[0], f.freq[-1], num=f.Nfreq, endpoint=True).tolist()
```
stevengj: This will just become `return f.freqs`

python/meep.i
```diff
@@ -1346,29 +1366,19 @@ void _get_eigenmode(meep::fields *f, double omega_src, meep::direction d, const
...
     %pythoncode %{
         def freqs(self):
             import math
             import numpy as np
-            start = self.omega_min / (2 * math.pi)
-            stop = start + (self.domega / (2 * math.pi)) * self.Nomega
-            return np.linspace(start, stop, num=self.Nomega, endpoint=False).tolist()
+            start = self.omega[0] / (2 * math.pi)
+            stop = self.omega[-1] / (2 * math.pi)
+            return np.linspace(start, stop, num=self.Nomega, endpoint=True).tolist()
```
stevengj: After the change to `std::vector`, just `return (self.omega / (2*math.pi)).tolist()`

python/meep.i
```diff
@@ -1346,29 +1366,19 @@ void _get_eigenmode(meep::fields *f, double omega_src, meep::direction d, const
...
 %extend meep::dft_ldos {
 
-    double get_omega_min() {
-        return $self->omega_min;
-    }
-    double get_domega() {
-        return $self->domega;
-    }
     int get_Nomega() {
         return $self->Nomega;
```
stevengj: This will be `return $self->freqs.size()`

stevengj: We might not need `get_Nomega` anymore, since this was only used internally

stevengj:
The `dft_ldos` accessor functions were added by @ChristopherHogan in #581.  It's a bit weird that we have these for `dft_ldos` but not for flux objects etcetera?   Presumably this code will simplify a *lot* (and maybe eliminate the SWIG 4 incompatibility #965) if we just have a `std::vector` object for the frequency array.

Along the way, we probably want to change the `dft_ldos` object to use `freqs` rather than `omega` to be consistent with the rest of the code.

stevengj:
Note that for each function there should be two interfaces: the old interface that takes `(freq_min, freq_max, Nfreqs)` and the new interface that takes `freqs`.   However, the implementation of the old interface should just be a one-liner that calls the new interface by constructing an array from `(freq_min, freq_max, Nfreqs)` .   You'll probably want to add a helper function `meep::linspace(freq_min, freq_max, Nfreqs)` that creates the `freqs`.

oskooi:
`double *freq` is changed to `std::vector<double> freq` and all C tests are passing. However, it seems that SWIG is *not* automatically generating typemaps for `std::vector<double> freq` in Python since `flux.freq.tolist()` is not being recognized as indicated by many of the Python tests which use `get_flux_freqs` are failing with:

```
AttributeError: 'DoubleVector' object has no attribute 'tolist'
```

oskooi:
After removing the `.tolist` suffix to `freq` as well as some other minor fixes, *all* tests are now passing. I also fixed a warning bug in `python/tests/geom.py` caused by [`python/simulation.py:725`](https://github.com/NanoComp/meep/blob/master/python/simulation.py#L725). If everything looks good, the final thing to add is a C test.

python/simulation.py
```diff
@@ -2104,9 +2107,9 @@ def get_eigenmode_coefficients(self, flux, bands, eig_parity=mp.NO_PARITY, eig_v
             direction = flux.normal_direction
 
         num_bands = len(bands)
-        coeffs = np.zeros(2 * num_bands * flux.Nfreq, dtype=np.complex128)
-        vgrp = np.zeros(num_bands * flux.Nfreq)
-        cscale = np.zeros(num_bands * flux.Nfreq)
+        coeffs = np.zeros(2 * num_bands * flux.freq.size(), dtype=np.complex128)
+        vgrp = np.zeros(num_bands * flux.freq.size())
+        cscale = np.zeros(num_bands * flux.freq.size())
```
stevengj: Does every call to `flux.freq.size()` first create a numpy array out of `flux.freq`?

stevengj: I guess not, because `.size()` is not a NumPy array method.

python/simulation.py
```diff
@@ -2967,7 +2970,7 @@ def scale_flux_fields(s, flux):
 
 
 def get_flux_freqs(f):
-    return np.linspace(f.freq_min, f.freq_min + f.dfreq * f.Nfreq, num=f.Nfreq, endpoint=False).tolist()
+    return f.freq
```
stevengj: What type does this return?

stevengj: (i.e. what is `type(f.freq)`?)

oskooi:
```
>>> type(freqs)
<class 'meep.DoubleVector'>
```

stevengj: Probably want to convert it to a Python list if it is a Meep array.

src/dft.cpp
```diff
@@ -338,17 +367,23 @@ void load_dft_hdf5(dft_chunk *dft_chunks, component c, h5file *file, const char
 dft_flux::dft_flux(const component cE_, const component cH_, dft_chunk *E_, dft_chunk *H_,
                    double fmin, double fmax, int Nf, const volume &where_,
                    direction normal_direction_, bool use_symmetry_)
-    : Nfreq(Nf), E(E_), H(H_), cE(cE_), cH(cH_), where(where_), normal_direction(normal_direction_),
+    : E(E_), H(H_), cE(cE_), cH(cH_), where(where_), normal_direction(normal_direction_),
+      use_symmetry(use_symmetry_) {
```
stevengj: Can't you use the initializer for this?   i.e. have `freq(linspace(fmin, fmax, Nf)) { }`

python/simulation.py
stevengj: Try
```py
[freqs[i] for i in range(freqs.size())]
```
to make a Python list.

src/near2far.cpp
```diff
@@ -46,8 +59,9 @@ dft_near2far::dft_near2far(dft_chunk *F_, double fmin, double fmax, int Nf, doub
 }
 
 dft_near2far::dft_near2far(const dft_near2far &f)
-    : freq_min(f.freq_min), dfreq(f.dfreq), Nfreq(f.Nfreq), F(f.F), eps(f.eps), mu(f.mu),
-      where(f.where) {
+    : F(f.F), eps(f.eps), mu(f.mu), where(f.where) {
```
stevengj: You can just do `freq(f.freq)` as part of the initializers and it should make a copy — no need to copy the elements one by 1.

src/stress.cpp
```diff
@@ -27,20 +27,27 @@ namespace meep {
...
+  freq = meep::linspace(fmin, fmax, Nf);
+  offdiag1 = offdiag1_;
+  offdiag2 = offdiag2_;
+  diag = diag_;
+  // where = new volume(where_.get_min_corner(), where_.get_max_corner());
+}
+
+dft_force::dft_force(dft_chunk *offdiag1_, dft_chunk *offdiag2_, dft_chunk *diag_,
+                     const std::vector<double> freq_, const volume &where_)
+    : where(where_) {
```
stevengj: As above: just add `freq(freq_)` to the initializers list

src/stress.cpp
```diff
@@ -27,20 +27,27 @@ namespace meep {
...
   diag = diag_;
   // where = new volume(where_.get_min_corner(), where_.get_max_corner());
 }
 
 dft_force::dft_force(const dft_force &f) : where(f.where) {
-  freq_min = f.freq_min;
-  Nfreq = f.Nfreq;
-  dfreq = f.dfreq;
+  for (int i = 0; i < f.freq.size(); ++i)
+    freq.push_back(f.freq[i]);
```
stevengj: `freq(f.freq)` in the initializers list

stevengj: Or just `freq = f.freq`

Issue #548, opened 2018-10-08
todo: improve CW solver convergence properties using preconditioning


oskooi:
Currently, the [frequency-domain solver](https://meep.readthedocs.io/en/latest/Python_User_Interface/#frequency-domain-solver) does not use any [preconditioning](https://en.wikipedia.org/wiki/Preconditioner) to improve the convergence properties of its [biconjugate gradient](https://en.wikipedia.org/wiki/Biconjugate_gradient_method) iterative method. Thus, there is potential for performance gains which could enable this feature to become more widely used; particularly for cases involving e.g., oblique planewaves which require narrow-band sources (as described in #471).

stevengj:
@wsshin has done a lot of work on this sort of thing.  Preconditioning wave equations (Helmholtz equations) is notoriously hard, and there aren't that many good techniques.  However, Wonseok has identified two things that work pretty well: changing the PML (or doing an equivalent field preconditioner) and adding an operator to shift quasistatic eigenvalues.  Diagonal (Jacobi) preconditioners, on the other hand, don't work that well.  See:

* http://www.mit.edu/~wsshin/pdf/shin2013oe.pdf
* http://www.mit.edu/~wsshin/pdf/shin2012jcp.pdf

wsshin:
On top of the two techniques @stevengj  mentioned, you can get an additional 2X improvement by making the matrix symmetric in BiCG.  Then you can modify the original BiCG algorithm to something that does not require application of Aᵀ (because Aᵀ = A).

Making the operator symmetric is actually not trivial.  As far as I remember MEEP supports only the square grid, but the stretched-coordinate PML stretches the edges of the squares along the Cartesian directions inside PML and makes the squares effectively rectangles.  On such a rectangular grid, the finite-difference double-curl operator ∇×∇× is no longer symmetric.  To make it symmetric, you need to apply special symmetrizing preconditioners.

There are infinitely many symmetrizing preconditioners, but not all of them are good because most of them ill-condition the matrix though symmetrizing it.  One way to construct symmetrizing preconditioners that preserve the conditioning is described in https://patents.google.com/patent/US20140207426A1/en.  

If the boundary condition is Bloch with nonzero phase differences (to implement oblique incidence), you cannot make the operator symmetric even with this technique.

[...]

PR #1158, opened 2020-03-25
shift-and-invert eigensolver based on solve_cw

### LINKED ISSUE #1153: shift-and-invert eigensolver
We have a frequency-domain solver that computes the fields from a given current source at a given frequency.  It would be nice to be able to iterate this — using the resulting fields as the source in a subsequent frequency-domain solve — in order to implement the "shift-and-invert" power iteration to compute resonant modes.

(In the longer run, we could also do Arnoldi, Rayleigh-quotient iteration, etcetera…  If we go that route we should probably just call ARPACK or similar.)

stevengj:
Closes #1153.  Adds a `sim.solve_eigfreq` method similar to `solve_cw` except that it returns an estimated eigenfrequency (and sets the fields to the corresponding eigenmode) based on shift-and-invert iteration around the source frequency.

~~Currently untested, just an initial stab at this.~~  Needs:
- [x] Tests
- [x] Documentation

### DIFF
--- a/doc/docs/Eigensolver_Math.md
+++ b/doc/docs/Eigensolver_Math.md
@@ -1,0 +1,67 @@
+---
+# Frequency-domain eigensolvers for Meep
+---
+
+In these notes, we describe the mathematical background for the frequency-domain eigensolver algorithms in Meep
+(used by the `solve_eigfreq` function), which build on the [frequency-domain solver](Python_Tutorials/Frequency_Domain_Solver.md) algorithm described in the
+[Meep paper](http://doi.org/10.1016/j.cpc.2009.11.008).   See also [these notes](https://github.com/mitmath/18369/blob/9d61d1731af4ad32ea924e1af57b89e7e6a6c488/notes/time-evolution.pdf) for more background on the notation for Maxwell's equations used below.
+
+Maxwell frequency-domain equations
+----------------------------------
+
+The Meep frequency-domain solver (at a frequency $\omega$) is essentially solving the linear system of equations $\hat{M}\psi=\xi$, where:
+
+$$\hat{M}(\omega)\psi=\overbrace{\left[\underbrace{\left(\begin{array}{cc}
+ & \nabla\times\\
+-\nabla\times
+\end{array}\right)}_{\hat{C}}+i\omega(1+\hat{\chi})\right]}^{\hat{M}(\omega)}\underbrace{\left(\begin{array}{c}
+\mathbf{E}\\
+\mathbf{H}
+\end{array}\right)}_{\psi}=\underbrace{\left(\begin{array}{c}
+\mathbf{J}\\
+\mathbf{K}
+\end{array}\right)}_{\xi}.$$ That is, $\psi$ is the six-component field state, $\hat{\chi}(\omega)$ is the $6\times6$ susceptibility at $\omega$, $\xi$ is the six-component current, and $\hat{C}$ is the $6\times6$ "curl" operator. (For these notes I am using "natural" units in which $\varepsilon_{0}=\mu_{0}=1$.) Furthermore, in most physical circumstances the matrix $\hat{\chi}$ block-diagonalizes, and $1+\hat{\chi}$ is related to the electric permittivity $\varepsilon$ and the magnetic permeability $\mu$: $$1+\hat{\chi}(\omega)=\left(\begin{array}{cc}
+\varepsilon(\omega)\\
+ & \mu(\omega)
+\end{array}\right).$$
+
+Equivalently, we can write things in terms of the $\mathbf{D}$ and $\mathbf{B}$ fields $$\Psi=\left(\begin{array}{c}
+\mathbf{D}\\
+\mathbf{B}
+\end{array}\right)=(1+\hat{\chi})\psi,$$ yielding $$\overbrace{\left[\hat{C}(1+\hat{\chi})^{-1}+i\omega\right]}^{\hat{A}(\omega)}\Psi=\underbrace{(1+\hat{\chi})^{-1}\left(\begin{array}{c}
+\mathbf{J}\\
+\mathbf{K}
+\end{array}\right)}_{\Xi}.$$ In fact, this equation $\hat{A}\Psi=\Xi$ is precisely the linear system of equations that Meep is actually solving internally.
+
+Maxwell eigenproblems
+---------------------
+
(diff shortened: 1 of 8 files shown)


stevengj:
Update: *it seems to work*!

I tried it on the "holey waveguide" tutorial" to solve for the resonance mode, and I got `eigfreq = 0.23454890203771478+0.0005581153769037464j` (Q ≈ 210) to about 6–7 decimal places after 7 iterations of shift-and-invert.

(Caveat: I'm not sure why it's giving the eigenfrequency as having a positive imaginary part, when it should be negative, so I'll need to check this.  But at least it is converging, and plotting the field pattern confirms that it seems to be finding the resonant mode.)

```py
import meep as mp

resolution = 20   # pixels/um

eps = 13          # dielectric constant of waveguide
w = 1.2           # width of waveguide
r = 0.36          # radius of holes
d = 1.4           # defect spacing (ordinary spacing = 1)
N = 5        # number of holes on either side of defect

sy = 6            # size of cell in y direction (perpendicular to wvg.)
pad = 2           # padding between last hole and PML edge
dpml = 1          # PML thickness

sx = 2*(pad+dpml+N)+d-1  # size of cell in x direction

cell = mp.Vector3(sx,sy,0)

blk = mp.Block(size=mp.Vector3(mp.inf,w,mp.inf), material=mp.Medium(epsilon=eps))
geometry = [blk]

for i in range(N):
        geometry.append(mp.Cylinder(r, center=mp.Vector3(d/2+i)))
        geometry.append(mp.Cylinder(r, center=mp.Vector3(-(d/2+i))))
        
pml_layers = [mp.PML(1.0)]

fcen = 0.25
df = 0.2

src = [mp.Source(mp.GaussianSource(fcen, fwidth=df),
                 component=mp.Hz,
                 center=mp.Vector3(0),
                 size=mp.Vector3(0,0))]

sym = [mp.Mirror(mp.X, phase=-1), mp.Mirror(mp.Y, phase=-1)]

sim = mp.Simulation(cell_size=cell, force_complex_fields=True,
                    geometry=geometry,
                    boundary_layers=pml_layers,
                    sources=src,
                    symmetries=sym,
                    resolution=resolution)

eigfreq = sim.solve_eigfreq(L=20, tol=1e-11, eigtol=1e-6)
```

stevengj:
Hmm, now that I think about it, I'm not estimating the eigenvalue correctly, because the operator here is the discrete time-difference operator and not the ∂/∂t operator.

stevengj:
Okay, I've pushed a patch that corrects the eigenvalue estimate for the time discretization.  It still converges in 7 shift-and-invert iterations, but now the estimated frequency is `0.23453881435117674-3.0147532052022876e-05j` (Q ≈ 3889.85).

If I give it the initial guess frequency computed by Harminv in the tutorial, then it converges to nearly the same eigenvalue estimate `0.23453876201337692-3.014712616184812e-05j`, but in only 3 shift-and-invert iterations.

(Note that the Q is higher than in the tutorial because I have more holes surrounding the cavity.)

stevengj:
If I change the number of holes to `N = 3` to match the tutorial, then I get a frequency of `0.23445413145107227-0.0003147776872265489j` (Q = 372.4), versus `0.235109393214226--3.14979827803982e-4j` (Q ≈ 373.2) from Harminv.  (Lowering `eigtol` to `1e-8`, I get the same eigenvalue to 9 digits.)

So, this is consistent with the Harminv result.

oskooi:
I have noticed that `solve_eigfreq` takes considerably longer for a parallel simulation than a serial run. The following holey-waveguide cavity example takes `41.9494 s` using a single-core Kaby Lake 4.2 GHz but `320.6739 s` using two cores. More significantly, the eigenfrequency results are different: `0.2344541320888902-0.0003147773149478817j` (single core) vs. `0.22236758070320453-0.018193974796342j` (two cores).

```py
import meep as mp

w = 1.2           # width of waveguide                                                                                                             
r = 0.36          # radius of holes                                                                                                                
d = 1.4           # defect spacing (ordinary spacing = 1)                                                                                          
N = 3        # number of holes on either side of defect                                                                                            
sy = 6            # size of cell in y direction (perpendicular to wvg.)                                                                            
pad = 2           # padding between last hole and PML edge                                                                                         
dpml = 1          # PML thickness                                                                                                                  
sx = 2*(pad+dpml+N)+d-1  # size of cell in x direction                                                                                             

geometry = [mp.Block(size=mp.Vector3(mp.inf,w,mp.inf), material=mp.Medium(epsilon=13))]
for i in range(N):
    geometry.append(mp.Cylinder(r, center=mp.Vector3(d/2+i)))
    geometry.append(mp.Cylinder(r, center=mp.Vector3(-(d/2+i))))

fcen = 0.25
src = [mp.Source(mp.ContinuousSource(fcen),
                 component=mp.Hz,
                 center=mp.Vector3(0),
                 size=mp.Vector3(0,0))]

                    force_complex_fields=True,
                    geometry=geometry,
                    boundary_layers=[mp.PML(1.0)],
                    sources=src,
                    symmetries=[mp.Mirror(mp.X, phase=-1), mp.Mirror(mp.Y, phase=-1)],
                    resolution=20)
sim.init_sim()
eigfreq = sim.solve_eigfreq(tol=1e-7)

if mp.am_master():
    print("eigfreq: {}".format(eigfreq))
```

stevengj:
Regarding the results being different — this is a bug, I forgot to call `sum_to_all` in a couple of places.  It should be fixed now.

stevengj:
One of the tests is timing out, but that is because it was taking [49 minutes and 51 seconds before](https://travis-ci.org/github/NanoComp/meep/jobs/667230107) and so even a few seconds more exceeds the 50-minute time limit.  We'll need to split up that test.

stevengj:
Rebased; the tests should run faster now thanks to #1161.

smartalecH:
Does this work in cylindrical coordinates too (e.g. for calculating bent waveguide modes?)

What's the advantage of this approach vs the typical harminv approach? If I set up a broadband simulation, I could even "capture" the mode profiles using a DFT monitor at the resonance frequency and normalize out the source (as long as it runs long enough to converge...), right?

stevengj:
I think cylindrical coordinates should be fine, but I don't recall trying it.

The `solve_eigfreq` was kind of an experiment, and currently we haven't found any case where it was a clear advantage.   Maybe it would be better if we implemented an Arnoldi method on top of it, but the basic difficulty is that the CW solver is slow to converge (though it could potentially be sped up too: #548).

PR #1163, opened 2020-03-29
arbitrary frequency grid for DFT functions in Python interface

### LINKED ISSUE #1070: allow unequal DFT frequencies

oskooi:
Closes #1070.

Follows the outline proposed by @ChristopherHogan in #1070:[comment](https://github.com/NanoComp/meep/issues/1070#issuecomment-604762275) and described in more detail by @stevengj in #1070:[comment](https://github.com/NanoComp/meep/issues/1070#issuecomment-605375194). All DFT functions can accept either `fcen,df,nfreq` (as 3 numbers) or a Numpy array `freq`.

Also includes a test (`python/tests/bend_flux.py`) as well as documentation.

ToDo: extend this functionality to the Scheme interface?


stevengj:
(Please keep your documentation PRs on a separate branch, to make PRs easier to review.)

python/simulation.py
```diff
@@ -1633,22 +1633,36 @@ def _evaluate_dft_objects(self):
...
+        components = args[0]
+        if isinstance(args[1],(int,float)) and isinstance(args[2],(int,float)) and isinstance(args[3],(int,float)):
+            freq_min = args[1]
+            freq_max = args[2]
+            nfreq = args[3]
+            freq = [0.5*(freq_min+freq_max)] if nfreq == 1 else np.linspace(freq_min,freq_max,nfreq)
+        elif isinstance(args[1],np.ndarray):
+            freq = args[1]
+        else:
+            raise TypeError("add_dft_fields only accepts freq_min,freq_max,nfreq (3 numbers) or freq (numpy array)")
```
stevengj: Rather than have these code copy-and-pasted multiple times, it would be better to have a single function that fixes the argument list.

e.g.
```py
args = fix_freq_args(args, 1)
freq = args[1]
```
where you have something like:
```py
def fix_freq_args(args, i):
     if args.len > i+2 and isinstance(args[i],(int,float)) and isinstance(args[i+1],(int,float)) and isinstance(args[i+2],int):
        freq_min = args[i]
        freq_max = args[i+1]
        nfreq = args[i+2]
        freq = [0.5*(freq_min+freq_max)] if nfreq == 1 else np.linspace(freq_min,freq_max,nfreq)
        del args[i:i+2]
        args[i] = freq
    elseif not isinstance(args[1], (np.ndarray, list)):
        raise TypeError("add_dft functions only accept freq_min,freq_max,nfreq (3 numbers) or freq (array/list)")
```
Note also that `nfreq` has to be an integer and `freq` can be a `list` as well as a numpy array.

python/tests/bend_flux.py
```diff
@@ -68,7 +68,7 @@ def init(self, no_bend=False, gdsii=False):
         refl_fr = mp.FluxRegion(center=mp.Vector3((-0.5 * sx) + 1.5, wvg_ycen),
                                 size=mp.Vector3(0, w * 2))
 
-        self.refl = self.sim.add_flux(fcen, df, nfreq, refl_fr)
+        self.refl = self.sim.add_flux(np.linspace(fcen-0.5*df,fcen+0.5*df,nfreq), refl_fr)
```
stevengj: Doesn't the old interface still work?

stevengj: I guess you just wanted to test the new API.

src/dft.cpp
```diff
@@ -30,14 +30,15 @@ typedef complex<double> cdouble;
 
 namespace meep {
 
-std::vector<double> linspace(double freq_min, double freq_max, int Nfreq) {
+std::vector<double> linspace(double freq_min, double freq_max, size_t Nfreq) {
   double dfreq = Nfreq <= 1 ? 0.0 : (freq_max - freq_min) / (Nfreq - 1);
   std::vector<double> freq;
+  freq.resize(Nfreq);
   if (Nfreq <= 1)
```
stevengj: Can't you do:
```cxx
std::vector<double> freq(Nfreq);
```
rather than calling `resize`?

src/dft.cpp
```diff
@@ -379,6 +359,16 @@ dft_flux::dft_flux(const component cE_, const component cH_, dft_chunk *E_, dft_
   freq = freq_;
 }
 
+dft_flux::dft_flux(const component cE_, const component cH_, dft_chunk *E_, dft_chunk *H_,
+                   const double *freq_, size_t Nfreq, const volume &where_,
+                   direction normal_direction_, bool use_symmetry_)
+    : E(E_), H(H_), cE(cE_), cH(cH_), where(where_), normal_direction(normal_direction_),
+      use_symmetry(use_symmetry_) {
```
stevengj: similar to above, do `freq(Nfreq)` in the initializer list?

src/dft.cpp
```diff
@@ -726,6 +706,14 @@ dft_fields::dft_fields(dft_chunk *chunks_, const std::vector<double> freq_, cons
   freq = freq_;
 }
 
+dft_fields::dft_fields(dft_chunk *chunks_, const double *freq_, size_t Nfreq, const volume &where_)
+    : where(where_) {
+  chunks = chunks_;
+  freq.resize(Nfreq);
```
stevengj: as above…

src/near2far.cpp
```diff
@@ -57,6 +57,22 @@ dft_near2far::dft_near2far(dft_chunk *F_, const std::vector<double> freq_, doubl
   }
 }
 
+dft_near2far::dft_near2far(dft_chunk *F_, const double *freq_, size_t Nfreq, double eps_, double mu_,
+                           const volume &where_, const direction periodic_d_[2],
+                           const int periodic_n_[2], const double periodic_k_[2],
+                           const double period_[2])
+    : F(F_), eps(eps_), mu(mu_), where(where_) {
+  freq.resize(Nfreq);
```
stevengj: as above

python/simulation.py
oskooi: Unfortunately, this is not possible singe `args` is a tuple which is immutable. We could copy `args` into a list within `fix_freq_args` and return that but that is not so nice. Also, note that because `add_dft_fields` accepts `freq_min`,`freq_max`,`nfreq` (rather than `fcen`,`df`,`nfreq` like all the other `add_dft` functions), it would need to have its own function to set up the `freq` array.

stevengj: Then you can do:
```py
args = fix_freq_args(args, 1)
```py
def fix_freq_args(args, i):
     if args.len > i+2 and isinstance(args[i],(int,float)) and isinstance(args[i+1],(int,float)) and isinstance(args[i+2],int):
        freq_min = args[i]
        freq_max = args[i+1]
        nfreq = args[i+2]
        freq = [0.5*(freq_min+freq_max)] if nfreq == 1 else np.linspace(freq_min,freq_max,nfreq)
        return args[:i] + (freq,) + args[i+3:]
    elseif not isinstance(args[1], (np.ndarray, list)):
        raise TypeError("add_dft functions only accept freq_min,freq_max,nfreq (3 numbers) or freq (array/list)")
    else:
        return args
```

And we should really just change `add_dft_fields`.  It's not acceptable that it is inconsistent with every other DFT function in Meep.

oskooi: The arguments for `add_dft_fields` have been changed from `freq_min`,`freq_max` to `fcen`, `df` to be consistent with the other `add_dft` functions. Some of the `python/tests` which involved `add_dft_fields` have also been updated. (This is only for the Python interface, not Scheme.)

stevengj: Unfortunately, this is a breaking change.   We probably want to emit a BIG WARNING if `df ≥ fcen` in this function for the next few releases to catch people who are using the old API.

python/simulation.py
```diff
@@ -1633,22 +1632,36 @@ def _evaluate_dft_objects(self):
...
+        components = args[0]
+        if len(args) == 4 and isinstance(args[1],(int,float)) and isinstance(args[2],(int,float)) and isinstance(args[3],int):
+            fcen = args[1]
+            df = args[2]
+            nfreq = args[3]
+            freq = [fcen] if nfreq == 1 else np.linspace(fcen-0.5*df,fcen+0.5*df,nfreq)
+        elif len(args) == 2 and isinstance(args[1],(np.ndarray,list)):
+            freq = args[1]
+        else:
+            raise TypeError("add_dft_fields only accepts fcen,df,nfreq (3 numbers) or freq (array/list)")
```
stevengj: Can we please use my `fix_dft_args` suggestion so that we don't have these same 9 lines copy-and-pasted in to many different functions?

oskooi: Done.

python/simulation.py
oskooi: Done.

python/simulation.py
```diff
@@ -44,6 +44,19 @@
...
+    if len(args) > i+2 and isinstance(args[i],(int,float)) and isinstance(args[i+1],(int,float)) and isinstance(args[i+2],int):
+        fcen = args[i]
+        df = args[i+1]
+        nfreq = args[i+2]
+        freq = [fcen] if nfreq == 1 else np.linspace(fcen-0.5*df,fcen+0.5*df,nfreq)
+        if i == 1 and df >= fcen:
+            warnings.warn("add_dft_fields: df >= fcen ({} >= {}). input arguments are (fcen,df) rather than (freq_min,freq_max)".format(df,fcen), RuntimeWarning)
+        return args[:i] + (freq,) + args[i+3:]
+    elif not isinstance(args[i], (np.ndarray, list)):
+        raise TypeError("add_dft functions only accept freq_min,freq_max,nfreq (3 numbers) or freq (array/list)")
```
stevengj: This should say `fcen,df`

PR #1167, opened 2020-04-03
Introducing Adjoint Solver 2.0

smartalecH:
This is an initial stab at the new and improved adjoint solver. This update is meant to enable an adjoint solver that is simple, flexible, and accurate.

**Major differences and features from previous version:**
- Support for broadband dispersive materials
- Calculates gradients for multiple frequency points simultaneously -- even if the objective function has multiple arguments
- Plug and play support for filters, projections, "robust" implementations, etc.
- Automated differentiation of objective functions using `autograd`
- Accurate gradients rigorously benchmarked against their finite differences
- Simplified API with minimal code overhead
- Built-in bilinear interpolation basis

The quickest and easiest way to get up to speed with the new features is to review the accompanying notebooks:
- [Introduction](https://nbviewer.jupyter.org/github/smartalecH/meep/blob/adjoint/python/examples/adjoint_optimization/01-Introduction.ipynb)
- [Waveguide bend optimization](https://nbviewer.jupyter.org/github/smartalecH/meep/blob/adjoint/python/examples/adjoint_optimization/02-Waveguide_Bend.ipynb)
- [Filters and projections](https://nbviewer.jupyter.org/github/smartalecH/meep/blob/adjoint/python/examples/adjoint_optimization/03-Filtered_Waveguide_Bend.ipynb)
- [Broadband simulations](https://nbviewer.jupyter.org/github/smartalecH/meep/blob/adjoint/python/examples/adjoint_optimization/04-Splitter.ipynb)

I'm sure several changes will have to be made before merging. Some of those include:
- [ ] Docs
- [ ] Tests
- [ ] Fix interpolation gradient with lots of design parameters per pixel
- [x] Supply typical filters as built-in functions.
- [x] Specify a minimum run time, if you need to force it to run longer.
- [x] Pass an freq array (like other features recently added).
- [x] Instead of passing a single point to determine when to stop the simulation, look at the objective arguments and/or the Fourier transforms of the fields in the design region.
- [x] Maybe call the variable `design_variables = [...]` instead of `basis = [...]`.
- [x] Would be good to support passing a list of objective functions (all with the same arguments).
- [x] Add a tutorial that designs a fabricatable structure (using a simple design like the bend)
- [ ] Extend a basic tutorial (like the phase grating) to calculate the sensitivity
- [ ] Simple 3d example


stevengj:
The bend optimization results look like they are converging towards:
image

stevengj:
Comments:

* Supply typical filters as built-in.
* Specify a minimum run time, if you need to force it to run longer.
* Pass an `freq` array (like other features recently added).
* Instead of passing a single point to determine when to stop the simulation, look at the objective arguments and/or the Fourier transforms of the fields in the design region.
* Maybe call the variable `design_variables = [...]` instead of `basis = [...]`.

stevengj:
Would be good to support passing a *list* of objective functions (all with the same arguments).   Then you do one forward simulation to get all of the objective functions (at each frequency), but then do a separate adjoint simulation for each objective function.

The reason to do this is to support arbitrary constrained optimization problems.  For example, suppose we are maximizing power in mode 1 with a constraint that the power in mode 2 is ≤ 0.1 × (power in mode 1).

stevengj:
If you plot the gradient of an optimized structure, it should be positive where ε=12 and negative where ε=1 (or vice versa, depending on whether you are maximizing or minimizing), and nearly zero at points where ε takes on intermediate values.  (This is just the KKT conditions.)

smartalecH:
I've made the requested changes but discovered an annoying bug with my interpolator. It seems that if I have far more design parameters than meep pixels in the design region, the gradient doesn't match up with its finite difference approximate.

You can view this effect in [this](https://nbviewer.jupyter.org/github/smartalecH/meep/blob/adjoint/python/examples/adjoint_optimization/debug_base.ipynb) notebook.

Interestingly I can still optimize fairly well (as demonstrated in the [filtering](https://nbviewer.jupyter.org/github/smartalecH/meep/blob/adjoint/python/examples/adjoint_optimization/03-Filtered_Waveguide_Bend.ipynb) example).

smartalecH:
Tutorials are fixed.

[Filtered Waveguide Bend](https://nbviewer.jupyter.org/github/smartalecH/meep/blob/adjoint/python/examples/adjoint_optimization/03-Filtered_Waveguide_Bend.ipynb) designs a fully binarized, broadband bend with a minimum length scale constraint:

image

with a decent response:

image

I'd say it's ready for merging. I can pull docs and "official" tests together in another PR.

franklee90:
Hi, I tried to run the introduction script but got this error:

AttributeError: module 'meep.adjoint' has no attribute 'BilinearInterpolationBasis'

I updated my meep using the simpetus nightly build before trying. Has the adjoint solver 2.0 been included in it?

stevengj:
No, it was just merged, so it will take 24 hours or so to appear in the nightly builds.

PR #1170, opened 2020-04-03
Apparent Typo

aarogon:
I assume that the sentence refers to ω as frequency, not μ.


oskooi:
Good catch.

aarogon:
I was trying to be consistent with the other unicode Greek characters there. Would it be better to have them all changed to LaTeX-style names?

oskooi:
Latex is preferred although unfortunately the documentation is not as consistent as it should be. For example, there are occurrences of ε and μ when these should be $\varepsilon$ and $\mu$. If you can go through and make all these changes as part of this PR, that would be helpful.

When editing the markdown files of the documentation, make sure you verify that it renders correctly by viewing the page locally in your browser using `mkdocs`. From a shell terminal, you simply run `mkdocs serve` from the source tree and then paste the URL that is shown into your browser.

PR #1177, opened 2020-04-10
Source shape as Volume for Python interface

### LINKED ISSUE #1176: source shape as Volume as alternative to center and size
Currently, the `Source` shape must be defined via the `center` and `size` properties. As an alternative, it would be useful to assign a `Volume` to the shape directly rather than doing something roundabout like `center=myvol.center` and `size=myvol.size` which is used in [Tutorial/GDSII Import/S-Parameters of a Direction Coupler](https://meep.readthedocs.io/en/latest/Python_Tutorials/GDSII_Import/#s-parameters-of-a-directional-coupler) (because `meep.GDSII_vol` returns a `Volume`). This kind of interface would be consistent with other functions which also contain `center` and `size` properties such as `FluxRegion` and `get_farfields`.

oskooi:
Closes #1176.

Also includes updates to the two examples in [Tutorial/GDSII Import](https://meep.readthedocs.io/en/latest/Python_Tutorials/GDSII_Import).

Add the same feature to the Scheme interface in a separate PR?


stevengj:
To do this in Scheme you might need to mess around in libctl at a lower level…

PR #1192, opened 2020-04-23
merge_subgroup_data

### LINKED ISSUE #1178: Using MPI across multiple sims
Typically if I have a python script with one simulation object, I can easily parallelize it using mpi:

```bash
mpirun -np 16 python3 my_script.py
```

Now let's say my script has four simulation objects and I want to run all of them in parallel and each with multiple processes from the same script.  Using the example above with 16 allocated processes, is there a clever way to assign 4 processes to each of my simulation objects that all run concurrently (and only communicate between "subgroups")?

This would be especially useful for multiobjective adjoint optimization.

I realize I could do this with some clever bash scripting but it would be nice to keep everything "in the loop" of a single python script.

smartalecH:
Closes #1178 

Adds a `merge_subgroup_data` method. Takes a numpy array (int, float, complex, etc) and returns a new array with an extra dimension corresponding to the concatenated input from each subgroup.

An example script `test.py`:

```python3
import meep as mp
import numpy as np

num_groups = 3
a = np.zeros((5,7),dtype=np.complex128)

n = mp.divide_parallel_processes(num_groups)
a[:] = (1+n) + 1j*(1+n)

b = mp.merge_subgroup_data(a,num_groups)
print(b.shape)
print(b)
```

Execute using:

```bash
mpirun -n 6 python3 test.py
```

Produces the expected output:
```
(5, 7, 3)
[[[1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]]

 [[1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]]

 [[1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]]

 [[1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]]

 [[1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]
  [1.+1.j 2.+2.j 3.+3.j]]]
```


python/simulation.py
```diff
@@ -3169,3 +3169,52 @@ def quiet(quietval=True):
 
 def verbosity(verbose_val):
     mp.cvar.verbosity = verbose_val
+
+def merge_subgroup_data(data,num_groups):
```
stevengj, line 3173: Would be nice to compute `num_groups` automatically.   We don't currently store it, but you can compute it via something like:
```py
num_groups = comm.allreduce(int(my_rank() == 0), op=MPI.SUM)
```

python/simulation.py
```diff
@@ -3169,3 +3169,52 @@ def quiet(quietval=True):
...
+    dtype=data.dtype.char
+    
+    # Initialize new input and output datasets
+    input=np.zeros(shape,dtype,order='F')
+    input[:]=data
+    output=np.zeros(out_shape,dtype,order='F')
+    
+    # Specify how much talking each proc will do
+    count_in = np.array([size] * num_workers)
+    count_out = np.array([size] * num_workers)
```
stevengj, line 3192: So it looks like all of the workers are sending data, but multiple workers overwrite the same section of the output array?   Shouldn't we just send from `myrank() == 0` in each group?

python/simulation.py
```diff
@@ -3169,3 +3169,52 @@ def quiet(quietval=True):
...
+    comm = MPI.COMM_WORLD
+    num_workers = comm.Get_size()
+    shape=data.shape
+    size = data.size
+    out_shape = shape + (num_groups,)
+    dtype=data.dtype.char
+    
+    # Initialize new input and output datasets
+    input=np.zeros(shape,dtype,order='F')
+    input[:]=data
```
stevengj, line 3187: Seems like it would be better to do
```py
input = np.array(data, order='F')
```
and then compute `shape` and `size` from `input`.  That way it will work even if `data` is not a numpy array (e.g. for Python lists).

python/simulation.py
```diff
@@ -3169,3 +3169,52 @@ def quiet(quietval=True):
...
+            i=MPI.INT,
+            l=MPI.LONG,
+            q=MPI.LONG_LONG,
+            f=MPI.FLOAT,
+            d=MPI.DOUBLE,
+            F=MPI.C_COMPLEX,
+            D=MPI.C_DOUBLE_COMPLEX)
+    mpi_dtype = TypeMap[dtype]
+    sdt = [mpi_dtype] *num_workers
+    rdt = [mpi_dtype] *num_workers
```
stevengj, line 3207: unused?

python/simulation.py
```diff
@@ -3169,3 +3169,52 @@ def quiet(quietval=True):
...
+    rdsp = [i for i in range(0, size*num_groups,size) for _ in range(int(num_workers/num_groups))]
+
+    # Formulate send and receive packets
+    smsg = [input, (count_in, sdsp), mpi_dtype]
+    rmsg = [output, (count_out, rdsp), mpi_dtype]
+    
+    # Send and receive
+    comm.Alltoallv(smsg, rmsg)
+
+    return output
```
stevengj, line 3220: missing newline

python/simulation.py
smartalecH: Seems like mpi4py has built-in type conversion so I'll omit the typemap.

python/simulation.py
smartalecH: Agreed.

python/simulation.py
smartalecH: I had trouble preventing deadlocks, which is why I originally did it this way. I've overcome that issue though. Now only group masters send their data, but every process receives data.

smartalecH:
New script:

```python
import meep as mp
import numpy as np
from mpi4py import MPI

num_groups = 3
a = np.zeros((5,7),dtype=np.complex128)

n = mp.divide_parallel_processes(num_groups)
a[:] = (1+n) + 1j*(1+n)

b = mp.merge_subgroup_data(a)
print(b.shape)
print(b)
```

Produces the same output as above, but now more efficiently (no more overwriting redundant data). 

I also added helper functions that return the number of groups (`get_num_groups()`) and an array of group master ranks (`get_group_masters()`).

stevengj:
You can put a test in the documentation PR.

zlin-opt:
@stevengj @smartalecH 
If the num_groups does not divide the total number of processes, what happens? 
do the extra processors get assigned to the first group?
for example, num_groups=3 and total procs is 13, then the first group has 13 procs, the second 12 and the third 12?

oskooi:
Yes, those three groupings are correct with the ranks of each process divided into (0-4), (5-8), and (9-12).

The relevant line in the function `divide_parallel_processes` which assigns the group number to each process (based on its rank) is:

https://github.com/NanoComp/meep/blob/14fe4e03d0581ff9f209ba265e3da43e90cb2036/src/mympi.cpp#L706

Issue #1195, opened 2020-04-26
coupler notbook example results not as expected


ValerioGiove:
Hi,
I'm new on Meep.

Trying to run the coupler notebook example, without any change on code, and the results seem not consistent to what expected:

MPB solved for frequency_1(0.645161,0,0) = 0.645161 after 11 iters
Dominant planewave for band 1: (0.645161,-0.000000,0.000000)
MPB solved for frequency_1(0.645161,0,0) = 0.645161 after 11 iters
Dominant planewave for band 1: (0.645161,-0.000000,0.000000)MPB solved for frequency_1(0.645161,0,0) = 0.645161 after 11 iters
Dominant planewave for band 1: (0.645161,-0.000000,0.000000)
MPB solved for frequency_1(0.645161,0,0) = 0.645161 after 11 iters
Dominant planewave for band 1: (0.645161,-0.000000,0.000000)
MPB solved for frequency_1(0.645161,0,0) = 0.645161 after 11 iters
Dominant planewave for band 1: (0.645161,-0.000000,0.000000)
MPB solved for frequency_1(0.645161,0,0) = 0.645161 after 11 iters
Dominant planewave for band 1: (0.645161,-0.000000,0.000000)
trans:, 0.12, 0.000042, 0.079574, 0.069195
MPB solved for frequency_1(0.645161,0,0) = 0.645161 after 11 iters
Dominant planewave for band 1: (0.645161,-0.000000,0.000000)
MPB solved for frequency_1(0.645161,0,0) = 0.645161 after 11 iters
Dominant planewave for band 1: (0.645161,-0.000000,0.000000)
**trans:, 0.12, 0.000042, 0.079574, 0.069195**

The p3_trans,p4_trans should sum to 1, for energy conservation, instead are low as you can read here above. The Cell scale seem right to 34.4 x 8 um (with slight modification for d sweep accomodation) 

![image](https://user-images.githubusercontent.com/41860080/80311966-eb675d80-87e2-11ea-9f28-d59b3c58c417.png)

Furthermore the fields image seems to not take into account the coupler structure, even if the geometry seems right as attached below.

Thank you for your time
Regards
Valerio

-----------
Initializing structure...
time for choose_chunkdivision = 0.00104499 s
Working in 2D dimensions.
Computational cell is 34.4 x 7.82 x 0 with resolution 50
     prism, center = (-9.09425,1.32149,0)
          height 0, axis (0,0,1), sidewall angle: 0 radians, 174 vertices:
[...]

PR #1196, opened 2020-04-26
change Prism height to finite value in 2d cell for GDSII coupler tutorial

oskooi:
As reported in #1195, [Tutorial/GDSII Import/S-Parameters of a Directional Coupler](https://meep.readthedocs.io/en/latest/Python_Tutorials/GDSII_Import/#s-parameters-of-a-directional-coupler) produces incorrect results because a `Prism` object with a `height` of `0` in a 2d cell is effectively no object. #1127 fixed several tests with this change but somehow this tutorial example was left out.


stevengj:
I'll merge this, but we should really allow 0-height prisms in 2d.

PR #1204, opened 2020-04-30
specify direction in eigenmode source in optical forces tutorial

### LINKED ISSUE #1202: forcing sim to work in 3D rather than 2D
I'm running the posted example code for optical forces.  I'm comparing my output (errors and all) with that of the posted jupyter notebook at 
https://github.com/NanoComp/meep/blob/master/python/examples/parallel-wvgs-force.ipynb
The output that is shown on the example notebook 
   Working in 3D dimensions.
   Computational cell is 7 x 5 x 0.0333333 with resolution 30
I copy and pasted the exact same code into my notebook, and confirmed the same behavior from terminal.  In both situations I get the same error:
RuntimeError: meep: Could not determine normal direction for given grid_volume.
I noticed the initial output on my notebook and terminal a major difference:
    Working in 2D dimensions.
    Computational cell is 7 x 5 x 0 with resolution 30

I'm suspecting this is the reason why I'm getting the RuntimeError.  How do I force my simulation to run in 3D rather than defaulting to 2D?  I see in the example code the cell is defined as
    cell = mp.Vector3(sx+2*dpml,sy+2*dpml,0)
Would this not force a 2D simulation since the 3rd dimension is 0?

Below is the entire traceback
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-9-ba5c58f74146> in <module>
      7
      8 for k in range(len(s)):
----> 9     fluxes_odd[k], forces_odd[k] = parallel_waveguide(s[k],True)
     10     fluxes_even[k], forces_even[k] = parallel_waveguide(s[k],False)

<ipython-input-8-ae2b1ab9b4c1> in parallel_waveguide(s, xodd)
     82     wvg_force = sim.add_force(f, 0, 1, force_reg1, force_reg2)
     83
---> 84     sim.run(until_after_sources=500)
     85
     86     flux = mp.get_fluxes(wvg_flux)[0]

~/miniconda3/envs/mp/lib/python3.7/site-packages/meep/simulation.py in run(self, *step_funcs, **kwargs)
   2283
   2284         if self.fields is None:
-> 2285             self.init_sim()
   2286
   2287         self._evaluate_dft_objects()

~/miniconda3/envs/mp/lib/python3.7/site-packages/meep/simulation.py in init_sim(self)
   1335
   1336         for s in self.sources:
-> 1337             self.add_source(s)
   1338
   1339         for hook in self.init_sim_hooks:

~/miniconda3/envs/mp/lib/python3.7/site-packages/meep/simulation.py in add_source(self, src)
   1597         if isinstance(src, EigenModeSource):
   1598             if src.direction < 0:
-> 1599                 direction = self.fields.normal_direction(where)
   1600             else:
   1601                 direction = src.direction

~/miniconda3/envs/mp/lib/python3.7/site-packages/meep/__init__.py in normal_direction(self, where)
   4049
   4050     def normal_direction(self, where):
-> 4051         return _meep.fields_normal_direction(self, where)
   4052
   4053     def casimir_stress_dct_integral(self, dforce, dsource, mx, my, mz, ft, where, is_bloch=False):

RuntimeError: meep: Could not determine normal direction for given grid_volume.

oskooi:
Fixes #1202.

#1035 changed the default behavior when a 2d cell is combined with a `k_point` with a non-zero `z` component from a 3d to a 2d simulation. With this change, the 2d eigenmode source with `eig_freq_match=True` in [Tutorial/Optical Forces](https://meep.readthedocs.io/en/latest/Python_Tutorials/Optical_Forces/) was aborting with the error `Could not determine normal direction for given grid_volume.` To fix this, the `direction` property of the eigenmode source must be set to `z` from the default `AUTOMATIC`. (Note also that `NO_DIRECTION` will not work.) 

The Jupyter notebook and the Scheme version are also updated with this change.


stevengj:
In a 2d simulation with a nonzero kz, I think we should be able to automate this.

stevengj:
Arguable, we should be setting `eig_match_freq = False` here, because the kz is fixed by the simulation.

If `eig_match_freq == False` and we have a 2d simulation and a 2d region, with a nonzero kz (β), I think that Meep should automatically infer that the k direction is kz.

PR #1217, opened 2020-05-13
FAQ on modeling incoherent dipoles in cavities with long lifetimes

oskooi:
new FAQ:
- "Is it possible to model a dipole emitter in a resonant cavity for the case in which the cavity lifetime is longer than the coherence time of the dipole?"


stevengj:
I don't think this should be a FAQ.

Instead, it might be nice to have a tutorial with e.g. a 2d LED-like structure with a thin (1d) light-emitting layer, showing several ways to solve it:

1. Brute force: just a layer of random dipole emitters.   By linearity, we can make them white noise and then scale by a desired spectrum later.  (Similar to Alejandro's first near-field heat-transfer paper.)  You'll need lots of simulations (and/or a long time) to get the spectrum to good accuracy (low variance).

2. Better: since the dipoles are uncorrelated, we can just do a sequence of deterministic calculations, with one dipole at a time, for dipoles at different points — uncorrelated dipoles at different points just add their powers.  So, you compute the powers and (suitably normalized, similar to how LDOS is normalized) add them up, scaling by the desired spectrum. 

We could mention other techniques:

3. Even better: instead of summing dipoles, we could use a different set of basis functions, e.g. a  cosine series.  See Alex McCauley's Casimir FDTD paper.

4. Even better, for the case of volumetric emitters: use Alejandro and Homer's tricks to transform a volume of emitters into equivalent surface emitters.  Or just use SCUFF.

Issue #1084, opened 2019-12-21
improve memory usage of susceptibility objects


oskooi:
The storage of susceptibility objects for simulations involving [Drude-Lorentzian materials](https://meep.readthedocs.io/en/latest/Materials/#material-dispersion) is currently inefficient (see `susceptibility::needs_P` below). This results in much more memory being used than is actually necessary which is problematic for large simulations (as reported in #1001, #1082).

https://github.com/NanoComp/meep/blob/a3b92bb58a45f76f7808d22770e9b6879e2055f9/src/susceptibility.cpp#L66-L82

stevengj:
I'm guessing there could be addressed with a post-processing stop in boundaries.cpp, where it checks if any of the chunks it wants to communicate P with don't have that susceptibility, and if so it replaces the communication with a zero boundary condition for that polarization component.

PR #1219, opened 2020-05-14
For Bug#1117

### LINKED ISSUE #1117: Gyrotropic Fails on Multiple materials (process finished with exit code 139 (interrupted by signal 11: SIGSEGV)
hi, i've been trying to use the new gyrotropic feature, but it seems not to work when used for 2 or more materials. the following example is just a simple modification of the [faraday_rotation](https://github.com/seewhydee/meep/blob/master/python/examples/faraday-rotation.py) found the meep repository:

```
import meep as mp
from meep import materials as mt

b0 = 0.15 # magnitude of bias vector

## Set up and run the Meep simulation:
tmax = 100
L = 20.0
cell = mp.Vector3(0, 0, L)
fsrc, src_z = 0.8, -8.5
pml_layers = [mp.PML(thickness=1.0, direction=mp.Z)]

######## With this block works good ##########
BK7 = gyrotropic_conversion(mt.BK7, bias=mp.Vector3(0, 0, b0))
geometry = [
    mp.Block(material=BK7, size=mp.Vector3(mp.inf, mp.inf, L), center=mp.Vector3(0, 0, 0)),
]
#############################################

sources = [mp.Source(mp.ContinuousSource(frequency=fsrc),
                     component=mp.Ex, center=mp.Vector3(0, 0, src_z))]

sim = mp.Simulation(cell_size=cell, geometry=geometry, sources=sources,
                    boundary_layers=pml_layers, resolution=50)
sim.run(until=tmax)

## Plot results:
import numpy as np

ex_data = sim.get_efield_x().real
ey_data = sim.get_efield_y().real

z = np.linspace(-L/2, L/2, len(ex_data))
plt.figure(1)
plt.plot(z, ex_data, label='Ex')
plt.plot(z, ey_data, label='Ey')
plt.xlim(-L/2, L/2); plt.xlabel('z')
plt.legend()


plt.figure(2)
plt.subplot(2,1,1)
plt.plot(z, ex_data, label='Ex (MEEP)')
plt.xlim(-L/2, L/2); plt.xlabel('z')
plt.legend(loc='lower right')

plt.subplot(2,1,2)
plt.plot(z, ey_data, label='Ey (MEEP)')
plt.xlim(-L/2, L/2); plt.xlabel('z')
plt.legend(loc='lower right')
plt.tight_layout()
plt.show()

```
Ths previous code runs good and the only different thing is the function "gyrotropic_conversion" i created for convert existing meep.materials in gyrotropic.

```
def gyrotropic_conversion (material, bias):
    material_suc = []
    for susceptibility in material.E_susceptibilities:
        if type(susceptibility).__name__ == 'DrudeSusceptibility':
            material_suc.append(
                mp.GyrotropicDrudeSusceptibility(frequency=susceptibility.frequency, gamma=susceptibility.gamma,
                                                 sigma=susceptibility.sigma_diag.x, bias=bias))
        elif type(susceptibility).__name__ == 'LorentzianSusceptibility':
            material_suc.append(
                mp.GyrotropicLorentzianSusceptibility(frequency=susceptibility.frequency, gamma=susceptibility.gamma,
                                                 sigma=susceptibility.sigma_diag.x, bias=bias))

    return mp.Medium(epsilon=material.epsilon_diag.x, E_susceptibilities=material_suc, valid_freq_range=material.valid_freq_range)
```

Now if I add an extra material to the main code like this:

```
######## With this block doesnt work ##########
BK7 = gy.gyrotropic_conversion(mt.BK7, bias=mp.Vector3(0, 0, b0))
AU = gy.gyrotropic_conversion(mt.Au, bias=mp.Vector3(0, 0, b0))
geometry = [
    mp.Block(material=BK7, size=mp.Vector3(mp.inf, mp.inf, L), center=mp.Vector3(0, 0, 0)),
    mp.Block(material=AU, size=mp.Vector3(mp.inf, mp.inf, 2), center=mp.Vector3(0, 0, 0)),
]
#############################################
```

The execution stops with the error 

> process finished with exit code 139 (interrupted by signal 11: SIGSEGV)

or even if we just want to pretend an empty space with a small layer of any material (in this case BK7) like in the following, it won't also compile.

```
######## With this block doesnt work ##########
BK7 = gy.gyrotropic_conversion(mt.BK7, bias=mp.Vector3(0, 0, b0))
geometry = [
    mp.Block(material=BK7, size=mp.Vector3(mp.inf, mp.inf, 2), center=mp.Vector3(0, 0, 0)),
]
#############################################
```
best regards

seewhydee:
This switches gyromagnetic media to use the default needs_P function, and modifies gyrotropic_susceptibility::update_P to handle the cases where sigma is NULL.

The Faraday rotation test case (https://meep.readthedocs.io/en/latest/Python_Tutorials/Gyrotropic_Media/) is unaffected, and the test case by @mojv doesn't crash anymore. I haven't performed any additional tests.

Fixes #1117.


src/susceptibility.cpp
```diff
@@ -431,13 +431,6 @@ void *gyrotropic_susceptibility::copy_internal_data(void *data) const {
   return (void *)dnew;
 }
 
-bool gyrotropic_susceptibility::needs_P(component c, int cmp,
-                                        realnum *W[NUM_FIELD_COMPONENTS][2]) const {
-  if (!is_electric(c) && !is_magnetic(c)) return false;
-  direction d0 = component_direction(c);
-  return (d0 == X || d0 == Y || d0 == Z) && sigma[c][d0] && W[c][cmp];
-}
-
```
stevengj, line 440: It seems like we may be better off keeping this simpler `needs_P` criterion, and changing `gyrotropic_susceptibility::num_cinternal_notowned_needed` to return `0`.

Reason: since no offdiagonal averaging is done for the P components (https://github.com/NanoComp/meep/issues/1117#issuecomment-628591902), we never need to communicate the "not owned" P grid points from neighboring chunks.   So, we only need to allocate P if *this* chunk has the corresponding sigma.

oskooi:
Does this PR also address #1084?

stevengj:
@oskooi, no.

src/susceptibility.cpp
seewhydee: Yes, it seems that this is the correct fix (6310d75).

I verified that the Faraday rotation example still works, and the test case from @mojv no longer crashes.

src/susceptibility.cpp
```diff
@@ -471,8 +470,8 @@ void gyrotropic_susceptibility::update_P(realnum *W[NUM_FIELD_COMPONENTS][2],
                                  invdet * (gd * gd + gz * gz)}};
 
       FOR_COMPONENTS(c) DOCMP2 {
+	const direction d0 = component_direction(c);
         if (d->P[c][cmp][0]) {
-          const direction d0 = component_direction(c);
```
stevengj, line 475: Was this an accidental change?

stevengj:
Thanks for looking into this, @seewhydee!

PR #1231, opened 2020-05-26
TO interpolation and Multi-objective opt routines

smartalecH:
WIP; not yet ready for merging.

Adds new material interpolation methods and multi-objective optimization routines (fully differentiable using autograd).

The goal is to enable support for fully dispersive materials.


smartalecH:
No longer relevant.

PR #1234, opened 2020-06-01
WIP: multithreading support for grid loops

### LINKED ISSUE #228: OpenMP parallelism

oskooi:
Closes #228.

This is a slightly cleaned up version of @HomerReid's #565. The benchmarking routines (i.e. `checkpoint, print_benchmarks`) have been removed and the formatting has been made consistent with clang.

The good news is that it compiles and all checks pass for `MEEP_NUM_THREADS=0` (the default). The bad news is that some tests are failing (fields are blowing up) whenever `MEEP_NUM_THREADS` is a non-zero value.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -7,16 +7,16 @@ AM_CPPFLAGS = -I$(top_srcdir)/src
 BUILT_SOURCES = sphere-quad.h step_generic_stride1.cpp
 
 HDRS = meep.hpp meep_internals.hpp meep/mympi.hpp meep/vec.hpp	\
-bicgstab.hpp meepgeom.hpp material_data.hpp
+bicgstab.hpp meepgeom.hpp material_data.hpp multithreading.hpp
 
 libmeep_la_SOURCES = array_slice.cpp anisotropic_averaging.cpp 		\
 bands.cpp boundaries.cpp bicgstab.cpp casimir.cpp 	\
 cw_fields.cpp dft.cpp dft_ldos.cpp energy_and_flux.cpp 	\
 fields.cpp loop_in_chunks.cpp h5fields.cpp h5file.cpp 	\
 initialize.cpp integrate.cpp integrate2.cpp monitor.cpp mympi.cpp 	\
-multilevel-atom.cpp near2far.cpp output_directory.cpp random.cpp 	\
-sources.cpp step.cpp step_db.cpp stress.cpp structure.cpp structure_dump.cpp		\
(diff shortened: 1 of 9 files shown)


smartalecH:
This implementation seems overly complicated. 

The bulk of the work each timestep is always done within the `LOOP_OVER_IVECS` macro (as described in #565). The assumption is that folding the loop and wrapping in the proper directive isn't sufficient -- we need an entirely new class and bookkeeping mechanism. Specifically,

> the loop initialization and increment statements are much more complicated than the bare-bones possibililities (sic) allowable in OpenMP-parallelized loops.

Is this really true though? I don't think the loop bound statements are too complicated for openMP. In fact, I think it would be much cleaner and even more performant to simply collapse the loop.

Furthermore, there seems to be a "vectorized" macro, `S1LOOP_OVER_IVECS`, that isn't used anywhere in the codebase (even though most uses of `LOOP_OVER_IVECS` should satisfy the requirement of stride-1). Is there a reason for this?

Are there other macros/loops that experience significant overhead? It seems most of the other macros loop over very small quantities (e.g. fields, directions, etc.) whereas `LOOP_OVER_IVECS` often loops over the entire simulation domain (or at least what's owned by the proc).

stevengj:
I don't think there is an easy way to collapse the loops into a single loop without some kind of custom iterator.  The basic issues are that (1) we are not looping over a contiguous subarray in memory and (2) we need the three loop indices anyway in order to compute the (x,y,z) coordinate in each loop iteration (via `IVEC_LOOP_LOC` etcetera).   (In principle, we could have a single loop counter and then use a bunch of tricky modular-arithmetic operations to extract the 3 loop counters, but that is messy and expensive too.)

However, I think we should have only one version of all the loops, not two.

stevengj:
`S1LOOP_OVER_IVECS` is used in `step_generic_stride1.cpp`, which is generated by the Makefile — it is the same as `step_generic.cpp` but only works when the stride of the innermost loop is 1 and hence multiplication by the stride can be omitted.  This, in turn, is called by the `STEP_CURL` and `STEP_UPDATE_EDHB` macros in `meep_internals.hpp` for timestepping in cases where the stride is 1.

IIRC, this led to a 20–30% speedup due to increased SIMD utilization, but it is would be worthwhile checking whether this is still true.  Just change the `LOOPS_ARE_STRIDE1(gv)` macro to always return `false` and see how the performance changes.

PR #1240, opened 2020-06-05
Add Python docstrings

### LINKED ISSUE #1236: Python docstrings and docsystem integration
It would be good to transfer the Python [API reference](https://meep.readthedocs.io/en/latest/Python_User_Interface/) into docstrings attached to the classes/methods/functions themselves, so that you could e.g. get online help in interactive Jupyter sessions.

But we don't want to make two copies of every function's documentation, so that would mean filling in the reference manual from the docstrings.  Fortunately, there are several documentation-generator systems in Python for this.   (Note that we would want to be able to generate the *formatted* text, including LaTeX equations, so this would require some digging to see what is the best way to proceed.)

RobinD42:
Fixes #1236 

First pass implementation of a tool to pull docstrings in markdown format from the Python code, and add them to the main documentation tree.

TODOs:
- [x] Finish copying existing class docs into docstrings
- [x] Prettify the docstrings so they're more readable in the code
- [x] Reorder methods so `__init__` always comes first
- [x] Indent class members in the output (perhaps by adding a new CSS style)
- [x] Finish support for top-level function docstrings, add their docstrings
- [x] Add `__repr__` methods for classes that are used as default parameter values
- [ ] ~~Find way to auto-exclude SWIG and other lower-level or non-public things.~~
- [ ] ~~Handle auto documenting other types of items, such as properties??~~
- [ ] ~~Update intra-doc links to accomodate new document and anchor names~~
- [ ] ~~Remove the current Python Interface doc. (And rename the new one?)~~

Not part of this PR, but IMO "{Language} User Interface" name is very misleading, "{Language} User API" would be better.


python/geom.py
```diff
@@ -178,6 +287,54 @@ def __init__(self, epsilon_diag=Vector3(1, 1, 1),
                  H_chi2=None,
                  H_chi3=None,
                  valid_freq_range=FreqRange(min=-mp.inf, max=mp.inf)):
+        """
+        Creates a `Measure` object.
```
stevengj, line 291: Medium?

stevengj:
I like where this is heading!

One possibility would be to emulate the [Documenter.jl](https://juliadocs.github.io/Documenter.jl/stable/man/syntax/) package for Julia, which inserts Julia docstrings into Markdown documentation.    Essentially, we would take our current Python API markdown document, and replace the documentation for e.g. `Medium` with
````
```@docs
Medium
```
````
and then the postprocessing tool would insert the docstring for `Medium` there. 

The advantages of this, rather than generating the Python API docs wholesale, are:

1. We manually decide which things go in the manual, and in what order.  No need to filter out SWIG-generated code or undocumented features, for example.

2. We can still surround the docstring code with additional explanations, tutorials, etcetera.

File: python/geom.py (hunk 1 above)
Comment (RobinD42) [reply python/geom.py:291]: Oops. I kept making that switch in my head, not sure why I had "Measure" on my mind.

**Comment by RobinD42** [comment]
Yes, that's a good idea. I was thinking about adding some boolean class-level attribute to indicate if something should be added to the documentation, (defaulting to `False` if it wasn't present) but using a tag to indicate where to insert the doc is much cleaner, and would also allow to use a non-alphabetic ordering like what is in the original document.

**Comment by stevengj** [comment]
Is the generated `Python_User_Interface.md` supposed to be checked in too?

**Comment by RobinD42** [comment]
I wasn't sure what the preference/policy on that would be. Since it's now generated some projects would just remove it from git. Others would keep it in and developers would need to ensure it is rebuilt and checked in when any changes are made that affect it.  What makes the most sense to you?

**Comment by stevengj** [comment]
It would be nicer to not check it in, and build it automatically on commits.  `make dist` should certainly build it into the tarballs.

Unfortunately, since it requires Meep to be compiled, it may be too complicated for the mkdocs script.  One option would be to build it as part of Travis, and store it with Travis build artifacts for later use during documentation generation.   That probably requires some digging through the Travis docs to understand the artifacts system.

File: Makefile.am [hunk 2]
```diff
@@ -17,7 +17,9 @@ endif
   SUBDIRS += python
 endif
 
-EXTRA_DIST = NEWS.md LICENSE COPYRIGHT m4 meep-pkgconfig.in
+EXTRA_DIST = NEWS.md LICENSE COPYRIGHT m4 meep-pkgconfig.in \
+             doc generate_py_api.py mkdocs.yml
```
stevengj, line 21: Might want to just put `generate_py_api.py` into the `doc` directory?

Makefile.am
```diff
@@ -28,4 +30,19 @@ meep.pc: meep-pkgconfig
 libmeep:
 	cd src && $(MAKE)
 
+# Generate the Python API Document
+# Set PYTHONPATH to use the in-place build of meep
+python_api_doc: all
+	test -d doc || cp -rf $(top_srcdir)/doc . && chmod -R +rw doc
+	if test -z $$SKIP_BUILD_DOCS; then PYTHONPATH=./python $(PYTHON) generate_py_api.py; else true; fi
```
stevengj, line 37: then this will be `$(top_srcdir)/doc/generate_py_api.py`

configure.ac
```diff
@@ -587,6 +587,9 @@ if test "x$with_python" = xyes; then
    AC_CONFIG_FILES([libpympb/Makefile python/Makefile])
 fi
 
+AC_CONFIG_LINKS([generate_py_api.py:generate_py_api.py
+                 mkdocs.yml:mkdocs.yml])
```
stevengj, line 591: Probably not needed if we refer to `$(srcdir)` as needed.

PR #1242, opened 2020-06-09
WIP:MaterialGrid

### LINKED ISSUE #1199: material_grid API
Might be nice to have something like the `material_grid` feature in MPB where you can have a material function based on a numpy array of ε or u ∈ [0,1] values and it would linearly interpolate onto the Yee grid (implemented in C++ for performance) and then interpolate between two given materials as a function of u.

We might even implement the feature in MPB where you can have overlapping material grids, which is useful for projecting onto symmetry groups during optimization (http://arxiv.org/abs/arXiv:1405.4350) and potentially other applications (e.g. I initially implemented it to model overlapping lithography steps).

smartalecH:
closes #1199

Initial stab. The current iteration has memory leaks and struggles with dense meshes near the border of the object containing the material grid.

As of now, the user can create a `MaterialGrid` (in python). The routine accepts as arguments the number of design parameters, an optional initialization array, and the two `Mediums` that are to be interpolated. The user can then assign this new `MaterialGrid` object to a geometry's (e.g. a block) material.

Example waveguides:
mg
mg2

Swig and C++ routines are also in place. It also uses the existing libctl TO routines, so it's pretty efficient.

Todo:
- [x] Fix memory leaks
- [x] Fix issue on geometry interfaces
- [x] Interpolate lorentzians
- [x] Add function and interface to update material grid
- [x] Add gradients
- [x] Fix gradients of grids with dispersive materials
- [x] Enable ability to sum/multiply overlapping material grids
- [x] Enable mirror symmetries
- [x] Resolve failing python tests
- [x] Update notebooks
- [x] Add test
- [ ] ~~Add support for lossy materials~~?


src/meepgeom.cpp
```diff
@@ -748,6 +795,9 @@ void geom_epsilon::eff_chi1inv_matrix(meep::component c, symmetric_matrix *chi1i
     }
   }
 
+  if (mat->which_subclass == material_data::MATERIAL_GRID) {
+    goto noavg;
+  }
```
stevengj, line 800: If instead you return `true` for `is_variable` on `MATERIAL_GRID`, then `get_front_object` will return `false` and you can use that branch instead.

src/meepgeom.cpp
```diff
@@ -270,13 +280,116 @@ bool is_metal(meep::field_type ft, const material_type *material) {
...
+{
+  meep::realnum uprod = 1.0, umin = 1.0, usum = 0.0, udefault = 0.0, u;
+  matgrid_val_count = 0;
+  
+  // iterate through object tree at current point
+  if (tp) {
+    do {
+        u = material_grid_val(to_geom_box_coords(p, &tp->objects[oi]),(material_data *)tp->objects[oi].o->material);
+        if (matgrid_val_count == 0) udefault = u;
+        if (u < umin) umin = u;
```
smartalecH, line 311: I'm following the MPB convention by using `to_geom_box_coords`, rather than the more general `to_geom_object_coords`. When we enable polygons for material grids, we should consider this.

smartalecH:
The overlap and add/multiply method works great for rotational symmetries:
Figure_1
 Is there an easy way to do mirror symmetries using the same machinery?

src/fields.cpp
```diff
@@ -678,9 +678,9 @@ realnum linear_interpolate(realnum rx, realnum ry, realnum rz, realnum *data, in
...
 
   /* get the point corresponding to r in the epsilon array grid: */
-  x = pmod(int(rx * nx), nx);
-  y = pmod(int(ry * ny), ny);
-  z = pmod(int(rz * nz), nz);
+  x = rx == 1.0 ? nx-1 : pmod(int(rx * nx), nx);
+  y = ry == 1.0 ? ny-1 : pmod(int(ry * ny), ny);
+  z = rz == 1.0 ? nz-1 : pmod(int(rz * nz), nz);
 
   /* get the difference between (x,y,z) and the actual point
```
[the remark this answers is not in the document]
stevengj: If we don't want periodic, we shouldn't be using pmod at all...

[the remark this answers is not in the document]
stevengj: It would make more sense here to use `int(rx * (nx - 1))` maybe?  We have to make sure the rest of the routine is consistent...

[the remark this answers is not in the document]
stevengj: Maybe:
```diff
diff --git a/src/fields.cpp b/src/fields.cpp
index a5b979d7..cfd0d36d 100644
--- a/src/fields.cpp
+++ b/src/fields.cpp
@@ -654,6 +654,9 @@ void fields::unset_solve_cw_omega() {
     chunks[i]->unset_solve_cw_omega();
 }
 
+/* implement mirror boundary conditions for i outside 0..n-1: */
+static int mirrorindex(int i, int n) { return i >= n ? 2*n-1-i : (i < 0 ? -1-i : i); }
+
 /* Linearly interpolate a given point in a 3d grid of data.  The point
    coordinates should be in the range [0,1], or at the very least [-1,2]
    ... anything outside [0,1] is *mirror* reflected into [0,1] */
@@ -664,34 +667,24 @@ realnum linear_interpolate(realnum rx, realnum ry, realnum rz, realnum *data, in
   realnum dx, dy, dz;
 
   /* mirror boundary conditions for r just beyond the boundary */
-  if (rx < 0.0)
-    rx = -rx;
-  else if (rx > 1.0)
-    rx = 1.0 - rx;
-  if (ry < 0.0)
-    ry = -ry;
-  else if (ry > 1.0)
-    ry = 1.0 - ry;
-  if (rz < 0.0)
-    rz = -rz;
-  else if (rz > 1.0)
-    rz = 1.0 - rz;
+  rx = rx < 0.0 ? -rx : (rx > 1.0 ? 1.0 - rx : rx);
+  ry = ry < 0.0 ? -ry : (ry > 1.0 ? 1.0 - ry : ry);
+  rz = rz < 0.0 ? -rz : (rz > 1.0 ? 1.0 - rz : rz);
 
   /* get the point corresponding to r in the epsilon array grid: */
-  x = pmod(int(rx * nx), nx);
-  y = pmod(int(ry * ny), ny);
-  z = pmod(int(rz * nz), nz);
+  x = mirrorindex(int(rx * nx), nx);
+  y = mirrorindex(int(ry * ny), ny);
+  z = mirrorindex(int(rz * nz), nz);
 
-  /* get the difference between (x,y,z) and the actual point
-     ... we shift by 0.5 to center the data points in the pixels */
-  dx = rx * nx - x - 0.5;
-  dy = ry * ny - y - 0.5;
-  dz = rz * nz - z - 0.5;
+  /* get the difference between (x,y,z) and the actual point */
+  dx = rx * nx - x;
+  dy = ry * ny - y;
+  dz = rz * nz - z;
 
   /* get the other closest point in the grid, with mirror boundaries: */
-  x2 = pmod((dx >= 0.0 ? x + 1 : x - 1), nx);
-  y2 = pmod((dy >= 0.0 ? y + 1 : y - 1), ny);
-  z2 = pmod((dz >= 0.0 ? z + 1 : z - 1), nz);
+  x2 = mirrorindex(dx >= 0.0 ? x + 1 : x - 1, nx);
+  y2 = mirrorindex(dy >= 0.0 ? y + 1 : y - 1, ny);
+  z2 = mirrorindex(dz >= 0.0 ? z + 1 : z - 1, nz);
 
   /* take abs(d{xyz}) to get weights for {xyz} and {xyz}2: */
   dx = fabs(dx);
diff --git a/src/meep.hpp b/src/meep.hpp
index 5ee0b244..967ea37a 100644
--- a/src/meep.hpp
+++ b/src/meep.hpp
@@ -2079,12 +2079,6 @@ vec get_k(void *vedata);
 realnum linear_interpolate(realnum rx, realnum ry, realnum rz, realnum *data, int nx, int ny,
                            int nz, int stride);
 
-// utility routine for modular arithmetic that always returns a nonnegative integer
-inline int pmod(int n, int modulus) {
-  n = n % modulus;
-  if (n < 0) n += modulus;
-  return n;
-}
 } /* namespace meep */
 
 #endif /* MEEP_H */
diff --git a/src/mpb.cpp b/src/mpb.cpp
index e4fa3875..83a29f62 100644
--- a/src/mpb.cpp
+++ b/src/mpb.cpp
@@ -108,6 +108,13 @@ typedef struct eigenmode_data {
 
 #define TWOPI 6.2831853071795864769252867665590057683943388
 
+// utility routine for modular arithmetic that always returns a nonnegative integer
+static int pmod(int n, int modulus) {
+  n = n % modulus;
+  if (n < 0) n += modulus;
+  return n;
+}
+
 /*******************************************************************/
 /* compute position-dependent amplitude for eigenmode source       */
 /*  (similar to the routine formerly called meep_mpb_A)            */
```

src/meepgeom.cpp
stevengj: `to_geom_box_coords` just calls `to_geom_object_coords` after shifting the center…

stevengj: Since we have an `geom_box_object` taken from a tree, I think `to_geom_box_coords` is the right thing.

stevengj:
> Is there an easy way to do mirror symmetries using the same machinery?

Yes, you just mirror the block object by flipping one or more of its basis vectors.

smartalecH:
The gradients are slightly less accurate than our previous approach:

Old method:
image

New method:
image

(plots taken from the bend jupyter notebook).

stevengj:
To compare errors, it might be more reliable to look at the convergence with resolution rather than the numbers at a particular resolution.

smartalecH:
I've fixed most of the gradient errors. If the two mediums in the grid are dispersionless, then the results look great:
no_dispersion

Dispersive media aren't as straightforward. If I strictly interpolate the Lorentzians (silicon and silicon dioxide in this case) without adding the conductivity damping term, the results also look good (note, I dropped the Courant factor to prevent aliasing/field instabilities):
no_cond

But as discussed, the conductivity factor is important to prevent instabilities as well. If I also try to interpolate the conductivity then the gradients start to fall apart:
dispersion

[Here](https://github.com/NanoComp/meep/files/4806147/material_grid.pdf) is a write-up that describes my implementation. I'll highlight important code pieces below.

I still need to figure out why `mpb.py` and `simulation.py` are failing. It has something to do with the `default_material` interpolating from a NumPy array.

src/meepgeom.cpp
```diff
@@ -270,13 +295,121 @@ bool is_metal(meep::field_type ft, const material_type *material) {
...
+    // calculate mean harmonic frequency
+    double omega_mean = 0;
+    for (int i = 0; i < m1->E_susceptibilities.num_items; i++) { omega_mean += m1->E_susceptibilities.items[i].frequency;}
+    for (int i = 0; i < m2->E_susceptibilities.num_items; i++) { omega_mean += m2->E_susceptibilities.items[i].frequency;}
+    omega_mean = omega_mean / (m1->E_susceptibilities.num_items + m2->E_susceptibilities.num_items);
+
+    // assign interpolated, nondimensionalized conductivity term
+    mm->D_conductivity_diag.x = mm->D_conductivity_diag.y = mm->D_conductivity_diag.z = u*(1-u) * omega_mean;
+  }
+}
```
smartalecH, line 411: Materials are interpolated in this function.

src/meepgeom.cpp
```diff
@@ -1968,4 +2144,372 @@ void fragment_stats::print_stats() const {
...
+  for (int i = 0; i < 3; i++) {
+    for (int j = 0; j < 3; j++) {
+      int idx = i * 3 + j;
+      dummy[i] += dA_du[idx] * fields_f[j];
+    }
+  }
+  
+  // inner product
+  std::complex<double> result = dummy[0] * fields_a[0] + dummy[1] * fields_a[1]  + dummy[2] * fields_a[2] ;
+  return 2*result.real();
```
smartalecH, line 2322: Gradient considering material interpolation is computed here.

stevengj:
Maybe try to debug the conductivity term by itself, with no Lorentzians.

smartalecH:
All tests are passing now.

After testing *just* the conductivity (i.e. no dispersion), I was unable to produce an accurate gradient. Consequently, I think the issue is with lossy materials in general.

For now, lossy support is disabled. The user can still interpolate Lorentzians, but they better choose profiles that won't cross zero when being turned on and off (Si in the materials library should be fine, for example). Once we can figure out how to calculate the gradients with lossy materials, we can add the Lorentzian damping parameter.

smartalecH:
Ready to merge once checks pass.

Note that gradients of FOM with lossy materials still don't converge. Everything else works well, however. I think the next step is to use a simpler objective function to debug (i.e. just a DFT component).

The current test only checks that the `material_grid` object initializes properly. Once we work out the lossy materials bug, we can think of some more comprehensive tests.

You can view the [updated notebooks](https://nbviewer.jupyter.org/github/smartalecH/meep/tree/material_grid/python/examples/adjoint_optimization/) to see how the material grid works.

PR #1244, opened 2020-06-10
ensure highly factorizable MPB sizes

### LINKED ISSUE #1243: pass optimized grid size to MPB mode solver
When calling the MPB mode solver, the resolution of the MPB solve is somewhat arbitrary compared to the Meep grid.  Since we can choose it freely, we should "round" the size up to the next highly factorizable size (power of 2,3,5,7) similar to the `optimize-grid-size!` function in MPB.

stevengj:
Fixes #1243.

### DIFF
--- a/src/mpb.cpp
+++ b/src/mpb.cpp
@@ -233,6 +233,20 @@ static double dot_product(const mpb_real a[3], const mpb_real b[3]) {
   return a[0] * b[0] + a[1] * b[1] + a[2] * b[2];
 }
 
+// return the next number factorizable into powers of 2,3,5,7
+// for efficient FFTs, similar to optimize-grid-size! in MPB:
+static int nextpow2357(int n) {
+  while (1) {
+    int m = n;
+    while (m % 3 == 0) m /= 3;
+    while (m % 5 == 0) m /= 5;
+    while (m % 7 == 0) m /= 7;
+    if ((m & (m - 1)) == 0) // if m is a power of 2
+      return n;
+    n += 1;
+  }
+}
+
 /****************************************************************/
 /* call MPB to get the band_numth eigenmode at freq frequency.  */
 /*                                                              */
@@ -332,6 +346,7 @@ void *fields::get_eigenmode(double frequency, direction d, const volume where, c
   for (int i = 0; i < 3; ++i) {
     n[i] = int(resolution * s[i] + 0.5);
     if (n[i] == 0) n[i] = 1;
+    n[i] = nextpow2357(n[i]);
     if (s[i] != 0)
       R[i][i] = s[i];
     else {


oskooi:
This PR does seem to generally provide a slight improvement compared to master in reducing the time for the MPB calculation in cases where the resolution is not factorizable (although the time reduction is not anywhere near an order of magnitude or more). This is demonstrated in the following test involving an eigenmode source for a 3d ridge waveguide. The wall-clock time for the eigenmode source as a function of the resolution in the integer range of 10 to 100 is shown using a linear and log-log plot.

Another feature of this PR apparent in the plots is that the MPB wall-clock time shows less variation than master with changes in the resolution.

eigsrc_mpb_time_comparison_linear

eigsrc_mpb_time_comparison_loglog

```py
import meep as mp
import argparse

def main(args):
    resolution = args.res

    h = 0.22       # waveguide height                                                                                                                              
    w = 0.50       # waveguide width                                                                                                                               

    Si = mp.Medium(index=3.5)
    cladding = mp.Medium(index=1.5)

    sxy = 4
    sz = 4
    cell_size = mp.Vector3(sxy,sxy,sz)

    # input waveguide                                                                                                                                               
    geometry = [mp.Block(material=Si,
                         center=mp.Vector3(),
                         size=mp.Vector3(mp.inf,w,h))]

    dpml = 1.0
    boundary_layers = [mp.PML(dpml)]

    # mode frequency                                                                                                                                                
    fcen = 1/1.55

    # source width                                                                                                                                                  
    df = 0.2*fcen

    symmetries = [mp.Mirror(mp.Y,-1),mp.Mirror(mp.Z,+1)]               

    eig_parity = mp.ODD_Y+mp.EVEN_Z

    sources = [mp.EigenModeSource(src=mp.GaussianSource(fcen,fwidth=df),
                                  size=mp.Vector3(0,sxy-2*dpml,sz-2*dpml),
                                  center=mp.Vector3(-0.5*sxy+dpml,0,0),
                                  eig_match_freq=True,
                                  eig_parity=eig_parity)]

			cell_size=cell_size,
			geometry=geometry,
			default_material=cladding,
			sources=sources,
			symmetries=symmetries)

    sim.init_sim()

    print("mpb-time:, {}, {}".format(resolution,sim.time_spent_on(6)[0]))

if __name__ == '__main__':
    parser.add_argument('-res', type=int, default=20, help='resolution (pixels/um)')
    args = parser.parse_args()
    main(args)
```

stevengj:
Note that this will only make a big difference if the number of pixels you are passing to MPB is a prime number.

If you resolution and the width of the eigenmode source are both nice "round" numbers, then their product isn't going to be prime.

oskooi:
Indeed, by modifying the resolution to be less "rounded" via `resolution = args.res + random.random()`, the improvement in the MPB wall-clock time enabled by this PR compared to master becomes more pronounced than using integer values (approaching a factor of ~2 at a resolution of around 100).

eigsrc_mpb_time_comparison3

PR #1250, opened 2020-06-16
fix runtime warning for add_dft_fields

### LINKED ISSUE #1249: Incorrect warning from add_dft_fields()?
I'm adding a single-point DFT monitor for a list of fields with the following line:

`dft_monitor = self.sim.add_dft_fields([mp.Ex, mp.Ey, mp.Ez, mp.Hx, mp.Hy, mp.Hz],
                                              self.f_meep_center, self.f_meep_delta, self.n_freq_flux,
                                              center=pt_active_monitor, size=mp.Vector3())`

It throws the following warning:

`/home/vinc/meep_conda/envs/pmp_nightly/lib/python3.6/site-packages/meep/simulation.py:54: RuntimeWarning: add_dft_fields: df >= fcen (0.023349486663870645 >= 0.018346025235898365). input arguments are (fcen,df) rather than (freq_min,freq_max)
  warnings.warn("add_dft_fields: df >= fcen ({} >= {}). input arguments are (fcen,df) rather than (freq_min,freq_max)".format(df,fcen), RuntimeWarning)
`

As you see, my _df_ is indeed greater than _fcen_, but why should this be a problem? Maybe a warning should only be thrown if _df/2_ > _fcen_?

oskooi:
Fixes #1249.

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -50,8 +50,8 @@ def fix_dft_args(args, i):
         df = args[i+1]
         nfreq = args[i+2]
         freq = [fcen] if nfreq == 1 else np.linspace(fcen-0.5*df,fcen+0.5*df,nfreq)
-        if i == 1 and df >= fcen:
-            warnings.warn("add_dft_fields: df >= fcen ({} >= {}). input arguments are (fcen,df) rather than (freq_min,freq_max)".format(df,fcen), RuntimeWarning)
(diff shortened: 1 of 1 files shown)


stevengj:
Probably best to get rid of the warning at this point, since the new version is not correct either and the old interface has been deprecated for a while

PR #1264, opened 2020-06-26
Fix dft fields memory leak

### LINKED ISSUE #1262: Non-reproducible outputs from adjoint solver (memory leak?)
Under certain conditions, when I create an an `OptimizationProblem` instance `opt = mpa.OptimizationProblem(...)` and then run `opt([x])` multiple times, the result can be very different for the first run (often `nan`), e.g. if I run

```python
x1 = np.ones(dv.Nx*dv.Ny) * (n_SiO2**2 + n_air**2)/2
print("Run 1")
(J0_a, dJ_a, dg_a) = opt([x1])
print("Run 2")
(J0_b, dJ_b, dg_b) = opt([x1])
print("Run 3")
(J0_c, dJ_c, dg_c) = opt([x1])
```

I find that `dJ_a` differs from `dJ_b`, `dJ_c` and is often very large, `inf`, or `nan`.  It's only the calculation of `dJ` that gets messed up; the actual simulations look fine (figures below).  One possibility is `meep.adjoint` is reading from the array memory that hasn't been initialized or is already freed (I've run into this with autograd elsewhere).

[Note14b-dJ.pdf](https://github.com/NanoComp/meep/files/4823092/Note14b-dJ.pdf)
[Note14b-sims.pdf](https://github.com/NanoComp/meep/files/4823094/Note14b-sims.pdf)

Full code below:

```python
import numpy as np
import meep as mp
import sys
sys.path.append("/Users/ryan.hamerly/Documents/Research/Code/meep/python")
import adjoint as mpa
import autograd.numpy as npa

# Global Constants
lam = 0.532;     f0 = 1/lam;      df = 0.2*f0;       res = 20

n_SiO2 = 1.45;   SiO2 = mp.Medium(index=n_SiO2)
n_air  = 1.00;   air  = mp.Medium(index=n_air)
n_Si   = 3.50;   Si   = mp.Medium(index=n_Si)

dpml = 1
(sx, sy) = (10, 7.7)
(Sx, Sy) = (sx+2*dpml, sy+2*dpml)

#%%

dr_res = 2*res
dr = mp.Volume(center=mp.Vector3(0, 2.4), size=mp.Vector3(sx, 0.9))
dv = mpa.BilinearInterpolationBasis(volume=dr, resolution=dr_res)

geom = [mp.Block(size=dr.size, center=dr.center, epsilon_func=dv.func())]
src_gauss = mp.Source(mp.GaussianSource(f0, fwidth=df),
                      center=mp.Vector3(0, sy/2-0.1), size=mp.Vector3(sx, 0), component=mp.Ez)

sim = mp.Simulation(cell_size=mp.Vector3(Sx, Sy),
                    resolution=res,
                    boundary_layers=[mp.PML(dpml)],
                    geometry=geom,
                    sources=[src_gauss],
                    default_material=SiO2,
                    eps_averaging=False)

TE0 = mpa.EigenmodeCoefficient(sim, mp.Volume(mp.Vector3(0, -sy/2+1), size=mp.Vector3(lam*0.7, 0)), mode=1,
                               eig_parity=mp.ODD_Z, k0=mp.Vector3(0, -1))
ob_list = [TE0]
def J(alpha):
    return npa.abs(alpha)**2

opt = mpa.OptimizationProblem(sim, J, ob_list, [dv], fcen=f0, df=0, nf=1, decay_fields=[mp.Ez])

#%%  Now run the simulation 3 times with the same data.

print("************************* First run of opt([x1]) *************************")
x1 = np.ones(dv.Nx*dv.Ny) * (n_SiO2**2 + n_air**2)/2
(J0_a, dJ_a, dg_a) = opt([x1])

print("************************* Second run of opt([x1]) *************************")
x1 = np.ones(dv.Nx*dv.Ny) * (n_SiO2**2 + n_air**2)/2
(J0_b, dJ_b, dg_b) = opt([x1])

print("************************* Third run of opt([x1]) *************************")
x1 = np.ones(dv.Nx*dv.Ny) * (n_SiO2**2 + n_air**2)/2
(J0_c, dJ_c, dg_c) = opt([x1])

#%%  Show the derivatives dJ/d_epsilon for the three (identical) trials

(f, axList) = plt.subplots(3, 1, sharex=True, sharey=True)
for (ax, dJ, ttl) in zip(axList, [dJ_a, dJ_b, dJ_c], ["Run 1", "Run 2", "Run 3"]):
    cax = ax.imshow(np.squeeze(dJ).T, origin='lower', aspect='auto', cmap='seismic');
    ax.set_xticks([]); ax.set_yticks([]); ax.set_title(ttl)
    f.colorbar(cax, ax=ax)
plt.savefig("Note14b-dJ.pdf")
#plt.show()

#%%  Simulation setup, forward and adjoint runs

print("************************* Forward run *************************")
(f, (ax1, ax2)) = plt.subplots(1, 2, figsize=(8, 4), sharex=True, sharey=True)
opt.prepare_forward_run()
opt.sim.run(until=20)
opt.sim.plot2D(ax=ax1, fields=mp.Ez, eps_parameters=dict(cmap='binary_r'))
print("************************* Adjoint run *************************")
opt.adjoint_run()
opt.sim.reset_meep()
opt.sim.run(until=20)
opt.sim.plot2D(ax=ax2, fields=mp.Ez, eps_parameters=dict(cmap='binary_r'))
ax1.set_title("Forward Run, $t=20$"); ax2.set_title("Adjoint Run, $t=20$"); ax2.set_ylabel("")
f.tight_layout(); f.savefig("Note14b-sims.pdf")
#plt.show()
```

smartalecH:
Fixes #1262 

Fixes three issues currently found in the `_get_dft_array` routine.
1. Initializes the dimension object `dim`. 2D simulations don't set the third dimension by default, consequently whatever garbage is in memory becomes the third dimension. This is obviously problematic when creating a python array from the dimension object.
2. When a dft array is requested from a component that isn't stored due to symmetry, source specifications, etc, the routine used to return a python array size 0. Python doesn't like this, and produces an array size 1 with whatever garbage is in memory there. Now, the routine returns an array with a size of 1 and with a complex `0`. This is especially important for adjoint calculations, which need to store all field components and don't have knowledge about which components meep is actually calculating. The current adjoint toolbox will already broadcast the `0` accordingly.
3. I am now using `PyArray_SimpleNewFromData` to create and copy the array data.

### DIFF
--- a/python/meep.i
+++ b/python/meep.i
@@ -418,11 +418,14 @@ meep::volume_list *make_volume_list(const meep::volume &v, int c,
 template<typename dft_type>
 PyObject *_get_dft_array(meep::fields *f, dft_type dft, meep::component c, int num_freq) {
     int rank;
-    size_t dims[3];
+    size_t dims[3] = {1,1,1};
     std::complex<double> *dft_arr = f->get_dft_array(dft, c, num_freq, &rank, dims);
 
-    if (rank==0 || dft_arr==NULL) // this can happen e.g. if component c vanishes by symmetry
-     return PyArray_SimpleNew(0, 0, NPY_CDOUBLE);
+    if (rank==0 || dft_arr==NULL) { // this can happen e.g. if component c vanishes by symmetry
+        npy_intp di[1] = {1};
+        std::complex<double> d[1] = {std::complex<double>(0,0)};
+        return PyArray_SimpleNewFromData(1, di, NPY_CDOUBLE, d);
+    }
 
     size_t length = 1;
     npy_intp *arr_dims = new npy_intp[rank];
--- a/python/meep.i
+++ b/python/meep.i
@@ -422,9 +422,8 @@ PyObject *_get_dft_array(meep::fields *f, dft_type dft, meep::component c, int n
     std::complex<double> *dft_arr = f->get_dft_array(dft, c, num_freq, &rank, dims);
 
     if (rank==0 || dft_arr==NULL) { // this can happen e.g. if component c vanishes by symmetry
-        npy_intp di[1] = {1};
         std::complex<double> d[1] = {std::complex<double>(0,0)};
-        return PyArray_SimpleNewFromData(1, di, NPY_CDOUBLE, d);
+        return PyArray_SimpleNewFromData(0, 0, NPY_CDOUBLE, d);
     }
 
     size_t length = 1;
--- a/python/meep.i
+++ b/python/meep.i
@@ -421,10 +421,8 @@ PyObject *_get_dft_array(meep::fields *f, dft_type dft, meep::component c, int n
     size_t dims[3] = {1,1,1};
     std::complex<double> *dft_arr = f->get_dft_array(dft, c, num_freq, &rank, dims);
 
-    if (rank==0 || dft_arr==NULL) { // this can happen e.g. if component c vanishes by symmetry
-        std::complex<double> d[1] = {std::complex<double>(0,0)};
-        return PyArray_SimpleNewFromData(0, 0, NPY_CDOUBLE, d);
-    }
+    if (rank==0 || dft_arr==NULL) // this can happen e.g. if component c vanishes by symmetry
+        return PyArray_SimpleNewFromData(0, 0, NPY_CDOUBLE, &std::complex<double>(0,0));
 
     size_t length = 1;
     npy_intp *arr_dims = new npy_intp[rank];


python/meep.i
```diff
@@ -418,11 +418,14 @@ meep::volume_list *make_volume_list(const meep::volume &v, int c,
...
-    size_t dims[3];
+    size_t dims[3] = {1,1,1};
     std::complex<double> *dft_arr = f->get_dft_array(dft, c, num_freq, &rank, dims);
 
-    if (rank==0 || dft_arr==NULL) // this can happen e.g. if component c vanishes by symmetry
-     return PyArray_SimpleNew(0, 0, NPY_CDOUBLE);
+    if (rank==0 || dft_arr==NULL) { // this can happen e.g. if component c vanishes by symmetry
+        npy_intp di[1] = {1};
+        std::complex<double> d[1] = {std::complex<double>(0,0)};
+        return PyArray_SimpleNewFromData(1, di, NPY_CDOUBLE, d);
```
stevengj, line 427: We could still return a 0-dimensional array here (pass `0,0` for the first 2 args), since that has 1 element.

python/meep.i
```diff
@@ -418,11 +418,14 @@ meep::volume_list *make_volume_list(const meep::volume &v, int c,
 template<typename dft_type>
 PyObject *_get_dft_array(meep::fields *f, dft_type dft, meep::component c, int num_freq) {
     int rank;
-    size_t dims[3];
+    size_t dims[3] = {1,1,1};
```
stevengj, line 421: This is fine, but the old code seemed safe — `get_dft_array` initializes only `rank` elements of `dims`, but those are all of the elements that we actually use.

python/meep.i
```diff
@@ -418,11 +418,13 @@ meep::volume_list *make_volume_list(const meep::volume &v, int c,
...
     std::complex<double> *dft_arr = f->get_dft_array(dft, c, num_freq, &rank, dims);
 
-    if (rank==0 || dft_arr==NULL) // this can happen e.g. if component c vanishes by symmetry
-     return PyArray_SimpleNew(0, 0, NPY_CDOUBLE);
+    if (rank==0 || dft_arr==NULL) { // this can happen e.g. if component c vanishes by symmetry
+        std::complex<double> d[1] = {std::complex<double>(0,0)};
+        return PyArray_SimpleNewFromData(0, 0, NPY_CDOUBLE, d);
```
stevengj, line 426: We could pass just `&std::complex<double>(0,0)` here, I think.

smartalecH:
Implemented in #1242.

PR #1265, opened 2020-06-26
benchmarking results for runtime scaling on MPI clusters for docs

### LINKED ISSUE #1256: parallel scaling tutorial/demo
Might be nice to have a tutorial showing the scaling with # processors P on a large cluster with a large simulations.   The runtime should decrease with P, reach a minimum (where communication costs start to take over), and then grow with P.

oskooi:
Closes #1256.

Adds a new section "Runtime Scaling on MPI Clusters" to [Features/Parallel Meep](https://meep.readthedocs.io/en/latest/Parallel_Meep/).


stevengj:
It would be nice to add some rough analysis of the scaling, but this is a bit more complicated than I would like.  It depends on whether the communication scales like the *area per process* (in the limit where every process-to-process communication occurs over a separate "pipe" with its own bandwidth) or the *total area of all processes* (in the limit where they are all sharing a single "pipe") or something in between depending on the network topology.   Put another way, it depends on how the usable bandwidth scales with the number of processors.

Issue #1222, opened 2020-05-16
LED tutorial


stevengj:
As discussed in #1217, it might be nice to have a tutorial with e.g. a 2d LED-like structure with a thin (1d) light-emitting layer, showing several ways to solve it:

1. Brute force: just a layer of random dipole emitters.   By linearity, we can make them white noise and then scale by a desired spectrum later.  (Similar to Alejandro's first near-field heat-transfer paper.)  You'll need lots of simulations (and/or a long time) to get the spectrum to good accuracy (low variance).

2. Better: since the dipoles are uncorrelated, we can just do a sequence of deterministic calculations, with one dipole at a time, for dipoles at different points — uncorrelated dipoles at different points just add their powers.  So, you compute the powers and (suitably normalized, similar to how LDOS is normalized) add them up, scaling by the desired spectrum. 

We could mention other techniques:

3. Even better: instead of summing dipoles, we could use a different set of basis functions, e.g. a  cosine series.  See Alex McCauley's Casimir FDTD paper.

4. Even better, for the case of volumetric emitters: use Alejandro and Homer's tricks to transform a volume of emitters into equivalent surface emitters.  Or just use SCUFF.

For example, you could look at a structure like [this one](https://www.semanticscholar.org/paper/Radiation-pattern-design-of-photonic-crystal-LED-by-Merikhi-Mirjalili/95164b09e20e50c86612bec7ff5065b25a87d33a):

![image](https://user-images.githubusercontent.com/2913679/82107411-642c4c00-96f5-11ea-913c-04801fb794fe.png)

but with all widths W equal so that it's periodic.  Compute the emission in the normal direction by treating the emitting layer as infinitesimally thin.    This allows you to model a single unit cell with periodic (k=0) boundary conditions.  (Emission into other directions can be computed by changing k.)

oskooi:
[...]

PR #1266, opened 2020-06-28
amplitude function tutorial for custom source

oskooi:
Extends [Tutorial/Custom Source/Stochastic Dipole Emission in Light Emitting Diodes](https://meep.readthedocs.io/en/latest/Python_Tutorials/Custom_Source/#stochastic-dipole-emission-in-light-emitting-diodes) to include a demonstration of Method 3 from #1222 involving a line source with cosine Fourier series amplitude function.


stevengj:
Looks good.  Would be nice to be able to compare the unnormalized runs from Method 2 and Method 3 as well, since they should actually be identical if we get our scaling factors correct.

oskooi:
For the unnormalized radiated flux, Method 2 is larger than Method 3 by *exactly* a factor of 10 as shown in the figures below.

In this example, the resolution is 50 and the number of frequencies is 500. Method 2 involves 75 dipoles. Method 3 involves 15 cosine terms.

method2_3_res50_nfreq500_nsrc15_spectra

method2_3_res50_nfreq500_nsrc15_spectra2

stevengj:
(Couldn't you have used 50 dipoles, i.e. the same number as the resolution?)

stevengj:
Some generalizations:

1. In 3d, with a planar source layer (Lx by Ly), then you have the *product* of two cosine series.  That is, each source is a product of two cosines cos(π n1 x / Lx) cos(π n2 y / Ly)  (and the product of the normalization factors), and you need to sum over n1 and n2 (two nested loops.  That is, if n1 and n2 go from 0 to 9 each, then you have 100 simulations.

2. If you want to look at emission in *all* directions for a periodic surface, then you need to integrate over kx and ky (the Bloch wavevectors of the unit cell) as well.   However, if you've designed your periodic emitter to emit mostly vertically (via a resonance at kx=ky=0), then the emission probably drops off rapidly at other k points and you can probably estimate the width of the resonance by sampling only a few k points near (0,0).

3. For a periodic emitter like this one where the emitter extends the whole width of the unit cell, it is probably more efficient to use a Fourier sine+cosine series.  That is, you do simulations with *both* sine and cosine sources *separately*, where the spatial frequency **2** π n x / Lx rather than π n x / Lx as now.  (This should converge exponentially rather than polynomially as for the cosine-only series).  The cosine series is more general, however, as it is a better choice for emitters that don't extend the whole width of a periodic unit cell.

4. For a 3d emission volume, you would need a product of 3 cosine series, which will have a lot of terms (1000 simulations even if you only do 10 terms each).  Here is where "method 4", using mathematical tricks from surface-integral equations to transform the sources from volume sources to surface sources, should help.

stevengj:
Note that with method (3), you have to do sine and cosine sources *separately*.    It might be instructive to show a plot of the error vs. the number of terms, because that is where you would see the difference between a cosine series (power-law convergence) and a Fourier (sine & cosines) series (exponential convergence).

PR #1272, opened 2020-07-01
additional benchmarking results for runtime scaling on MPI clusters

oskooi:
Four additional figures for [Features/Parallel Meep/Runtime Scaling on MPI Clusters](https://meep.readthedocs.io/en/latest/Parallel_Meep/#runtime-scaling-on-mpi-clusters) from #1265.

(Also includes a couple of minor unrelated doc fixes.)


stevengj:
For the DFT and timestepping time, I would plot 1/time vs P, so that you can see the linear scaling (or lack thereof).

oskooi:
Inverting the time for timestepping reveals the linear scaling trend but not so for DFT which seems to converge to a constant value in the limit of increasing number of processors. This is not surprising given that we used equal-sized chunks in the test example. A single chunk or a group of chunks is likely dominating the DFT runtime; this is indeed verified by the bar plot (see below) showing the times for the individual processors for the case of 35 processors or 7 cluster nodes. 

Perhaps we should therefore only include the timestepping result in the demonstration for the user manual?

timestep_averages1

timestep_averages2

DFT_averages1

DFT_averages2

oled_barplot

stevengj:
Does split-by-cost improve the DFT scaling?  If so, this could help to illustrate when split-by-cost is useful.

PR #1281, opened 2020-07-07
tutorial for absorbed power density

### LINKED ISSUE #1276: absorbed power density tutorial
Someone recently asked about computing the absorbed power density in a dispersive material using the polarization fields.   I think actually there is an easy way to do it with our existing features.

The absorbed power density is `Re[E* dP/dt]` where P is the total polarization field, or in Fourier domain it is `Re[E* (-iωP)] = ωIm[E*P]` (where `E*P` denotes the dot product of `conj(E)` with `P`).    However, since `D = E + P`, this is equivalent to `ωIm[E*(D-E)]=ωΙm[E*D]` (since `E*E = |E|²` is purely real).

So, you just use the `dft_fields` feature to compute the Fourier-transformed `E` and `D` fields in a region (which is already supported), and then in post-processing compute `ωΙm[E*D]`.

A simple test case would be to have a cylinder of a Lorentzian absorbing material with an incident planewave, and then map out the absorbed power density.   Would be good to also compare to the net absorbed power (computed via `dft_flux`).

Note that `ωΙm[E*D]` will *not* include the absorption by a conductivity (if any), because of how conductivity is implemented in Meep.   However, you could include conductivity in post-processing by calculating ωΙm[(1+iσ/ω) E*D]`.

oskooi:
Closes #1276.

Would be good to also add a Scheme version ~~but this does not seem to be entirely possible due to the lack of array metadata routines (i.e., `get_array_metadata`) for obtaining the interpolation weights~~.


doc/docs/Python_Tutorials/Basics.md
```diff
@@ -896,6 +896,111 @@ For `resolution = 25`, the error decreases (as expected) to 1.5%.
...
+print("flux:, {} (dft_fields), {} (dft_flux), {} (error)".format(absorbed_power,absorbed_flux,err))
+```
+
+A schematic of the simulation layout generated using [`plot2D`](Python_User_Interface.md#data-visualization) shows the line source (red), PMLs (green hatch region), `dft_flux` box (solid blue contour line), and `dft_fields` surface (blue hatch region).
+
+<center>
+![](../images/power_density_cell.png)
+</center>
+
+The spatial map of the absorbed power density shows that, as expected, most of the absorption occurs on the surface of the cylinder facing the incident planewave. Note that the labeled $x$ and $y$ coordinate axes of the 2d region are obtained using [`get_array_metadata`](Python_User_Interface.md#array-metadata) which also provides the interpolation weights (`w`) used to compute the integral of the absorbed power density.
```
stevengj, line 992: It seems to be absorbed in a pretty thin layer.   What is the absorption length λ/Im[sqrt(ε)] at this wavelength for this material?   If the absorption length is comparable to a pixel, maybe you should change the material or the wavelength so that the absorption length is a bit larger and easier to resolve in the plot?

oskooi: It's unclear why the absorption is confined to such a thin layer on the front surface. The pixel width for a resolution of 50 is 0.02 μm and is much smaller than the absorption length (see below).

SiO<sub>2</sub> at λ=1 μm has an absorption length of ~3000 μm (i.e., 3000X the radius of the cylinder).
```py
>>> from meep.materials import SiO2
>>> SiO2.epsilon(1.0)[0][0]
(2.132084213667966+0.0009852435550391251j)
>>> 1.0/np.imag(np.sqrt(SiO2.epsilon(1.0)[0][0]))
2964.070860833652
```

power_density_map_SiO2_wvl1 0

By comparison, [ITO](https://github.com/NanoComp/meep/blob/master/python/materials.py#L67-L80) at λ=0.3 μm has a much smaller absorption length (Labs) of ~4 μm.
```py
>>> from meep.materials import ITO
>>> ITO.epsilon(1/0.3)[0][0]
(4.932295718818409+0.3550131776059032j)
>>> 0.3/np.imag(np.sqrt(ITO.epsilon(1/0.3)[0][0]))
3.7558827876760446
```

The spatial map for the ITO cylinder shows that the absorption is more strongly confined to its front surface than that of SiO2.

power_density_map_ITO_wvl0 3

stevengj:
It certainly looks like there is some kind of discretization artifact happening at the surface…

oskooi:
Could the discretization artifact be due to field discontinuities at the surface of the dispersive material? (This should be mentioned in the tutorial.)

Increasing the resolution (50, 100, 200) shows that there is a fixed one-pixel thick absorption layer on the surface.

power_density_map_SiO2_wvl0 3_res50

power_density_map_SiO2_wvl0 3_res100

power_density_map_SiO2_wvl0 3_res200

A plot of the relative error in the total flux vs. resolution shows that the convergence is slightly *less* than quadratic which indicates that there is indeed a discontinuity present.

SiO2_wvl0 3_convergence2

stevengj:
What would happen if you got D and E on the Yee grid?

oskooi:
Setting:
```
dft_fields = sim.add_dft_fields([mp.Dz,mp.Ez], fcen, 0, 1, center=mp.Vector3(), size=mp.Vector3(2*r,2*r), yee_grid=True)
```
produces the following spatial map of the absorbed power density with *no* absorption on the surface:

power_density_map_SiO2_wvl0 3_res100

stevengj:
Until we have a method to compute the E⋅D product on the Yee grid and then interpolate to the centered grid, we can't use the get_array_metadata routine (which expects the centered grid).   A workaround is just to weight by Δx² when summing (since they only places the weights differ from this are on the edges of the rectangle, where the absorbed power is zero in this example).

stevengj:
Note that this power density is (Meep power)/a².     To convert to a physical power for a given input power, you would multiply by (actual power)/(Meep power flux), where (actual power) is the known physical input power and (meep power flux) is the corresponding power measured in Meep.

For example, if you plan to have an incident planewave with a power of (actual power)=1mW incident on the cylinder cross-section, then you would compute (Meep power flux) in a separate normalization run with just vacuum, measuring the (DFT) power on a line segment corresponding to the cylinder diameter.

PR #1233, opened 2020-05-27
automatic kpoint guess update

stevengj:
Implements strategy 1 from #1223.

PR #1285, opened 2020-07-09
Yee grid gradients

### LINKED ISSUE #1270: More accurate gradients
Currently, the adjoint solver interpolates the forward and adjoint fields at the center of each yee voxel and uses those values to calculate the gradient. Consequently, the gradient is less accurate. While the accuracy of the gradient converges with the resolution, there are some applications that benefit from lower resolution accuracy (especially since the gradient converges differently for certain problems).

An easy solution for simple problems that only care about one DFT component is to evaluate the DFT field points directly on the yee grid and multiply accordingly. If the FOM is even slightly more complicated, however, then this approach doesn't work since each component lies on a different grid point. There doesn't seem to be a straightforward way to evaluate λAₚx in this scenario.

Is there a way to either account for the interpolation in the gradient itself or to somehow combine the different DFT components into the gradient without interpolating at all?

smartalecH:
Closes #1270

Before:
image
image

After:
image
image


stevengj:
To automatically fix merge conflicts due to `clang-format`, see #662:
```
# update repo
git checkout master && git pull origin master # update master

# switch to your branch
git checkout MYBRANCH

# rebase onto PREVIOUS commit
git rebase ad8986b622631232324fc7a100fb726a297f22f3

# reformat and rebase to clang-format commit
find . -name '*.*pp' | xargs clang-format -i
git commit -a -m "reformat"
git rebase -Xtheirs c363d2e609b7a970279a503d5cb85cb6201ef138

# finish rebase
git rebase master
```

stevengj:
@zlin-opt has also been playing around with this, and recently set up an adjoint calculation "by hand" where he directly populated the Yee grid for his example problem and was able to get very good accuracy.  It would be good to compare your (more general) code to Zin's for his special case to see if you get equivalent accuracy or if one of you is missing a trick.

smartalecH:
Sounds great. Do you have a link?

zlin-opt:
> Sounds great. Do you have a link?

I will post my example case soon.

zlin-opt:
@stevengj @smartalecH 
Here is my test case for adjoint gradients on a 3d structure. The DOFs here are the variable heights.
image
adj_diff
adj_diff_err

my approach is to directly populate the Yee-grid through epsilon_func and amp_func for any given epsilon and current sources and also use yee-grid dft. I have made a repo for my codes.

smartalecH:
Looks great, Zin. Thanks for taking the time to pull this together.

I carefully went through your [repo](https://github.com/zlin-opt/meep-adjoint-3d) (linked for posterity). It seems very consistent with how I implemented meep's "general" adjoint solver. In particular, it's important to use the yee grid for the *adjoint source* and the *forward/adjoint fields* in the design region.

This PR enables evaluating the forward/adjoint fields directly on the yee grid, a major improvement that's long overdue. A future PR needs to implement the first half of Zin's efforts (also discussed in #1270): using the amp_func for the adjoint source. 

Currently, since the objective arguments are eigenmode coefficients, I simply use another eigenmode source pointing in the opposite direction as the adjoint source (with the correct scaling of course). There's some obvious handwaving here, especially when it comes to the yee grid.

 It would be better to "generalize" this by storing the fields and directly implementing the adjoint source on the yee grid. (Homer's original code did this to some extent). This is somewhat problematic, however, for eigenmode arguments, Poynting flux, or anything with a cross product because the corresponding adjoint source uses the "opposite" field quantities (i.e. J sources use H fields and M sources use E fields). This may not play nicely with the yee grid. Is it possible to put a point J source on an H yee grid point? Or are sources agnostic to the yee grid?

stevengj:
@smartalecH, I think the situation with eigenmode sources may be easier than you think; the main thing is to make sure that the overlap integral (for the coefficient extraction) occurs on the Yee grid rather than interpolating the DFT fields to the voxel centers first.  Then the subsequent current sources should do the right thing.

smartalecH:
I've tried evaluating the eigenmode coefficient using the yee grid directly, but the results are actually worse:
image
Note I also use the DFT fields directly for the adjoin source, but got identical results as and eigenmode source pointed in the opposite direction; so it's definitely something going on with the eigenmode coefficient.

python/adjoint/objective.py
```diff
@@ -88,30 +103,69 @@ def place_adjoint_source(self,dJ,dt):
...
+        if (((self.volume.size.y == 0) and (self.volume.size.z == 0)) or 
+            ((self.volume.size.x == 0) and (self.volume.size.z == 0)) or
+            ((self.volume.size.x == 0) and (self.volume.size.y == 0))):
+            self.dx = self.sim.resolution ** (-1)
+        elif (((self.volume.size.y != 0) and (self.volume.size.z != 0)) or 
+            ((self.volume.size.x != 0) and (self.volume.size.z != 0)) or
+            ((self.volume.size.x != 0) and (self.volume.size.y != 0))):
+            self.dx = self.sim.resolution ** (-2)
+        else:
+            self.dx = self.sim.resolution ** (-3)
```
smartalecH, line 166: eigenmode routine here

stevengj:
I think we need to add a `centered_grid` flag (defaults to `true`) to `add_dft_flux` so that you can passes `centered_grid=false` to `add_dft` [here](https://github.com/NanoComp/meep/blob/60076eed77f974fc08eb2277ace556031c6ea432/src/dft.cpp#L458-L459) when you are adding DFT regions for mode coefficient calculations.

(The issue is that the `add_dft_flux` regions are designed for computing ExH of Meep fields, so it defaults to interpolating to the centered grid.   Here, we don't want to do that, because the eigenmode sources do no centered-grid interpolation.   It looks like the subsequent `get_mode_flux_overlap` calculations will be correct—they will use whatever grid and weights were employed when `add_dft_flux` was called.)

smartalecH:
> I think we need to add a centered_grid flag

This seemed to have done the trick (and was much simpler than my original fix -- which I also learned works after a bugfix). As discussed before, the gradients improve as the Courant factor decreases:

image

image

Note that I already implemented the iω discrete-time transform (so it's not that). Rather, I think the issue is with the forward source normalization. The forward source is typically a gaussian and normalized such that the center frequency has unit power. However, the normalization constant is performed using a continuous approximation of the gaussian, rather than the DTFT of the truncated gaussian (remember -- all of the Fourier transformed fields in meep are calculated using the DTFT).

This complex amplitude factor is used to normalize out the Gaussian pulse in the adjoint source -- but it's wrong. It makes sense that the adjoints are more accurate with smaller timesteps (but equal resolution) because this factor more closely approximates the continuous transform.

It would be nice to analytically derive the dtft of the truncated gaussian. For now, I'll try a brute force calculation and use that in my normalization factor.

smartalecH:
Okay I think I got it. It's not the most elegant approach but it works well (Courant=0.5):
image

Next steps are to clean up the failing tests and resolve a few bugs in 3D I found during an unrelated exercise.

python/meep.i
```diff
@@ -508,15 +508,26 @@ kpoint_list get_eigenmode_coefficients_and_kpoints(meep::fields *f, meep::dft_fl
...
+    int rank = f->get_array_slice_dimensions(where, dims, dirs, collapse_empty_dimensions, snap_empty_dimensions, min_max_loc_vec_ptr, 0, cgrid);
 
     PyObject *py_dirs = PyList_New(3);
     for (Py_ssize_t i = 0; i < 3; ++i) {
         PyList_SetItem(py_dirs, i, PyInteger_FromLong(static_cast<int>(dirs[i])));
     }
 
+    if (min_max_loc){
+        PyList_Append(min_max_loc, vec2py(min_max_loc_vec[0],true));
+        PyList_Append(min_max_loc, vec2py(min_max_loc_vec[1],true));
```
stevengj, line 528: This seems like it has a memory leak — it needs to call `Py_DECREF` on the `vec2py` result after it is done with it.

(Actually, skimming over this file, it seems like a lot of the `vec2py` calls leak memory in this way.)

stevengj: (The `PyList_Append` documentation does not say that it steals the reference to the argument.  I double-checked and the [source code](https://github.com/python/cpython/blob/5241e189e77972d3a07acbbb3f0c0cbc2aeeb681/Objects/listobject.c#L313) does indeed increment the reference.  So we need to decrement the reference after appending or the vector will never be freed.)

python/meep.i
```diff
@@ -765,6 +776,7 @@ void _get_gradient(PyObject *grad, PyObject *fields_a, PyObject *fields_f, PyObj
     if (!PyArray_Check(pao_grad)) meep::abort("grad parameter must be numpy array.");
     if (!PyArray_ISCARRAY(pao_grad)) meep::abort("Numpy grad array must be C-style contiguous.");
     meep::realnum * grad_c = (meep::realnum *)PyArray_DATA(pao_grad);
+    int ng = PyArray_DIMS(pao_grad)[1];
```
stevengj, line 779: do we check that `PyArray_NDIM(pao_grad) == 1`?

python/meep.i
```diff
@@ -793,13 +805,7 @@ void _get_gradient(PyObject *grad, PyObject *fields_a, PyObject *fields_f, PyObj
     if (!PyArray_Check(pao_freqs)) meep::abort("frequencies parameter must be numpy array.");
     if (!PyArray_ISCARRAY(pao_freqs)) meep::abort("Numpy fields array must be C-style contiguous.");
     meep::realnum* frequencies_c = (meep::realnum *)PyArray_DATA(pao_freqs);
-
-    // get the proper dimensions of the fields array
-    if (PyArray_NDIM(pao_fields_a) !=5) {meep::abort("Fields array must have 5 dimensions.");}
```
stevengj, line 798: Do we still need a `PyArray_NDIM` check?

smartalecH:
Note that a grating coupler example still produces an incorrect gradient by a common scalefactor:

image

image

The scale factor decreases with resolution. I'll post more information once I have it.

python/adjoint/objective.py
```diff
@@ -82,19 +82,25 @@ def place_adjoint_source(self,dJ):
         
         # an ugly way to calcuate the scaled dtft of the forward source
         y = np.array([self.time_src.swigobj.current(t,dt) for t in np.arange(0,T,dt)]) # time domain signal
-        signal_dtft = np.matmul(np.exp(1j*2*np.pi*self.frequencies[:,np.newaxis]*np.arange(y.size)*dt), y)
-        source_amp = np.abs([self.time_src.fourier_transform(f) for f in self.frequencies]) * np.exp(2*1j*np.angle(signal_dtft)) # note the fudge factor of 2 in the phase...
+        signal_dtft = np.matmul(np.exp(1j*2*np.pi*self.frequencies[:,np.newaxis]*np.arange(y.size)*dt), y) # dtft
+        fwd_src_phase = np.exp(1.0*1j*np.angle(signal_dtft)) # phase component of dtft
+        fwd_src_amp = np.abs([self.time_src.fourier_transform(f) for f in self.frequencies]) # amplitude of source
```
stevengj, line 87: Is this a DTFT?

smartalecH: This is out of date...let me push an update

stevengj: Why not just use `np.abs(signal_dtft)` with the appropriate scale factor (1/2π or sqrt(1/2π))?

smartalecH: Yes, it's a dtft.

python/meep.i
smartalecH: No, we don't

python/meep.i
smartalecH: We dont but probably should

python/meep.i
stevengj: So the code requires `PyArray_NDIM(pao_fields_a) == 1`, then.

python/adjoint/objective.py
stevengj: What is the difference between `signal_dtft` and `self.time_src.fourier_transform(f)`, then? Why are we computing both?

smartalecH: There's an "amplitude" discrepancy when I manually compute the DTFT. For some reason, the actual amplitude of the source matches the continuous FT approximation, rather than the DTFT computation...

python/meep.i
smartalecH: Yes should be fixed now

python/meep.i
smartalecH: Should be fixed now

python/meep.i
smartalecH: Should be fixed now

python/adjoint/objective.py
```diff
@@ -76,28 +79,38 @@ def place_adjoint_source(self,dJ,dt):
             dV = 1/self.sim.resolution * 1/self.sim.resolution
         else:
             dV = 1/self.sim.resolution * 1/self.sim.resolution * 1/self.sim.resolution
-        da_dE = 0.5*(dV * self.cscale)
-        scale = da_dE * dJ * 1j * 2 * np.pi * self.frequencies / np.array([self.time_src.fourier_transform(f) for f in self.frequencies]) # final scale factor
+        
+        # an ugly way to calcuate the scaled dtft of the forward source
+        y = np.array([self.time_src.swigobj.current(t,dt) for t in np.arange(0,T,dt)]) # time domain signal
+        signal_dtft = np.matmul(np.exp(1j*2*np.pi*self.frequencies[:,np.newaxis]*np.arange(y.size)*dt), y) # dtft
```
stevengj, line 85: This is missing a factor of ∆t/sqrt(2π) — see [this line](https://github.com/NanoComp/meep/blob/5eeb01b2a27798daeda1ce0710d19f0cdbbbc261/src/dft.cpp#L181) and also the manual: https://meep.readthedocs.io/en/latest/Introduction/#transmittancereflectance-spectra

python/adjoint/objective.py
```diff
@@ -76,28 +79,39 @@ def place_adjoint_source(self,dJ,dt):
...
+            adj_src_phase = np.exp(1j*np.angle(src_center_dtft))
+            
+            amp = da_dE * dV * dJ * iomega / fwd_dtft / adj_src_phase # final scale factor
         else:
-            # TODO: In theory we should be able drive the source without normalizing out the time profile.
-            # But for some reason, there is a frequency dependent scaling discrepency. It works now for 
-            # multiple monitors and multiple sources, but we should figure out why this is.
-            src = FilteredSource(self.time_src.frequency,self.frequencies,scale,dt,self.time_src) # generate source from broadband response
+            # multi frequency simulations
+            scale = da_dE * dV * dJ * iomega # final scale factor
```
smartalecH, line 101: There's still a bug with this **broadband** scale factor (the single frequency case is fine). The amplitude is correct, but there is still a phase error. The resulting gradients are "less correlated" than they should be. The gradient accuracy actually *decreases* with resolution.

Here's a sample gradient distribution (adjoint vs finite difference) when resolution = 10
image


and when resolution = 30
image

I'm assuming I need to normalize the phase of the input Gaussian profile (i.e. divide by `fwd_dtft` like in the single frequency case) -- unfortunately, that only makes it worse.

Any thoughts?

smartalecH: Ah, the issue appears to either be with how the forward/adjoint fields are stored, or how the gradient is calculated (`meepgeom.py`). The gradient corresponding to the first frequency in the array of gradients is always correct, regardless of which frequency it is or what the resolution is. I/ve tried randomly choosing different frequencies, shuffling their order, etc. -- the first frequency is always correct and the others diverge as resolution increases.

I've tested the `FilteredSource` routine and that is consistently correct. My guess would be there is something wrong with how I'm iterating through the array of gradients in C++ before returning them back to python...

src/meepgeom.cpp
```diff
@@ -2436,61 +2450,64 @@ void material_grids_addgradient_point(meep::realnum *v, std::complex<double> *fi
...
+    n2 = dims[c * 3]; n3 = dims[c * 3 + 1]; n4 = dims[c * 3 + 2];
+    c1 = cen[c][0]; c2 = cen[c][1]; c3 = cen[c][2];
+    s1 = s[c][0]; s2 = s[c][1]; s3 = s[c][2];
+
+    for (int i1 = 0; i1 < nf; ++i1) {       // freq
+      meep::realnum *v_cur = &v[ng * i1];
+      for (int i2 = 0; i2 < n2; ++i2) {     // x
+        for (int i3 = 0; i3 < n3; ++i3) {   // y
+          for (int i4 = 0; i4 < n4; ++i4) { // z
+            int xyz_index = c_offset + (((i1 * n2 + i2) * n3 + i3) * n4 + i4);
```
stevengj, line 2502: Might be clearer to do 
```c
int xyz_index = ((((c * nf + i1) * n2 + i2) * n3 + i3) * n4 + i4);
```
here rather than computing `c_offset` separately.

stevengj: Nevermind, this won't work since each component has different-sized grids.

python/adjoint/objective.py
smartalecH: Aha!

Commenting out the patch submitted in #1233 fixes the issues.

stevengj: Per our discussion, this should be resolved by #1306 — which is now merged so you should rebase this PR onto master.

PR #1286, opened 2020-07-10
enh: enabled libctl-dir to specify installation path

zerothi:
The configure option libctl-dir now accepts installation
directory, in addition to the standard prefix/share/libctl
directory.

Same as https://github.com/NanoComp/mpb/pull/106


stevengj:
Thanks!

PR #1287, opened 2020-07-12
documentation for mirror symmetry and grid discretization

### LINKED ISSUE #1277: artifacts from grid discretization
There are unexpectedly large discretization artifacts when using a amplitude function for the `Source` object. This effect can be demonstrated using an example based on "Method 3" of [Tutorial/Custom Source/Stochastic Dipole Emission in Light Emitting Diodes](https://meep.readthedocs.io/en/latest/Python_Tutorials/Custom_Source/#stochastic-dipole-emission-in-light-emitting-diodes) involving a cosine Fourier series (m=7) source amplitude function. The radiated flux from the LED structure at the pulse center frequency shows large variations due to tiny changes in the resolution.

This effect may be due to the way the source amplitude function is interpolated onto the Yee grid.

```py
import numpy as np
import meep as mp
from meep.materials import Ag

resolution = 50

dpml = 1.0
dair = 0.9
hrod = 0.6
wrod = 0.8
dsub = 5.4
dAg = 0.4

sx = 1.5
sy = dpml+dair+hrod+dsub+dAg

cell_size = mp.Vector3(sx,sy)

pml_layers = [mp.PML(direction=mp.Y,
                     thickness=dpml,
                     side=mp.High)]

fcen = 1.0
df = 0.2
ndipole = int(sx*resolution)
run_time = 200

geometry = [mp.Block(material=mp.Medium(index=3.45),
                     center=mp.Vector3(0,0.5*sy-dpml-dair-hrod-0.5*dsub),
                     size=mp.Vector3(mp.inf,dsub,mp.inf)),
            mp.Block(material=Ag,
                     center=mp.Vector3(0,-0.5*sy+0.5*dAg),
                     size=mp.Vector3(mp.inf,dAg,mp.inf)),
            mp.Block(material=mp.Medium(index=3.45),
                     center=mp.Vector3(0,0.5*sy-dpml-dair-0.5*hrod),
                     size=mp.Vector3(wrod,hrod,mp.inf))]

def src_amp_func(n):
    def _src_amp_func(p):
        if n == 0:
            return 1/np.sqrt(sx)
        else:
            return np.sqrt(2/sx) * np.cos(2*n*np.pi*(p.x+0.5*sx)/sx)
    return _src_amp_func

                     component=mp.Ez,
                     center=mp.Vector3(0,-0.5*sy+dAg+0.5*dsub),
                     size=mp.Vector3(sx,0),
                     amp_func=src_amp_func(7))]

                    resolution=resolution,
                    k_point=mp.Vector3(),
                    boundary_layers=pml_layers,
                    geometry=geometry,
                    sources=sources)

flux_mon = sim.add_flux(fcen, 0, 1,
                        mp.FluxRegion(center=mp.Vector3(0,0.5*sy-dpml),size=mp.Vector3(sx)))

sim.run(until=run_time)

flux = mp.get_fluxes(flux_mon)
print("flux:, {}".format(flux[0]))
```

**resolution = 50**
```
flux:, 2.8759697666698216e-18
```

**resolution = 51**
```
flux:, 8.409680174807422e-05
```

**resolution = 52**
```
flux:, 3.1591784346399953e-18
```

**resolution = 53**
```
flux:, 7.981497205970976e-05
```

**resolution = 54**
```
flux:, 1.657503397760138e-18
```

**resolution = 55**
```
flux:, 0.00010086263373814502
```

bug

Additional notes: this effect does not depend on the value of m or the presence of the Ag back reflector.

oskooi:
Closes #1277.

Adds a new section "Mirror Symmetry and Grid Discretization" to [Features/Exploiting Symmetry](https://meep.readthedocs.io/en/latest/Exploiting_Symmetry/) based on the discussion and example in #1277.


doc/docs/Exploiting_Symmetry.md
```diff
@@ -74,3 +74,87 @@ One subtlety that arises in specifying the symmetry of the system is that you ha
...
+![](images/grid_break_mirror_symmetry.png)
+</center>
+
+The radiated flux of the evanescent mode varies by nearly twelve orders of magnitude when the resolution changes by just one! Also, the radiated flux is "zero" (the correct/expected result) only when the resolution is an even number. These large variations are due to the cell dimensions in the periodic $x$ direction which are displayed in the simulation output:
+
+```
+Computational cell is 1.5 x 6 x 0 with resolution 50
+flux:, 6.788984186414434e-16
+...
+Computational cell is 1.5098 x 6 x 0 with resolution 51
```
stevengj, line 142: i.e. it is changing the period, as I said.

It seems like this is something that should be fixed.

stevengj:
I think we should just file an issue — we should make sure the cell size is *exactly* what the user specified (even if we pad the resolution slightly to get an even number of pixels in the presence of mirror symmetries).  Otherwise we are introducing a first-order error.

oskooi:
Replaced by #1291.

PR #1289, opened 2020-07-14
Python doc tool updates

### LINKED ISSUE #1284: documentation-generation improvements
Some things we discussed today:

* Require that methods be listed explicitly to have their documentation inserted.  This gives us manual control over what methods appear and in what order.
* Reduce the indentation.
* Remove the redundant header lines before function signatures.

RobinD42:
Fixes #1284 

Lots of changes for the API document generation templates and tools, mostly inline with #1284.

* I changed the order of some things to improve readability and searchability of the document
  - The"prose" sections are first, with the docstrings for the mentioned methods injected in this section along with the prose discussing that group of methods. This brings it back to be more like the original document.
  - The "reference" sections are moved to the end. This includes all the documented classes and functions, with their docstrings injected.

* Tags used for positioning the injected docstrings can be: 
  - `@@ top_level_function_name @@`
  - `@@ ClassName @@`
  - `@@ ClassName.method_name @@`
  - `@@ ClassName[all-methods] @@`
  - `@@ ClassName[methods-with-docstrings] @@`

  The latter is used in the "reference" section mentioned above and pulls in the docstrings for all methods in a class, but only for those which actually have a docstring.

* I left the markdown headers for the classes, although there is a code block immediately following with the class def, otherwise they would not get put into the TOC and the outline in the sidebar. The headers in the function and method templates were removed as discussed.

* I experimented with removing all indentation, but ended up putting some of it back because IMO it greatly aids readability, and subtly indicates what is docstring and what is prose text from the document. It's easy to experiment and fine tune the indentation, if needed, just by editing the `.../css/extra.css` file and reloading the page.

### DIFF
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -1374,7 +1374,7 @@ def add_energy(self, *args):
 
 <div class="method_docstring" markdown="1">
 
-add_energy(fcen, df, nfreq, freq, EnergyRegions...)
+`add_energy(fcen, df, nfreq, freq, EnergyRegions...)`
 
 Add a bunch of `EnergyRegion`s to the current simulation (initializing the fields
 if they have not yet been initialized), telling Meep to accumulate the appropriate
@@ -1600,7 +1600,7 @@ def add_force(self, *args):
 
 <div class="method_docstring" markdown="1">
 
-add_force(fcen, df, nfreq, freq, ForceRegions...)
+`add_force(fcen, df, nfreq, freq, ForceRegions...)`
 
 Add a bunch of `ForceRegion`s to the current simulation (initializing the fields
 if they have not yet been initialized), telling Meep to accumulate the appropriate
@@ -1807,7 +1807,7 @@ def Ldos(*args):
 
 <div class="function_docstring" markdown="1">
 
-Ldos(fcen, df, nfreq, freq)
+`Ldos(fcen, df, nfreq, freq)`
 
 Create an LDOS object with either frequency bandwidth `df` centered at `fcen` and
 `nfreq` equally spaced frequency points or an array/list `freq` for arbitrarily spaced
@@ -1837,7 +1837,7 @@ def dft_ldos(*args, **kwargs):
 
 <div class="function_docstring" markdown="1">
 
-dft_ldos(fcen=None, df=None, nfreq=None, freq=None, ldos=None)
+`dft_ldos(fcen=None, df=None, nfreq=None, freq=None, ldos=None)`
 
 Compute the power spectrum of the sources (usually a single point dipole source),
 normalized to correspond to the LDOS, in either a frequency bandwidth `df` centered at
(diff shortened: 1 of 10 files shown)


doc/docs/Python_User_Interface.md
```diff
@@ -1158,7337 +1335,6462 @@ If `match_frequency=False` or `kpoint` is not zero in the given `direction`, the
...
-                     resolution,
-                     where=None,
-                     center=None,
-                     size=None):
-```
+Finally, another run function, useful for computing ω(**k**) band diagrams, is available via these top-level functions:
 
-<div class="method_docstring" markdown="1">
+@@ meep.run_k_points @@
+@@ meep.run_k_point @@
```
oskooi, line 2583: The positioning tags for `meep_run_k_points` and `meep_run_k_point` are not being rendered correctly.

doc/docs/Python_User_Interface.md
```diff
@@ -1158,7337 +1335,6462 @@ If `match_frequency=False` or `kpoint` is not zero in the given `direction`, the
...
+def add_energy(self, *args):
 ```
 
 <div class="method_docstring" markdown="1">
 
-Given a frequency `frequency` and a `Vector3` `pt`, returns the average eigenvalue
-of the permittivity tensor at that location and frequency. If `frequency` is
-non-zero, the result is complex valued; otherwise it is the real,
-frequency-independent part of ε (the $\omega    o\infty$ limit).
+add_energy(fcen, df, nfreq, freq, EnergyRegions...)
```
oskooi, line 1377: Perhaps `add_energy(fcen, df, nfreq, freq, EnergyRegions...)` should appear in code blocks rather an prose?

doc/docs/Python_User_Interface.md
```diff
@@ -1158,7337 +1335,6462 @@ If `match_frequency=False` or `kpoint` is not zero in the given `direction`, the
...
+<div class="class_members" markdown="1">
 
 ```python
-def get_hfield_y(self):
+def add_force(self, *args):
 ```
 
 <div class="method_docstring" markdown="1">
 
+add_force(fcen, df, nfreq, freq, ForceRegions...)
```
oskooi, line 1603: Same issue here: `add_force(fcen, df, nfreq, freq, ForceRegions...)` should be a code block rather than part of the prose description?

doc/docs/Python_User_Interface.md
```diff
@@ -1158,7337 +1335,6462 @@ If `match_frequency=False` or `kpoint` is not zero in the given `direction`, the
...
 
 ```python
-def get_sfield_y(self):
+def Ldos(*args):
 ```
 
-<div class="method_docstring" markdown="1">
+<div class="function_docstring" markdown="1">
 
+Ldos(fcen, df, nfreq, freq)
```
oskooi, line 1810: Again: `Ldos(fcen, df, nfreq, freq)` which lists the arguments should be a code block?

doc/docs/Python_User_Interface.md
RobinD42: Oops, those are actually `Simulation` methods. Fixed.

doc/docs/Python_User_Interface.md
RobinD42: The best fix for things like this would be to change the code so it specifies the parameters instead of using`*args,**kwargs` but that's probably a task for another day. In any case, that text is part of the docstring so it can just be edited there.

I tried with a triple-backticks to get a full code block but that doesn't look very good next to the one already there, so I'll just wrap them with the single backticks for now.

Another possibility would be to create some way that the tool can recognise a different args string specified within the docstring, and put that in the document instead of, or maybe in addition to, the `def ...` block that is generated automatically. That would probably be best done in a separate PR though.

PR #1290, opened 2020-07-16
extend LED tutorial involving cosine series source terms to 2d

oskooi:
Extends the presentation of Method 3 of [Tutorial/Custom Sources/Stochastic Dipole Emission in Light Emitting Diodes](https://meep.readthedocs.io/en/latest/Python_Tutorials/Custom_Source/#stochastic-dipole-emission-in-light-emitting-diodes) to 2d/3d following @stevengj's [comment](https://github.com/NanoComp/meep/pull/1266#issuecomment-652139470) in #1266.


stevengj:
This is a symmetry-breaking effect, but I think you are breaking the Bloch-periodicity of the source somehow and not the mirror symmetry.

stevengj:
(If you want to show more data on convergence rates, just decrease the wavelength so that it converges more slowly.)

PR #1297, opened 2020-07-23
minor docs update

oskooi:


stevengj:
In the future, when you have something like `|m|>20`, do `$|m|>20` rather than `|$m$|>20` — the whole expression should be rendered by LaTeX, not just the variable name.

PR #1300, opened 2020-07-23
Near2far adjoint sensitivity

mochen4:
Adjoint method for near2far


python/adjoint/objective.py
```diff
@@ -132,4 +134,188 @@ def get_evaluation(self):
...
+        elif ary.ndim == 2:
+            result = ary[:, :, newaxis]
+        else:
+            result = ary
+        res.append(result)
+    if len(res) == 1:
+        return res[0]
+    else:
+        return res
+
```
smartalecH, line 253: Rather than redefine, why not import them from optimization_problem.py?

stevengj:
Closed by #1329.

PR #1302, opened 2020-07-27
Add verbosity flag for MPB

RobinD42:
See https://github.com/NanoComp/mpb/issues/116 and https://github.com/NanoComp/mpb/pull/119

* Adds a verbosity flag for the mpb code. 
* All prints, whether in C++ or Python are now conditional with a verbosity check
* Adds a Python verbosity manager class to make setting and checking the verbosity from Python more pythonic, it is used for both Meep and MPB.

Currently all verbosity checks are checking for `>=2` so more fine-tuning will be needed to adjust the levels tested for the specific messages.


RobinD42:
In addition...

* There are a few print statements in libctl that are still showing up, so similar verbosity checking will probably be needed there.

* Does it make sense to have to set multiple verbosity flags from Python? It would not be hard to adjust the `Verbosity` class to be able to manage more than one of the C/C++ verbosity flags, and then set them all to the same value with one statement. The `Verbosity` object could then be adjusted to be used as a Singleton.

RobinD42:
I'm looking into the test failure now.

libpympb/pympb.hpp
```diff
@@ -9,6 +9,16 @@
...
 namespace py_mpb {
 
+// A global "verbosity" level.
+//    0: minimal output
+//    1: a little
+//    2: a lot (default)
+//    3: debugging
+//
+// This is redeclared here so SWIG will see it.
+extern "C" int mpb_verbosity;
```
stevengj, line 19: Now that https://github.com/NanoComp/mpb/commit/3108059ef78b57ba466fb3dce240c26c6844045e is merged, you can protect this with
```c
#if MPB_VERSION_MAJOR > 1 || (MPB_VERSION_MAJOR == 1 && MPB_VERSION_MINOR >= 11)
```

stevengj:
When running interactively from Python, e.g. in Jupyter notebooks, we probably want to set verbosity to 1 by default.

libpympb/pympb.hpp
RobinD42: Since SWIG currently doesn't see the values of those macros I made a corresponding change in `pympb.cpp` instead. Although it's possible to make those available to SWIG, this seemed to be the simplest change to make things work with old and new libmpb.

PR #1303, opened 2020-07-29
complex source point Gaussian beam

### LINKED ISSUE #711: complex point source gaussian beam
It might be nice to have a built-in method to create a Gaussian-beam source.  Not only is this a common request, but to create a really high-quality Gaussian beam is a bit tricky.

One nice method to define the appropriate current source is to apply the equivalence principle to the [complex point source](https://www.osapublishing.org/josaa/abstract.cfm?uri=josaa-3-12-2005) method.  SCUFF-EM already has an [implementation of this](https://github.com/HomerReid/scuff-em/blob/master/libs/libIncField/GaussianBeam.cc) that we could probably adapt.

oskooi:
Fixes #711.

This PR adds a new function `gaussianbeam_3d` to the `fields` class which computes the fields for a Gaussian beam using the complex source point method. The new function is based on @HomerReid's implementation from SCUFF-EM in [`GaussianBeam::GetFields`](https://github.com/HomerReid/scuff-em/blob/master/libs/libIncField/GaussianBeam.cc#L78-L218). The main difference is that the user-specified field-strength vector `E0` is real (because it is based on Meep's `vec` object which only supports [real-valued elements](https://github.com/NanoComp/meep/blob/master/src/meep/vec.hpp#L600)) whereas Homer supported the more general case of complex numbers. We could support a complex field-strength vector but this does not seem to be necessary.

The next thing to do is to add the function `add_gaussianbeam_source` which calls `gaussianbeam_3d` for every point in a given user-specified region `where` and uses these fields to compute the equivalent currents. 

A couple of issues regarding the API for `add_gaussianbeam_source`:
- for a 3d simulation involving a 2d source object, should the center of the Gaussian beam be an input parameter (with arbitrary values) or just be set to the geometric center of the user-specified `where` object? Under what scenario would the beam center *not* lie in the 2d plane?
- is there a requirement that the source object be positioned within a lossless medium or should the imaginary part of ε be ignored along with a warning message?
- should we support a 3d source object?
- how can quantify how much more accurate generating a Gaussian beam is using this complex source point method versus just an amplitude function? this could form the basis for a test that we can add to the `make check` suite.

### DIFF
--- a/src/meep.hpp
+++ b/src/meep.hpp
@@ -1490,6 +1490,7 @@ public:
   gaussianbeam(const vec &x0, const vec &kdir, double w0, double freq,
                double eps, double mu, std::complex<double> EO[3]);
   void get_fields(std::complex<double> *EH, const vec &x);
+  std::complex<double> get_E0(int n) { return E0[n]; };
 
 private:
   vec x0;           // beam center
@@ -2009,6 +2010,10 @@ private:
   void step_source(field_type ft, bool including_integrated = false);
   void update_pols(field_type ft);
   void calc_sources(double tim);
+  // mpb.cpp
+  void add_volume_source_check(component c, const src_time &src, const volume &where,
+                               std::complex<double> A(const vec &), std::complex<double> amp,
+                               component c0, direction d, int has_tm, int has_te);
 
 public:
   // monitor.cpp
--- a/src/meep.hpp
+++ b/src/meep.hpp
@@ -1697,7 +1697,7 @@ public:
                          std::complex<double> A(const vec &), std::complex<double> amp = 1.0);
   void add_volume_source(component c, const src_time &src, const volume &,
                          std::complex<double> amp = 1.0);
-  void add_volume_source(const src_time &src, direction d, const volume &, gaussianbeam beam);
+  void add_volume_source(const src_time &src, const volume &, gaussianbeam beam);
   void require_component(component c);
 
   // mpb.cpp
--- a/src/meep.hpp
+++ b/src/meep.hpp
@@ -1484,6 +1484,22 @@ public:
   std::vector<std::complex<double> > values; // updated by update_values(idx)
 };
 
+class gaussianbeam {
+
+public:
+  gaussianbeam(const vec &x0, const vec &kdir, double w0, double freq,
+               double eps, double mu, std::complex<double> EO[3]);
+  void get_fields(std::complex<double> *EH, const vec &x);
+
+private:
+  vec x0;           // beam center
+  vec kdir;         // beam propagation direction
+  double w0;        // beam waist radius
+  double freq;      // beam frequency
+  double eps, mu;   // permittivity/permeability of homogeneous medium
+  std::complex<double> E0[3]; // electric field-strength vector
+};
+
 /***************************************************************/
 /* prototype for optional user-supplied function to provide an */
 /* initial estimate of the wavevector of mode #mode at         */
@@ -1681,10 +1697,8 @@ public:
                          std::complex<double> A(const vec &), std::complex<double> amp = 1.0);
   void add_volume_source(component c, const src_time &src, const volume &,
                          std::complex<double> amp = 1.0);
+  void add_volume_source(const src_time &src, direction d, const volume &, gaussianbeam beam);
   void require_component(component c);
-  void gaussianbeam_3d(std::complex<double> *EH, const vec &x0, const vec &x,
-                       const vec &kdir, double w0, double freq, double eps,
-                       double mu, const vec &E0);
 
   // mpb.cpp
 
--- a/src/meep.hpp
+++ b/src/meep.hpp
@@ -2010,7 +2010,7 @@ private:
   void step_source(field_type ft, bool including_integrated = false);
   void update_pols(field_type ft);
   void calc_sources(double tim);
-  // mpb.cpp
+  // sources.cpp
   void add_volume_source_check(component c, const src_time &src, const volume &where,
                                std::complex<double> A(const vec &), std::complex<double> amp,
                                component c0, direction d, int has_tm, int has_te);
--- a/src/meep.hpp
+++ b/src/meep.hpp
@@ -1489,8 +1489,8 @@ class gaussianbeam {
 public:
   gaussianbeam(const vec &x0, const vec &kdir, double w0, double freq,
                double eps, double mu, std::complex<double> EO[3]);
-  void get_fields(std::complex<double> *EH, const vec &x);
-  std::complex<double> get_E0(int n) { return E0[n]; };
+  void get_fields(std::complex<double> *EH, const vec &x) const;
+  std::complex<double> get_E0(int n) const { return E0[n]; };
 
 private:
   vec x0;           // beam center
@@ -1498,7 +1498,7 @@ private:
   double w0;        // beam waist radius
   double freq;      // beam frequency
   double eps, mu;   // permittivity/permeability of homogeneous medium
-  std::complex<double> E0[3]; // electric field-strength vector
+  std::complex<double> E0[3]; // polarization vector
 };
 
 /***************************************************************/
--- a/src/meep.hpp
+++ b/src/meep.hpp
@@ -1682,6 +1682,9 @@ public:
   void add_volume_source(component c, const src_time &src, const volume &,
                          std::complex<double> amp = 1.0);
   void require_component(component c);
+  void gaussianbeam_3d(std::complex<double> *EH, const vec &x0, const vec &x,
+                       const vec &kdir, double w0, double freq, double eps,
+                       double mu, const vec &E0);
 
   // mpb.cpp
 
--- a/src/sources.cpp
+++ b/src/sources.cpp
@@ -436,6 +436,24 @@ void fields::add_volume_source(component c, const src_time &src, const volume &w
   require_component(c);
 }
 
+/***************************************************************/
+/* helper routine for add_eigenmode_source that calls          */
+/* add_volume_source only if certain conditions are met        */
+/***************************************************************/
+void fields::add_volume_source_check(component c, const src_time &src, const volume &where,
+                                     complex<double> A(const vec &), complex<double> amp, component c0,
+                                     direction d, int has_tm, int has_te) {
+  if (!gv.has_field(c)) return;
+  if (c0 != Centered && c0 != c) return;
+  if (component_direction(c) == d) return;
+  if (gv.dim == D2) // parity checks
+  {
+    if (has_te && is_tm(c)) return;
+    if (has_tm && !is_tm(c)) return;
+  };
+  add_volume_source(c, src, where, A, amp);
+}
+
 static gaussianbeam *global_gaussianbeam_object = 0;
 static component global_gaussianbeam_component;
 
--- a/src/sources.cpp
+++ b/src/sources.cpp
@@ -453,6 +453,14 @@ static std::complex<double> gaussianbeam_ampfunc(const vec &p) {
   }
 }
 
+static int gaussianbeam_has_tm() {
+  return abs(global_gaussianbeam_object->get_E0(2)) > 0;
+}
+
+static int gaussianbeam_has_te() {
+  return abs(global_gaussianbeam_object->get_E0(0)) > 0 || abs(global_gaussianbeam_object->get_E0(1)) > 0;
+}
+
 void fields::add_volume_source(const src_time &src, const volume &where, gaussianbeam beam) {
   global_gaussianbeam_object = &beam;
   direction d = normal_direction(where);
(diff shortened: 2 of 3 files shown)


src/sources.cpp
```diff
@@ -436,4 +436,102 @@ void fields::add_volume_source(component c, const src_time &src, const volume &w
   require_component(c);
 }
 
+/* Given the field-strength vector E0 for a Gaussian beam centered at x0 with beam waist w0 and propagation direction kdir,
+   compute the E/H fields EH[6] (6 components) at x for a frequency freq in the homogeneous 3d medium eps and mu.
+
+   Adapted from code by M. T. Homer Reid in his SCUFF-EM package
+   (file scuff-em/src/libs/libIncField/GaussianBeam.cc), which is GPL v2+. */
+
+void fields::gaussianbeam_3d(std::complex<double> *EH, const vec &x0, const vec &x,
```
stevengj, line 445: Probably this doesn't need to be a method of the `fields` class since it doesn't use the fields.

src/meep.hpp
```diff
@@ -1682,6 +1682,9 @@ class fields {
   void add_volume_source(component c, const src_time &src, const volume &,
                          std::complex<double> amp = 1.0);
   void require_component(component c);
+  void gaussianbeam_3d(std::complex<double> *EH, const vec &x0, const vec &x,
+                       const vec &kdir, double w0, double freq, double eps,
+                       double mu, const vec &E0);
```
stevengj, line 1687: I would use `std::complex<double> E0[3]`

stevengj:
I would define two things:

1. a `gaussianbeam` class that encapsulates the parameters (freq, E0, etc) of the beam, and a method to compute the fields `E` at a point `x`.

2. a new method `fields::add_volume_source(const src_time &src, const volume &v, gaussianbeam beam);` that  then calls `add_volume_source` several times on the same volume `v` with different transverse components `c` and an amplitude function `A` that calls the `beam` to compute the corresponding field component of the beam.

stevengj:
Note that the only purpose of this is to create an equivalent current source, which must be a *surface* (a line in 2d or a plane in 3d).

At the C++ level, the user will have to specify `kdir, eps, mu` to the `gaussianbeam` constructor.  (In the higher-level Python API, we might infer defaults for these from the source volume.)

stevengj:
Note that because our source volume is finite, the user will need to ensure that the gaussian beam (evaluated in the source volume) is significantly narrower than the source volume (so that the currents/fields decay to a negligible value at the edge of the source).   

(We could even print out a warning if this is not the case, e.g. if the currents at the edge of the source volume are > 1% of the peak current.)

oskooi:
A new `gaussianbeam` class and `fields::add_volume_source(const src_time &src, direction d, const volume &v, gaussianbeam beam)` function have been added. While everything compiles, it has *not* been tested yet. Perhaps that can wait until we add the Python API since it will simplify things a bit (i.e., computing the normal to the source volume object for the `direction d` argument, etc).

oskooi:
The following C++ example tries to launch a Gaussian beam for three test cases involving different values for the beam waist radius and propagation direction. As shown in the accompanying figures, the beam profile is not correct although it does appear that the mode is unidirectional.

```cpp
#include <math.h>
#include <stdio.h>
#include <stdlib.h>
#include <meep.hpp>

using namespace meep;

double one(const vec &) { return 1.0; }

  meep::initialize mpi(argc, argv);
  int sxy = 10;
  int resolution = 50;
  int dpml = 1;

  grid_volume gv = voltwo(sxy, sxy, resolution);
  gv.center_origin();
  structure the_structure(gv, one, pml(dpml));

  fields f(&the_structure);

  double fcen = 1.0;
  continuous_src_time src(fcen);
  src.is_integrated = true;

  vec x0(0,-0.5*sxy+dpml);         // beam center                                                                                                              
  double theta = -35 * pi/180;
  vec kdir(sin(theta),cos(theta)); // beam propagation direction                                                                                               
  double w0 = 1.0;                 // beam waist radius                                                                                                        
  std::complex<double> EO[3] = {1.0,1.0,1.0}; // electric field-strength vector                                                                                
  gaussianbeam beam(x0, kdir, w0, fcen, 1.0, 1.0, EO);

  volume vol(vec(-0.5*sxy,-0.5*sxy+dpml), vec(0.5*sxy,-0.5*sxy+dpml));

  f.add_volume_source(src, vol, beam);

  while (f.time() < 50.0)
    f.step();

  f.output_hdf5(Ez, gv.surroundings());

  return 0;
}
```
**1. w0 = 0.5, theta = 0**

beam_w0 5_theta0

**2. w0 = 1.0, theta = 25 * pi/180**
beam_w1 0_theta20

**3. w0 = 1.0, theta = -35 * pi/180**

beam_w1 0_theta35m

stevengj:
To use `add_volume_source_check`, you will need to add it as a private method of `fields` so that you can call it from other methods.

You will need to change the `parity` argument to two arguments, say `int has_tm` and `int has_te`.   Then when you call it from mpb.cpp, pass `parity & EVEN_Z_PARITY` and `parity & ODD_Z_PARITY` for those two parameters; in contrast, when you call it from the gaussian-beam code, pass `E0[2] != 0` for the `has_tm` parameter and `E0[0] != 0 || E0[1] != 0` for the `has_te` parameter.

oskooi:
Working now: unidirectional Gaussian beam. Tested for both polarizations in 2d.

gaussian_beam_profiles

oskooi:
An example of a Gaussian beam source (green dotted line) and focus (black dot) which are at different locations. (The previous examples had the source and focus at the same location.)

gaussian_beam_focus

smartalecH:
Have you checked 3D yet?

oskooi:
Yes, this is working in 3d as well. 

(I also tried removing the field scaling that Homer had included in his `GaussianBeam::GetFields` to account for potential numerical stability issues related to large complex values for the argument to the sin function and found that it did not affect the results. We may therefore want to remove this from our implementation.)

smartalecH:
Are you planning to add swig/python support in this PR too? Seems pretty straightforward given all the current machinery.

oskooi:
I was planning to create a separate PR for the Python API which would also include a test, documentation, and tutorial example.

oskooi:
Results for 3d Gaussian beams at two different angles: 0° and 30° (rotation about *x* axis, 0° along +*z* axis).

beam3d_theta0

beam3d_theta30

```cpp
#include <math.h>
#include <stdio.h>
#include <stdlib.h>
#include <meep.hpp>

using namespace meep;

double one(const vec &) { return 1.0; }

  meep::initialize mpi(argc, argv);
  int s = 14;
  int resolution = 30;
  int dpml = 2;

  grid_volume gv = vol3d(s, s, s, resolution);
  gv.center_origin();
  structure the_structure(gv, one, pml(dpml));

  fields f(&the_structure);

  double fcen = 1.0;
  continuous_src_time src(fcen);

  vec x0(0,0,0);             // beam center
  double theta = 30 * pi/180; // rotation angle around +x axis (0: +z axis)
  vec kdir(0,sin(theta),cos(theta)); // beam propagation direction
  double w0 = 1.0;                   // beam waist radius
  std::complex<double> EO[3] = {1.0,0,0}; // polarization vector
  gaussianbeam beam(x0, kdir, w0, fcen, 1.0, 1.0, EO);

  volume vol(vec(-0.5*s,-0.5*s,0), vec(0.5*s,0.5*s,0));

  f.add_volume_source(src, vol, beam);

  while (f.time() < 10.0)
    f.step();

  volume output_vol(vec(-0.5*s+dpml,-0.5*s+dpml,-0.5*s+dpml),
                    vec(0.5*s-dpml,0.5*s-dpml,0.5*s-dpml));
  f.output_hdf5(Ex, output_vol);

  return 0;
}
```

```
$ h5topng -vZc bluered -o beam3d-theta30-xz.png -d ex.r -y 151 ex-000010.00.h5
$ h5topng -vZc bluered -o beam3d-theta30-xy.png -d ex.r -z 151 ex-000010.00.h5
$ h5topng -vZc bluered -o beam3d-theta30-xy.png -d ex.r -z 151 ex-000010.00.h5
```

src/sources.cpp
```diff
@@ -470,14 +478,18 @@ void fields::add_volume_source(const src_time &src, const volume &where, gaussia
   int np2 = (n + 2) % 3;
   // Kx = -Hy, Ky = Hx   (for d==Z)
   global_gaussianbeam_component = cH[np1];
-  add_volume_source(cE[np2], src, where, gaussianbeam_ampfunc, +1.0);
+  add_volume_source_check(cE[np2], src, where, gaussianbeam_ampfunc, +1.0, cE[np2], d,
+                          gaussianbeam_has_tm(), gaussianbeam_has_te());
```
stevengj, line 500: Rather than define functions for these, maybe just define variables?
```cxx
bool has_te = abs(beam.get_E0(2)) > 0;
bool has_tm = abs(beam.get_E0(0)) > 0 || abs(beam.get_E0(1)) > 0;
```
and then just pass `has_te, has_tm`?

src/meep.hpp
```diff
@@ -1490,6 +1490,7 @@ class gaussianbeam {
   gaussianbeam(const vec &x0, const vec &kdir, double w0, double freq,
                double eps, double mu, std::complex<double> EO[3]);
   void get_fields(std::complex<double> *EH, const vec &x);
+  std::complex<double> get_E0(int n) { return E0[n]; };
```
stevengj, line 1493: Both `get_fields` and `get_E0` should be `const` methods since they don't modify the object:
```
void get_fields(std::complex<double> *EH, const vec &x) const;
std::complex<double> get_E0(int n) const { return E0[n]; };
```

src/sources.cpp
```diff
@@ -436,4 +436,211 @@ void fields::add_volume_source(component c, const src_time &src, const volume &w
...
+  mu = mu_;
+  for (int j=0; j<3; ++j)
+    E0[j] = E0_[j];
+}
+
+/* Compute the E/H fields EH[6] (6 components) at point x in 3d.
+
+   Adapted from code by M. T. Homer Reid in his SCUFF-EM package
+   (file scuff-em/src/libs/libIncField/GaussianBeam.cc), which is GPL v2+. */
+
```
stevengj, line 531: You should copy and paste the copyright notice from that source file (https://github.com/HomerReid/scuff-em/blob/9c6d0cb7695463af803dee8d04cdae939740cdcc/libs/libIncField/GaussianBeam.cc#L1-L18) verbatim here.

stevengj: also copy and paste the description with the reference (https://github.com/HomerReid/scuff-em/blob/9c6d0cb7695463af803dee8d04cdae939740cdcc/libs/libIncField/GaussianBeam.cc#L20-L43)

and link the URL https://github.com/HomerReid/scuff-em/blob/master/libs/libIncField/GaussianBeam.cc

stevengj: We should also copy and paste the full copyright notice into https://github.com/NanoComp/meep/blob/master/src/near2far.cpp above the `green3d` function, for that matter.

oskooi: Fixed.

src/meep.hpp
oskooi: Fixed.

src/sources.cpp
oskooi: Fixed.

src/sources.cpp
```diff
@@ -520,20 +514,65 @@ gaussianbeam::gaussianbeam(const vec &x0_, const vec &kdir_, double w0_, double
...
+ */
+
+void gaussianbeam::get_fields(std::complex<double> *EH, const vec &x) const {
   double n = sqrt(eps * mu);
   double k = 2 * pi * freq * n;
-  double impedance = sqrt(mu/eps);
-  double z0 = k * w0 * w0 / 2; // Rayleigh range
+  double ZR = sqrt(mu/eps);
+  double z0 = k * w0 * w0 / 2;
   double kz0 = k*z0;
```
stevengj, line 576: Maybe all of these parameters should be computed once in the `gaussianbeam` constructor, rather than each time `gaussianbeam::get_fields` is called?

It probably makes little difference in performance, however, so we might as well leave it as-is for simplicity.

stevengj:
Would be nice to add a 2d test.    Maybe creating a focus like your example above (https://github.com/NanoComp/meep/pull/1303#issuecomment-667585563) but at a nonzero angle.  Since the code seems to be working now, the test can just check to make sure that the intensity at the focal spot doesn't change.

stevengj:
This is not quite right for 2d, because we are just taking a slice of a circular-profile Gaussian beam.

In principle, we need *two* width parameters (and axes thereof), so that we can set the z width to infinity for 2d calculations (and more generally this would allow an elliptical focus).

stevengj:
In Python, we'll want a `GaussianBeam` object that takes the same parameters as `gaussianbeam`.   Then you would pass this for the `amp_func` parameter of the source object.   Then, when Python calls the `add_volume_source` method, it will check if `amp_func` is a `GaussianBeam`, and if so it will instantiate a `meep::gaussianbeam` object with the same parameters and pass it.

`E0` can be a `mp.Vector3` (which can be complex).  When creating the `meep::gaussianbeam` object, you will need to convert it to a list `[E0.x, E0.y, E0.z]` to pass to C++, and you will need something like the following typemap:
```cxx
%typecheck(SWIG_TYPECHECK_POINTER, fragment="NumPy_Fragments") std::complex<double> E0[3] {
    $1 = is_array($input);
}

%typemap(in) std::complex<double> E0[3] {
    std::complex<double> *pyE0  = (std::complex<double> *)array_data($input);
    $1[0] = pyE0[0]; $1[1] = pyE0[1]; $1[2] = pyE0[2];
}
```

stevengj:
At some point in the future, we could include a function to compute the beam power, and maybe normalize the beam to unit power.  Homer provided a routine for this.  However, that function won't work in 2d (#1308), where we'd either have to re-derive it or do a numerical (2d) quadrature to get the power.  It's probably not worth the trouble, since it will be more reliable to do what you normally do in Meep — perform a separate normalization run.

oskooi:
> Then, when Python calls the add_volume_source method, it will check if amp_func is a GaussianBeam, and if so it will instantiate a meep::gaussianbeam object with the same parameters and pass it.

How would this be set up in `python/meep.i`?

oskooi:
An alternative to passing a `GaussianBeam` object as the `amp_func` parameter of the `Source` object (which is somewhat awkward) is to specify the source using a new `GaussianBeamSource` class object (a subclass of `Source`) as e.g.:
```py
source = [mp.GaussianBeamSource(src=mp.GaussianSource(fcen,df),
                                center=mp.Vector3(),
                                size=mp.Vector3(sx,sy,0),
                                x0=mp.Vector3(bx,by,bz),
                                kdir=mp.Vector3(kx,ky,kz),
                                w0=w0,
                                E0=mp.Vector3(cx,cy,cz))]
```
The remaining parameters `eps`, `mu`, and `freq` of the `GaussianBeam` object can be automatically inferred from `center` and `src`.

This approach was described in #711:[comment](https://github.com/NanoComp/meep/issues/711#issuecomment-463242022) and is more consistent with the API for [`EigenModeSource`](https://meep.readthedocs.io/en/latest/Python_User_Interface/#eigenmodesource).

stevengj:
That's fine.

PR #1315, opened 2020-08-13
Add warnings for non-whole number usage with grid_volume

### LINKED ISSUE #1291: ensuring that user-specified cell dimensions are equivalent to the actual cell dimensions
As originally reported in #1277 and #1287, simulations involving Bloch-periodic boundary conditions can undergo symmetry breaking due to small changes in the cell dimensions and thereby the Bloch periodicity as a result of discretization errors introduced by the choice of the resolution. The effect of this symmetry breaking is to introduce first-order errors.

This effect can be demonstrated using a 1d line source with a sinusoidal amplitude function in a 2d cell with periodic boundaries. The line source spans the entire length *L* of the cell in the *x* direction and is defined using a cosine profile: cos(k<sub>x</sub>x) where k<sub>x</sub> = 2πm/L and m is an integer. The wavevector k<sub>y</sub> of the propagating modes can be computed analytically: for a frequency of ω (in c=1 units), the propagating modes are the **real** solutions of sqrt(ω<sup>2</sup>n<sup>2</sup>-k<sub>x</sub><sup>2</sup>) where n is the refractive index of the propagating medium. (This is analogous to the [diffractions orders of a binary grating](Python_Tutorials/Mode_Decomposition.md#transmittance-spectra-for-planewave-at-normal-incidence).) PMLs surround the cell in the *y* direction in order to absorb the outgoing flux. In this example, L=1.5, ω=2π, and n=3.5 for a total of |m|=5 propagating modes. This can be verified by computing the total flux radiated by the source in the *+y* direction for the evanescent m=6 mode and checking that it is "zero" (i.e., close to [machine precision](https://en.wikipedia.org/wiki/Machine_epsilon)).

```py
￼import numpy as np
￼import meep as mp
￼
￼resolution = 50 # pixels/um
￼
￼sx = 1.5
￼sy = 6.0
￼cell_size = mp.Vector3(sx,sy)
￼
￼dpml = 1.0
￼pml_layers = [mp.PML(direction=mp.Y,
￼                     thickness=dpml)]
￼
￼fcen = 1.0
￼
￼def src_amp_func(m):
￼    def _src_amp_func(p):
￼        return np.cos(2*np.pi*m*(p.x+0.5*sx)/sx)
￼    return _src_amp_func
￼
￼sources = [mp.Source(mp.GaussianSource(fcen,fwidth=0.2*fcen),
￼                     component=mp.Ez,
￼                     center=mp.Vector3(0,-0.5*sy+dpml),
￼                     size=mp.Vector3(sx,0),
￼                     amp_func=src_amp_func(6))]
￼
￼sim = mp.Simulation(cell_size=cell_size,
￼                    resolution=resolution,
￼                    k_point=mp.Vector3(),
￼                    boundary_layers=pml_layers,
￼                    sources=sources,
￼                    default_material=mp.Medium(index=3.5))
￼
￼flux_mon = sim.add_flux(fcen, 0, 1,
￼                        mp.FluxRegion(center=mp.Vector3(0,0.5*sy-dpml),size=mp.Vector3(sx)))
￼
￼sim.run(until_after_sources=20)
￼
￼flux = mp.get_fluxes(flux_mon)
￼print("flux:, {}".format(flux[0]))
```

Plotting the radiated flux as a function of the resolution (using a semi-logarithmic scale) reveals an unexpected finding.

bug2

The radiated flux of the evanescent mode varies by nearly twelve orders of magnitude when the resolution changes by just one! Also, the radiated flux is "zero" (the correct/expected result) only when the resolution is an even number. These large variations are due to the cell dimensions in the periodic *x* direction which are displayed in the simulation output:

```
￼Computational cell is 1.5 x 6 x 0 with resolution 50
￼flux:, 6.788984186414434e-16
￼...
￼Computational cell is 1.5098 x 6 x 0 with resolution 51
flux:, 0.0004094502264147733
￼...
￼Computational cell is 1.5 x 6 x 0 with resolution 52
￼flux:, 5.484924186312514e-16
￼...
￼Computational cell is 1.50943 x 6 x 0 with resolution 53
￼flux:, 0.00035958138720120446
￼...
￼Computational cell is 1.5 x 6 x 0 with resolution 54
￼flux:, 4.543947667089586e-16
￼...
￼Computational cell is 1.50909 x 6 x 0 with resolution 55
￼flux:, 0.00033641577621399885
```
In this particular example, an even grid resolution results in a cell size in the *x* direction of 1.5 but something slightly larger whenever the resolution is odd. When the cell size in *x* is exactly 1.5 (even resolutions), the source does not radiate and the flux is therefore ~10<sup>-16</sup> (i.e., [round-off errors](https://en.wikipedia.org/wiki/Round-off_error)). However, when the cell size in *x* is not exactly 1.5 (odd resolutions), the source *does* radiate a small but non-trivial amount (~10<sup>-4</sup>) which is only due to the different periodicity. (The same symmetry-breaking errors affect the propagating |m| <= 5 modes though they are less pronounced.)

Fixing this bug requires making sure that the user-specified cell dimensions are equivalent to the actual cell dimensions by slightly padding the resolution as necessary to obtain an even number of pixels.

RobinD42:
Fixes #1291 

I started with using Python Warnings as an experiment, which can be seen in 42be907 if interested. This has the advantage that warnings can be suppressed or promoted to exceptions, if the user is interested in doing so. The downside is that the actual warning for any one location is displayed only once by default, which is not very useful in an interactive environment like a notebook.

For the final implementation the warning is checked for and printed in the C++ code in the various `grid_volume` factory functions as discussed. I also added a `master_printf_stderr` function that forwards the text to `sys.stderr`, which has the nice effect of printing the text with a background color in the notebook.

Snap007

Also included in this PR is the addition of `__repr__` and related methods for `vec`, `ivec`, `volume` and `grid_volume`.

### DIFF
--- a/python/meep.i
+++ b/python/meep.i
@@ -199,24 +199,22 @@ static meep::vec py_kpoint_func_wrap(double freq, int mode, void *user_data) {
     return result;
 }
 
-void py_master_printf_wrap(const char *s) {
-    PySys_WriteStdout("%s", s);
-    static PyObject *py_stdout = NULL;
-    if (py_stdout == NULL) {
-      py_stdout = PySys_GetObject("stdout");
-    }
-    PyObject *result = PyObject_CallMethod(py_stdout, "flush", NULL);
+static void _do_master_printf(const char* stream_name, const char* text) {
+    PyObject* py_stream = PySys_GetObject(stream_name);
+    PyObject* result;
+
+    result = PyObject_CallMethod(py_stream, "write", "(s)", text);
+    Py_XDECREF(result);
+    result = PyObject_CallMethod(py_stream, "flush", NULL);
     Py_XDECREF(result);
 }
 
+void py_master_printf_wrap(const char *s) {
+    _do_master_printf("stdout", s);
+}
+
 void py_master_printf_stderr_wrap(const char *s) {
-    PySys_WriteStderr("%s", s);
-    static PyObject *py_stderr = NULL;
-    if (py_stderr == NULL) {
-      py_stderr = PySys_GetObject("stderr");
-    }
-    PyObject *result = PyObject_CallMethod(py_stderr, "flush", NULL);
-    Py_XDECREF(result);
+    _do_master_printf("stderr", s);
 }
 
 void set_ctl_printf_callback(void (*callback)(const char *s)) {
--- a/python/meep.i
+++ b/python/meep.i
@@ -200,7 +200,7 @@ static meep::vec py_kpoint_func_wrap(double freq, int mode, void *user_data) {
 }
 
 static void _do_master_printf(const char* stream_name, const char* text) {
-    PyObject* py_stream = PySys_GetObject(stream_name);
+    PyObject* py_stream = PySys_GetObject((char*)stream_name); // arg is non-const on Python2
     PyObject* result;
 
     result = PyObject_CallMethod(py_stream, "write", "(s)", text);
--- a/python/meep.i
+++ b/python/meep.i
@@ -209,6 +209,16 @@ void py_master_printf_wrap(const char *s) {
     Py_XDECREF(result);
 }
 
+void py_master_printf_stderr_wrap(const char *s) {
+    PySys_WriteStderr("%s", s);
+    static PyObject *py_stderr = NULL;
+    if (py_stderr == NULL) {
+      py_stderr = PySys_GetObject("stderr");
+    }
+    PyObject *result = PyObject_CallMethod(py_stderr, "flush", NULL);
+    Py_XDECREF(result);
+}
+
 void set_ctl_printf_callback(void (*callback)(const char *s)) {
 #if HAVE_CTL_PRINTF_CALLBACK
   ctl_printf_callback = callback;
@@ -508,14 +518,14 @@ kpoint_list get_eigenmode_coefficients_and_kpoints(meep::fields *f, meep::dft_fl
 }
 
 PyObject *_get_array_slice_dimensions(meep::fields *f, const meep::volume &where, size_t dims[3],
-                                      bool collapse_empty_dimensions, bool snap_empty_dimensions, 
+                                      bool collapse_empty_dimensions, bool snap_empty_dimensions,
                                       meep::component cgrid = Centered, PyObject *min_max_loc = NULL) {
     meep::direction dirs[3] = {meep::X, meep::X, meep::X};
 
     meep::vec min_max_loc_vec[2];
     meep::vec* min_max_loc_vec_ptr = min_max_loc_vec;
     if (!min_max_loc) min_max_loc_vec_ptr = NULL;
-    
+
     int rank = f->get_array_slice_dimensions(where, dims, dirs, collapse_empty_dimensions, snap_empty_dimensions, min_max_loc_vec_ptr, 0, cgrid);
 
     PyObject *py_dirs = PyList_New(3);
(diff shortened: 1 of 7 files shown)


python/meep.i
```diff
@@ -199,16 +199,24 @@ static meep::vec py_kpoint_func_wrap(double freq, int mode, void *user_data) {
...
-    }
-    PyObject *result = PyObject_CallMethod(py_stdout, "flush", NULL);
+static void _do_master_printf(const char* stream_name, const char* text) {
+    PyObject* py_stream = PySys_GetObject((char*)stream_name); // arg is non-const on Python2
+    PyObject* result;
+
+    result = PyObject_CallMethod(py_stream, "write", "(s)", text);
+    Py_XDECREF(result);
+    result = PyObject_CallMethod(py_stream, "flush", NULL);
     Py_XDECREF(result);
```
stevengj, line 209: Can probably get rid of the `result` variable and just call `Py_XDECREF(PyObject_CallMethod(...));` directly.

src/vec.cpp
```diff
@@ -27,6 +27,12 @@ using namespace std;
 
 namespace meep {
 
+static bool _wholenumber(double value) {
```
stevengj, line 30: I would just call it
```cxx
static bool isinteger(double x) { return x == floor(x); }
```
(We haven't generally been using underscores for `static` functions, since those are automatically private.)

src/vec.cpp
```diff
@@ -899,10 +905,16 @@ vec grid_volume::dz() const {
 }
 
 grid_volume volone(double zsize, double a) {
+  if (!_wholenumber(zsize) || !_wholenumber(a))
```
stevengj, line 908: It's not clear to me that we need `a` to be an integer.

stevengj: Isn't the criterion that `zsize * a` is an integer?

src/vec.cpp
```diff
@@ -899,10 +905,16 @@ vec grid_volume::dz() const {
...
+    master_printf_stderr(
+      "WARNING: Creating a grid_volume with non-whole numbers may result in rounding errors.\n");
```
stevengj, line 910: Maybe:
```
Warning: grid volume is not an integer number of pixels; cell size will be rounded to nearest pixel.
```

RobinD42: I ran a test case with only the resolution being a floating point value and while there is still some variance in the volume levels, it is very minor. So it looks like the warning is probably not needed for that parameter, or at least not as important.

src/vec.cpp
RobinD42: It looks like GitHub may have lost a comment I made here earlier, or perhaps I just didn't click the button before closing the tab...

I ran some experiments with `vol2d` where just `a` was a floating point value, and while there was still some variance in the volumes, they were very minor. So it looks like the warning may not be needed there, or at least it is less important. I went ahead and made the change to remove the check of the `a` parameter, but it's in a separate commit so it will be easy to revert if we decide otherwise.

Or were you just commenting about just the `volone` case?

RobinD42: Ah, I found my earlier comment in the other change discussion. The PR was probably updated before I started typing and I didn't notice that I was in the wrong field.

src/vec.cpp
```diff
@@ -899,10 +901,16 @@ vec grid_volume::dz() const {
 }
 
 grid_volume volone(double zsize, double a) {
+  if (!isinteger(zsize))
```
stevengj, line 904: Shouldn't this be `zsize * a` etcetera?

src/vec.cpp
RobinD42: Message changed.

src/vec.cpp
RobinD42: Yes, that makes sense.

Issue #291, opened 2018-04-19
Optimize mode coefficients for planewave case


stevengj:
For the mode-coefficient expansion, there is a special case worth optimizing.   In particular, if:

* The flux plane is in a homogeneous medium.  (Initially, I would just support homogeneous scalar ε and μ, with no anisotropy.)

* The flux plane fills the entire unit cell in those directions, and the boundary conditions in those directions are Bloch-periodic.

In this case, the propagating modes are a finite set of planewaves.   Computing the mode coefficients is just a discrete Fourier transform.  So one option would be to circumvent MPB entirely in this case and just perform a DFT to get the coefficients of the planewaves, appropriately normalized.  It could even compute the coefficients of the evanescent waves too.

Alternatively, since MPB is also super-efficient in this case there is no huge need to optimize the eigenmode calculation.   What would be nice, however, would be to:

* Automatically compute the correct *number* of (non-evanescent) modes for a given frequency.

* Give MPB an initial "guess" for each mode consisting of the analytically known k.

For example suppose we have a 2d calculation with period Λ in the x direction, bloch-periodic boundary conditions with a given kₓ, and have a flux plane of width Λ in a homogeneous region of index n.  Then the propagating modes at a frequency ω (in c=1 units) are the real solutions of k = sqrt(ω²n² - (kₓ+2πm/Λ)²) for integers m.   That is, you just (a) compute the integers m for which the square root is real = number of modes; and (b) compute the corresponding k values, passing (kₓ,k) to MPB to get the mode at that frequency.

cc @HomerReid

stevengj:
In mpb.cpp, after you call `set_maxwell_dielectric`, you can check for a uniform material by something like:
```c
bool ishomogeneous(maxwell_data *mdata) {
    symmetric_matrix *eps_inv = mdata->eps_inv;
    int n = mdata->fft_output_size;
    for (int i = 0; i < n;++i) if (eps_inv[i] != eps_inv[0]) return false;
    return true;
}
```

stevengj:
In particular, change `get_eigenmode` to:

1.  call `ishomogeneous` when the maxwell data is initialized to check whether it is a homogeneous region.

[...]

PR #1316, opened 2020-08-14
eigenmodes of diffracted planewaves

oskooi:
Follows the outline described by @stevengj in #291:[comment](https://github.com/NanoComp/meep/issues/291).

Documentation, tutorial, ~~and test~~ will likely be added in a separate PR once the API is working.

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -3283,7 +3283,12 @@ class Simulation(object):
         if direction is None or direction == mp.AUTOMATIC:
             direction = flux.normal_direction
 
-        if isinstance(bands, list):
+        try:
+            bands_list_range = isinstance(bands, (list,range))
+        except TypeError:
+            bands_list_range = isinstance(bands, list)
+
+        if bands_list_range:
             num_bands = len(bands)
             coeffs = np.zeros(2 * num_bands * flux.freq.size(), dtype=np.complex128)
             vgrp = np.zeros(num_bands * flux.freq.size())
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -3304,17 +3304,17 @@ class Simulation(object):
                 direction
             )
         elif isinstance(bands,DiffractedPlanewave):
-            coeffs = np.zeros(2 * flux.freq.size(), dtype=np.complex128)
-            vgrp = np.zeros(flux.freq.size())
-            cscale = np.zeros(flux.freq.size())
-
+            num_bands = 1
+            coeffs = np.zeros(2 * num_bands * flux.freq.size(), dtype=np.complex128)
+            vgrp = np.zeros(num_bands * flux.freq.size())
+            cscale = np.zeros(num_bands * flux.freq.size())
             diffractedplanewave_args = [
                 np.array(bands.g, dtype=np.intc),
-                py_v3_to_vec(self.dimensions, bands.axis, is_cylindrical=self.is_cylindrical),
-                bands.s,
-                bands.p
+                np.array([bands.axis.x, bands.axis.y, bands.axis.z], dtype=np.float64),
+                bands.s * 1.0,
+                bands.p * 1.0
                 ]
-            diffractedplanewave = mp.diffractedplanewave(*diffractedplane_args)
+            diffractedplanewave = mp.diffractedplanewave(*diffractedplanewave_args)
 
             kpoints, kdom = mp.get_eigenmode_coefficients_and_kpoints(
                 self.fields,
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -99,6 +99,32 @@ def py_v3_to_vec(dims, iterable, is_cylindrical=False):
     else:
         raise ValueError("Invalid dimensions in Volume: {}".format(dims))
 
+class DiffractedPlanewave(object):
+    def __init__(self,
+                 g=None,
+                 axis=None,
+                 s=None,
+                 p=None):
+        self._g = g
+        self._axis = axis
+        self._s = s
+        self._p = p
+
+    @property
+    def g(self):
+        return self._g
+
+    @property
+    def axis(self):
+        return self._axis
+
+    @property
+    def s(self):
+        return self._s
+
+    @property
+    def p(self):
+        return self._p
 
 class PML(object):
     """
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -106,9 +106,9 @@ class DiffractedPlanewave(object):
                  s=None,
                  p=None):
         self._g = g
-        self._axis = axis
-        self._s = s
-        self._p = p
+        self._axis = Vector3(*axis)
+        self._s = complex(s)
+        self._p = complex(p)
 
     @property
     def g(self):
@@ -3283,25 +3283,55 @@ class Simulation(object):
         if direction is None or direction == mp.AUTOMATIC:
             direction = flux.normal_direction
 
-        num_bands = len(bands)
-        coeffs = np.zeros(2 * num_bands * flux.freq.size(), dtype=np.complex128)
-        vgrp = np.zeros(num_bands * flux.freq.size())
-        cscale = np.zeros(num_bands * flux.freq.size())
+        if isinstance(bands,(list,range)):
+            num_bands = len(bands)
+            coeffs = np.zeros(2 * num_bands * flux.freq.size(), dtype=np.complex128)
+            vgrp = np.zeros(num_bands * flux.freq.size())
+            cscale = np.zeros(num_bands * flux.freq.size())
 
-        kpoints, kdom = mp.get_eigenmode_coefficients_and_kpoints(
-            self.fields,
-            flux.swigobj,
-            eig_vol,
-            np.array(bands, dtype=np.intc),
-            eig_parity,
-            eig_resolution,
-            eig_tolerance,
-            coeffs,
-            vgrp,
-            kpoint_func,
-            cscale,
-            direction
-        )
+            kpoints, kdom = mp.get_eigenmode_coefficients_and_kpoints(
+                self.fields,
+                flux.swigobj,
+                eig_vol,
+                np.array(bands, dtype=np.intc),
+                eig_parity,
+                eig_resolution,
+                eig_tolerance,
+                coeffs,
+                vgrp,
+                kpoint_func,
+                cscale,
+                direction
+            )
+        elif isinstance(bands,DiffractedPlanewave):
+            coeffs = np.zeros(2 * flux.freq.size(), dtype=np.complex128)
+            vgrp = np.zeros(flux.freq.size())
+            cscale = np.zeros(flux.freq.size())
+
+            diffractedplanewave_args = [
+                np.array(bands.g, dtype=np.intc),
+                py_v3_to_vec(self.dimensions, bands.axis, is_cylindrical=self.is_cylindrical),
+                bands.s,
+                bands.p
+                ]
+            diffractedplanewave = mp.diffractedplanewave(*diffractedplane_args)
+
+            kpoints, kdom = mp.get_eigenmode_coefficients_and_kpoints(
+                self.fields,
+                flux.swigobj,
+                eig_vol,
+                diffractedplanewave,
+                eig_parity,
+                eig_resolution,
+                eig_tolerance,
+                coeffs,
+                vgrp,
+                kpoint_func,
+                cscale,
+                direction
+            )
+        else:
+            raise TypeError("get_eigenmode_coefficients: bands must be either a list or DiffractedPlanewave object")
 
         return EigCoeffsResult(np.reshape(coeffs, (num_bands, flux.freq.size(), 2)), vgrp, kpoints, kdom, cscale)
 
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -3283,7 +3283,7 @@ class Simulation(object):
         if direction is None or direction == mp.AUTOMATIC:
             direction = flux.normal_direction
 
-        if isinstance(bands,(list,range)):
+        if isinstance(bands, list):
             num_bands = len(bands)
             coeffs = np.zeros(2 * num_bands * flux.freq.size(), dtype=np.complex128)
             vgrp = np.zeros(num_bands * flux.freq.size())
@@ -3303,7 +3303,7 @@ class Simulation(object):
                 cscale,
                 direction
             )
-        elif isinstance(bands,DiffractedPlanewave):
+        elif isinstance(bands, DiffractedPlanewave):
             num_bands = 1
             coeffs = np.zeros(2 * num_bands * flux.freq.size(), dtype=np.complex128)
             vgrp = np.zeros(num_bands * flux.freq.size())
--- a/src/mpb.cpp
+++ b/src/mpb.cpp
@@ -626,8 +626,11 @@ void *fields::get_eigenmode(double frequency, direction d, const volume where, c
           vec cen = eig_vol.center();
           double nn = sqrt(real(get_eps(cen, frequency)) * real(get_mu(cen, frequency)));
           double k2 = frequency*frequency*nn*nn - k2sum;
-          if (k2 < 0)
+          if (k2 < 0) {
+            master_printf("WARNING: diffraction order for g=(%d,%d,%d) is evanescent!\n",
+                          dp->get_axis()[0],dp->get_axis()[1],dp->get_axis()[2]);
             return NULL;
+          }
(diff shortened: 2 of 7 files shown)


python/simulation.py
```diff
@@ -98,6 +98,32 @@ def py_v3_to_vec(dims, iterable, is_cylindrical=False):
     else:
         raise ValueError("Invalid dimensions in Volume: {}".format(dims))
 
+class DiffractedPlanewave(object):
+    def __init__(self,
+                 g=None,
+                 axis=None,
```
stevengj, line 104: Eventually, we'll have the EigenmodeSource etcetera choose a reasonable default if `axis=None` when the source is added.

(You'll generally want `axis` to lie in the plane of the source, I think.)

python/simulation.py
```diff
@@ -98,6 +98,32 @@ def py_v3_to_vec(dims, iterable, is_cylindrical=False):
     else:
         raise ValueError("Invalid dimensions in Volume: {}".format(dims))
 
+class DiffractedPlanewave(object):
```
stevengj, line 101: Note: the current plan is to optionally pass this for the `band_num` parameter to eigenmode sources etcetera.

oskooi:
The basic framework now seems to be working although there are several issues that need to be resolved.

As an initial demonstration/test, the mode coefficients of the diffracted orders of the 2d binary phase grating (from [Tutorial/Mode Decomposition/Transmittance Spectra for Planewave at Normal Incidence](https://meep.readthedocs.io/en/latest/Python_Tutorials/Mode_Decomposition/#transmittance-spectra-for-planewave-at-normal-incidence)) computed using a `DiffractedPlanewave` object passed as the `bands` parameter to `get_eigenmode_coefficients` are equivalent to those from the tutorial as demonstrated below.

```py
import meep as mp
import math
import cmath
import numpy as np

resolution = 50        # pixels/μm      

dpad = 3.0             # length of padding between grating and PML   

sx = dpml+dsub+gh+dpad+dpml
sy = gp

ng = 1.5
glass = mp.Medium(index=ng)

# rotation angle of incident planewave; counter clockwise (CCW) about Z axis, 0 degrees along +X axis
theta_in = math.radians(0)

# k (in source medium) with correct length (plane of incidence: XY) 
k = mp.Vector3(fcen*ng).rotate(mp.Vector3(z=1), theta_in)

symmetries = []
eig_parity = mp.ODD_Z
if theta_in == 0:
  k = mp.Vector3(0,0,0)
  eig_parity += mp.EVEN_Y

def pw_amp(k,x0):
  def _pw_amp(x):
  return _pw_amp

src_pt = mp.Vector3(-0.5*sx+dpml+0.3*dsub,0,0)
                     component=mp.Ez,
                     center=src_pt,
                     size=mp.Vector3(0,sy,0),
                     amp_func=pw_amp(k,src_pt))]

                    cell_size=cell_size,
                    boundary_layers=pml_layers,
                    k_point=k,
                    default_material=glass,
                    sources=sources,
                    symmetries=symmetries)

tran_pt = mp.Vector3(0.5*sx-dpml-0.5*dpad,0,0)
tran_flux = sim.add_flux(fcen, 0, 1, mp.FluxRegion(center=tran_pt, size=mp.Vector3(0,sy,0)))

sim.run(until_after_sources=50)

input_flux = mp.get_fluxes(tran_flux)

sim.reset_meep()

geometry = [mp.Block(material=glass,
                     size=mp.Vector3(dpml+dsub,mp.inf,mp.inf),
                     center=mp.Vector3(-0.5*sx+0.5*(dpml+dsub),0,0)),
            mp.Block(material=glass,
                     size=mp.Vector3(gh,gdc*gp,mp.inf),
                     center=mp.Vector3(-0.5*sx+dpml+dsub+0.5*gh,0,0))]

                    cell_size=cell_size,
                    boundary_layers=pml_layers,
                    geometry=geometry,
                    k_point=k,
                    sources=sources,
                    symmetries=symmetries)

tran_flux = sim.add_mode_monitor(fcen, 0, 1, mp.FluxRegion(center=tran_pt, size=mp.Vector3(0,sy,0)))

## when using "bands" as a list:         
##      bands=[1] is 0th diffraction order, bands[2] is 1st diffraction order, ...   
## when using "bands" as a DiffractedPlanewave: 
##      (0,0,0) is 0th diffraction order, (0,1,0) is 1st diffraction order, ...
diff_order = 1

res = sim.get_eigenmode_coefficients(tran_flux, [diff_order+1], eig_parity=eig_parity)
t_ref = res.alpha

res = sim.get_eigenmode_coefficients(tran_flux, mp.DiffractedPlanewave((0,diff_order,0),mp.Vector3(1,1,0),1,0))
t_dp = res.alpha

print("tran:, {}, {:.8f} (eigensolver), {:.8f} (planewave)".format(diff_order,
                                                                   0.5*abs(t_ref[0,0,0])**2/input_flux[0],
                                                                   abs(t_dp[0,0,0])**2/input_flux[0]))
```

**first diffraction order (`diff_order=1`)**
```
tran:, 1, 0.38779358 (eigensolver), 0.38779517 (planewave)
```

**third diffraction order (`diff_order=3`)**
```
tran:, 3, 0.04276357 (eigensolver), 0.04276669 (planewave)
```

**fifth diffraction order (`diff_order=5`)**
```
tran:, 5, 0.01513773 (eigensolver), 0.01514239 (planewave)
```

Issues that need to be fixed:
- passing `eig_parity` to `get_eigenmode_coefficients` when using `DiffractedPlanewave` and replacing `sim.add_mode_monitor` with `sim.add_flux` (in order to exploit symmetries) produces incorrect results. Is there something missing when calling `maxwell_set_planewave` when symmetries are present?
- the dominant wavevector obtained from calling `maxwell_dominant_planewave` at the end of the `get_eigenmode` function call is incorrect when using a `DiffractedPlanewave` object because the `band_num` input parameter is always `1`. How should the call to `maxwell_dominant_planewave` be modified when using `DiffractedPlanewave`? Should the `band_num` parameter be set to the maximum absolute value of one of the three elements of its `g` member object?
- the serial build on Python 2.7 is failing for reasons that are not clear
- additional testing is required using a 3d example

oskooi:
The following is a 3d test based on extending the [2d binary-grating tutorial example](https://meep.readthedocs.io/en/latest/Python_Tutorials/Mode_Decomposition/#transmittance-spectra-for-planewave-at-normal-incidence). The test involves computing the mode coefficients of all transmitted diffraction orders based on individual calls to `get_eigenmode_coefficients` using `DiffractedPlanewave` and verifying that the sum of the power in all the orders are equivalent to the Poynting flux value obtained using `get_fluxes`. (This is similar to [python/tests/binary_grating.py](https://github.com/NanoComp/meep/blob/master/python/tests/binary_grating.py).)

```py
# binary grating on a semi-infinite glass substrate                                                                                                                  
# structure is 3d extrusion of tutorial example                                                                                                                      

import math
import cmath
import numpy as np
import meep as mp

resolution = 50        # pixels/μm                                                                                                                                   

dsub = 3.0             # substrate length                                                                                                                            
dpad = 3.0             # padding length between grating and pml                                                                                                      
gp = 2.0               # grating period                                                                                                                              

sx = dpml+dsub+gh+dpad+dpml
sy = gp
sz = gp

cell_size = mp.Vector3(sx,sy,sz)

wvl = 0.5              # wavelength                                                                                                                                  
fcen = 1/wvl           # frequency                                                                                                                                   

ng = 1.5
glass = mp.Medium(index=ng)

# angle of incident planewave; CCW about Z axis, 0 degrees along +X axis                                                                                             
theta_in = math.radians(0)

# k with correct length (plane of incidence: XY)                                                                                                                     
k = mp.Vector3(math.cos(theta_in),math.sin(theta_in)).scale(fcen)

## disabled: causes field instability                                                                                                                                
# symmetries = [mp.Mirror(mp.Z, phase=-1)]                                                                                                                           
symmetries = []
if theta_in == 0:
  k = mp.Vector3()
  symmetries.append(mp.Mirror(mp.Y))

def pw_amp(k,x0):
  def _pw_amp(x):
    return cmath.exp(1j*2*math.pi*ng*k.dot(x+x0))
  return _pw_amp

sources = [mp.Source(mp.GaussianSource(fcen, fwidth=0.2*fcen),
                     component=mp.Ez,
                     center=src_pt,
                     size=mp.Vector3(0,sy,sz),
                     amp_func=pw_amp(k,src_pt))]

                    cell_size=cell_size,
                    boundary_layers=pml_layers,
                    k_point=k,
                    default_material=glass,
                    sources=sources,
                    symmetries=symmetries)

flux_mon = sim.add_flux(fcen, 0, 1,
                        mp.FluxRegion(center=mp.Vector3(0.5*sx-dpml-0.5*dpad), size=mp.Vector3(0,sy,sz)))

sim.run(until_after_sources=10)

input_flux = mp.get_fluxes(flux_mon)

sim.reset_meep()

geometry = [mp.Block(material=glass,
                     size=mp.Vector3(dpml+dsub,mp.inf,mp.inf),
                     center=mp.Vector3(-0.5*sx+0.5*(dpml+dsub))),
            mp.Block(material=glass,
                     size=mp.Vector3(gh,gdc*gp,gdc*gp),
                     center=mp.Vector3(-0.5*sx+dpml+dsub+0.5*gh))]

                    cell_size=cell_size,
                    boundary_layers=pml_layers,
                    geometry=geometry,
                    k_point=k,
                    sources=sources,
                    symmetries=symmetries)

mode_mon = sim.add_flux(fcen, 0, 1,
                        mp.FluxRegion(center=mp.Vector3(0.5*sx-dpml-0.5*dpad), size=mp.Vector3(0,sy,sz)))

m_flux = 0
for ny in range(-4,5):
  for nz in range(-4,5):
    res = sim.get_eigenmode_coefficients(mode_mon,
                                         mp.DiffractedPlanewave((0,ny,nz),mp.Vector3(1.2,1.5,0),1,0),
                                         eig_parity=mp.EVEN_Y+mp.ODD_Z)
    tran = abs(res.alpha[0,0,0])**2/input_flux[0]
    print("tran:, {}, {}, {}".format(ny,nz,tran))
    m_flux += tran

flux = mp.get_fluxes(mode_mon)
t_flux = t_flux[0]/input_flux[0]

err = abs(m_flux_t_flux)/t_flux

print("flux:, {} (mode coeffs.), {} (flux), {} (error)".format(0.5*m_flux,t_flux,err))
```
The results show that the relative error is converging to zero with increasing resolution.

**resolution = 30**
```
flux:, 0.843164820063639 (mode coeffs.), 0.9033501462240456 (flux), 0.06662458229732734 (error)
```

**resolution = 40**
```
flux:, 0.871396520203924 (mode coeffs.), 0.9137573158711103 (flux), 0.04635891273472605 (error)
```

RobinD42:
@oskooi, If I'm reading the diffs and the current code correctly, it looks like you didn't add the new `diffractedplanewave *dp` parameter to the version of the `fields::get_` methods that are used when the build is done without MPB. Look inside the `#else // #ifdef HAVE_MPB` branch.

stevengj:
Diffracted planewaves are not mirror symmetric.   As explained in the documentation, you have to use `add_mode_monitor` when computing mode coefficients that don't match the `Symmetry` of the simulation — you can only use `add_flux` to compute coupling to modes of the same symmetry.

oskooi:
Output for the fifth diffraction order:
```
Dominant planewave for band 5: (0.987636,-1.739130,0.000000)
eigenmode group velocity: 0.49382
diffracted planewave: k = 2, 0, 0 for omega = 2
diffracted planewave: group velocity v=1.41421
Dominant planewave for band 1: (2.000000,1.739130,0.000000)
tran:, 4, 0.01082581 (eigensolver), 0.01240070 (planewave). 0.14547530 (error)
```

src/mpb.cpp
```diff
@@ -591,6 +600,28 @@ void *fields::get_eigenmode(double frequency, direction d, const volume where, c
     } while (match_frequency &&
              fabs(sqrt(eigvals[band_num - 1]) - frequency) > frequency * match_tol);
 
+  if (am_master() && dp) {
+    scalar_complex s = {real(dp->get_s()),imag(dp->get_s())};
+    scalar_complex p = {real(dp->get_p()),imag(dp->get_p())};
+
+    master_printf("diffracted planewave: k = %g, %g, %g for omega = %g\n", k[0],k[1],k[2], frequency);
```
stevengj, line 607: It needs to first compute the correct k vector from the frequency and `get_g()`.

src/mpb.cpp
```diff
@@ -591,6 +600,28 @@ void *fields::get_eigenmode(double frequency, direction d, const volume where, c
...
+    update_maxwell_data_k(mdata, k, G[0], G[1], G[2]);
+    maxwell_set_planewave(mdata, H, band_num, dp->get_g(), s, p, dp->get_axis());
+
+    eigvals[band_num - 1] = frequency;
```
stevengj, line 611: or if match_frequency is false then it needs to compute the `frequency` from `k` and `get_g()`.

oskooi:
The following code computes the correct "k" (which is actually equivalent to the dominant planewave) from the frequency and `get_g()`:

```cpp
    if (match_frequency) {
      double k2sum = 0;
      LOOP_OVER_DIRECTIONS(v.dim, dd) {
        int m = dp->get_g()[dd - X];
        if ((float(eig_vol.in_direction(dd)) == float(v.in_direction(dd))) &&
                (boundaries[High][dd] == Periodic && boundaries[Low][dd] == Periodic)) {
          k[dd - X] += m/v.in_direction(dd);
          k2sum += k[dd - X]*k[dd - X];
        }
      }
      vec cen = eig_vol.center();
      double nn = sqrt(real(get_eps(cen, frequency)) * real(get_mu(cen, frequency)));
      LOOP_OVER_DIRECTIONS(v.dim, dd) {
        if ((float(eig_vol.in_direction(dd)) != float(v.in_direction(dd))) &&
                (boundaries[High][dd] == Periodic && boundaries[Low][dd] == Periodic)) {
          k[dd - X] = sqrt(frequency*frequency*nn*nn - k2sum);
        }
      }
    }
```

While it produces the correct result for "k" as shown in the output below on the line `diffracted planewave: k = ...`, the dominant planewave (following the call to `maxwell_dominant_planewave` after this code block) and  the eigenmode coefficients are *incorrect*.

```
Dominant planewave for band 4: (1.977372,-0.300000,0.000000)
eigenmode group velocity: 0.98869
diffracted planewave: k = 1.97737, 0.3, 0 for omega = 2
diffracted planewave: group velocity v=1.39821
Dominant planewave for band 1: (1.977372,0.330000,0.000000)
tran:, 3, 0.04292408 (eigensolver), 0.00794525 (planewave), 0.81490000 (error)
```

The dominant planewave is reported as `(1.977372,0.330000,0.000000)`; the correct answer is `(1.97737, 0.3, 0)`.

Is the code computing an incorrect value for "k"?

stevengj:
Notes:

For the 3d case, things seem fine — the basic conceptual issue here is that in 3d, different diffraction orders have different planes of incidence (not just the xy plane!), and so the definition of s and p will depend on the choice of axis.  Normally you would want the axis to be orthogonal to the mode-monitor plane, however, I think — it would be good to make this the default.

The issue with symmetries seems likely to be unrelated to this PR, so probably it should be debugged separately.

oskooi:
I think I know why `DiffractedPlanewave` is not working with symmetries. It has to do with `fields::add_mode_monitor` which ignores symmetries when creating the mode monitor:

https://github.com/NanoComp/meep/blob/859d2198e43c9d3bec00858ceaae87aaa696d4f1/src/dft.cpp#L658-L660

Omitting symmetries in the mode monitor when symmetries are specified for the cell volume in the case of a normally incident planewave with `k_point=mp.Vector3(0,0,0)` therefore means that the length of the mode monitor in the periodic directions (which extends the entire length of the cell when computing diffracted planewaves) is *not* equivalent to the cell length in those directions (for which a mirror symmetry has been defined; the difference being a factor of ~2) which therefore causes these two checks in `fields::get_eigenmode` for the k<sub>⟂</sub> and k<sub>∥</sub> components of the `diffractedplanewave` object to fail:

https://github.com/NanoComp/meep/blob/859d2198e43c9d3bec00858ceaae87aaa696d4f1/src/mpb.cpp#L603-L604

https://github.com/NanoComp/meep/blob/859d2198e43c9d3bec00858ceaae87aaa696d4f1/src/mpb.cpp#L614-L615

We obviously do not want to modify `add_mode_monitor` to support symmetries since it is not intended for that purpose.

Unfortunately, there does not seem to be a simple workaround.

PR #1326, opened 2020-08-25
tutorial for diffracted planewave

oskooi:
Should be merged after #1316.


stevengj:
Do we want to choose a default axis first?

stevengj:
Probably should be the "first" direction that lies in the plane of the monitor/source (e.g. the Y direction for a YZ plane).

PR #1329, opened 2020-08-26
Adjoint near2far with vol_src

mochen4:


python/adjoint/objective.py
```diff
@@ -132,4 +133,216 @@ def get_evaluation(self):
         try:
             return self.eval
         except AttributeError:
-            raise RuntimeError("You must first run a forward simulation before resquesting an eigenmode coefficient.")
\ No newline at end of file
+            raise RuntimeError("You must first run a forward simulation before resquesting an eigenmode coefficient.")
+
+
+
+class Fourier_Coefficients(ObjectiveQuantitiy):
```
smartalecH, line 140: Maybe call it `FourierCoefficients` (omit the underscore) to match the rest of the codebase.

python/adjoint/objective.py
```diff
@@ -132,4 +133,216 @@ def get_evaluation(self):
...
+            return self.eval
+        except AttributeError:
+            raise RuntimeError("You must first run a forward simulation.")
+
+
+
+class Far_Coefficients(ObjectiveQuantitiy):
+    def __init__(self,sim,Near2FarRegions, far_pt):
+        self.sim = sim
+        self.Near2FarRegions=Near2FarRegions
```
smartalecH, line 213: This should always be a list, right? Maybe do a check so the user can supply a single region without a list.

python/adjoint/optimization_problem.py
```diff
@@ -229,12 +230,30 @@ def prepare_adjoint_run(self,objective_idx):
...
+        if self.adjoint_sources[0] == 'src_vol':
+            self.sim.change_sources([])
+            self.sim.init_sim()
+            for (lincoeff, basisfunc, srcdata) in self.adjoint_sources[1:]:
+                (num_basis, num_pts) = lincoeff.shape
+                for basis_i in range(num_basis):
+                    amps_arr_vec = mp.ComplexVector(num_pts)
+                    for i in range(num_pts):
+                        amps_arr_vec[i] = lincoeff[basis_i, i]
+                    self.sim.fields.add_srcdata(srcdata.fc, srcdata.near_fd_comp, basisfunc[basis_i].swigobj, num_pts, srcdata.idx_arr, amps_arr_vec)
```
smartalecH, line 250: Why aren't we just creating a volume source in the FarCoefficients class? This routine should simply update the sources and not have to do any "postprocessing" of the sources.

smartalecH: You might need to create a new python source that does all of the c++ interfaces once the simulation object is initialized (that looks like what you are trying to hack here). Maybe look at how the other sources are defined.

python/adjoint/optimization_problem.py
```diff
@@ -433,10 +462,22 @@ def _stop(sim):
...
+                    #relative_change_raw = np.abs(previous_fields[mi][ic] - current_fields[mi][ic]) / np.abs(previous_fields[mi][ic])
+                    #relative_change.append(np.mean(relative_change_raw.flatten())) # average across space and frequency
+
+                    if previous_fields[mi][ic] is not None:
+                        relative_change_raw = np.abs(previous_fields[mi][ic] - current_fields[mi][ic]) / np.abs(previous_fields[mi][ic])
+                        relative_change.append(np.mean(relative_change_raw.flatten())) # average across space and frequency
+                    else:
+                        relative_change.append(1)
+
+
```
smartalecH, line 480: This modification doesn't look right... what's the advantage of doing it this way?

python/adjoint/objective.py
```diff
@@ -132,4 +133,216 @@ def get_evaluation(self):
...
+                scale = -freq_scale * amp_arr
+            else:
+                scale = freq_scale * amp_arr
+
+            if self.num_freq == 1:
+                ##TODO
+                self.sources += [mp.Source(self.time_src, component=cur_comp, amplitude=scale[0], center=near_pt)]
+            else:
+                src = FilteredSource(self.time_src.frequency,self.frequencies,scale,dt,self.time_src)
+                self.sources+=[(src.nodes, src.time_src_bf, near_data)]
```
smartalecH, line 331: I think you should code the basis function in C++... otherwise, you're still calling back to python every timestep. It's not as bad as before since the entire volume has the same time profile, but if you have lot's of frequencies and multiple volumes it can add up.

Plus I think it will keep the interface cleaner (and easier to follow...)

python/adjoint/optimization_problem.py
smartalecH: For example, after you create a new class, you can just add an extra condition here that does all the hackery:

https://github.com/NanoComp/meep/blob/859d2198e43c9d3bec00858ceaae87aaa696d4f1/python/simulation.py#L2288-L2294

src/meep.hpp
```diff
@@ -1188,6 +1190,25 @@ class dft_force {
...
+  vec near_x;
+  component near_fd_comp;
+  std::vector<std::vector<std::complex<double> > > matrix_elt;
+};
+
+struct sourcedata{
+  component near_fd_comp;
+  std::vector<ptrdiff_t> idx_arr;
+  int fc_idx;
+  std::vector<std::vector<std::complex<double> > > amp_arr;
```
stevengj, line 1205: should be 1d array in row-major

src/meep.hpp
```diff
@@ -1188,6 +1190,25 @@ class dft_force {
   volume where;
 };
 
+
+
+struct near_data{
+  vec near_x;
+  component near_fd_comp;
+  std::vector<std::vector<std::complex<double> > > matrix_elt;
+};
```
stevengj, line 1199: delete this and other old code if it's no longer relevant

src/meep.hpp
stevengj: (storing multidimensional arrays as arrays of arrays will be much harder to get into Python — numpy arrays are just "flat" row-major arrays)

python/adjoint/objective.py
stevengj: Maybe FourierFields and Near2FarFields

python/adjoint/objective.py
```diff
@@ -132,4 +133,149 @@ def get_evaluation(self):
...
+
+        dJ = list(dJ.flatten())
+        dJ_c = mp.ComplexVector(len(dJ))
+        for i in range(len(dJ)):
+            dJ_c[i] = dJ[i]
+
+        #TODO far_pts in 3d or cylindrical, perhaps py_v3_to_vec from simulation.py
+        self.all_nearsrcdata = self.monitor.swigobj.near_sourcedata(mp.vec(self.far_pt.x, self.far_pt.y), dJ_c)
+
+
```
stevengj, line 248: Delete the extra blank lines

python/adjoint/objective.py
```diff
@@ -132,4 +133,149 @@ def get_evaluation(self):
...
+            if cur_comp in [mp.Ex, mp.Ey, mp.Ez]:
+                scale = -freq_scale * amp_arr
+            else:
+                scale = freq_scale * amp_arr
+
+            if self.num_freq == 1:
+                ##TODO
+                self.sources += [mp.Source(self.time_src, component=cur_comp, amplitude=scale[0], center=near_pt)]
```
stevengj, line 261: Seems like this should be an array of amplitudes for each spatial point, not a single point source?

python/adjoint/objective.py
```diff
@@ -132,4 +133,149 @@ def get_evaluation(self):
...
+            else:
+                src = FilteredSource(self.time_src.frequency,self.frequencies,scale,dt,self.time_src)
+                self.sources+=[(src.nodes, src.time_src_bf, near_data)]
```
stevengj, line 264: This should be a `Source` object of some sort.

python/adjoint/optimization_problem.py
```diff
@@ -229,12 +230,30 @@ def prepare_adjoint_run(self,objective_idx):
...
+        if self.adjoint_sources[0] == 'src_vol':
+            self.sim.change_sources([])
+            self.sim.init_sim()
+            for (lincoeff, basisfunc, srcdata) in self.adjoint_sources[1:]:
+                (num_basis, num_pts) = lincoeff.shape
+                for basis_i in range(num_basis):
+                    amps_arr_vec = mp.ComplexVector(num_pts)
+                    for i in range(num_pts):
+                        amps_arr_vec[i] = lincoeff[basis_i, i]
+                    self.sim.fields.add_srcdata(srcdata, basisfunc[basis_i].swigobj, num_pts, amps_arr_vec)
```
stevengj, line 250: This should be in the near2far class?

python/adjoint/objective.py
```diff
@@ -132,4 +133,149 @@ def get_evaluation(self):
...
+
+        for near_data in self.all_nearsrcdata:
+            cur_comp = near_data.near_fd_comp
+            amp_arr = near_data.amp_arr.reshape(self.num_freq, -1)
+
+
+            if cur_comp in [mp.Ex, mp.Ey, mp.Ez]:
+                scale = -freq_scale * amp_arr
+            else:
+                scale = freq_scale * amp_arr
```
stevengj, line 257: It seems like this should be done in the near_sourcedata function?

python/adjoint/objective.py
```diff
@@ -132,4 +133,149 @@ def get_evaluation(self):
...
+            dV = 1/self.sim.resolution
+        elif self.sim.cell_size.z == 0:
+            dV = 1/self.sim.resolution * 1/self.sim.resolution
+        else:
+            dV = 1/self.sim.resolution * 1/self.sim.resolution * 1/self.sim.resolution
+
+        self.sources = ['src_vol']
+
+
+        freq_scale = 1j * 2 * np.pi * self.frequencies / np.array([self.time_src.fourier_transform(f) for f in self.frequencies])
```
stevengj, line 238: Why is this scaling necessary?

python/adjoint/objective.py
```diff
@@ -166,16 +166,26 @@ def place_adjoint_source(self,dJ):
...
+        iomega = (1.0 - np.exp(-1j * (2 * np.pi * self.frequencies) * dt)) * (1.0 / dt) # scaled frequency factor with discrete time derivative fix
+        src = self.time_src
+
+        # an ugly way to calcuate the scaled dtft of the forward source
+        y = np.array([src.swigobj.current(t,dt) for t in np.arange(0,T,dt)]) # time domain signal
+        fwd_dtft = np.matmul(np.exp(1j*2*np.pi*self.frequencies[:,np.newaxis]*np.arange(y.size)*dt), y)*dt/np.sqrt(2*np.pi) # dtft
+
+        # we need to compensate for the phase added by the time envelope at our freq of interest
+        src_center_dtft = np.matmul(np.exp(1j*2*np.pi*np.array([src.frequency])[:,np.newaxis]*np.arange(y.size)*dt), y)*dt/np.sqrt(2*np.pi)
+        adj_src_phase = np.exp(1j*np.angle(src_center_dtft))
```
stevengj, line 174: This copy-and-paste code should be refactored into a function.

See e.g. https://www.beyondjava.net/the-art-of-refactoring-removing-code-duplications

python/source.py
```diff
@@ -583,3 +582,12 @@ def beam_w0(self):
     @property
     def beam_E0(self):
         return self._beam_E0
+
+class IndexedSource(Source):
+    def __init__(self, src, num_pts, srcdata, amp_arr):
+        self.src = src
+        self.num_pts = num_pts
+        self.srcdata = srcdata
+        self.amp_arr_vec = mp.ComplexVector(num_pts)
```
stevengj, line 591: Can't you just do `self.amp_arr_vec = amp_arr.copy()`?

stevengj: Can't `ComplexVector` just be a numpy array?

python/source.py
```diff
@@ -583,3 +582,12 @@ def beam_w0(self):
     @property
     def beam_E0(self):
         return self._beam_E0
+
+class IndexedSource(Source):
+    def __init__(self, src, num_pts, srcdata, amp_arr):
+        self.src = src
+        self.num_pts = num_pts
```
stevengj, line 589: We don't need `num_pts` since this is just `len(amp_arr)`.

python/source.py
```diff
@@ -583,3 +582,12 @@ def beam_w0(self):
...
+        self.srcdata = srcdata
```
stevengj, line 590: Is `srcdata` a `meep::srcdata*` or something?  Would be good to have a comment here.

stevengj:
Note: issue #1365 will make this even more efficient.

src/meep.hpp
```diff
@@ -1188,6 +1188,16 @@ class dft_force {
   volume where;
 };
 
+
+struct sourcedata{
+  component near_fd_comp;
+  std::vector<ptrdiff_t> idx_arr;
+  int fc_idx;
+  std::vector<std::complex<double> > amp_arr;
+
```
stevengj, line 1197: no blank line

src/sources.cpp
```diff
@@ -304,6 +304,23 @@ static void src_vol_chunkloop(fields_chunk *fc, int ichunk, component c, ivec is
...
+    index_arr[i] = cur_data.idx_arr[i];
+  component c = cur_data.near_fd_comp;
+  complex<double> *amps_arr = new complex<double>[n];
+  for (size_t i = 0; i < n; i++){
+    amps_arr[i] = amps_arr_vec[i];
+  }
+  src_vol *tmp = new src_vol(c, src, n, index_arr, amps_arr);
+  field_type ft = is_magnetic(c) ? B_stuff : D_stuff;
+  fields_chunk *fc = chunks[cur_data.fc_idx];
+  fc->sources[ft] = tmp->add_to(fc->sources[ft]);
```
stevengj, line 320: Should `abort` if `cur_data.fc_idx` is not `0 < cur_data.fc_idx < num_chunks` or `!fc->is_mine()`

src/sources.cpp
```diff
@@ -304,6 +304,23 @@ static void src_vol_chunkloop(fields_chunk *fc, int ichunk, component c, ivec is
   fc->sources[ft] = tmp->add_to(fc->sources[ft]);
 }
 
+void fields::add_srcdata(struct sourcedata cur_data, src_time *src, size_t n, std::vector<std::complex<double> > amps_arr_vec){
+  sources = src->add_to(sources, &src);
+  ptrdiff_t* index_arr = new ptrdiff_t[n];
+  for (size_t i = 0; i < n; i++)
+    index_arr[i] = cur_data.idx_arr[i];
```
stevengj, line 311: can combine this into a single loop over `i` below.

python/meep.i
```diff
@@ -1464,6 +1472,16 @@ void _get_gradient(PyObject *grad, PyObject *fields_a, PyObject *fields_f, PyObj
...
 
+%include "typemaps.i"
+%template(near_src_data) std::vector<meep::sourcedata>;
+
+%include "std_complex.i"
+%template(ComplexVector) std::vector<std::complex<double> >;
+
+std::vector<struct meep::sourcedata> meep::dft_near2far::near_sourcedata(const meep::vec &x, std::complex<double>* dJ);
+
+void meep::fields::add_srcdata(struct meep::sourcedata cur_data, meep::src_time *src, size_t n, std::complex<double>* amp_arr);
```
stevengj, line 1483: @mochen4, which did you add these declaration here?    SWIG should already pull in the declarations of `dft_near2far::near_sourcedata` and `fields::add_srcdata` from `meep.hpp`, no?

mochen4: I guess you are right. I wasn't familiar with SWIG and meep.i, so I wrote those explicitly in meep.i to make sure. I will double check to see if we don't need them, thanks!

src/meep.hpp
```diff
@@ -1346,11 +1357,12 @@ class fields_chunk {
   structure_chunk *new_s;
   structure_chunk *s;
   const char *outdir;
+  int chunk_idx;
```
oskooi, line 1360: Why was it necessary to add a new member variable `chunk_idx` to the `fields_chunk` class? It doesn't seem to be actually used for anything. There is a member function `n_proc()` which already returns the chunk ID (equivalent to the MPI rank) and thus adding a `chunk_idx` variable seems redundant.

smartalecH: Can't a proc have multiple chunks?

oskooi: Yes for the PML regions but still `chunk_idx` doesn't seem to be used for anything.

mochen4: ``chunk_idx`` is first used in ``dft_near2far::near_sourcedata``, as the ``fc_idx`` in the structure ``sourcedata``. That information is later used in`` fields::add_srcdata`` to find the right field chunk to add source.

oskooi: `dft_near2far::near_sourcedata` seems to be creating a `sourcedata` object for every `dft_chunk` and since `dft_chunk`s are only ever created in *non-PML* regions it should be unnecessary to add a separate `chunk_idx` member variable to the `fields_chunk` class since `n_proc()` already provides this information.

PR #1330, opened 2020-08-26
symmetry support for diffractedplanewave

oskooi:
Adds support for symmetries to the recently added `diffractedplanewave` class from #1316.

When computing the k<sub>∥</sub> component of the `diffractedplanewave` object, rather than identify the parallel direction(s) of the mode monitor by checking whether its length in that direction matches the cell length (i.e., the monitor extends the entire length of the cell), it just checks whether the user-specified diffraction order `m` is non-zero. This avoids any issues involving symmetries in which the cell length is halved but the length of the mode monitor is not.

Similarly, the direction of the k<sub>⟂</sub> component is identified by simply checking that the length of the mode monitor in that direction is 0 (i.e., line in 2d, plane in 3d).

The test and tutorial have been updated to include symmetries.

(Two of the four subtests in `python/tests/diffracted_planewave.py` which are redundant were removed to provide a speedup.)


src/mpb.cpp
```diff
@@ -596,22 +596,21 @@ void *fields::get_eigenmode(double frequency, direction d, const volume where, c
...
+        ktmp += m/eig_vol.in_direction(dd);
+      k2sum += ktmp*ktmp;
     }
 
     // compute kperp (if it is non evanescent) OR
     // frequency from kperp^2 and sum of (kparallel+G)^2
     LOOP_OVER_DIRECTIONS(v.dim, dd) {
-      if ((float(eig_vol.in_direction(dd)) != float(v.in_direction(dd))) &&
+      if ((float(eig_vol.in_direction(dd)) == 0) &&
           (boundaries[High][dd] == Periodic && boundaries[Low][dd] == Periodic)) {
```
stevengj, line 614: Probably would be better to call `normal_direction` to get the perpendicular direction?

stevengj: i.e. `eig_vol.normal_direction()`.

stevengj: (also the cell doesn't need to be periodic in that direction)

src/mpb.cpp
```diff
@@ -596,22 +596,21 @@ void *fields::get_eigenmode(double frequency, direction d, const volume where, c
...
-      int m = 0;
-      if ((float(eig_vol.in_direction(dd)) == float(v.in_direction(dd))) &&
-          (boundaries[High][dd] == Periodic && boundaries[Low][dd] == Periodic)) {
-        m = dp->get_g()[dd - X];
-        ktmp = kpoint.in_direction(dd) + m/v.in_direction(dd);
-        k2sum += ktmp*ktmp;
-      }
+      m = dp->get_g()[dd - X];
+      ktmp = kpoint.in_direction(dd);
+      if ((m != 0) &&
```
stevengj, line 604: I think this should just be:
```cxx
if (eig_vol.in_direction(dd) != 0)
```

src/mpb.cpp
```diff
@@ -596,23 +596,20 @@ void *fields::get_eigenmode(double frequency, direction d, const volume where, c
...
+        ktmp += m/eig_vol.in_direction(dd);
+      k2sum += ktmp*ktmp;
     }
 
     // compute kperp (if it is non evanescent) OR
     // frequency from kperp^2 and sum of (kparallel+G)^2
     LOOP_OVER_DIRECTIONS(v.dim, dd) {
-      if ((float(eig_vol.in_direction(dd)) != float(v.in_direction(dd))) &&
-          (boundaries[High][dd] == Periodic && boundaries[Low][dd] == Periodic)) {
+      if (eig_vol.normal_direction() == dd) {
```
stevengj, line 612: You don't need a loop at all.  Just do:
```cxx
{
    direction dd = eig_vol.in_direction(dd);
   ...
}
```

oskooi:
This is ready to be merged.

PR #1331, opened 2020-08-27
Version += git revision

### LINKED ISSUE #1317: append git commit to version number for alpha releases
When the version number contains "alpha", it would be nice if the `configure` script appended the first 6 digits of the commit to the version number (assuming the source dir is a git repo).

cc @RobinD42

RobinD42:
Closes #1317 

Add the git revision to the version number for 'alpha' builds.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,6 +1,6 @@
 # Process this file with autoconf to produce a configure script.
 
-AC_INIT(meep, m4_esyscmd([version.sh 1.16.0 alpha]))
+AC_INIT(meep, m4_esyscmd([./version.sh 1.16.0 alpha]))
 AC_CONFIG_SRCDIR(src/step.cpp)
 
 # Shared-library version number; indicates api compatibility, and is
--- a/configure.ac
+++ b/configure.ac
@@ -1,5 +1,6 @@
 # Process this file with autoconf to produce a configure script.
-AC_INIT(meep, 1.16.0-alpha)
+
+AC_INIT(meep, m4_esyscmd([version.sh 1.16.0 alpha]))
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -1,5 +1,6 @@
 # Process this file with autoconf to produce a configure script.
-AC_INIT(meep, 1.16.0-alpha)
+
+AC_INIT(meep, m4_esyscmd([./version.sh 1.16.0 alpha]))
```
stevengj, line 3: won't work for vpath builds

stevengj: I think just `$srcdir/version.sh` though you have to be careful with quoting if the source directory contains spaces (though our Makefiles may break in that case anyway).

RobinD42: NB: The reason it works currently in vpath builds (on TravisCI at least) with using just `./version.sh` is because the script is run when `autoreconf` is run, which is before the the build folder is created and `../configure` is run.

RobinD42: And it looks like `$srcdir` does not have a value while `autoreconf` is running:

```
$ autoreconf --verbose --install --symlink --force
autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal --force -I m4
sh: /version.sh: No such file or directory
configure.ac:12: error: AC_INIT should be called with package and version arguments
/usr/local/Cellar/automake/1.16.2/share/aclocal-1.16/init.m4:171: AM_INIT_AUTOMAKE is expanded from...
configure.ac:12: the top level
autom4te: /usr/bin/m4 failed with exit status: 1
aclocal: error: echo failed with exit status: 1
autoreconf: aclocal failed with exit status: 1
```

PR #1339, opened 2020-09-01
launching diffracted planewave using eigenmode source

oskooi:
Extends the `DiffractedPlanewave` feature from #1316 to the eigenmode source as described in #291:[comment](https://github.com/NanoComp/meep/issues/291#issuecomment-672444352). Also includes documentation.

As a validation, the following test involves launching a diffracted planewave for three different diffraction orders and verifying that the resulting field profiles are as expected.
```py
import meep as mp
import numpy as np

import matplotlib
matplotlib.use('agg')

resolution = 25 # pixels/μm

cell_size = mp.Vector3(14,14)

pml_layers = [mp.PML(thickness=2)]

fsrc = 1.0  # frequency   
order = -5  # diffraction order   

sources = [mp.EigenModeSource(src=mp.ContinuousSource(fsrc,is_integrated=True),
                              center=mp.Vector3(),
	                      size=mp.Vector3(y=14),
                              eig_band=mp.DiffractedPlanewave((0,order,0),mp.Vector3(0,1,0),1,0))]

                    resolution=resolution,
                    boundary_layers=pml_layers,
                    sources=sources,
                    symmetries=[mp.Mirror(mp.Y)] if order == 0 else [])

sim.run(until=20)
sim.plot2D(output_plane=mp.Volume(center=mp.Vector3(), size=mp.Vector3(10,10)),
           fields=mp.Ez)
plt.show()
```

**1. `order = 4`**

diffpw-source-order4

**2. `order = 0`**

diffpw-source-order0

**3. `order = -5`**

diffpw-source-order-5

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -99,9 +99,31 @@ def py_v3_to_vec(dims, iterable, is_cylindrical=False):
     else:
         raise ValueError("Invalid dimensions in Volume: {}".format(dims))
 
+def bands_to_diffractedplanewave(where,bands):
+    if bands.axis is None:
+        if where.in_direction(mp.X) != 0:
+            axis = np.array([1, 0, 0], dtype=np.float64)
+        elif where.in_direction(mp.Y) != 0:
+            axis = np.array([0, 1, 0], dtype=np.float64)
+        elif where.in_direction(mp.Z) != 0:
+            axis = np.array([0, 0, 1], dtype=np.float64)
+        else:
+            raise ValueError("axis parameter of DiffractedPlanewave must be a non-zero Vector3")
+    elif isinstance(bands.axis,mp.Vector3):
+        axis = np.array([bands.axis.x, bands.axis.y, bands.axis.z], dtype=np.float64)
+    else:
+        raise TypeError("axis parameter of DiffractedPlanewave must be a Vector3")
+    diffractedplanewave_args = [
+        np.array(bands.g, dtype=np.intc),
+        axis,
+        bands.s * 1.0,
+        bands.p * 1.0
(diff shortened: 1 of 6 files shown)


python/simulation.py
```diff
@@ -2284,10 +2312,16 @@ def add_source(self, src):
             ]
             add_eig_src = functools.partial(self.fields.add_eigenmode_source, *add_eig_src_args)
 
-            if src.amp_func is None:
-                add_eig_src()
+            if isinstance(src.eig_band, DiffractedPlanewave):
+                if src.amp_func is None:
+                    add_eig_src(lambda v: 1.0, diffractedplanewave)
+                else:
+                    add_eig_src(src.amp_func, diffractedplanewave)
```
stevengj, line 2319: You should just be able to pass `src.amp_func` even if it is `None`, if you [edit line 699 of meep.i](https://github.com/NanoComp/meep/blob/98a49d09863c57062df2f239123bb0b4587decc1/python/meep.i#L699 to
```
$1 = PyCallable_Check($input) || $input == Py_None;
```
(then you don't even need the `if src.amp_func is None`)

stevengj: You'll also need to change [this typemap](https://github.com/NanoComp/meep/blob/98a49d09863c57062df2f239123bb0b4587decc1/python/meep.i#L703) to:
```
%typemap(in) std::complex<double> (*)(const meep::vec &) {
    if ($input) {
        $1 = py_amp_func_wrap;
        py_amp_func = $input;
        Py_INCREF(py_amp_func);
    }
    else
        $1 = NULL;
}
```

PR #1346, opened 2020-09-04
expand size of PML conductivity arrays to prevent out of bounds error

### LINKED ISSUE #1340: out of bounds error in step_db.cpp
[`meep/tests/stress_tensor.cpp`](https://github.com/NanoComp/meep/blob/master/tests/stress_tensor.cpp) is triggering an out of bounds error:

```
meep/src/step_db.cpp:114:7: runtime error: index 5 out of bounds for type 'double *[5]'
```

with the following stack trace:

```
in meep::fields_chunk::step_db(meep::field_type) meep/src/step_db.cpp:114:7
in meep::fields::step_db(meep::field_type) meep/src/step_db.cpp:35:22
in meep::fields::step() meep/src/step.cpp:63:3
in main meep/tests/stress_tensor.cpp:76:7
```

The offending line in `step_db.cpp` is:

https://github.com/NanoComp/meep/blob/a3fdff0037980151aead3677eb1f8246f7a2f70c/src/step_db.cpp#L114-L117

`dsig` which is of type `direction` is set to `NO_DIRECTION` (since it is equal to `5` in the call) which may be causing a problem in accessing the array values `s->sig[dig]`, `s->kap[dsig]`, and `s->siginv[dsig]`.

https://github.com/NanoComp/meep/blob/a3fdff0037980151aead3677eb1f8246f7a2f70c/src/meep/vec.hpp#L79

oskooi:
Fixes #1340.

### DIFF
--- a/src/structure.cpp
+++ b/src/structure.cpp
@@ -619,7 +619,7 @@ structure_chunk::~structure_chunk() {
     delete[] chi2[c];
     delete[] chi3[c];
   }
-  for (int d = 0; d < 6; ++d) {
+  FOR_DIRECTIONS(d) {
     delete[] sig[d];
     delete[] kap[d];
     delete[] siginv[d];
@@ -820,7 +820,7 @@ structure_chunk::structure_chunk(const structure_chunk *o) : v(o->v) {
   }
   condinv_stale = o->condinv_stale;
   // Allocate the PML conductivity arrays:
-  FOR_DIRECTIONS(d) {
+  for (int d = 0; d < 6; ++d) {
     sig[d] = NULL;
     kap[d] = NULL;
     siginv[d] = NULL;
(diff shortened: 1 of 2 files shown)


src/structure.cpp
```diff
@@ -619,7 +619,7 @@ structure_chunk::~structure_chunk() {
     delete[] chi2[c];
     delete[] chi3[c];
   }
-  FOR_DIRECTIONS(d) {
```
stevengj, line 622: This line can stay the same — only valid directions will ever be allocated.

stevengj:
Also need to change here: https://github.com/NanoComp/meep/blob/5ad4b8407084a839133e92a7638bb6d94b140b38/src/structure.cpp#L823

src/structure.cpp
oskooi: Fixed.

oskooi:
This is ready to be merged.

PR #1347, opened 2020-09-06
clarify computation of reflection coefficient S11 in directional-coupler tutorial

### LINKED ISSUE #1342: clarify S-matrix computation in directional-coupler calculation
Might be nice to update [this tutorial](https://meep.readthedocs.io/en/latest/Python_Tutorials/GDSII_Import/) to explictly calculate the S-matrix elements like
```py
S12 = p2_coeff / p1_coeff
p2_trans = abs(S12)**2
```
etcetera, and to clarify why the eigenmode-coefficient feature makes it unnecessary to do a separate normalization run (as long as port 1 is strictly to the right of the source — you probably want it to be at least a couple of pixels to the right to avoid overlaps due to discretization).

If you are computing the reflection coefficient S11, then you will see a "noise floor" below which you cannot measure reflection in this way, due to the slight discretization error in the eigenmode-coefficient extraction — in that case, you would have to do a separate normalization run and subtract the Fourier-transformed fields in port 1 (as we do in flux calculations) in order to accurately extract a tiny backscattered field.

oskooi:
Closes #1342.

Adds a new subsection "When computing the reflection coefficient S11, is it necessary to perform a separate normalization run to obtain the incident fields?" to [Tutorial/GDSII Import/S-Parameters of a Directional Coupler](https://meep.readthedocs.io/en/latest/Python_Tutorials/GDSII_Import/#s-parameters-of-a-directional-coupler) based on the results and discussion in #1342.


stevengj:
I would show the subtraction example in the directional-coupler case.   

In the straight-waveguide case, the reflection is zero *even at a finite resolution* (because the discrete system is still reflectionless).   So what you are seeing are just floating-point roundoff errors, which are on the order of (machine precision)² ≈ 10⁻³⁰.

stevengj:
Note that the source of the "noise floor" here is the mismatch in the discretization between Meep and MPB (we are computing the overlap with slightly the "wrong modes" for the discrete system at a finite resolution).  This is discussed in paragraph 3 of section 4.2.2 of https://arxiv.org/abs/1301.5366

stevengj:
Note that S11 is the complex reflection coefficient, whereas reflectance is |S11|².

oskooi:
When subtracting the incident fields in the directional-coupler case, the reflectance |S11|² is consistently 0 at all resolutions. This is somewhat unexpected because in the straight waveguide case, the reflectance was not exactly 0 but ~10<sup>-32</sup>.


Here is the output from the script below for resolutions in `seq 50 25 150`. The first column is resolution, second column is |S11|², and third column is |S12|².

```
 50, 0.0000000000000000, 0.0000000419820603
 75, 0.0000000000000000, 0.0000000695924934
 100, 0.0000000000000000, 0.0000001426181377
 125, 0.0000000000000000, 0.0000000089120082
 150, 0.0000000000000000, 0.0000000099710251
```

```py
    if args.norm:
                            cell_size=cell.size,
                            boundary_layers=[mp.PML(dpml)],
                            sources=sources,
                            geometry=geometry)

        mode1 = sim.add_mode_monitor(fcen, 0, 1, mp.ModeRegion(volume=p1))
        input_data = sim.get_flux_data(mode1)
        input_flux = mp.get_fluxes(mode1)[0]

        sim.reset_meep()

                        cell_size=cell.size,
                        boundary_layers=[mp.PML(dpml)],
                        sources=sources,
                        geometry=geometry)

    mode1 = sim.add_mode_monitor(fcen, 0, 1, mp.ModeRegion(volume=p1))
    mode2 = sim.add_mode_monitor(fcen, 0, 1, mp.ModeRegion(volume=p2))
    mode3 = sim.add_mode_monitor(fcen, 0, 1, mp.ModeRegion(volume=p3))
    mode4 = sim.add_mode_monitor(fcen, 0, 1, mp.ModeRegion(volume=p4))

    if args.norm:
        sim.load_minus_flux_data(mode1, input_data)

    eig_bands = [1]
    eig_parity = mp.EVEN_Y+mp.ODD_Z

    # mode coefficients (S parameters)  
    S11 = sim.get_eigenmode_coefficients(mode1,eig_bands,eig_parity).alpha[0,0,1]
    S12 = sim.get_eigenmode_coefficients(mode2,eig_bands,eig_parity).alpha[0,0,1]

    if not args.norm:
        input_flux = abs(sim.get_eigenmode_coefficients(mode1,eig_bands,eig_parity).alpha[0,0,0])**2

    # reflectance
    R11 = abs(S11)**2/input_flux
    R12 = abs(S12)**2/input_flux

    print("trans:, {}, {:.16f}, {:.16f}".format(args.res,R11,R12))
```

stevengj:
1. To be clear, the "noise floor" only applies at a fixed resolution — as you increase resolution, the discretization error in the mode-coefficient calculation goes away, and S11 should approach the true reflection.

2. You would do the normalization approach in order to decrease the resolution required to resolve a small reflection coefficient.   In this example, a resolution of 100 suffices for S21 but you need 2-3 times that for S11 — with a normalization run, probably you could get away with a smaller resolution.   (You still won't be able to resolve reflections on the order of "machine precision", because floating-point precision will spoil the normalization approach at that level).

3. For many practical devices, people don't care about reflectance less than 40dB (e.g. telecom multi-path interference tolerances).    But there will be other applications, especially theoretical studies, where you are trying to resolve much smallre reflections.

PR #1348, opened 2020-09-07
API generator updates

### LINKED ISSUE #1318: some api documentation generation improvements
From Ardavan:

> I noticed a few minor improvements we can make to the generate_py_api.py
> which required me to manually edit Python_User_Interface.md:
> 
> - in the entry for "EigenModeSource" and "GaussianBeamSource" in
> meep/python/source.py, the line "component=mp.ALL_COMPONENTS" is being
> replaced with "component=20" (it would be good to preserve the original
> "mp.ALL_COMPONENTS"). Same for "direction=mp.AUTOMATIC" and
> "eig_kpoint=mp.Vector3()"
> 
> - in the __init__ code block of the PML object, the line
> "pml_profile=lambda u: u * u)" in meep/python/simulation.py is showing
> up as "pml_profile=<function PML.<lambda> at 0x7f4f93f958c8>)" (it would
> be good to preserve the text showing the inline function definition)
> 
> - in some cases (i.e., "add_mode_monitor(self, *args, **kwargs):"),
> an unnecessary blank space is inserted after the colon
> 
> - the line "See help(type(self)) for accurate signature." was
> changed to "See `help(type(self))` for accurate signature." (i.e.,
> backticks were necessary to place the text inside a code block but I
> couldn't find where this line actually appears in either
> meep/python/*.py, Python_User_Interface.md.in, or generate_py_api.py)

RobinD42:
Fixes #1318 (mostly)

* Remove some extra whitespace in the function and method templates

* Add a docstring for some `__init__` methods that don't have one, so they do not inherit default one from `object`

* Use Python's `ast` module (Abstract Syntax Tree) to extract the text for default parameters from the source code instead of using the evaluated value as `inspect` does. This means that the default values in the documentation will be more meaningful, for example: `mp.ALL_COMPONENTS` instead of `20`. 

This extraction via the `ast` works well for simple values and names, but some possible types of default values can result a complex syntax tree, so the code will fall back to using the `inspect` module when it runs into something that is not currently supported, rather than trying to implement something that might end up being fragile and more effort to maintain than it's worth. The most pertinent for our source code are:

* Lambdas. There is currently one example of using a lambda as a default parameter value, in `PML.__init__`. It results in an AST tree with nodes for the lambda's parameters and also a tree for the expression. So rather than trying to support any possible expression there I just changed the source and added a `DefaultPMLProfile` global which could then be used as the default value for the parameter, resulting in just one node in the AST.

* Function calls. This type is what you get when the default parameter is a call to some callable object, like `wedge_start=Vector3()`. I'm letting this one fall back to use `inspect` to get the value of the parameter. As long as the object being created there has a good `__repr__` method then this isn't too bad. But it might make sense in the future to also add something like `DefaultVector3 = Vector3()` and using that everywhere as the default value for the parameters.

* Math expressions. There are some cases where the default value is an expression, like `wedge_angle=2 * math.pi`. WIth some more work at least some of these could be supported, but for now I'm letting them fall back to being evaluated by `inspect`. Of course, adding globals for some of these for use as the default parameter values would also work with things as they are currently in this PR.

### DIFF
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -127,7 +127,7 @@ def __init__(self,
              force_all_components=False,
              split_chunks_evenly=True,
              chunk_layout=None,
-             collect_stats=False): 
+             collect_stats=False):
 ```
 
 <div class="method_docstring" markdown="1">
@@ -179,12 +179,12 @@ Python. `Vector3` is a `meep` class.
   as the default material, use the `material_function` keyword argument (below).
 
 + **`material_function` [ function ]** — A Python function that takes a `Vector3`
-  and returns a `Medium`. See also [Material Function](#medium).
+  and returns a `Medium`. See also [Material Function](#material-function).
   Defaults to `None`.
 
 + **`epsilon_func` [ function ]** — A Python function that takes a `Vector3` and
   returns the dielectric constant at that point. See also [Material
-  Function](#medium). Defaults to `None`.
+  Function](#material-function). Defaults to `None`.
 
 + **`epsilon_input_file` [`string`]** — If this string is not empty (the default),
   then it should be the name of an HDF5 file whose first/only dataset defines a
@@ -254,7 +254,7 @@ Python. `Vector3` is a `meep` class.
 
 + **`eps_averaging` [`boolean`]** — If `True` (the default), then [subpixel
   averaging](Subpixel_Smoothing.md) is used when initializing the dielectric
-  function. For simulations involving a [material function](#medium),
+  function. For simulations involving a [material function](#material-function),
   `eps_averaging` is `False` (the default) and must be
   [enabled](Subpixel_Smoothing.md#enabling-averaging-for-material-function) in
   which case the input variables `subpixel_maxeval` (default 10<sup>4</sup>) and
@@ -368,7 +368,7 @@ use. See also [SWIG Wrappers](#swig-wrappers).
 <div class="class_members" markdown="1">
 
 ```python
-def run(self, *step_funcs, **kwargs): 
+def run(self):
 def run(step_functions..., until=condition/time):
 def run(step_functions..., until_after_sources=condition/time):
 ```
@@ -413,7 +413,7 @@ or set the output folder, with these methods of the `Simulation` class:
 <div class="class_members" markdown="1">
 
 ```python
-def get_filename_prefix(self): 
+def get_filename_prefix(self):
 ```
 
 <div class="method_docstring" markdown="1">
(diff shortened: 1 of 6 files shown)


doc/docs/Python_User_Interface.md
```diff
@@ -1081,7 +1081,7 @@ Once you have called `add_flux`, the Fourier transforms of the fields are accumu
 <div class="class_members" markdown="1">
 
 ```python
-def display_fluxes(self, *fluxes): 
+def display_fluxes(self):
```
oskooi, line 1084: The function `display_fluxes` has an input parameter `*fluxes` that is described in the documentation which should therefore not be removed.

doc/docs/Python_User_Interface.md
```diff
@@ -3561,7 +3560,7 @@ Rather than writing a brand-new step function every time something a bit differe
 <a id="combine_step_funcs"></a>
 
 ```python
-def combine_step_funcs(*step_funcs): 
+def combine_step_funcs():
```
oskooi, line 3563: `*step_funcs` is the only input parameter to `combine_step_funcs` and should not be removed from the documentation.

doc/docs/Python_User_Interface.md
```diff
@@ -7022,26 +7007,13 @@ value will persist across runs within the same script.
...
-```
-
-<div class="function_docstring" markdown="1">
-
-Given a number `v`, specify the degree of Meep's output: `0` is quiet mode, `1` (the
-default) is ordinary output, `2` is extra debugging output, and `3` is all debugging
-output.
-
-</div>
+@@ verbosity @@
```
oskooi, line 7010: Seems the id tag for `verbosity` is not being recognized or is misplaced in `Python_User_Interace.md`?

doc/docs/Python_User_Interface.md
```diff
@@ -1471,7 +1471,7 @@ Once you have called `add_energy`, the Fourier transforms of the fields are accu
 <div class="class_members" markdown="1">
 
 ```python
-def display_electric_energy(self, *energys): 
+def display_electric_energy(self):
```
oskooi, line 1474: Again, `*energys` should not be removed from the function header. Same for the related functions `display_magnetic_energy` and `display_total_energy`.

doc/docs/Python_User_Interface.md
```diff
@@ -1697,7 +1697,7 @@ Once you have called `add_force`, the Fourier transforms of the fields are accum
 <div class="class_members" markdown="1">
 
 ```python
-def display_forces(self, *forces): 
+def display_forces(self):
```
oskooi, line 1700: `*forces` should not be removed from the function header.

oskooi (changes requested):

doc/docs/Python_User_Interface.md
RobinD42: It's because it's not a function any more. I added tags for the `Verbosity` class in my other PR, and added text to the docstring to mention that there is an instance of the class named `verbosity`

oskooi (approved):

PR #1359, opened 2020-09-14
tweaks to graphene FAQ

stevengj:
@oskooi, isn't graphene described by an anistropic conductivity, i.e. it is only conductive in the plane of the graphene?  That's what we [wrote in the paper](http://math.mit.edu/~stevenj/papers/MillerIl17.pdf) with @odmiller.

Also, I don't think it's true that the pixel size must be comparable to an atomic thickness.   As long as the layer thickness is much smaller than the wavelength, the solution to Maxwell's equation should be nearly that of a delta-function conductivity.


stevengj:
It might be nice to make a little study comparing differing methods of modeling the delta function:

1. 1 pixel thick, resolution, amplitude ~ 1/Δx

2. thickness = sqrt(Δx), amplitude = 1/thickness

3. method (1), but with some subpixel smoothing

PR #1368, opened 2020-09-22
minor fixes and tweaks, and small doc updates

oskooi:
Should be merged *before* #1367.

- fixes a bug in `scale_near2far_fields`
- adds a ValueError to `get_dft_array` if the `num_freq` argument exceeds the maximum index of the dft array
- converts unicode to Latex symbols
- minor tweaks to docs


python/simulation.py
```diff
@@ -3201,6 +3206,9 @@ def get_dft_array(self, dft_obj, component, num_freq):
         else:
             dft_swigobj = dft_obj
 
+        if num_freq > dft_obj.nfreqs-1:
+            raise ValueError("get_dft_array: frequency index {} exceeds max index {}".format(num_freq,dft_obj.nfreqs-1))
```
stevengj, line 3210: Arguably, this should be fixed in the C++ code in `process_dft_component`

oskooi: This `ValueError` exception has been removed from this PR.

PR #1373, opened 2020-09-23
abort get_dft_array if user-supplied frequency index is invalid

oskooi:
Currently, if the frequency index parameter `num_freq` of [`get_dft_array`](https://meep.readthedocs.io/en/latest/Python_User_Interface/#array-slices) is outside the valid range (e.g., if `num_freq` is larger than the size of the array), the function returns an array of zero values which is incorrect. This PR adds a check in `dft_chunk::process_dft_component` to verify that `num_freq` is valid, otherwise it aborts with an error.


stevengj:
(Following this comment: https://github.com/NanoComp/meep/pull/1368#discussion_r493118149)

PR #1387, opened 2020-10-06
add notes to LED tutorial on cutoff value for line source with cosine amplitude function

oskooi:
This PR provides some additional details to the [LED tutorial](https://meep.readthedocs.io/en/latest/Python_Tutorials/Custom_Source/) for the case involving a line source with a cosine amplitude function.


doc/docs/Python_Tutorials/Custom_Source.md
```diff
@@ -304,7 +304,7 @@ if mp.am_master():
...
 
-Method 3 requires a convergence check in which $M$ (`nsrc` in the script) is repeatedly doubled until the change in the results are within a desired tolerance of e.g., < 1%. For this example, $M=15$ was found to be sufficient.
+Method 3 requires a convergence check in which $M$ (`nsrc` in the script) is repeatedly doubled until the change in the results are within a desired tolerance of e.g., < 1%. For this example, $M=11$ was found to be sufficient. Note that because a line source with a cosine amplitude function is analogous to generating a planewave at a discrete angle, at each frequency $\omega$ there exists a cutoff $M$ beyond which there are *no* propagating planewaves. The cutoff $M$ can be computed analytically using the grating equation: $\sqrt{\omega^2n^2 - \left(k_x+\frac{M\pi}{L}\right)^2} = 0$, where $n$ is the refractive index of the source medium and $k_x$ is the Bloch-periodic wavevector in the $x$ direction. In this example, $\omega=2\pi$, $L=1.5$, $k_x=0$, and $n=3.45$ for which the cutoff $M$ is 11.
```
stevengj, line 307: Technically, what happens is that for M > 11 the cosine source produces evanescent waves *in the material*, but it scatters into other Fourier components (including propagating waves) once it hits the grating.   So, it still produces propagating waves, but the amplitude of the propagating waves (and hence the contribution to the power) decreases exponentially for M > 11 (as the coupling between the source and the grating decreases).   (This only happens for periodic boundary conditions where the line source extends for the whole period.)

stevengj: Might be helpful to show the convergence plot with M

oskooi: Fixed.

stevengj:
Plot might look clearer on a semilogy scale?

I'm also not sure why the error is "saturating" at such a large value?

oskooi:
It is not the error that is being plotted but rather the L2-norm of the error (since the normalized flux is over a range of frequencies). The figure below shows the flux spectra for the point-dipole (N=75) and cosine (M=15) source which has an L2-norm error of `0.256966`.

method2_3_res50_nfreq500_nsrc15_normalized_spectra_

The plot above and the L2-norm error are computed using the following script:
```py
import matplotlib
matplotlib.use('agg')

import numpy as np

res = 50
nfreq = 500
ndipole = 75

method2_f0 = np.load('method2_flat_res{}_nfreq{}_ndipole{}.npz'.format(res,nfreq,ndipole))
method2_f1 = np.load('method2_textured_res{}_nfreq{}_ndipole{}.npz'.format(res,nfreq,ndipole))

method2_freq = method2_f0['freqs']
method2_f0_mean = np.mean(method2_f0['fluxes'][:,::1],axis=1)
method2_f1_mean = np.mean(method2_f1['fluxes'][:,::1],axis=1)

res = 50
nsrc = 15
method3_f0 = np.load('method3_flat_res{}_nfreq{}_nsrc{}.npz'.format(res,nfreq,nsrc))
method3_f1 = np.load('method3_textured_res{}_nfreq{}_nsrc{}.npz'.format(res,nfreq,nsrc))

method3_freq = method3_f0['freqs']
method3_f0_mean = np.mean(method3_f0['fluxes'][:,:15],axis=1)
method3_f1_mean = np.mean(method3_f1['fluxes'][:,:15],axis=1)

method2_flux = method2_f1_mean/method2_f0_mean
method3_flux = method3_f1_mean/method3_f0_mean

plt.figure()
plt.semilogy(method2_freq,method2_flux,'b-',label='Point Source (ndipole={})'.format(ndipole))
plt.semilogy(method3_freq,method3_flux,'r-',label='Line Source (nsrc={})'.format(nsrc))
plt.xlabel('frequency')
plt.ylabel('normalized flux')
plt.legend()
plt.savefig('norm_flux.png'.format(res,nfreq,nsrc),dpi=200,bbox_inches='tight')

rel_err = (method3_flux-method2_flux)/method2_flux
print("norm:, {}".format(np.linalg.norm(rel_err))
```

stevengj:
This should be:
```
rel_err = np.linalg.norm(method3_flux-method2_flux)/np.linalg.norm(method2_flux)
```
(i.e. you should compute the norms *before* dividing)

oskooi:
The convergence plot for the relative error computed from the norm before dividing: 

method3_costerms_norm

It shows similar features to the plot which is used `np.linalg.norm((method3_flux-method2_flux)/method2_flux)` but the saturation point is smaller by an order of magnitude.

oskooi:
For reference (in case of future discussions), the following is a plot of the difference in the normalized flux of the cosine source (method 3) and dipole source (method 2) from the results in the [figure above](https://github.com/NanoComp/meep/pull/1387#issuecomment-705823307) on a *linear* scale:

flux_diff_method3_method2

Note that the difference in the flux at certain frequencies is not small which is why computing `rel_err = np.linalg.norm(method3_flux-method2_flux)/np.linalg.norm(method2_flux)` yields values ~0.03. The use of a log-scale in the figure above makes the discrepancies in the flux from the two methods appear smaller than it really is.

PR #1388, opened 2020-10-06
Verbosity improvements

### LINKED ISSUE #1362: more verbosity tweaks
Possible changes:

1. Allow the verbosity singleton class to separately set Meep and MPB verbosity, e.g. with ` verbosity.mpb = 0`.

2. When calling MPB via Meep (in src/mpb.cpp), temporarily set the MPB verbosity to `meep::verbosity - 1`.

3. In MPB, make all of the LINMIN warnings only appear for `verbosity > 1`.

RobinD42:
Fixes #1362 
See also https://github.com/NanoComp/mpb/pull/125

* Make it possible to access the individual verbosity levels by name in the `Verbosity` class
* Switch verbosity checks in Python code to `verbosity.meep` or `verbosity.mpb` 
* Make MPB's version numbers be accessible from Python in the `meep.mpb` package
* When calling mpb code from meep code set `mpb_verbosity` to `meep::verbosity - 1`
* Switch most `mpb_verbosity` checks to `>=1`, to be closer to the original intent of the new verbosity functionality
* Added a unit test module for testing the `Verbosity` class

### DIFF
--- a/libpympb/pympb.cpp
+++ b/libpympb/pympb.cpp
@@ -696,12 +696,12 @@ void mode_solver::init(int p, bool reset_fields, geometric_object_list geometry,
   n[1] = grid_size.y;
   n[2] = grid_size.z;
 
-  if (target_freq != 0.0 && mpb_verbosity >= 2) { meep::master_printf("Target frequency is %g\n", target_freq); }
+  if (target_freq != 0.0 && mpb_verbosity >= 1) { meep::master_printf("Target frequency is %g\n", target_freq); }
 
   int true_rank = n[2] > 1 ? 3 : (n[1] > 1 ? 2 : 1);
   if (true_rank < dimensions) { dimensions = true_rank; }
   else if (true_rank > dimensions) {
-    if (mpb_verbosity >= 2)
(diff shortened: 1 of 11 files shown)


RobinD42:
**NOTE**: After playing with and testing this change for a couple days I'm not so sure that the feature of setting `mpb_verbosity` to `meep::verbosity - 1` won't cause more confusion than it's worth. Maybe it won't happen often in real life usage, but it took me several minutes to realize why this wasn't working, and I'm the one who made the change:

```python
verbosity.meep = 2
verbosity.mpb = 3
a_solver.run_te()
```
Since the `mpb_verbosity` was set to 1, I didn't see the messages I was expecting.

stevengj:
> After playing with and testing this change for a couple days I'm not so sure that the feature of setting mpb_verbosity to meep::verbosity - 1 won't cause more confusion than it's worth.

It should only be done in `src/mpb.cpp`, i.e. when MPB is called from Meep, and not when you are calling functions from pympb like `run_te()`.

RobinD42:
Okay. I've reverted the applicable changes in `pympb.cpp`

PR #1396, opened 2020-10-08
Ignore some C++ str() methods to avoid SWIG warnings

### LINKED ISSUE #1391: 'str' conflicts with built-in
Do something about the swig warnings like this:

```
src/meep/vec.hpp:872: Warning 321: 'str' conflicts with a built-in name in python
```

IIRC they are not needed in Python code so a `%ignore` is probably sufficient.

RobinD42:
Fixes #1391

### DIFF
--- a/python/vec.i
+++ b/python/vec.i
@@ -70,6 +70,12 @@
 %rename(vec_dim_val) meep::vec::vec(ndim, double);
 %rename(vec_dim) meep::vec::vec(ndim);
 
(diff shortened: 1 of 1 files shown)


RobinD42:
Oops I just saw where it is actually used in Python. I'll redo this with a `%rename` instead.

RobinD42:
False alarm. It's the C++ str() being used inside of some `%extend` additions, so they are not being used directly from Python afterall.

Issue #1299, opened 2020-07-23
gray_indicator() returns incorrect type


Chuban:
A typo in gray_indicator() ([python/adjoint/filters.py](https://github.com/NanoComp/meep/blob/7ff6c89de264408723891831a96ca67e53ae2115/python/adjoint/filters.py#L863)) is causing it to return an array. The doc string indicates that this was intended to return a float. I suspect that
`return np.mean(4 * x.flatten()) * (1-x.flatten()) * 100 `
should be
`return np.mean(4 * x.flatten() * (1-x.flatten())) * 100 `

stevengj:
That sounds right.  Care to contribute a PR?

PR #1398, opened 2020-10-09
WIP: Adjoint poynting flux

### LINKED ISSUE #1307: Adjoint optimization through Poynting Flux
Is it possible to perform adjoint inverse design through the Poynting Flux over a certain range of frequencies? The **meep.adjoint.EigenmodeCoefficient** object has a **register_monitors** function to track the power across specified frequencies, but I'm unsure as to how one could use this to inverse design a device.

I tried to use it in this tutorial, where I added the following lines in the adjoint optimization tutorial:

frequencies = np.linspace(1/1.55 - 0.1, 1/1.55 + 0.1, 100)
TE_top.register_monitors(frequencies)

https://nbviewer.jupyter.org/github/NanoComp/meep/blob/master/python/examples/adjoint_optimization/01-Introduction.ipynb

I replaced the **TE0** variable with **TE0.monitor**, but this ended up giving me the following error:

AttributeError: 'DftFlux' object has no attribute 'register_monitors'.

smartalecH:
Closes #1307 and #1299.


smartalecH:
I'm slightly confused by the default behavior of the `get_dft_array()` routine. Note that the C implementation of `get_dft_array()` contains two basic steps: grab the DFT fields themselves (`process_dft_component()`) and format the data array with `collapse_array()`.

https://github.com/NanoComp/meep/blob/d29f21121b938fded5458a454aa927e336e42b83/src/dft.cpp#L1082-L1090

When the user specifies a volume other than a plane/volume (e.g. point or line) _and_ the user is pulling the interpolated fields from the center of each voxel, there's some odd behavior.

For a simple point, `process_dft_component()` returns a 2x2 array. This array is then passed to `collapse_array()`, which seems to grab the first point and throw the rest away.

In the case of a line monitor across 10 pixels, `process_dft_component()` returns a 10x2 array. Again, `collapse_array()` grabs the first column and throws the rest away.

Is this the expected behavior?

This makes it tricky when trying to use the indices of the DFT array data values for a broadband adjoint source (for both Poynting fluxes and basic Fourier fields).

smartalecH:
There are two options for placing the adjoint source in this case:

1. In the C++ code, *uncollapse* the jacobian passed from python, combine it with all the *uncollapsed* dft chunk data, and add the result is an index source.
2. In the python code, use the *collapsed* dft fields and the *collapsed* Jacobian to add a source using the `amp_data` functionality for any arbitrary source. The python source will automatically distribute the fields onto the correct yee grid points (assuming the interpolation is taken into account).

Option 1 is rather difficult -- I don't see a reliable way to uncollapse the Jacobian and map it to the dft fields (which are just stored as sequential indices in the C++ code). Option 2 requires some slight modifications to the `amp_data` routine, specifically with how the array is interpolated onto the grid.

I implemented option 2 for the broadband Poynting flux (10 frequency points) and got the following result:

Figure_1

The result isn't perfect (1%-10% error) but it's a good start. I've incorporated the interpolation weights into the adjoint scale factor (as we've discussed numerous times) but they only slightly improve the accuracy. Perhaps I still need to take into account the interpolation of the final source object and `amp_array` routine?

PR #1399, opened 2020-10-10
subpixel averaging for boundaries of variable-material objects

stevengj:
This PR implements subpixel averaging for the *boundaries* of objects with variable materials (e.g. user-defined material functions).    I realized that we can still handle this case accurately and efficiently, as long as the material function is *continuous inside the object*.

For example, here is a simulation of transmitted power through a cylinder of linearly-varying ε:
image

```py
def linear_eps(p):
    return 1 + 11 * (p.x + 0.4) / 0.8

def flux2(res):
    sim = mp.Simulation(cell_size=(7,1), resolution=res, boundary_layers=[mp.PML(direction=mp.X, thickness=0.5)],
                    k_point=mp.Vector3(0,0), symmetries=[mp.Mirror(direction=mp.Y)],
                    sources=[mp.Source(mp.GaussianSource(frequency=0.7, fwidth=0.2, is_integrated=True),
                                       component=mp.Ez, center=(-3,0), size=(0,1))],
                    geometry=[mp.Cylinder(radius=0.4, height=mp.inf, epsilon_func=linear_eps)])

    tran = sim.add_flux(0.7, 0.2, 1, mp.FluxRegion(center=(3,0), size=(0,1)))
    sim.run(until=200)
    return tran.flux()[0]
```

If I compute and plot the convergence with respect to resolution using this code:
```py
import numpy as np
r = [10,20,40,80,160,320,640]
f2 = np.asarray([flux2(r) for r in r])
r = np.asarray(r)
plt.loglog(r[0:-1], np.abs(f2[0:-1] - f2[-1]) / f2[-1], "ro-")
plt.loglog(r[0:-1], 30 / r[0:-1]**2, "k--")
plt.loglog(r[0:-1], 1 / r[0:-1], "b:")
plt.legend(["flux2 error", "$1/r^2$ reference", "$1/r$ reference"])
plt.xlabel("resolution $r$")
plt.ylabel("fractional error")
```
then **before** this PR I get first-order convergence due to the lack of subpixel averaging at the boundaries of the cylinder:
image
but **after** this PR I get second-order convergence:
image

### DIFF
--- a/src/meepgeom.cpp
+++ b/src/meepgeom.cpp
@@ -267,14 +267,13 @@ bool is_material_grid(material_type mt) {
 }
 bool is_material_grid(void *md) { return is_material_grid((material_type)md); }
 
+// return whether mt is spatially varying
 bool is_variable(material_type mt) {
   return (mt->which_subclass == material_data::MATERIAL_USER) ||
-         (mt->which_subclass == material_data::MATERIAL_GRID);
+         (mt->which_subclass == material_data::MATERIAL_GRID) ||
+         (mt->which_subclass == material_data::MATERIAL_FILE);
 }
 bool is_variable(void *md) { return is_variable((material_type)md); }
-
-bool is_file(material_type md) { return (md->which_subclass == material_data::MATERIAL_FILE); }
-bool is_file(void *md) { return is_file((material_type)md); }
 
 bool is_medium(material_type md, medium_struct **m) {
   if (md->which_subclass == material_data::MEDIUM) {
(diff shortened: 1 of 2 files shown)


smartalecH:
Once we debug the issues with the `default_material`, there is one more (sneaky) bug in the main material grid routine:

https://github.com/NanoComp/meep/blob/32788d7569887edfa1193211680e73d4bfac6e47/src/meepgeom.cpp#L325-L351

This piece of code is called whenever there is a material grid present in a tree. The first if statement traverses through the tree and will interpolate as needed. The second if statement is a special case, which is called if the tree is done being traversed and the `default_material` is a material grid.

We need to account for the final case, where the current point doesn't have any material grids (e.g. it could be another medium or the `default_material`).

it will require some debugging and trial and error.

src/meepgeom.cpp
```diff
@@ -914,18 +930,25 @@ void geom_epsilon::eff_chi1inv_matrix(meep::component c, symmetric_matrix *chi1i
...
+  // for variable materials with do_averaging == true, switch over to slow fallback integration method
+  if ((is_variable(mat) && mat->do_averaging) || (is_variable(mat_behind) && mat_behind->do_averaging)) {
+    fallback = true;
+    return;
   }
 
   /* check for trivial case of only one object/material */
+  if (material_type_equal(mat, mat_behind)) {
+    if (is_variable(mat)) eval_material_pt(mat, vec_to_vector3(v.center()));
+    goto trivial;
```
stevengj, line 942: In #1539, this will branch to the new averaging procedure if `mat` is a variable material.

PR #1383, opened 2020-09-27
offline computation of cost statistics

oskooi:
Closes #1150.

Following @stevengj's outline in #1150, this PR creates a new `choose_chunkdivision` function which does not belong to any class and is based on refactoring `structure::choose_chunkdivision`. All that is required to print the costs of each chunk *without* allocating the  `structure` and `fields` objects is to set the environment variable `MEEP_STATS` to an integer value. Note that the `MEEP_STATS` environment variable is currently used to print the fragment statistics as defined in `python/simulation.py`:

https://github.com/NanoComp/meep/blob/29a92c0ab074259fcfa66b9d09e6c6254032ce46/python/simulation.py#L1247

https://github.com/NanoComp/meep/blob/29a92c0ab074259fcfa66b9d09e6c6254032ce46/python/simulation.py#L1656-L1669

As a demonstration, here is the output for a test script involving 25 and 34 requested chunks:
```
$ MEEP_STATS=1 mpirun -np 25 python3 foo.py
Using MPI version 3.1, 25 processes
-----------
Initializing structure...
Warning: grid volume is not an integer number of pixels; cell size will be rounded to nearest pixel.
STATS:, aniso_eps:, 0
STATS:, anis_mu:, 0
STATS:, nonlinear:, 0
STATS:, susceptibility:, 21688389
STATS:, conductivity:, 20935276
STATS:, pml_1d:, 17031619
STATS:, pml_2d:, 0
STATS:, pml_3d:, 0
STATS:, dft:, 9920976000
STATS:, total_pixels:, 51248959
STATS:, num_cores:, 25
Splitting into 25 chunks by cost
CHUNK:,  0, 58678.791325
CHUNK:,  1, 58466.975063
CHUNK:,  2, 59117.904559
CHUNK:,  3, 58337.339266
CHUNK:,  4, 60545.863842
CHUNK:,  5, 58684.960462
CHUNK:,  6, 58684.960462
CHUNK:,  7, 59497.770752
CHUNK:,  8, 59497.770752
CHUNK:,  9, 60314.116779
CHUNK:, 10, 59175.846869
CHUNK:, 11, 59175.846869
CHUNK:, 12, 58974.471592
CHUNK:, 13, 58974.471592
CHUNK:, 14, 60448.832289
CHUNK:, 15, 58686.790644
CHUNK:, 16, 58686.790644
CHUNK:, 17, 59909.432557
CHUNK:, 18, 59912.967719
CHUNK:, 19, 59912.967719
[...]

PR #1400, opened 2020-10-10
Add surface area function to grid_volume class

oskooi:
This PR adds a new function `size_t grid_volume::surfacearea()` as originally proposed by @stevengj in #1069. The computation of the surface area is based on that of a simple cuboid. No distinction is made between counting only "outer" boundaries in periodic directions (as suggested in this [comment](https://github.com/NanoComp/meep/issues/1069#issuecomment-564351650)) since the boundary information is available from the `structure` or `fields` class. Unfortunately, this means the surface area results may not be accurate for computing the communication cost but at least it provides a rough estimate while keeping the implementation simple.

For offline computations (following #1383), the surface area of each chunk volume (in which PMLs are excluded) is output along with its cost.

### DIFF
--- a/src/vec.cpp
+++ b/src/vec.cpp
@@ -860,6 +860,16 @@ ivec grid_volume::iloc(component c, ptrdiff_t ind) const {
   return out + iyee_shift(c) + io;
 }
 
+size_t grid_volume::surface_area() const {
+  switch(dim) {
+    case Dcyl: return 2*(nr()+nz());
+    case D1: abort("grid_volume::surface_area() does not support 1d\n");
+    case D2: return 2*(nx()+ny());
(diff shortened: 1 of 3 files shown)


src/vec.cpp
```diff
@@ -860,6 +860,16 @@ ivec grid_volume::iloc(component c, ptrdiff_t ind) const {
   return out + iyee_shift(c) + io;
 }
 
+size_t grid_volume::surface_area() const {
+  switch(dim) {
+    case Dcyl: return 2*(nr()+nz());
+    case D1: abort("grid_volume::surface_area() does not support 1d\n");
```
stevengj, line 866: Shouldn't it just `return 2` in 1d?

oskooi: Fixed.

PR #1409, opened 2020-10-21
output timing statistics to CSV file

oskooi:
This PR adds a new function `fields::output_times` which outputs the timing statistics to an ~~HDF5~~ 
CSV file as an alternative to `fields::print_times` and `fields::time_spent_on` which either print to standard output or return the values.

### DIFF
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -547,6 +547,23 @@ other.
 
 </div>
 
+<a id="Simulation.output_times"></a>
+
+<div class="class_members" markdown="1">
+
+```python
+def output_times(self, fname=''):
+```
+
+<div class="method_docstring" markdown="1">
+
+Call after running a simulation to output to a file with filename `fname` the
+times spent on various types of work as CSV (comma separated values) with headers
+for each column.
+
+</div>
+
+</div>
 
 ### Field Computations
 
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -559,7 +559,7 @@ def output_times(self, fname):
 
 Call after running a simulation to output to a file with filename `fname` the
 times spent on various types of work as CSV (comma separated values) with headers
-for each column.
+for each column and one row per process.
 
 </div>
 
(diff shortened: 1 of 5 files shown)


stevengj:
For this sort of data, I would just output CSV (comma separated values), with headers for each column as needed.  It will be a lot easier to import into e.g. a spreadsheet.

oskooi:
The output has been changed to a CSV file. Python interface and documentation have also been added.

As a demonstration, here are the contents of the output file for the [direction coupler tutorial](https://meep.readthedocs.io/en/latest/Python_Tutorials/GDSII_Import/#s-parameters-of-a-directional-coupler) with six processors:

```
connecting chunks, time stepping, copying borders, communicating, outputting fields, Fourier transforming, MPB, getting farfields, everything else
0.0548074, 2.14366, 0, 1.07554, 0, 0.0533044, 0.0448153, 0, 0.174047
0.0303695, 1.35704, 0, 2.03147, 0, 0.000870471, 0.00304685, 0, 0.123424
0.0518182, 1.77557, 0, 1.51457, 0, 0.000991946, 0.00489046, 0, 0.198386
0.0410307, 1.76686, 0, 1.5286, 0, 0.00103788, 0.00480071, 0, 0.203897
0.0508296, 1.83852, 0, 1.44166, 0, 0.00105197, 0.00483134, 0, 0.209327
0.0525085, 1.65281, 0, 1.68401, 0, 0.0393224, 0.00472933, 0, 0.11285
```
The first row contains the headers for the type of work. The timings for each process are listed on six separate rows.

doc/docs/Python_User_Interface.md
```diff
@@ -547,6 +547,23 @@ other.
...
+
+```python
+def output_times(self, fname):
+```
+
+<div class="method_docstring" markdown="1">
+
+Call after running a simulation to output to a file with filename `fname` the
+times spent on various types of work as CSV (comma separated values) with headers
+for each column.
```
stevengj, line 562: and one row per process

stevengj:
Might be convenient to have a `time_spent_first` array mirroring `time_spent`.    In step.cpp, at the beginning of timestep 2, it would copy over the `time_spent` data` to `time_spent_first`.   Then, in output, you could optionally output the difference of the two, or output both.

PR #1414, opened 2020-10-28
documentation for optimization studies of parallel simulations

oskooi:


doc/docs/Parallel_Meep.md
```diff
@@ -53,6 +53,10 @@ However, there is an alternative strategy for parallelization. If you have many
 
 Meep also supports [thread-level parallelism](https://en.wikipedia.org/wiki/Task_parallelism) (i.e., multi-threading) on a single, shared-memory, multi-core machine for multi-frequency [near-to-far field](Python_User_Interface.md#near-to-far-field-spectra) computations. Meep does not currently use thread-level parallelism for the time stepping although this feature may be added in the future (see [Issue \#228](https://github.com/NanoComp/meep/issues/228)).
 
+### Random Initial Conditions for Optimization
+
+When running Meep simulations as part of an optimization study (e.g., via the [adjoint solver](Python_Tutorials/AdjointSolver.md)), every process stores the *same* copy of the optimization variables. The overhead due to this storage redundancy is minimal. For random initial conditions, the seed of the random number generator must be specified otherwise each process will have a different initial condition which will cause a crash.
```
stevengj, line 58: Be specific: it's only the seed of the random number generator used for the initial conditions, e.g. if you are initializing the design variables with `numpy.random.rand`, then you should call `numpy.random.seed(some number)` to set the same `numpy.random` seed on every process.

stevengj: Having every process run the same optimization algorithm on the same optimization variables is by far the easiest way to keep them all synchronized.

stevengj: The computational cost of the optimization algorithm and the storage of the design variables is negligible compared to those of the FDTD simulation, so duplicating these on every process is not normally a concern.

oskooi: Fixed.

Issue #1418, opened 2020-11-05
MPB test failures


stevengj:
We are getting test failures on a bunch of `mpb.py` tests.  I'm guessing it's due to NanoComp/mpb#127 by @thchr, which slightly changed the subpixel averaging in some cases.

@thchr, I thought your changes mostly affected averaging of user-specified material functions?   But it looks like we are getting changes even in tests like https://github.com/NanoComp/meep/blob/master/python/examples/mpb_diamond.py that only use geometric objects.

Before that MPB PR, the `mpb_diamond.py` script outputs:
```
Gap from band 2 (0.39676580457429544) to band 3 (0.44237076818414445), 10.869497311965496%
```
but after that PR I get slightly different results:
```
Gap from band 2 (0.39715066104943847) to band 3 (0.4426866269013708), 10.843997165936608%
```

@thchr, is this to be expected, and why?

thchr:
Hi Steven,

Ah, that's unfortunate :(.

Regarding the test failures for `mpb_diamond.py`:
As I mentioned in [this comment](https://github.com/NanoComp/mpb/pull/127#issuecomment-717284485), while initially surprising, this is probably a result of the spheres being close-packed: if a voxel intersects two spheres, there's actually three materials in that voxel, in which case the averaging function for libctl geometries will give up and fall back to the same code that handles the user-specified material functions. So, the averaging now ends up different for those voxels.

This was actually the only test case that ended up needing changing for NanoComp/mpb#127 as well (see [this bit of the diff](https://github.com/NanoComp/mpb/pull/127/files#diff-f64effba14fd5f4d466e94be451e7a74eaedc3d640ce14daab35da913d8ecbf2L263-R263)).

So overall, this seems expected (though I hadn't anticipated that it would impact meep, unfortunately).

PR #1419, opened 2020-11-05
updated documentation for non-linear saturable media.

acerjan:
updated saturable medium documentation that we discussed.


stevengj:
Thanks!  Test failure is unrelated (#1418).

PR #1437, opened 2020-11-28
add option to plot2D to plot contour of geometry rather than its image

### LINKED ISSUE #1435: overlaying fields on geometry contour using plot2D
When visualizing the fields overlaid on a geometry, the [`plot2D`](https://meep.readthedocs.io/en/latest/Python_User_Interface/#data-visualization) routine provides options to set the transparency of each object type separately via the `alpha` setting of `fields_parameters` and `eps_parameters`. However, a useful feature that is currently missing is to overlay the fields on the *contour* of the geometry similar to the `-C` setting of `h5topng`.

As a demonstration, compare the output from `plot2D` and `h5topng` for a demultiplexer:

```
sim.plot2D(fields=mp.Hz,
           field_parameters={'alpha':0.9},
           plot_sources_flag=False,
           plot_monitors_flag=False,
           plot_boundaries_flag=False)
```

demux_cw_fields_1 31um_transparency0 9

The geometry is noticeable in the dark background features of this figure but this adds a shading effect which reduces the contrast in the field profile.

```
$ h5topng -Zc bluered -C epsilon.h5 output.h5
```

cwfields_RdBu

For a geometry comprised of only two materials, overlaying the fields on the geometry contour improves the visual presentation by not degrading the contrast.

oskooi:
Fixes #1435.

Adds new options `contour` and  `contour_linewidth` (for Matplotlib's [`contour`](https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.pyplot.contour.html)) to `eps_parameters` of `plot2D`.

Documentation is also provided.

Example usage:
```
sim.plot2D(fields=mp.Hz,
           field_parameters={'alpha':1.0},
           eps_parameters={'contour':True},
           plot_sources_flag=True,
           plot_monitors_flag=True,
           plot_boundaries_flag=True)
```

demux_cw_fields_1 31um

For comparison, the default `imshow` rendering:

```
sim.plot2D(fields=mp.Hz,
           field_parameters={'alpha':0.7},
           plot_sources_flag=True,
	   plot_monitors_flag=True,
           plot_boundaries_flag=True)
```

demux_cw_fields_1 31um_imshow

### DIFF
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -2484,8 +2484,8 @@ plt.savefig('sim_domain.png')
     - `interpolation='spline36'`: interpolation algorithm used to upsample the pixels.
     - `cmap='binary'`: the color map of the geometry
     - `alpha=1.0`: transparency of geometry
-    - `contour=False`: if True, plot a contour of the geometry rather than its image
-    - `contour_linewidth=0.5`: width of the contour lines if `contour=True`
+    - `contour=False`: if `True`, plot a contour of the geometry rather than its image
+    - `contour_linewidth=1`: line width of the contour lines if `contour=True`
 * `boundary_parameters`: a `dict` of optional plotting parameters that override
   the default parameters for the boundary layers.
     - `alpha=1.0`: transparency of boundary layers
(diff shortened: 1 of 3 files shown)


smartalecH:
Looks good on my end.

oskooi:
Note that the `origin` keyword parameter of `contour` is set to `lower` by default which is *different* from [`imshow`](https://matplotlib.org/3.3.3/api/_as_gen/matplotlib.pyplot.imshow.html) which is `upper`. The default `origin` value for `contour` must therefore be changed to `upper` for consistency.

PR #1446, opened 2020-12-08
minor updates and fixes to docs

oskooi:
- add a small note for `plot2D` in [Python User Interface/Data Visualization](https://meep.readthedocs.io/en/latest/Python_User_Interface/#data-visualization) to not use this function in combination with `meep.am_master()`
- add a demonstration of the recently added contour plot capability for `plot2D` to [Tutorial/GDSII Import/Modes of a Ring Resonator](https://meep.readthedocs.io/en/latest/Python_Tutorials/GDSII_Import/#modes-of-a-ring-resonator)
- fix some minor formatting errors


python/simulation.py
```diff
@@ -3849,8 +3849,8 @@ def plot2D(self, ax=None, output_plane=None, fields=None, labels=False,
         plt.savefig('sim_domain.png')
         ```
 
-        Warning: When running a [parallel simulation](Parallel_Meep.md), do *not* call `plot2D` from within an
-        `if meep.am_master():` statement since this will cause one of the MPI processes to deadlock.
+        Note: When running a [parallel simulation](Parallel_Meep.md), the `plot2D` function expects to be called
+        on all processes, but only generates a plot on the master process.
```
oskooi, line 3853: Changing the docstrings here should have been followed by executing `meep/doc/generate_py_api.py` so that the changes also appeared in the markdown page `Python_User_Interface.md`. This is why 8af7808 was necessary after this PR was merged. There is nothing wrong with readthedocs.io.

Issue #1431, opened 2020-11-24
Heap buffer overflow and segmentation fault in collapse_array


ianwilliamson:
I am getting a heap buffer overflow error when running address sanitizer on a 3D version of one of the adjoint optimization examples. It seems to be pointing to line 698 of array_slice.cpp:

https://github.com/NanoComp/meep/blob/41548a3df131ce7fb5497cf8e59ecd34f48c9866/src/array_slice.cpp#L698

Below is the address sanitizer output, where it looks like something resembling an off-by-one issue. I've been getting segmentation faults when trying to run an optimization under MPI, with stack traces resembling the one shown below.

```
==2021035==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x7f85f88f5000 at pc 0x5627c03fd546 bp 0x7fffcbec3f70 sp 0x7fffcbec3f68
READ of size 8 at 0x7f85f88f5000 thread T0
    #0 0x5627c03fd545 in meep::collapse_array(double*, int*, unsigned long*, meep::direction*, meep::volume, int) meep/src/array_slice.cpp:698:48
    #1 0x5627c03fdb5c in meep::fields::get_array_metadata(meep::volume const&, bool, bool) meep/src/array_slice.cpp:733:44
    #2 0x7f864da01861  (___/@0xab1e8000+0x5fa861)
    #3 0x5627c4a1686a in _PyCFunction_FastCallDict python_runtime/v3_6/Objects/methodobject.c:234:22
    #4 0x5627c4aba521 in call_function python_runtime/v3_6/Python/ceval.c:4851:9
    #5 0x5627c4ab3b25 in _PyEval_EvalFrameDefault python_runtime/v3_6/Python/ceval.c:3335:19
    #6 0x5627c4abaf87 in PyEval_EvalFrameEx python_runtime/v3_6/Python/ceval.c:754:12
    #7 0x5627c4abaf87 in _PyEval_EvalCodeWithName python_runtime/v3_6/Python/ceval.c:4166:14
    #8 0x5627c4abbc13 in fast_function python_runtime/v3_6/Python/ceval.c:4992:12
    #9 0x5627c4aba4e5 in call_function python_runtime/v3_6/Python/ceval.c:4872:17
    #10 0x5627c4ab3b25 in _PyEval_EvalFrameDefault python_runtime/v3_6/Python/ceval.c:3335:19
[...]

PR #1447, opened 2020-12-09
fix memory leak in get_array_metadata

stevengj:
May help with #1431?

### DIFF
--- a/src/array_slice.cpp
+++ b/src/array_slice.cpp
@@ -732,13 +732,7 @@ std::vector<double> fields::get_array_metadata(const volume &where, bool collaps
     full_dirs[fr] = dirs[fr];
 
   double *weights = get_array_slice(where, NO_COMPONENT);
-  if (collapse_empty_dimensions) {
-    double *newweights = collapse_array(weights, &rank, dims, dirs, where);
-    if (weights != newweights) {
-      delete[] weights;
-      weights = newweights;
-    }
-  }
+  if (collapse_empty_dimensions) weights = collapse_array(weights, &rank, dims, dirs, where);
 
   /* get length and endpoints of x,y,z tics arrays */
   size_t nxyz[3] = {1, 1, 1};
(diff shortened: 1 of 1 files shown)


ianwilliamson:
> May help with #1431?

My understanding of the output in #1431 is that the heap overflow is occurring somewhere inside of `collapse_array`. Rerunning asan with the fix introduced here still produces the same output as in #1431. At the moment, I can not confirm whether this fixes a separate memory leak (though it certainly looks like weights should be freed).

src/array_slice.cpp
```diff
@@ -732,7 +732,13 @@ std::vector<double> fields::get_array_metadata(const volume &where, bool collaps
     full_dirs[fr] = dirs[fr];
 
   double *weights = get_array_slice(where, NO_COMPONENT);
-  if (collapse_empty_dimensions) weights = collapse_array(weights, &rank, dims, dirs, where);
+  if (collapse_empty_dimensions) {
+    double *newweights = collapse_array(weights, &rank, dims, dirs, where);
+    if (weights != newweights) {
```
oskooi, line 737: This line *always* evaluates to false because in the line just before this one, `weights` is passed by reference to `collapse_array` which deletes it at the end:

https://github.com/NanoComp/meep/blob/8af7808bf674481a71a1c04a2fc27c1a624e5e4e/src/array_slice.cpp#L708-L710

Thus, the only thing that is necessary to fix the memory leak is to include the line `delete[] newweights;` replacing `delete[] weights;` in this PR.

smartalecH: > This line always evaluates to false because in the line just before this one, weights is passed by reference to collapse_array which deletes it at the end:

Actually, it only deletes it if `collapse_array` creates a new array. There are certain cases where it simply returns the existing `weights` array, which is why it's important to check if it's the same as before.

I think this is correct.

oskooi: I modified my original comment to point out that the `delete[] weights;` line inside the if statement is actually unnecessary since whenever the `weights` array *is* collapsed, it is deleted inside `collapse_array`.

stevengj: Right, I missed that `collapse_array` was already deleting the old array in the case where it allocates a new one.  Should be fixed.

PR #1456, opened 2020-12-20
collapse empty dimensions of time-domain grid slices by default and remove snap argument from get_array_metadata

oskooi:
Fixes #759.

Fixes #1431 based on the source of the bug described in [\#1431 (comment)](https://github.com/NanoComp/meep/issues/1431#issuecomment-748430786).

The `snap` parameter of `get_array_metadata` has been removed since it does not really serve a practical purpose compared with the argument `collapse` which does interpolation. Also, documentation has been added for the `collapse` argument which was missing.

The tests described in [\#1431 (comment)](https://github.com/NanoComp/meep/issues/1431#issuecomment-747848794) now produce the correct results:

```py
import meep as mp
import numpy as np

sim = mp.Simulation(cell_size=mp.Vector3(5,5),
                    resolution=10)

sim.init_sim()

ds = np.linspace(0,0.1,11)

print("sweep grid slice along y")
for d in ds:
    (x,y,z,w) = sim.get_array_metadata(center=mp.Vector3(0,d), size=mp.Vector3(0.5,0), collapse=False)
    if (w.ndim == 2):
	print('{:.3f}, {:.3f}, {:.3f}'.format(d,np.sum(w[:,0]),np.sum(w[:,1])))
    elif (w.ndim == 1):
	print('{:.3f}, {:.3f}'.format(d,np.sum(w)))

print("sweep grid slice along x")
for d in ds:
    (x,y,z,w) = sim.get_array_metadata(center=mp.Vector3(d,0), size=mp.Vector3(0,0.5), collapse=False)
    if (w.ndim == 2):
	print('{:.3f}, {:.3f}, {:.3f}'.format(d,np.sum(w[0,:]),np.sum(w[1,:])))
    elif (w.ndim == 1):
	print('{:.3f}, {:.3f}'.format(d,np.sum(w)))
```

```
sweep grid slice along y
0.000, 0.250, 0.250
0.010, 0.200, 0.300
0.020, 0.150, 0.350
0.030, 0.100, 0.400
0.040, 0.050, 0.450
0.050, 0.500
0.060, 0.450, 0.050
0.070, 0.400, 0.100
0.080, 0.350, 0.150
0.090, 0.300, 0.200
0.100, 0.250, 0.250
sweep grid slice along x
0.000, 0.250, 0.250
0.010, 0.200, 0.300
0.020, 0.150, 0.350
0.030, 0.100, 0.400
0.040, 0.050, 0.450
0.050, 0.500
0.060, 0.450, 0.050
[...]

PR #1458, opened 2020-12-29
return numpy arrays from std::vectors

stevengj:
Fixes the issue discussed in https://github.com/NanoComp/meep/pull/1456#discussion_r546324832


stevengj:
Passes tests locally.  Now if we do
```py
[x,y,z,w] = sim.get_array_metadata(dft_cell=near_fields)
type(x),type(y),type(z),type(w)
```
I get 
```
(numpy.ndarray, numpy.ndarray, numpy.ndarray, numpy.ndarray)
```
as desired.

PR #1459, opened 2020-12-30
fix issue relating to file IO in simulation.py

HayleyDeckers:
fixes two small bugs.
first a tiny fix for a hard-fail bug: A misnamed variable in load_minus_near2far that lead to an error when called.

Second a small change to `use_output_directory`. The file prefix was being overwritten with 'None' when `use_output_directory` was used.

This was causing issues for me as I wanted to use both a file prefix (specifying simulation parameters) and an output directory (specifying the particular experiment/to keep all the data neatly together) for later analysis.
Because the file prefix was overwritten with `None` when the output directory was set, the analysis script tried to open `$FOLDER/analysis-$FILE.h5` whereas the simulation script wrote to files `$FOLDER/simulation-$FILE.h5` regardless of the value of 'filename_prefix' in the simulations constructor.

This change sets the filename prefix to an empty string when `use_output_directory` is called without specifying a directory name, keeping it set otherwise. This leads to slightly different behaviour when no file prefix is set but `sim.use_output_directory()` is called (with not arguments). If the script is called `my_script.py` This wil now generate files as `my_script-out/$FILENAME.h5` instead of `my_script-out/my_script-$FILENAME.h5` which seems preferable.

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -2638,7 +2638,7 @@ class Simulation(object):
         loaded. This means that they will be *subtracted* from any future field Fourier
         transforms that are accumulated.
         """
-        self.load_near2far(fname, near2far)
+        self.load_near2far(fname, n2f)
         near2far.scale_dfts(-1.0)
 
     def get_near2far_data(self, near2far):
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -2124,6 +2124,7 @@ class Simulation(object):
         """
         if not dname:
             dname = self.get_filename_prefix() + '-out'
+            self.filename_prefix = ''
 
         closure = {'trashed': False}
 
@@ -2139,8 +2140,7 @@ class Simulation(object):
 
         if self.fields is not None:
             hook()
-        self.filename_prefix = None
-
+       
         return dname
 
     def _run_until(self, cond, step_funcs):
@@ -2638,7 +2638,7 @@ class Simulation(object):
         loaded. This means that they will be *subtracted* from any future field Fourier
         transforms that are accumulated.
         """
-        self.load_near2far(fname, n2f)
+        self.load_near2far(fname, near2far)
         near2far.scale_dfts(-1.0)
 
     def get_near2far_data(self, near2far):


stevengj:
Could you separate the near2far bugfix into a separate PR, so we can merge that separately from discussing the behavior change in `use_output_directory`?

HayleyDeckers:
Done https://github.com/NanoComp/meep/pull/1461

oskooi:
Rather than modify the default behavior, why not just specify the `filename_prefix` to whatever you want *after* calling `use_output_directory`:
```py
>>> sim = mp.Simulation(...)
>>> sim.use_output_directory("foo") ## sets filename_prefix = None
'foo'
>>> sim.filename_prefix = "bar"
```
Note that you can verify the value of `filename_prefix` using `sim.get_filename_prefix()`:
```py
>>> sim.get_filename_prefix()
'bar'
```

oskooi:
The documentation is correct: if you call `sim.use_output_directory()` with *no* arguments then the directory name is specified using the Python filename.

Setting the `filename_prefix` to `None` inside `use_output_directory` is just a matter of convention since we figured that users would specify one or the other (the filename prefix *or* the name of the output directory) but not both. In any event, as I mentioned previously, the simple workaround that does not require any changes is to just specify `filename_prefix` on your own after calling `use_output_directory`.

HayleyDeckers:
> The documentation is correct: if you call sim.use_output_directory() with no arguments then the directory name is specified using the Python filename.

I'm not on my work PC at the moment, so I cannot check but looking at the source it calls `dname = self.get_filename_prefix() + '-out'` before setting `self.filename_prefix = None`. Which means that if it the filename prefix was set, before calling `use_output_directory()`, it will use that prefix instead. 

Therefore the documentation should read that it uses whatever value the `filename_prefix` is set to, only using the filename if it is `None`. 

 > Setting the filename_prefix to None inside use_output_directory is just a matter of convention since we figured that users would specify one or the other (the filename prefix or the name of the output directory) but not both.
 
 Though the suggested solution is fine, it might be good to explicitly incorperate the incompatibility of the two in the documenation to prevent confusion if these weren't intended to be combined.
 
p.s. The old behaviour in the case of not setting a `filename_prefix` _and_ calling `use_output_directory` with no arguments can be preserved in the patch by replacing the `self.filename_prefix = ''` with `self.filename_prefix = None`.  Then it would only modify the behaviour in the case where a `filename_prefix` is set before `use_output_directory` is called with an argument. It would then independently set the folder and prefix, as if the filename prefix was set after `use_output_directory`. Instead of ignoring and resetting the initial value of `filename_prefix`, as is currently the case.

The documentation of `use_output_directory` could then say that it
> Willl use the value `dname` as a directory name if provided. If no value is provided it will use the value of `self.get_filename_prefix()+'-output'` as a directory name and reset the value of `self.filename_prefix` to `None`.

oskooi:
Updating the documentation for `use_output_directory` to make it clearer what happens when the `filename_prefix` is also set would be helpful. Note that you will need to first update the docstrings in `python/simulation.py` and then run the `doc/generate_py_api.py` script which generates the markdown pages for the user manual. See the description at the top of that file for how this is done.

PR #1460, opened 2020-12-31
update interpolation weight in get_array_slice

oskooi:
Addresses [#1456 (comment)](https://github.com/NanoComp/meep/pull/1456#issuecomment-753022745).

### DIFF
--- a/src/array_slice.cpp
+++ b/src/array_slice.cpp
@@ -320,7 +320,7 @@ static void get_array_slice_chunkloop(fields_chunk *fc, int ichnk, component cgr
     for (int i = 0; i < num_components; ++i) {
       // special case for fetching grid point coordinates and weights
       if (cS[i] == NO_COMPONENT) {
-        fields[i] = IVEC_LOOP_WEIGHT(s0, s1, e0, e1, dV0 + dV1 * loop_i2);
+        fields[i] = dV0 + dV1 * loop_i2;
       }
       else if (cS[i] == Dielectric) {
         complex<double> tr(0.0, 0.0);


stevengj:
I assume it passes the tests locally?

oskooi:
`python/tests/array_metadata.py` is failing with and without including the weights array `W` in computing the integral:

https://github.com/NanoComp/meep/blob/27d4512d8783629d70ffd5e7c7a6e01734ad24b7/python/tests/array_metadata.py#L75-L81

The error from this branch:
```
FAIL: test_array_metadata (__main__.TestArrayMetadata)
  File "python/tests/array_metadata.py", line 83, in test_array_metadata
    self.assertAlmostEqual(cw_modal_volume/pulse_modal_volume, 1.00, places=2)
AssertionError: (0.9768618126950315+0j) != 1.0 within 2 places
```

oskooi:
Not including the weights from `get_array_metadata` in computing the test integral `np.sum(EpsE2)/np.max(EpsE2) * np.sum(xm**2 + 2*ym**2)` from `array_metadata.py` makes the result much larger than the reference value computing using the field function. Thus, even with the change to the weights in this PR, the weights must still be included in computing the integral which is not what we expected.

Fixing the failing test just requires reducing the tolerance slightly.

stevengj:
Actually, it looks like the collapsed arrays do *not* include the edge weights, so we don't need to remove them: https://github.com/NanoComp/meep/blob/4d4ccb7ee497d79317bdf9f52fb3de9d30b8d298/src/dft.cpp#L786-L800

PR #1464, opened 2021-01-06
re-instate undocumented get_array_slice_dimensions in Python interface

oskooi:
#1456 removed `get_array_slice_dimensions` from the Python interface which unfortunately was discovered afterwards to be used by the adjoint solver in a couple of places: 

https://github.com/NanoComp/meep/blob/37117fe527866093bc0713fc21fd34d4aacd448b/python/adjoint/optimization_problem.py#L191-L195

https://github.com/NanoComp/meep/blob/37117fe527866093bc0713fc21fd34d4aacd448b/python/adjoint/optimization_problem.py#L408-L415

Rather than modify the above lines to use the low-level `_get_array_slice_dimensions`, this PR simply re-instates `get_array_slice_dimensions` (in its original form). Note that this is an internal function used only by the adjoint solver that is not documented in the user manual.

We will need to do a version 1.17.1 release after this is merged since 1.17.0 broke the adjoint solver. 

Also, it would be good to add some unit tests for the adjoint solver to prevent these types of fixes in the future.

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -3299,6 +3299,28 @@ class Simulation(object):
             return points,weights
         return tuple(tics) + (weights,)
 
+    def get_array_slice_dimensions(self, component, vol=None, center=None, size=None):
+        """
+        Computes the dimensions of an array slice for a particular `component` (`mp.Ez`, `mp.Ey`, etc.).
+
+        Accepts either a volume object (`vol`), or a `center` and `size` `Vector3` pair.
+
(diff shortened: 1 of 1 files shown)


smartalecH (approved):
Looks good to me. I ran locally just to make sure we weren't missing anything else:

image

The bend tutorial gradients look fine.

I agree though -- a simple adjoint unit test would really help (it's on my list of things to do).

PR #1472, opened 2021-01-15
unit test for adjoint solver

oskooi:
Adds a new unit test for the adjoint solver which is based on computing the gradient of the scattering matrix element S<sub>12</sub> of a straight waveguide with respect to a square region of random ε (shown below) and comparing to the brute-force result based on a finite difference. The test verifies that the gradient at two randomly chosen positions in the design region computed using the adjoint solver and finite differences are equivalent to within three decimal places. The total runtime is ~20 seconds using a single Intel Xeon 4.20 GHz core.

test_matgrid

Note that `yee_grid=True` is used in the `add_mode_monitor` definition of the brute-force simulations in order to be consistent with its use in the adjoint solver:

https://github.com/NanoComp/meep/blob/36942d3356c37f630e60b4970da3420dd3ec2026/python/adjoint/objective.py#L43

Before this PR can be merged, [`python/Makefile.am`](https://github.com/NanoComp/meep/blob/master/python/Makefile.am) needs to be modified so that the `meep.adjoint` module can be loaded from the `python/adjoint` subdirectory otherwise `make check` will fail for this test. Note that `autograd` is a dependency for this test.

### DIFF
--- a/python/tests/adjoint_solver.py
+++ b/python/tests/adjoint_solver.py
@@ -1,5 +1,5 @@
 import meep as mp
-import meep.adjoint as mpa
+import adjoint as mpa
 import numpy as np
 from autograd import numpy as npa
 from copy import deepcopy
@@ -9,8 +9,6 @@ class TestAdjointSolver(unittest.TestCase):
 
     def test_adjoint_solver(self):
         np.random.seed(9861548)
-        mp.verbosity(1)
-
         resolution = 25
 
         silicon = mp.Medium(epsilon=12)
@@ -99,7 +97,9 @@ class TestAdjointSolver(unittest.TestCase):
                                 sources=sources,
                                 geometry=geometry)
 
-            obj_list = [mpa.EigenmodeCoefficient(sim,mp.Volume(center=mp.Vector3(0.5*sxy-dpml),size=mp.Vector3(0,sxy,0)),1)]
+            obj_list = [mpa.EigenmodeCoefficient(sim,
+                                                 mp.Volume(center=mp.Vector3(0.5*sxy-dpml),
+                                                           size=mp.Vector3(0,sxy,0)),1)]
 
             def J(mode_mon):
                 return npa.abs(mode_mon)**2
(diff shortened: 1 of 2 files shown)


smartalecH:
Rather than compute the FD gradient each test, compute it for several points on your machine and make sure it matches the adjoint gradient. 

Then, just cache the adjoint gradients (either hard code some of them in or save them to a file) and use that to verify the adjoint solve (we do this with several other tests). This way, you don't have to compute a FD approximation each time. Plus, two points probably aren't enough to reliably confirm that the gradients are accurate. 

As we add more features, we can update the test and its gradient comparison if needed.

stevengj:
Rather than checking a single point, maybe compute the directional derivative in a random direction.  That is, if `p` is a random parameter vector and `Δp` is a random small perturbation, check that `Δp⋅∇f ≈ (f(p + Δp) - f(p)) / ‖Δp‖`

python/tests/adjoint_solver.py
```diff
@@ -0,0 +1,150 @@
+import meep as mp
+import meep.adjoint as mpa
```
stevengj, line 2: Just `import adjoint as mpa` to import from the local path.

oskooi: Done but unfortunately now this test can only be run as part of `make check`. Trying to run the test independently produces an error when loading the `adjoint` module.

oskooi:
Following @stevengj's suggestion above, the test has been modified to verify the *directional derivative* rather than the gradient at two arbitrary points. The test also verifies that the value for the "objective function" returned by the forward simulation of the adjoint solver matches that of the unperturbed brute-force simulation (the slight differences are due to the different termination criteria).

Running `make check` works fine.

python/tests/adjoint_solver.py
stevengj: Do:
```py
try:
    import meep.adjoint as mpa
except:
    import adjoint as mpa
```

PR #1478, opened 2021-01-22
unit test for adjoint solver involving backpropagation of gradients using filters

### LINKED ISSUE #1402: adjoint-solver test
We just need some tests that test the derivatives (i.e. two simulations).  No need to test optimization, although we should test derivatives including filters.

oskooi:
Fixes #1402.

Adds a second unit test for the adjoint solver to verify the backpropagation of the gradients using the filters for a random structure by comparing the directional derivative with the brute-force finite-difference result (similar to #1472).

`python/tests/adjoint_solver.py` is now comprised of two separate tests which required some reorganization to prevent duplicating code.

### DIFF
--- a/python/tests/adjoint_solver.py
+++ b/python/tests/adjoint_solver.py
@@ -5,123 +5,130 @@ except:
     import adjoint as mpa
 import numpy as np
 from autograd import numpy as npa
-from copy import deepcopy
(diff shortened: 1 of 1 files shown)


smartalecH (approved):
Looks good to me.

PR #1479, opened 2021-01-24
more granular MPI timing statistics

### LINKED ISSUE #1474: more granular MPI timing
Would be nice to keep a separate timer for all-process communications (including sum-to-all, broadcast, MPI_Barrier) vs send/recv between individual pairs of processors.

(Since the send/recv communications only occur in the timestepping code for [`boundary_communications`](https://github.com/NanoComp/meep/blob/f842b52f3c0b40729e903db2b91c2246ce158d15/src/mympi.cpp#L525-L570), maybe that type should get a separate timer, so we can leave the other places unchanged.)

oskooi:
Fixes #1474.

Splits the timing statistics for MPI into two separate bins for (1) all-to-all communication/synchronization and (2) one-to-one communication.

Also, the timing bin `Boundaries` which was never actually used is now used to time the portion of `fields::step_boundaries` that involves copying the chunk boundaries. This occurs in two places: before and after `boundary_communications`.


ianwilliamson:
Since this will silently break code that calls this function (because the values extracted for a given index will have a different meaning), what do you think about instead using string or enum keys for extracting these measurements? If these measurements are ever extended in the future that would make such changes more sustainable. Rather than changing the existing interface, you could add a new one that accepts these keys and maintain backwards compatibility.

stevengj:
It's true that this is a breaking change, but in practice I suspect that only the Meep developers (i.e. us) are using the API, so it's not such a bit deal. 

I agree that using string keys, maybe stored as a `std::map`, would be more flexible.

src/mpb.cpp
```diff
@@ -440,7 +440,7 @@ void *fields::get_eigenmode(double frequency, direction d, const volume where, c
     // then, synchronize the data
     eps_data.ncache = eps_data.icache; // actual amount of cached data
     double *summed_cache = (double *)malloc(sizeof(double) * 6 * eps_data.ncache);
-    am_now_working_on(MpiTime);
+    am_now_working_on(MPI_all);
```
stevengj, line 443: I would use `MpiAll` or something like that, since `MPI_*` functions are names use by `mpi.h` itself.

src/step.cpp
```diff
@@ -157,10 +156,14 @@ void fields::step_boundaries(field_type ft) {
         }
       }
     }
+  finished_working();
 
+  am_now_working_on(MPI_one);
```
stevengj, line 161: similarly, use `MpiOne`.

src/mpb.cpp
stevengj: Better yet `MpiAllTime`

PR #1480, opened 2021-01-25
Freed the memory of two arrays in structure.cpp

jianguan210:
Freed the memory of two arrays after use.


src/structure.cpp
```diff
@@ -245,6 +245,8 @@ std::vector<grid_volume> choose_chunkdivision(grid_volume &gv, volume &v, int de
       master_printf("Splitting into %d chunks by cost\n", my_num_chunks);
     split_by_cost(prime_factors, gv, chunk_volumes);
   }
+  delete [] effort_volumes;
+  delete [] effort;
```
stevengj, line 249: Looks good.   The `double *effort` in `grid_volume::split_by_effort` should also probably be declared as `const` to indicate that the function doesn't take ownership of the pointer.

PR #1481, opened 2021-01-25
Clean runtime warnings

jianguan210:
CONTEXT: We observe some warnings related to meep c++ source code when unit tests are performed with MSan and ASan. We would like to clean the runtime warnings.

SCOPE:
1. Fixed a logic bug in `update_stats_from_material()`
2. Removed the fall-through warnings in switch statements and if statements.
3. Deleted unused variables.
4. Initialized variables and an array created by `new`.
5. Fixed a bug in the constructor function of `gaussianbeam` class.

### DIFF
--- a/src/meepgeom.cpp
+++ b/src/meepgeom.cpp
@@ -2198,32 +2198,76 @@ static void print_tensor(std::complex<double> *in) {
 
 /* convenience routine to get element of material tensors */
 static std::complex<double> cvec_to_value(vector3 diag, cvector3 offdiag, int idx) {
+  std::complex<double> val = std::complex<double>(0, 0);
   switch (idx) {
-    case 0: return std::complex<double>(diag.x, 0);
-    case 1: return std::complex<double>(offdiag.x.re, offdiag.x.im);
-    case 2: return std::complex<double>(offdiag.y.re, offdiag.y.im);
-    case 3: return std::complex<double>(offdiag.x.re, -offdiag.x.im);
-    case 4: return std::complex<double>(diag.y, 0);
-    case 5: return std::complex<double>(offdiag.z.re, offdiag.z.im);
-    case 6: return std::complex<double>(offdiag.y.re, -offdiag.y.im);
-    case 7: return std::complex<double>(offdiag.z.re, -offdiag.z.im);
-    case 8: return std::complex<double>(diag.z, 0);
+    case 0:
+      val = std::complex<double>(diag.x, 0);
+      break;
+    case 1:
+      val = std::complex<double>(offdiag.x.re, offdiag.x.im);
+      break;
+    case 2:
+      val = std::complex<double>(offdiag.y.re, offdiag.y.im);
+      break;
+    case 3:
+      val = std::complex<double>(offdiag.x.re, -offdiag.x.im);
+      break;
+    case 4:
+      val = std::complex<double>(diag.y, 0);
+      break;
+    case 5:
+      val = std::complex<double>(offdiag.z.re, offdiag.z.im);
+      break;
+    case 6:
+      val = std::complex<double>(offdiag.y.re, -offdiag.y.im);
+      break;
+    case 7:
+      val = std::complex<double>(offdiag.z.re, -offdiag.z.im);
+      break;
+    case 8:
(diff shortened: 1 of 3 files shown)


src/meepgeom.cpp
```diff
@@ -1978,8 +1978,8 @@ void fragment_stats::update_stats_from_material(material_type mat, size_t pixels
           nonzero_conductivity_pixels_extmat_already_added =
               count_nonzero_conductivity_pixels(med, pixels);
         }
-        break;
       }
+      break;
```
stevengj, line 1982: Thanks!  Technically, the fall-through was safe since it just goes to the `return` statement, but the old code would cause a problem if we ever added more cases to this `switch` statement.

src/meepgeom.cpp
```diff
@@ -2198,32 +2198,76 @@ static void print_tensor(std::complex<double> *in) {
...
+      val = std::complex<double>(offdiag.y.re, -offdiag.y.im);
+      break;
+    case 7:
+      val = std::complex<double>(offdiag.z.re, -offdiag.z.im);
+      break;
+    case 8:
+      val = std::complex<double>(diag.z, 0);
+      break;
+    default:
+      meep::abort("Invalid value in switch statement.");
```
stevengj, line 2231: Probably we should a [`noreturn` attribute](https://en.cppreference.com/w/cpp/language/attributes/noreturn) to `meep::abort` (requires C++11, or could be done with a [gcc attribute](https://gcc.gnu.org/onlinedocs/gcc-4.7.2/gcc/Function-Attributes.html#Function-Attributes), which would presumably gotten rid of the compiler warning you are addressing here.  But this change is fine in any case.

src/multilevel-atom.cpp
```diff
@@ -98,11 +98,11 @@ static bool invert(realnum *S, int p) {
   } // singular
 
   int lwork = -1;
-  realnum work1;
+  realnum work1 = 0.0;
   DGETRI(&p, S, &p, ipiv, &work1, &lwork, &info);
   if (info != 0) abort("error %d in DGETRI workspace query", info);
   lwork = int(work1);
-  realnum *work = new realnum[lwork];
+  realnum *work = new realnum[lwork]();
```
stevengj, line 105: I don't understand these parens?

stevengj: Oh, I see, it initializes the array to zero.

Why is this needed?  `DGETRI` doesn't require the `work` array to be initialized.

stevengj:
Thanks!  In general I'm in favor of eliminating compiler warnings.

I'm not sure why you are initializing the `work` input to [DGETRI](http://www.netlib.org/lapack/explore-html/dd/d9a/group__double_g_ecomputational_ga56d9c860ce4ce42ded7f914fdb0683ff.html), though?

src/meepgeom.cpp
jianguan210: Thanks for the feedback. Here the position change for the "break" is for a logic issue. The original version will only executes `i=0`, and then escape the `for` loop.

src/meepgeom.cpp
jianguan210: Thanks for feedback.

src/multilevel-atom.cpp
jianguan210: Yes. I agree that it will not cause a problem as it is the output variable. But the MSan will complain the non-initialization before using the variable. Usually we always initialize variables when they are defined. If I remember correctly, this is the same reason for the other PR submitted for NanoComp/mpb.

jianguan210:
Prof. Johnson, Thanks for the review. I added comments accordingly. Please let me know if you have any suggestions or feedbacks.

src/meepgeom.cpp
stevengj: Oh, I see, it is in a `loop`; I was looking at the wrong line, thanks.

src/multilevel-atom.cpp
stevengj: Fine; I wish MSan was a bit more smart about this, but I agree that it doesn't hurt to silence the warning here.

Issue #1492, opened 2021-02-03
"cache-oblivious" MPI chunk subdivision


stevengj:
In a parallel MPI run with N nodes and M cores per node, for a total of NM processes, we ideally want to divide a simulation so that *adjacent* chunks are assigned to a *single* node as much as possible (so that we exploit fast intra-node communication for the chunk boundary conditions).

MPI provides us only limited information about this (with some very limited facilities for virtual "process topologies").  However, `mpirun` can easily be configured to ensure that *consecutive* process ranks are assigned within each node.  If we do this, then our problem becomes: *divide the chunks so that adjacent chunks have nearby ranks* as much as possible.

Stated this way, the problem becomes very similar to maximizing "cache locality" with an *unknown cache size* M (and indeed, the intra-node memory can be thought of as a kind of "cache").     A classic approach to this problem is a [cache-oblivious algorithm](https://en.wikipedia.org/wiki/Cache-oblivious_algorithm).   In our case, that should essentially boil down to partitioning the simulation grid **recursively**, assigning MPI ranks in *depth first* order.

Fortunately, we already do such recursive partitioning in the `split_by_cost` algorithm, and the `split_by_effort` algorithm is similarly recursive although the algorithm is a bit weirder.

It would be good to go through this more carefully and determine whether there is anything to improve here, or whether we should simply document that MPI runs should be set up with consecutive ranks within shared-memory nodes.

stevengj:
Note that MPI ranks are assigned to chunks [here](https://github.com/NanoComp/meep/blob/24cdcc51978fa9ab9ee72f4bc3d36dc2f694f883/src/structure.cpp#L154): first we split the grid volume into an array of chunks, and then assign processes to those chunks consecutively.

stevengj:
Of course, it's not clear how much this matters, since the performance should presumably be limited by the *slowest* link, i.e. by the inter-node communications (which will always be present to some extent no matter how we order the nodes).

stevengj:
Looking closely at `split_by_effort`, it is exactly a depth-first algorithm, so it already corresponds to the cache-oblivious algorithm.

PR #1499, opened 2021-02-12
unify split_by_effort and split_by_cost

### LINKED ISSUE #1498: unify split_by_effort and split_by_cost
Steps:

1.  Change `split_by_cost` to not split by prime factors anymore — instead, like `split_by_effort`, always divide into two pieces along some axis, except that the "halves" are unequal (weighed by the number of processes that will go into each half) for an odd number of processes.  (Recursively.)  (Hence it will always give the requested number of chunks, regardless of the prime factorization.)

2. Give `split_by_cost` an option to use a simpler cost metric of just the number of pixels in a subvolume, similar to `split_by_effort`.

At this point, `split_by_effort` should be redundant (and slower) and can be eliminated.

oskooi:
Closes #1498.

This PR unifies `split_by_cost` and `split_by_effort` based on the outline in #1498. `split_by_effort` has been removed entirely and been made a subset of `split_by_cost` which now contains a boolean flag to specify whether to use either the fragment statistics or the pixel count as the cost function for the given `grid_volume`. The original recursive bisection to chunk subdivision is retained (which is cache oblivious as described in #1492). `split_by_cost` now works for an arbitrary number of chunks and is no longer based on prime factors. The runtime performance of `split_by_effort` is improved as it computes the entire chunk division in one go rather than the previous approach which involved separate recursive calls to determine each chunk.

A comparison of the chunk layouts for the two different methods using a 2d square cell with just vacuum is shown below. The two methods produce nearly identical results in most cases although there are small differences (which are highlighted in the figures). These differences are the result of rounding errors in the cost function using the fragment statistics. 

```py
import meep as mp

import matplotlib
matplotlib.use('agg')

sim = mp.Simulation(resolution=10,
                    cell_size=mp.Vector3(10,10,0),
                    split_chunks_evenly=False)

sim.visualize_chunks()
```

chunk_layout_2d_batch1

chunk_layout_2d_batch2

All tests in the `make check` suite pass *except* for the subtest `test_cyl_metal_mirror_nonlinear` in `meep/tests/symmetry.cpp` which has been disabled:

https://github.com/NanoComp/meep/blob/7067536d9db2d229f7717b49f8bdb5abab0c3e10/tests/symmetry.cpp#L135-L176

This failing test involves a z-mirror symmetry plane in a cell with cylindrical coordinates which seems to be affected by the small change in the `split_by_effort` chunk division method.

```
meep: error in test_cyl_metal_mirror nonlinear vacuum
meep: error in test_cyl_metal_mirror nonlinear vacuum
imaginary part = 0 differs by 1.51084e-09 from -1.51084e-09
This gives a fractional error of 1
ez differs by 6.69695e-13-1.51084e-09i from 6.69695e-13-1.51084e-09i
This comes out to a fractional error of 1
Right now I'm looking at z = 0.5, r = 0.501, time 0.0625
```


oskooi:
Changing to `ttot = 10.0` from `ttot = 3.0` and `field_check_time = 8.0`.
```py
meep: error in test_cyl_metal_mirror nonlinear vacuum
meep: error in test_cyl_metal_mirror nonlinear vacuum
real part = 0.00432538 differs by -8.74538e-05 from 0.00441284
This gives a fractional error of 0.019818
er differs by 8.74538e-05-8.30032e-05i from 0.00441284-0.00354327i
This comes out to a fractional error of 0.0213051
Right now I'm looking at z = 0.5, r = 0.01, time 8
```

stevengj:
Since the only thing that this PR should have changed, I would print out the ivecs of the chunk coordinates in this PR vs master for the failing test, to see what changed.

oskooi:
For the failing test, this PR and master are producing *different* chunk divisions for the 2d cell using cylindrical coordinates as shown in the figure below. The figure shows the `ivecs` of the `litte_corner` (blue) and `big_corner` (red) for the `grid_volume` of each chunk.

PR1499

The change in the chunk division introduced by this PR is due to this line:

https://github.com/NanoComp/meep/blob/761e31b738e4e867a46d5253d2a1ea2b27dafe01/src/vec.cpp#L1101-L1102

However, it seems that the chunk division is *not* the source of the test failure because removing the nonlinearities from the test (deleting the lines with `set_chi3` or setting `chi3` internally to 0) causes it to pass:

https://github.com/NanoComp/meep/blob/761e31b738e4e867a46d5253d2a1ea2b27dafe01/tests/symmetry.cpp#L145-L146

Perhaps this suggests there is an unrelated bug somewhere in `structure_chunk::set_chi3` for cylindrical coordinates?

https://github.com/NanoComp/meep/blob/761e31b738e4e867a46d5253d2a1ea2b27dafe01/src/structure.cpp#L871-L905

oskooi:
Outputting the `split_measure`, `longest_axis` during the chunk splitting:

```
Testing Z mirror symmetry in Cylindrical...
Halving computational cell along direction z
split: z (direction), 80.00000 (split_measure), r (longest axis)
split: r (direction), 80.00000 (split_measure), r (longest axis)
```
```
split: z (direction), 128.00000 (split_measure), z (longest axis)
split: r (direction), 128.00000 (split_measure), z (longest axis)

```

src/vec.cpp
```diff
@@ -1094,46 +1019,20 @@ void grid_volume::find_best_split(int desired_chunks, int &best_split_point,
...
-    double split_measure = max(left_cost * (desired_chunks - 1), right_cost);
+    double split_measure = max(left_cost / (desired_chunks / 2), right_cost / (desired_chunks - (desired_chunks / 2)));
+    if (d == longest_axis) split_measure *= 0.7;
     if (split_measure < best_split_measure) {
-      if (d == longest_axis || split_measure < (best_split_measure - (0.3 * best_split_measure))) {
-        // Only use this split_measure if we're on the longest_axis, or if the split_measure is
-        // more than 30% better than the best_split_measure. This is a heuristic to prefer lower
-        // communication costs when the split_measure is somewhat close.
-        // TODO: Use machine learning to get a cost function for the communication instead of hard
-        // coding 0.3
```
stevengj, line 1107: Probably we should keep this comment (and put it before the `d == longest_axis` check).

oskooi:
The failing tests described in #1506 have been disabled for the time being because they are unrelated to this PR.

src/vec.cpp
oskooi: Fixed.

Issue #1229, opened 2020-05-23
support level-set material type


stevengj:
If the user could supply a (differentiable) level-set function φ(x) (such that it uses material A for φ<0 and material B for φ>0), instead of the current material function, that would provide enough information to do efficient subpixel smoothing even for user-defined functions.

In particular, suppose we are in 3d and are evaluating the material of a voxel.   We evaluate φ at the 8 corners of the voxel (possibly caching function evaluations from neighboring voxels).  If they are all the same sign, it is material A or B.  If they differ in sign, then we can do subpixel smoothing as follows:

1. Using the 8 points, fit φ(x) to an affine function α+aᵀx.   This is an 8x4 least-squares problem.

2. Now we know the normal direction (a) and the intersection plane α+aᵀx=0, we can compute the fill-fractions of A and B.   One approach is to just position a block to coincide with this plane and call our existing routine.    Another is to use formulas like the ones [here](https://github.com/stevengj/GeometryPrimitives.jl/blob/master/src/vxlcut.jl) to find the intersection of a planar half-space with the voxel.

Note that the user could interpolate a level-set function from a sufficiently high-resolution image by something like [fast marching](https://pythonhosted.org/scikit-fmm/).

Note also that we could eventually incorporate this into our topology optimization routine — once we have sufficiently binarized the structure and the topology has stopped changing, we could switch to a level-set representation.  This would probably require us to completely change how we compute the gradients, however.

stevengj:
(This would be in src/meepgeom.cpp — it would be be a new material type just like MATERIAL_FUNCTION, but would switch over to a different subpixel-averaging routine.

smartalecH:
Doesn't the prism object already support subpixel smoothing? If so, it might be easier to do the discrete shape optimization over various prism object vertex points, rather than a level-set.

For most practical applications, the design needs to eventually be projected onto some sort of polygonal basis anyway. Several libraries (e.g. [shapely](https://shapely.readthedocs.io/en/latest/manual.html)) can even do erosions and dilations on the polygons needed for continued robust optimization.

I could see this being an issue for applications that optimize in all 3 dimensions (e.g. 3D printing). In that case, a level-set would be ideal.

stevengj:
One basic drawback of optimizing over vertex locations is that, if an object expands too much in some region, you may need to increase the number of vertices.

thchr:
[...]

PR #1503, opened 2021-02-15
enable support for subpixel smoothing of MaterialGrid

### LINKED ISSUE #1500: enabling support for subpixel smoothing of MaterialGrid
The `MaterialGrid` feature added by @smartalecH in #1242 (which is not documented in the [User Interface](https://meep.readthedocs.io/en/latest/Python_User_Interface/) and is intended mainly for use with the [adjoint solver](https://meep.readthedocs.io/en/latest/Python_Tutorials/AdjointSolver/)) provides support for importing the material geometry as a NumPy array. One limitation of the current implementation is that it is based on a bilinear interpolation of the user-specified pixel grid onto Meep's Yee grid which is therefore only *first-order* accurate (rather than second order) with respect to the resolution:

https://github.com/NanoComp/meep/blob/7067536d9db2d229f7717b49f8bdb5abab0c3e10/src/meepgeom.cpp#L693-L701

The accuracy of the `MaterialGrid` feature can be improved by replacing the bilinear interpolation with anisotropic subpixel smoothing using a quadrature scheme similar to what [already exists for the material function](https://github.com/NanoComp/meep/blob/7067536d9db2d229f7717b49f8bdb5abab0c3e10/src/meepgeom.cpp#L1093-L1180) but without the slow callbacks to Python since it would be implemented entirely in C/C++.

As a demonstration of the convergence properties of the current implementation of the `MaterialGrid`, we can compute the relative error in the resonant mode of a 2d photonic crystal using a unit cell comprised of a cylinder with n=3.5 in air and non-zero Bloch-periodic boundaries (so that the mode is propagating at an oblique angle to the Cartesian grid). The source polarization is H<sub>z</sub> (i.e., in-plane electric field) in order to generate non-parallel fields along the discontinuous dielectric interface. (This type of convergence test was originally demonstrated in [Optics Letters, vol. 31, pp. 2972-74 (2006)](http://www-math.mit.edu/~stevenj/papers/FarjadpourRo06.pdf)). For comparison, we can also compute the same result using a [`Cylinder`](https://meep.readthedocs.io/en/latest/Python_User_Interface/#cylinder) geometric object which already supports subpixel smoothing.

The following schematic shows the H<sub>z</sub> profile for the `Cylinder` and `MaterialGrid` geometry representations at three different resolutions. The discrete artifacts of the `MaterialGrid` are slightly more apparent than the `Cylinder` representation. (These images were generated using [`plot2D`](https://meep.readthedocs.io/en/latest/Python_User_Interface/#data-visualization) and the red dots are the source locations.)

Cylinder_MaterialGrid_epsilon_compare

The simulation involves using [`Harminv`](https://meep.readthedocs.io/en/latest/Python_User_Interface/#Harminv) to compute the frequency of the resonant mode over a range of resolutions (`20-200`). The "exact' result is computed at a resolution of `300`.

```py
import numpy as np
import argparse
import meep as mp
import matplotlib

parser.add_argument('-res',
                    type=int,
                    default=20,
                    help='resolution (default: 20 pixels/um)')
parser.add_argument('-geom_type',
                    type=int,
                    choices=[1,2],
                    default=1,
                    help='type of geometry: 1: Cylinder (default), 2: material grid')
args = parser.parse_args()

cell_size = mp.Vector3(1,1,0)

rad = 0.301943

if args.geom_type == 1:
    geometry = [mp.Cylinder(radius=rad,
                            center=mp.Vector3(),
                            height=mp.inf,
else:
    design_region_resolution = int(2*args.res)
    xv, yv = np.meshgrid(x,y)

                              mp.air,
                              mp.Medium(index=3.5),
                              design_parameters=design_params,
                              grid_type='U_SUM')

                         material=matgrid)]

fcen = 0.3
df = 0.2*fcen
                     component=mp.Hz,
                     center=mp.Vector3(-0.1057,0.2094,0))]

k_point = mp.Vector3(0.3892,0.1597,0)

                    cell_size=cell_size,
                    geometry=geometry,
                    sources=sources,
                    k_point=k_point)

h = mp.Harminv(mp.Hz, mp.Vector3(0.3718,-0.2076), fcen, df)
sim.run(mp.after_sources(h),
        until_after_sources=200)

for m in h.modes:
    print("harminv:, {}, {}, {}".format(args.res,m.freq,m.Q))
```

A plot of the relative error vs. resolution for the two methods demonstrates that the `MaterialGrid` is indeed first-order accurate whereas the `Cylinder` is second order.

relerr_vs_resolution

To verify that anisotropic subpixel smoothing has been correctly set up for the `MaterialGrid`, we can use this same convergence test to verify that it is second-order accurate.

oskooi:
Closes #1500.

After some recent consultation with @smartalecH, it turns out enabling subpixel smoothing for the `MaterialGrid` is actually much easier than expected. All that is required is to make sure that the parameter `fallback` is set to `true` inside `geom_epsilon::eff_chi1inv_matrix` in order to use the *existing* adaptive quadrature routine from libctl:

https://github.com/NanoComp/meep/blob/7067536d9db2d229f7717b49f8bdb5abab0c3e10/src/meepgeom.cpp#L874

Because the routine `geom_epsilon::fallback_chi1inv_row` invokes `geom_epsilon::get_material_pt(material_type &material, const meep::vec &r)` to obtain ε at a given point `r` for which the use case of `material` as a material grid is *already* supported, no additional functions need to be added. In fact, setting it up this way means that subpixel smoothing also automatically supports symmetries for the material grid.

This PR simply adds a new `do_averaging` boolean property to the `MaterialGrid` object (default is `False`) which is passed from the Python interface into `geom_epsilon::eff_chi1inv_matrix` in C/C++. This means that the existing parameters `subpixel_maxeval` and `subpixel_tol` for the `Simulation` constructor which are used to specify the properties of the quadrature scheme can be used as-is.

The convergence test described in #1500 (shown below) verifies that the quadrature scheme for the `MaterialGrid` is second-order accurate (green line). This requires making sure that the resolution of the `MaterialGrid` is suffficiently higher than Meep's Yee grid (in this example, the ratio is `100`). If the resolution of the `MaterialGrid` is small relative to the Meep resolution (e.g., 2X), the convergence is closer to first-order accurate.

The next feature to add is the gradient of the subpixel smoothing for back propagation.

relerr_vs_resolution

```py
import numpy as np
import argparse
import meep as mp

parser.add_argument('-res',
                    type=int,
                    default=20,
                    help='resolution (default: 20 pixels/um)')
parser.add_argument('-geom_type',
                    type=int,
                    choices=[1,2],
                    default=1,
                    help='type of geometry: 1: Cylinder (default), 2: material grid')
args = parser.parse_args()

cell_size = mp.Vector3(1,1,0)

rad = 0.301943

if args.geom_type == 1:
    geometry = [mp.Cylinder(radius=rad,
                            center=mp.Vector3(),
                            height=mp.inf,
else:
    design_region_resolution = int(100*args.res)
    xv, yv = np.meshgrid(x,y)

                              mp.air,
                              mp.Medium(index=3.5),
                              design_parameters=design_params,
                              grid_type='U_SUM',
                              do_averaging=True)

                         material=matgrid)]

fcen = 0.3
df = 0.2*fcen
                     component=mp.Hz,
                     center=mp.Vector3(-0.1057,0.2094,0))]

k_point = mp.Vector3(0.3892,0.1597,0)

                    cell_size=cell_size,
                    geometry=geometry,
                    sources=sources,
                    k_point=k_point)

h = mp.Harminv(mp.Hz, mp.Vector3(0.3718,-0.2076), fcen, df)
sim.run(mp.after_sources(h),
        until_after_sources=200)

for m in h.modes:
    print("harminv:, {}, {}, {}".format(args.res,m.freq,m.Q))
```

### DIFF
--- a/python/geom.py
+++ b/python/geom.py
@@ -523,7 +523,7 @@ class Medium(object):
         return np.squeeze(epsmu)
 
 class MaterialGrid(object):
-    def __init__(self,grid_size,medium1,medium2,design_parameters=None,grid_type="U_DEFAULT",do_averaging=False):
+    def __init__(self,grid_size,medium1,medium2,design_parameters=None,grid_type="U_DEFAULT",do_averaging=False,beta=0,eta=0):
         self.grid_size = mp.Vector3(*grid_size)
         self.medium1 = medium1
         self.medium2 = medium2
@@ -537,6 +537,8 @@ class MaterialGrid(object):
             self.grid_size.z = 1
         self.num_params=int(self.grid_size.x*self.grid_size.y*self.grid_size.z)
         self.do_averaging = do_averaging
+        self.beta = beta
+        self.eta = eta
         if design_parameters is None:
             self.design_parameters = np.zeros((self.num_params,))
         elif design_parameters.size != self.num_params:
--- a/python/geom.py
+++ b/python/geom.py
@@ -523,7 +523,7 @@ class Medium(object):
         return np.squeeze(epsmu)
 
 class MaterialGrid(object):
-    def __init__(self,grid_size,medium1,medium2,design_parameters=None,grid_type="U_DEFAULT",do_averaging=False,beta=0,eta=0):
+    def __init__(self,grid_size,medium1,medium2,design_parameters=None,grid_type="U_DEFAULT",do_averaging=False,beta=0,eta=0.5):
         self.grid_size = mp.Vector3(*grid_size)
         self.medium1 = medium1
         self.medium2 = medium2
--- a/python/geom.py
+++ b/python/geom.py
@@ -523,7 +523,7 @@ class Medium(object):
         return np.squeeze(epsmu)
 
 class MaterialGrid(object):
-    def __init__(self,grid_size,medium1,medium2,design_parameters=None,grid_type="U_DEFAULT"):
+    def __init__(self,grid_size,medium1,medium2,design_parameters=None,grid_type="U_DEFAULT",do_averaging=False):
         self.grid_size = mp.Vector3(*grid_size)
         self.medium1 = medium1
         self.medium2 = medium2
@@ -536,7 +536,7 @@ class MaterialGrid(object):
         if isclose(self.grid_size.z,0):
             self.grid_size.z = 1
         self.num_params=int(self.grid_size.x*self.grid_size.y*self.grid_size.z)
-
+        self.do_averaging = do_averaging
         if design_parameters is None:
             self.design_parameters = np.zeros((self.num_params,))
         elif design_parameters.size != self.num_params:
--- a/python/typemap_utils.cpp
+++ b/python/typemap_utils.cpp
@@ -525,7 +525,13 @@ static int pymaterial_to_material(PyObject *po, material_type *mt) {
     PyObject *py_do_averaging = PyObject_GetAttrString(po, "do_averaging");
     bool do_averaging = false;
     if (py_do_averaging) { do_averaging = PyObject_IsTrue(py_do_averaging); }
-    md = make_material_grid(do_averaging);
+    PyObject *py_beta = PyObject_GetAttrString(po, "beta");
+    double beta = 0;
+    if (py_beta) { beta = PyFloat_AsDouble(py_beta); }
+    PyObject *py_eta = PyObject_GetAttrString(po, "eta");
+    double eta = 0;
+    if (py_eta) { eta = PyFloat_AsDouble(py_eta); }
+    md = make_material_grid(do_averaging,beta,eta);
     if (!pymaterial_grid_to_material_grid(po, md)) { return 0; }
     Py_XDECREF(py_do_averaging);
   }
--- a/python/typemap_utils.cpp
+++ b/python/typemap_utils.cpp
@@ -522,8 +522,12 @@ static int pymaterial_to_material(PyObject *po, material_type *mt) {
     if (!pymedium_to_medium(po, &md->medium)) { return 0; }
   }
   else if (PyObject_IsInstance(po, py_material_grid_object())) { // Material grid subclass
-    md = make_material_grid();
+    PyObject *py_do_averaging = PyObject_GetAttrString(po, "do_averaging");
+    bool do_averaging = false;
+    if (py_do_averaging) { do_averaging = PyObject_IsTrue(py_do_averaging); }
+    md = make_material_grid(do_averaging);
     if (!pymaterial_grid_to_material_grid(po, md)) { return 0; }
+    Py_XDECREF(py_do_averaging);
   }
   else if (PyFunction_Check(po)) {
     PyObject *eps = PyObject_GetAttrString(po, "eps");
--- a/python/typemap_utils.cpp
+++ b/python/typemap_utils.cpp
@@ -534,6 +534,8 @@ static int pymaterial_to_material(PyObject *po, material_type *mt) {
     md = make_material_grid(do_averaging,beta,eta);
     if (!pymaterial_grid_to_material_grid(po, md)) { return 0; }
     Py_XDECREF(py_do_averaging);
+    Py_XDECREF(py_beta);
+    Py_XDECREF(py_eta);
   }
   else if (PyFunction_Check(po)) {
     PyObject *eps = PyObject_GetAttrString(po, "eps");
--- a/src/meepgeom.cpp
+++ b/src/meepgeom.cpp
@@ -359,9 +359,11 @@ meep::realnum matgrid_val(vector3 p, geom_box_tree tp, int oi, material_data *md
 
   // project interpolated grid point
   meep::realnum u_proj;
-  if (md->beta)
-    u_proj = (tanh(md->beta*md->eta) + tanh(md->beta*(u_interp-md->eta))) /
-      (tanh(md->beta*md->eta) + tanh(md->beta*(1-md->eta)));
+  if (md->beta != 0) {
+    double tanh_beta_eta = tanh(md->beta*md->eta);
+    u_proj = (tanh_beta_eta + tanh(md->beta*(u_interp-md->eta))) /
+      (tanh_beta_eta + tanh(md->beta*(1-md->eta)));
+  }
 
   return md->beta ? u_proj : u_interp;
 }
@@ -1808,7 +1810,7 @@ material_type make_file_material(const char *eps_input_file) {
 /******************************************************************************/
 /* Material grid functions                                                    */
 /******************************************************************************/
-  material_type make_material_grid(bool do_averaging, double beta, double eta) {
+material_type make_material_grid(bool do_averaging, double beta, double eta) {
   material_data *md = new material_data();
   md->which_subclass = material_data::MATERIAL_GRID;
   md->do_averaging = do_averaging;
--- a/src/meepgeom.cpp
+++ b/src/meepgeom.cpp
@@ -869,7 +869,6 @@ void geom_epsilon::eff_chi1inv_row(meep::component c, double chi1inv_row[3], con
   symmetric_matrix meps_inv;
   bool fallback;
   eff_chi1inv_matrix(c, &meps_inv, v, tol, maxeval, fallback);
-  ;
 
   if (fallback) { fallback_chi1inv_row(c, chi1inv_row, v, tol, maxeval); }
   else {
@@ -916,7 +915,9 @@ void geom_epsilon::eff_chi1inv_matrix(meep::component c, symmetric_matrix *chi1i
 
   if (!get_front_object(v, geometry_tree, p, &o, shiftby, mat, mat_behind)) {
     get_material_pt(mat, v.center());
-    if (mat && mat->which_subclass == material_data::MATERIAL_USER && mat->do_averaging) {
+    if (mat && (mat->which_subclass == material_data::MATERIAL_USER ||
+                mat->which_subclass == material_data::MATERIAL_GRID)
+        && mat->do_averaging) {
       fallback = true;
       return;
     }
@@ -1797,10 +1798,10 @@ material_type make_file_material(const char *eps_input_file) {
 /******************************************************************************/
 /* Material grid functions                                                    */
 /******************************************************************************/
-material_type make_material_grid() {
+material_type make_material_grid(bool do_averaging) {
   material_data *md = new material_data();
   md->which_subclass = material_data::MATERIAL_GRID;
-  md->do_averaging = false;
+  md->do_averaging = do_averaging;
   return md;
 }
 
--- a/src/meepgeom.cpp
+++ b/src/meepgeom.cpp
@@ -350,12 +350,20 @@ meep::realnum matgrid_val(vector3 p, geom_box_tree tp, int oi, material_data *md
     ++matgrid_val_count;
   }
 
-  return (md->material_grid_kinds == material_data::U_MIN
-              ? umin
-              : (md->material_grid_kinds == material_data::U_PROD
-                     ? uprod
-                     : (md->material_grid_kinds == material_data::U_SUM ? usum / matgrid_val_count
-                                                                        : udefault)));
+  meep::realnum u_interp = (md->material_grid_kinds == material_data::U_MIN
+                            ? umin
+                            : (md->material_grid_kinds == material_data::U_PROD
+                               ? uprod
+                               : (md->material_grid_kinds == material_data::U_SUM ? usum / matgrid_val_count
+                                  : udefault)));
+
+  // project interpolated grid point
+  meep::realnum u_proj;
+  if (md->beta)
+    u_proj = (tanh(md->beta*md->eta) + tanh(md->beta*(u_interp-md->eta))) /
+      (tanh(md->beta*md->eta) + tanh(md->beta*(1-md->eta)));
+
+  return md->beta ? u_proj : u_interp;
 }
 static void cinterp_tensors(vector3 diag_in_1, cvector3 offdiag_in_1, vector3 diag_in_2,
                             cvector3 offdiag_in_2, vector3 *diag_out, cvector3 *offdiag_out,
@@ -697,8 +705,10 @@ void geom_epsilon::get_material_pt(material_type &material, const meep::vec &r) 
       geom_box_tree tp;
 
       tp = geom_tree_search(p, restricted_tree, &oi);
-      u = matgrid_val(p, tp, oi, md); // interpolate onto material grid
-      epsilon_material_grid(md, u);   // interpolate material from material grid point
+      // interpolate and (possibly) project onto material grid
+      u = matgrid_val(p, tp, oi, md);
+      // interpolate material from material grid point
+      epsilon_material_grid(md, u);
 
       return;
     // material read from file: interpolate to get properties at r
@@ -1798,10 +1808,12 @@ material_type make_file_material(const char *eps_input_file) {
 /******************************************************************************/
 /* Material grid functions                                                    */
 /******************************************************************************/
-material_type make_material_grid(bool do_averaging) {
+  material_type make_material_grid(bool do_averaging, double beta, double eta) {
   material_data *md = new material_data();
   md->which_subclass = material_data::MATERIAL_GRID;
   md->do_averaging = do_averaging;
+  md->beta = beta;
+  md->eta = eta;
   return md;
 }
 
--- a/src/meepgeom.cpp
+++ b/src/meepgeom.cpp
@@ -365,7 +365,7 @@ meep::realnum matgrid_val(vector3 p, geom_box_tree tp, int oi, material_data *md
       (tanh_beta_eta + tanh(md->beta*(1-md->eta)));
   }
 
-  return md->beta ? u_proj : u_interp;
+  return (md->beta != 0) ? u_proj : u_interp;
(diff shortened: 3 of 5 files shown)


oskooi:
Here is a plot of the relative error vs. resolution for the `MaterialGrid` with subpixel smoothing at different resolutions relative to the Meep grid. It's only when the `MaterialGrid` resolution is ~20X the Meep resolution (yellow line) and beyond that the quadratic convergence rate is distinguishable. Note that when the `MaterialGrid` resolution is 2X the Meep resolution (red line), the convergence rate is roughly linear. 

relerr_vs_resolution

In terms of runtime performance, the larger the resolution of the `MaterialGrid`, obviously the more memory storage that is required. However since the `MaterialGrid` is comprised of only *two* materials, the NumPy array comprising the `MaterialGrid` therefore contains only 2-bit booleans rather than 64-bit floating points. The reduced storage requirements combined with the C implementation of the adaptive quadrature routine (in libctl) means that the actual execution time for `set_epsilon` is fast. For example, at a Meep resolution of `300` and `MaterialGrid` resolution of `15000` (i.e., 50X), the NumPy array of `15000` x `15000` `bool` values (equivalent to ~0.4 GB of storage) executes in ~13 s using two cores of Intel Xeon 4.2 GHz.

smartalecH:
> However since the MaterialGrid is comprised of only two materials, the NumPy array comprising the MaterialGrid, therefore, contains only 2-bit booleans rather than 64-bit floating points.

Unfortunately, this isn't true. Remember, the material grid stores *continuous values* from 0 to 1, up to whatever precision you specify. In theory, you could use a binary array... but this not only kills your optimization, but the SWIG code still casts this as a 64-bit float array.

More importantly, the whole point of TO is to leverage continuous optimization methods on a large scale.

That being said, the storage requirements aren't terrible from meep's perspective. Most of the other data arrays are far bigger. Rather, the optimizer needs to be able to handle an array of that size. Normally MMA algorithms (e.g. CCSA in NLOPT) should work just fine. However, if you are using MPI, then this algorithm runs on each process, which means this array is *copied* to each process. You could of course get around this by only *optimizing* on one processor and broadcasting the results, but you just need to make sure your bookkeeping is correct.

smartalecH (approved):

stevengj:
It occurs to me that we may be able to do better if we implement the projection step as part of the MaterialGrid rather than in Python.  The basic reason for this is that, before projection, we have a nice smooth (low-pass-filtered) function, for which bilinear interpolation is a good description.

That is suppose g∈[0,1]ⁿ is our MaterialGrid data (*before*) projection and L(g,𝐱) ∈ [0,1] denotes the bilinear interpolation of g at 𝐱.    Then, we define ε at each point in space by ε(𝐱) = ε₁+(ε₂-ε₁)P(L(g,𝐱)) where P is our projection function. Our mean ⟨ε⟩ etcetera is then defined by integrals of ε(𝐱), which we can sample as finely in space as we want (e.g. at 30× the resolution, or using specialized quadrature schemes).

stevengj:
In particular, the MaterialGrid object would have two additional parameters β and η (for the tanh projection, taken from Python).  Then `matgrid_val` would apply this projection to the interpolated result (if β≠0 … if β=0 we should just skip the projection).

oskooi:
For reference, the hyperbolic tangent function that is to be replicated in C is:

https://github.com/NanoComp/meep/blob/7067536d9db2d229f7717b49f8bdb5abab0c3e10/python/adjoint/filters.py#L583

stevengj:
The point is that this then decouples the MaterialGrid resolution from the Meep resolution.    The MaterialGrid effectively defines an ε(x) at infinite resolution, which you can then *either* do subpixel averaging on (to get the Meep ε) *or* sample at any desired resolution to ship to the foundry (by upsampling & then projecting), so that the Meep calculation is consistent (to the degree allowed by the Meep resolution) with the foundry design.

Note also that you would test convergence by keeping the MaterialGrid resolution fixed and increasing the Meep resolution.

stevengj:
The MaterialGrid should have a high enough resolution that the "corners" induced in the interpolated+projected process (at infinite resolution you will have "corners" at the boundaries between MaterialGrid pixels) are shallow enough that they don't affect convergence much.

For a test case, I would probably use a smooth MaterialGrid input, similar to what we plan to use in the adjoint solver with filters.  You can either get this by taking your cylinder and filtering it. or just by putting in u = Gaussian "bump" or similar.

oskooi:
This seems to be working and producing quadratic convergence using a *fixed* resolution for the `MaterialGrid` as demonstrated below. The test involves specifying a 1.0 x 1.0 `MaterialGrid` with a fixed resolution of `50` by applying a Gaussian filter to the initial binarized cylinder design. The projection operator uses β = 1000 and η = 0.5, as suggested.

The filtered design which is then interpolated is shown in the figure inset. Note that the boundaries have been blurred using SciPy's `gaussian_filter` function.

Two new parameters `beta` and `eta` for the projection/thresholding operator (hyperbolic tangent) have been added to the `MaterialGrid` class object which are then passed into `matgrid_val` of `meepgeom.cpp` to apply the projection *after* the interpolation.

interp_project_convergence

```py
import numpy as np
import argparse
import meep as mp
import matplotlib
import time

parser.add_argument('-res',
                    type=int,
                    default=20,
                    help='resolution (default: 20 pixels/um)')
parser.add_argument('-geom_type',
                    type=int,
                    choices=[1,2],
                    default=1,
                    help='type of geometry: 1: Cylinder (default), 2: material grid')
args = parser.parse_args()

cell_size = mp.Vector3(1,1,0)

rad = 0.301943

if args.geom_type == 1:
    geometry = [mp.Cylinder(radius=rad,
                            center=mp.Vector3(),
                            height=mp.inf,
else:
    design_region_resolution = 50
    xv, yv = np.meshgrid(x,y)
    filtered_design_params = gaussian_filter(design_params, sigma=3.0, mode='nearest', output=np.double)

                              mp.air,
                              mp.Medium(index=3.5),
                              design_parameters=filtered_design_params,
                              grid_type='U_SUM',
                              do_averaging=True,
                              beta=1000,
                              eta=0.5)

                         material=matgrid)]

fcen = 0.3
df = 0.2*fcen
                     component=mp.Hz,
                     center=mp.Vector3(-0.1057,0.2094,0))]

k_point = mp.Vector3(0.3892,0.1597,0)

                    cell_size=cell_size,
                    geometry=geometry,
                    sources=sources,
                    k_point=k_point,
                    subpixel_maxeval=100000,
                    subpixel_tol=1e-5)

start = time.time()
sim.init_sim()
print("set_epsilon:, {} s".format(time.time()-start))

h = mp.Harminv(mp.Hz, mp.Vector3(0.3718,-0.2076), fcen, df)
sim.run(mp.after_sources(h),
        until_after_sources=200)

for m in h.modes:
    print("harminv:, {}, {}, {}".format(args.res,m.freq,m.Q))
```

smartalecH:
Looks good! It might be nice to have...
* a python `MaterialGrid` helper function that can easily set the `beta`/`eta` values for the material grid. Something that we can plug into the `OptimizationProblem` interface rather easily.
* a python `MaterialGrid` helper function that can interpolate and project the material grid onto any size array. Use cases include generating a higher resolution design field for image segmentation and shipping out to a foundry. Something like `interpolate_and_project(Nx=None,Ny=None)` where `Nx` and `Ny` are the desired output sizes of the grid. The default `None` will use the default resolution.

And of course, we still need to add gradient support.

ianwilliamson:
This is an interesting approach to getting subpixel smoothing working with `MaterialGrid`. This change seems to move Meep's `MaterialGrid` interface in a direction where it is tightly coupled to a very specific three field parameterization.

I'm just wondering if it would be useful to keep a more general interface for `MaterialGrid`? What's nice about the current interface is that it is very general and the user may use any parameterization that they wish (or any flavor of three field). What do you think?

stevengj:
> This change seems to move Meep's MaterialGrid interface in a direction where it is tightly coupled to a very specific three field parameterization.

I'm not quite sure which three fields you're referring to?

The proposed change would tightly couple the material grid to a particular projection operation, but even then you could turn this projection off (set β=0) and do the projection separately in any way that you want (albeit without subpixel smoothing).

smartalecH:
> I'm just wondering if it would be useful to keep a more general interface for MaterialGrid?

If `beta=0` (the default) then you get the same behavior as before (i.e. no thresholding). So it's still perfectly generalizable.

Note that you can still parameterize however you want *before* you pass into the material grid. You can do as many filters, thresholds, or function fits as you have time for.

stevengj:
It also occurs to me that we can now compute the normal vector directly via the gradient ∇u of the interpolated field at the center of the voxel before projection, similar to #1229 (since it's essentially a level-set function), rather than by quadrature.

ianwilliamson:
> I'm not quite sure which three fields you're referring to?

I was thinking of different kinds and combinations of projection / filtering operations, along the lines of what @smartalecH mentioned. It initially struck me as less flexible to hardcode one of these operations into the Meep `MaterialGrid` interface, although as you both point out, this can effectively be disabled to retain the existing behavior.

> albeit without subpixel smoothing

Is there an approach that would allow for subpixel smoothing in the `MaterialGrid` interface _without_ assuming a particular three field representation? For example, maybe I have a very high resolution design that's very close to binarized.

It seems like we just need a way to provide Meep with a continuous (or at least effectively continuous) representation of a design here? Above, my very high resolution binarized design might be one example of an effectively continuous design. The smooth level set variable (after interpolation) might be the continuous representation in the three field case. Is there an approach that could support both of these cases?

stevengj:
> Is there an approach that would allow for subpixel smoothing in the MaterialGrid interface without assuming a particular three field representation? For example, maybe I have a very high resolution design that's very close to binarized.

Yes.   Just use a MaterialGrid with β=0 at a high resolution, which still works with this PR (at a cost in memory).

stevengj:
> a python MaterialGrid helper function that can easily set the beta/eta values for the material grid.

Why do you need a helper function here, instead of just directly accessing the fields, i.e. `mgrid.beta = something`?

python/typemap_utils.cpp
```diff
@@ -522,8 +522,18 @@ static int pymaterial_to_material(PyObject *po, material_type *mt) {
...
+    if (py_do_averaging) { do_averaging = PyObject_IsTrue(py_do_averaging); }
+    PyObject *py_beta = PyObject_GetAttrString(po, "beta");
+    double beta = 0;
+    if (py_beta) { beta = PyFloat_AsDouble(py_beta); }
+    PyObject *py_eta = PyObject_GetAttrString(po, "eta");
+    double eta = 0;
+    if (py_eta) { eta = PyFloat_AsDouble(py_eta); }
+    md = make_material_grid(do_averaging,beta,eta);
     if (!pymaterial_grid_to_material_grid(po, md)) { return 0; }
+    Py_XDECREF(py_do_averaging);
```
stevengj, line 536: Don't forget to `Py_XDECREF` the `py_beta` and `py_eta` objects.

src/meepgeom.cpp
```diff
@@ -350,12 +350,20 @@ meep::realnum matgrid_val(vector3 p, geom_box_tree tp, int oi, material_data *md
...
+  meep::realnum u_interp = (md->material_grid_kinds == material_data::U_MIN
+                            ? umin
+                            : (md->material_grid_kinds == material_data::U_PROD
+                               ? uprod
+                               : (md->material_grid_kinds == material_data::U_SUM ? usum / matgrid_val_count
+                                  : udefault)));
+
+  // project interpolated grid point
+  meep::realnum u_proj;
+  if (md->beta)
```
stevengj, line 362: I would explicitly check `md->beta != 0` for clarity.

src/meepgeom.cpp
```diff
@@ -350,12 +350,20 @@ meep::realnum matgrid_val(vector3 p, geom_box_tree tp, int oi, material_data *md
...
+    u_proj = (tanh(md->beta*md->eta) + tanh(md->beta*(u_interp-md->eta))) /
```
stevengj, line 363: Maybe factor out
```cxx
double tanh_beta_eta = tanh(md->beta*md->eta);
```

python/geom.py
```diff
@@ -523,7 +523,7 @@ def _get_epsmu(self, diag, offdiag, susceptibilities, conductivity_diag, conduct
         return np.squeeze(epsmu)
 
 class MaterialGrid(object):
-    def __init__(self,grid_size,medium1,medium2,design_parameters=None,grid_type="U_DEFAULT"):
+    def __init__(self,grid_size,medium1,medium2,design_parameters=None,grid_type="U_DEFAULT",do_averaging=False,beta=0,eta=0):
```
stevengj, line 526: Probably eta=0.5 should be the default

python/typemap_utils.cpp
oskooi: Done.

python/geom.py
oskooi: Fixed.

src/meepgeom.cpp

src/meepgeom.cpp

python/typemap_utils.cpp
oskooi: Fixed.

smartalecH:
> For example, maybe I have a very high resolution design that's very close to binarized.

Say, for example you already have a design that works pretty well, but maybe you just want to eek out a bit of performance with another round of optimization.

As @stevengj mentioned, you could still use sub-pixel averaging on this, provided the original design has a really high resolution and you turn off the threshold.

*Or* you could do a mini topology optimization problem, where you want to find the new set of design parameters that, after passing through a filter and projection, correspond to your initial starting guess. Because of the nonlinear projection, it's not going to be a convex problem, but should be pretty straightfoward to solve (I've done this quite a few times with meep's adjoint framework). In fact, you can use the existing autograd libraries to efficiently compute the gradients.

Then, with this new set of design parameters, you can continue to optimize *with* the projection step *and* take advantage of the subpixel averaging without having to hypersample your image (lots of memory) because you are passing it through a filter first.

This methodology can be applied to other use cases too.

ianwilliamson:
> _Or_ you could do a mini topology optimization problem, where you want to find the new set of design parameters that, after passing through a filter and projection, correspond to your initial starting guess. Because of the nonlinear projection, it's not going to be a convex problem, but should be pretty straightfoward to solve (I've done this quite a few times with meep's adjoint framework). In fact, you can use the existing autograd libraries to efficiently compute the gradients.
> 
> Then, with this new set of design parameters, you can continue to optimize _with_ the projection step _and_ take advantage of the subpixel averaging without having to hypersample your image (lots of memory) because you are passing it through a filter first.

Yeah, performing re-optimization is certainly an interesting use case there.

As I mentioned above, I think what you're describing is one way of creating an effectively continuous representation in the three field parameterization: the bilinear interpolation of a relatively coarse filtered density array will give Meep a smooth level set-like distribution that can be thresholded. That's all good, but what strikes me as odd is the choice of hardcoding this into the Meep interface, when you could instead have a Meep adjoint module that is much more general.

[Not all users](https://www.osapublishing.org/oe/fulltext.cfm?uri=oe-26-24-31717&id=401362) may be optimizing with a three field parameterization. A user could have some other function for generating their geometry (either that they want to optimize or that they just want to simulate with the benefit of subpixel smoothing). Being required to pass this geometry to Meep as a very high resolution numpy array seems less flexible and may not be the most convenient. For example, perhaps the geometry isn't coming from an array in the first place.

You could still support the three field use case you describe above but with an interface that is more extensible to enable other approaches. I think what I am imagining here actually looks very similar to the material function interface that already exists? You give Meep a way of querying density values at arbitrary (finely) sampled points. Maybe you need to assume some well-behavedness or smoothness of the function, but that seems fair. You would also need an interface to pass the sensitivity / gradient at those points back out, but you basically already have this. You could then build the three field-specific scheme with the bilinear interpolation on top of this API, but still allow flexibility for other use cases.

stevengj:
> You give Meep a way of querying density values at arbitrary (finely) sampled points. Maybe you need to assume some well-behavedness or smoothness of the function, but that seems fair

You already have a way of querying density values at arbitrarily finely sampled points with the MaterialGrid.   The basic issue here is precisely that this is a poor representation of discontinuous functions unless you have a very fine resolution, *or* unless the MaterialGrid is essentially treated as level set.

Incorporating the (optional) projection into the MaterialGrid **precisely corresponds to allowing you to treat it as a level set** (with optional smoothing β and offset η), which is a very general way to describe discontinuous materials that still allows efficient subpixel smoothing and other operations.

It is not specific to a "three-field" parameterization that you keep referring to.

ianwilliamson:
> We already have this feature, and you can already do subpixel smoothing with it.

Thanks for the clarification. I incorrectly thought that subpixel smoothing was not supported in that interface. I suppose there's also the issue of getting gradients if one wanted to do optimization using that but, more importantly, I can imagine that it is slow.

Just a thought... but could an alternative interface be one where Meep provides the coordinates where it wants to perform the evaluations? This might allow one to take advantage of vectorization and other speed ups on the Python side, depending on how the function is computed.

stevengj:
> could an alternative interface be one where Meep provides the coordinates of the points where it wants to perform the evaluations?

It's pretty hard to do subpixel smoothing that way, because the subpixel quadrature points required would be hard to compute in advance (e.g. with vectorized code) without massive memory usage.

PR #1508, opened 2021-02-22
unit test and documentation for MaterialGrid

oskooi:
Adds a unit test and documentation to the [User Interface](https://meep.readthedocs.io/en/latest/Python_User_Interface/) for `MaterialGrid`.

A test ([`python/tests/material_grid.py`](https://github.com/NanoComp/meep/blob/master/python/tests/material_grid.py)) was added in #1242 but it does not seem to actually be doing anything (other than verify that an instance of the `MaterialGrid` object can be instantiated). The test added in this PR is based on the resonant mode calculation from #1500 which was later modified to use subpixel smoothing in #1503.

Separately, it might be useful to rename the member variable `design_parameter` of the `MaterialGrid` class to e.g. `density_grid` to make it more general since usage of the `MaterialGrid` does not necessarily always have to involve the adjoint solver.


doc/docs/Python_User_Interface.md
```diff
@@ -4156,6 +4156,85 @@ media](FAQ.md#why-does-my-simulation-diverge-if-0).
...
+The input are two materials `medium1` and `medium2` which are linearly interpolated at each grid point using
+a NumPy array `design_parameters` of size `grid_size` (a 3-tuple or `Vector3` of integers) with floating-point values in
+the range [0,1] used to define a linear weight (i.e., 0 is `medium1` and 1 is `medium2`). Currently, only two material
+types are supported: (1) frequency-independent isotropic $\varepsilon$ or $\mu$ and (2) `LorentzianSusceptibility`.
+`medium1` and `medium2` must both be the same type.
+
+[Subpixel smoothing](Subpixel_Smoothing.md) can be enabled by specifying `do_averaging=True`. To improve the accuracy
+of the subpixel smoothing, a [level set](https://en.wikipedia.org/wiki/Level_set) of the grid can be generated by applying
+a projection operator to the grid values after Meep has bilinearly interpolated the user-specified grid onto the
+[Yee grid](Yee_Lattice.md). The projection is based on a hyperbolic tangent function and is specified using the parameters `beta`
```
stevengj, line 4207: When the projection feature is enabled, the design parameters u(x) can be though of as [level-set function](https://en.wikipedia.org/wiki/Level-set_method) defining an interface at u(x)=η with a smoothing factor β ($\beta=\infty$ gives an unsmoothed, discontinuous interface).

doc/docs/Python_User_Interface.md
```diff
@@ -4156,6 +4156,85 @@ media](FAQ.md#why-does-my-simulation-diverge-if-0).
...
+
+Creates a `MaterialGrid` object.
+
+The input are two materials `medium1` and `medium2` which are linearly interpolated at each grid point using
+a NumPy array `design_parameters` of size `grid_size` (a 3-tuple or `Vector3` of integers) with floating-point values in
+the range [0,1] used to define a linear weight (i.e., 0 is `medium1` and 1 is `medium2`). Currently, only two material
+types are supported: (1) frequency-independent isotropic $\varepsilon$ or $\mu$ and (2) `LorentzianSusceptibility`.
+`medium1` and `medium2` must both be the same type.
+
+[Subpixel smoothing](Subpixel_Smoothing.md) can be enabled by specifying `do_averaging=True`. To improve the accuracy
```
stevengj, line 4204: This is not to "improve the accuracy".   It's just different.   I would say something like:

If you want to use a material grid to define a (nearly) discontinuous, piecewise-constant material that is *either* `medium1` or `medium2` almost everywhere, you can optionally enable a (smoothed) *projection* feature by setting the parameter `beta` to a positive value.

doc/docs/Python_User_Interface.md
```diff
@@ -4156,6 +4156,85 @@ media](FAQ.md#why-does-my-simulation-diverge-if-0).
...
+[Yee grid](Yee_Lattice.md). The projection is based on a hyperbolic tangent function and is specified using the parameters `beta` 
+($\beta$: "smoothness" of the turn on) and `eta` ($\eta$: contraction/dilation). The level set provides a general approach for
+defining a *discontinuous* representation of the otherwise continuously varying (via the bilinear interpolation) grid values.
+The subpixel smoothing is based on an adaptive quadrature scheme with properties `subpixel_maxeval` and `subpixel_tol` which
+can be specified using the [`Simulation`](#Simulation) constructor.
+
+Grids which are symmetric (e.g., mirror, rotation) must be explicitly defined. This requires overlapping a given
+`MaterialGrid` object with a symmetrized copy of itself. In this case, the overlapping grid points are combined using
+the method `grid_type` which is one of `"U_MIN"` (minimum of two grid values), `"U_PROD"` (product), `"U_SUM"` (sum),
+`"U_DEFAULT"` (default material at grid point).
```
stevengj, line 4216: U_DEFAULT doesn't combine material grids — it just takes the topmost material grid.

In general, these U_* options allow you to combine *any* material grids that overlap in space with no intervening objects.

doc/docs/Python_User_Interface.md
```diff
@@ -4156,6 +4156,85 @@ media](FAQ.md#why-does-my-simulation-diverge-if-0).
...
+a NumPy array `design_parameters` of size `grid_size` (a 3-tuple or `Vector3` of integers) with floating-point values in
+the range [0,1] used to define a linear weight (i.e., 0 is `medium1` and 1 is `medium2`). Currently, only two material
+types are supported: (1) frequency-independent isotropic $\varepsilon$ or $\mu$ and (2) `LorentzianSusceptibility`.
+`medium1` and `medium2` must both be the same type.
+
+[Subpixel smoothing](Subpixel_Smoothing.md) can be enabled by specifying `do_averaging=True`. To improve the accuracy
+of the subpixel smoothing, a [level set](https://en.wikipedia.org/wiki/Level_set) of the grid can be generated by applying
+a projection operator to the grid values after Meep has bilinearly interpolated the user-specified grid onto the
+[Yee grid](Yee_Lattice.md). The projection is based on a hyperbolic tangent function and is specified using the parameters `beta` 
+($\beta$: "smoothness" of the turn on) and `eta` ($\eta$: contraction/dilation). The level set provides a general approach for
```
stevengj, line 4208: Probably we should give the formula for the projection function.

doc/docs/Python_User_Interface.md
```diff
@@ -4156,6 +4156,88 @@ media](FAQ.md#why-does-my-simulation-diverge-if-0).
...
+`medium1` and `medium2` must both be the same type.
+
+[Subpixel smoothing](Subpixel_Smoothing.md) can be enabled by specifying `do_averaging=True`. If you want to use a
+material grid to define a (nearly) discontinuous, piecewise-constant material that is either `medium1` or `medium2`
+almost everywhere, you can optionally enable a (smoothed) *projection* feature by setting the parameter `beta` to a
+positive value. When the projection feature is enabled, the design parameters $u(x)$ can be though of as level-set
+function defining an interface at $u(x)=\eta$ with a smoothing factor $\beta$ ($\beta=\infty$ gives an unsmoothed,
+discontinuous interface). The projection operator is `tanh(beta*eta) + tanh(beta*(x-eta))) / (tanh(beta*eta) + 
+tanh(beta*(1-eta)))` involving the parameters `beta` ($\beta$: "smoothness" of the turn on) and `eta` ($\eta$: 
+erosion/dilation). The level set provides a general approach for defining a *discontinuous* representation of the otherwise
```
stevengj, line 4211: s/representation of the otherwise/function from otherwise/

doc/docs/Python_User_Interface.md
```diff
@@ -4156,6 +4156,88 @@ media](FAQ.md#why-does-my-simulation-diverge-if-0).
...
+continuously varying (via the bilinear interpolation) grid values. The subpixel smoothing is based on an adaptive quadrature
+scheme with properties `subpixel_maxeval` and `subpixel_tol` which can be specified using the [`Simulation`](#Simulation)
+constructor.
+
+Grids which are symmetric (e.g., mirror, rotation) must be explicitly defined. This requires overlapping a given
```
stevengj, line 4216: s/This requires/One way to implement this is by/

doc/docs/Python_User_Interface.md
```diff
@@ -4156,6 +4156,88 @@ media](FAQ.md#why-does-my-simulation-diverge-if-0).
...
+`MaterialGrid` object with a symmetrized copy of itself. In this case, the overlapping grid points are combined using
```
stevengj, line 4217: In the case of spatially overlapping `MaterialGrid` objects (with no intervening objects), any overlapping points are computed using

doc/docs/Python_User_Interface.md
```diff
@@ -4156,6 +4156,88 @@ media](FAQ.md#why-does-my-simulation-diverge-if-0).
...
+the method `grid_type` which is one of `"U_MIN"` (minimum of two grid values), `"U_PROD"` (product), `"U_SUM"` (sum),
```
stevengj, line 4218: minimum of the overlapping grid values

stevengj: U_SUM is the mean, not the sum

stevengj: Maybe it should be renamed to U_MEAN

python/geom.py
```diff
@@ -523,7 +523,38 @@ def _get_epsmu(self, diag, offdiag, susceptibilities, conductivity_diag, conduct
...
+    This class is used to specify materials interpolated from a discrete Cartesian grid. A class object is passed as the
+    `material` argument of a [`GeometricObject`](#GeometricObject) or the `default_material` argument of the `Simulation`
+    constructor (similar to a material function).
+    """
     def __init__(self,grid_size,medium1,medium2,design_parameters=None,grid_type="U_DEFAULT",do_averaging=False,beta=0,eta=0.5):
+        """
+        Creates a `MaterialGrid` object.
+
+        The input are two materials `medium1` and `medium2` which are linearly interpolated at each grid point using
+        a NumPy array `design_parameters` of size `grid_size` (a 3-tuple or `Vector3` of integers) with floating-point values in
```
stevengj, line 536: Maybe call it `weights` and a "weight function"

python/tests/material_grid.py
```diff
@@ -1,63 +1,81 @@
...
+        res = [25, 50]
+        freq_matgrid = []
+        for r in res:
+            freq_matgrid.append(compute_resonant_mode(r))
+            ## verify that the resonant mode is approximately equivalent to
+            ## the reference value
+            self.assertAlmostEqual(freq_ref, freq_matgrid[-1], 2)
+
+        ## verify that the relative error is decreasing with increasing resolution
+        self.assertLess(abs(freq_matgrid[1]-freq_ref),abs(freq_matgrid[0]-freq_ref))
```
stevengj, line 78: We really should check that it's less by some factor, say by a factor of 3 (ideally 4, but…)

PR #1512, opened 2021-02-25
additional improvements for MaterialGrid

oskooi:
Additional improvements to `MaterialGrid` following @stevengj's two unresolved comments in #1508:
- rename `design_parameters` to `weights`
- rename `U_SUM` to `U_MEAN`

Also includes more fixes and tweaks for the docstrings.

Additionally, this PR fixes a bug in `libpympb/mpb.h` related to the name of an imported header file that was discovered during the course of development:

https://github.com/NanoComp/meep/blob/a079d9bc13c92f4e70724fe88a50807549334e59/libpympb/pympb.hpp#L8

The preprocessor for GCC was loading `meepgeom.hpp` found in the system directory `/usr/local/include/meep` rather than the file in the source tree (from the `meep/src` subdirectory). As a result, this was causing an error during `make` because `design_parameters` had been renamed to `weights` in `meepgeom.cpp` but not in `meepgeom.hpp`. The fix involved changing this line to:

```cpp
#include "../src/meepgeom.hpp"
```

### DIFF
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -4168,8 +4168,8 @@ class MaterialGrid(object):
 <div class="class_docstring" markdown="1">
 
 This class is used to specify materials interpolated from a discrete Cartesian grid. A class object is passed as the
-`material` argument of a [`GeometricObject`](#GeometricObject) or the `default_material` argument of the `Simulation`
-constructor (similar to a material function).
+`material` argument of a [`Block`](#block) geometric object or the `default_material` argument of the
+[`Simulation`](#Simulation) constructor (similar to a material function).
 
 </div>
 
@@ -4184,7 +4184,7 @@ def __init__(self,
              grid_size,
              medium1,
              medium2,
-             design_parameters=None,
+             weights=None,
              grid_type='U_DEFAULT',
              do_averaging=False,
              beta=0,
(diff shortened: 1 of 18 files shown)


oskooi:
Configure command:
```
./configure --enable-maintainer-mode --enable-shared --with-mpi PYTHON=python3.5 LDFLAGS=-L/usr/lib/x86_64-linux-gnu/hdf5/openmpi -L/usr/local/lib -Wl,-rpath,/usr/lib/x86_64-linux-gnu/hdf5/openmpi:/usr/local/lib CPPFLAGS=-I/usr/include/hdf5/openmpi -I/usr/local/include
```

oskooi:
```
RPATH_FLAGS="-Wl,-rpath,/usr/lib/x86_64-linux-gnu/hdf5/openmpi:/usr/local/lib"
MY_LDFLAGS="-L/usr/lib/x86_64-linux-gnu/hdf5/openmpi -L/usr/local/lib ${RPATH_FLAGS}"
MY_CPPFLAGS="-I/usr/include/hdf5/openmpi -I/usr/local/include"

sh autogen.sh --enable-shared --with-mpi PYTHON=python3.5 LDFLAGS="${MY_LDFLAGS}" CPPFLAGS="${MY_CPPFLAGS}"
make
```

stevengj:
`meepgeom.hpp` is installed in `/usr/local/include/meep`, not `/usr/local/include`, so `-I/usr/local/include` won't cause it to be found by `#include "meepgeom.h"`

Furthermore, gcc searches the `-I` paths in [left to right order](https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html).  This means that the `-I$(top_srcdir)/src` specified in `AM_CPPFLAGS` by `libpympb/Makefile.am` will take precedence over any `-I` flags you specify in `CPPFLAGS`, because `CPPFLAGS` appears *after* `AM_CPPFLAGS` [in automake's flag ordering](https://www.gnu.org/software/automake/manual/html_node/Flag-Variables-Ordering.html#:~:text=CPPFLAGS%20is%20the%20user%20variable,when%20compiling%20C%20sources%20files.).

So, I don't think that change is correct.

PR #1519, opened 2021-03-05
use SWIG arginit

### LINKED ISSUE #1518: Passing a Python float in the `subpixel_maxeval` arg causes a segfault
Inadvertently passing a Python float in the `subpixel_maxeval` arg of the `mp.Simulation` constructor (e.g. `1000.0` ) will cause Meep to segfault. Python is much looser with its typing, so it would be nice to either raise an exception or automatically cast to an integer.

stevengj:
Whenever we use a SWIG `freearg` typemap, we need to pair it with a SWIG `arginit` typemap to ensure that errors don't cause segfaults.  Fixes #1518.

### DIFF
--- a/python/meep.i
+++ b/python/meep.i
@@ -736,8 +736,12 @@ meep::volume_list *make_volume_list(const meep::volume &v, int c,
     }
 }
 
+%typemap(arginit) GEOMETRIC_OBJECT {
+    $1.material = NULL;
+}
+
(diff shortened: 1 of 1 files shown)


stevengj:
(With this PR, passing a floating-point value for `subpixel_maxeval` produces `TypeError: in method 'create_structure_and_set_materials', argument 14 of type 'int'`.)

Issue #1434, opened 2020-11-25
Occasional segfault with Near2Far adjoint and MPI


smartalecH:
When running some basic near2far adjoint examples with MPI, there is an occasional segfault -- but at different locations. For example, when running the same script 3 different times, the segfault occurs at three different lines:

https://github.com/NanoComp/meep/blob/c844fd80e86ea8fbcc161dfc87be7c5d8057dd28/src/step.cpp#L195

https://github.com/NanoComp/meep/blob/e89ad6b67c52a5b9cd4adac86354e929d84f99c7/src/boundaries.cpp#L260

https://github.com/NanoComp/meep/blob/c844fd80e86ea8fbcc161dfc87be7c5d8057dd28/src/step.cpp#L155

Unfortunately, the behavior is completely unpredictable.

PR #1521, opened 2021-03-05
fix deadlock in add_srcdata via new require_source_components() function

stevengj:
@mochen4 has been seeing some deadlocks when using near2far adjoints on multiple processors.  I think this PR fixes the underlying problem.

`fields::add_srcdata` was calling `fields::require_component(c)` to ensure that the fields corresponding to the source component were added, but `require_component` is a *collective* function — it must be called from all processes with the same `c`.  Because of the low-level way in which `add_srcdata` is constructed, however, the list of sources can be different on each processor, and in particular sources were only added on processors whose chunks overlap with the DFT near-field region.

To solve this, I did some refactoring and added a new `fields::require_source_components()` function which can be called collectively *after* all of the sources are added, allocating the necessary fields.  Python now calls this by default after it adds sources, just in case one of the sources used `add_srcdata` (an `IndexedSource` in Python).

### DIFF
--- a/src/fields.cpp
+++ b/src/fields.cpp
@@ -532,9 +532,6 @@ void fields::_require_component(component c, bool aniso2d) {
     figure_out_step_plan();
     chunk_connections_valid = false; // connect_chunks() will synchronize this for us
   }
-  am_now_working_on(MpiAllTime);
-  if (sum_to_all(need_to_reconnect)) chunk_connections_valid = false;
-  finished_working();
 }
 
 void fields_chunk::remove_sources() {
--- a/src/fields.cpp
+++ b/src/fields.cpp
@@ -484,7 +484,9 @@ void fields::require_source_components() {
     }
   }
   int allneeded[NUM_FIELD_COMPONENTS];
+  am_now_working_on(MpiAllTime);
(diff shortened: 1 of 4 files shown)


smartalecH:
Fixes #1434

src/fields.cpp
```diff
@@ -472,15 +472,28 @@ bool fields_chunk::alloc_f(component c) {
...
+  int needed[NUM_FIELD_COMPONENTS];
+  memset(needed, 0, sizeof(needed));
+  for (int i = 0; i < num_chunks; i++) {
+    FOR_FIELD_TYPES(ft) {
+      for (src_vol *src = chunks[i]->sources[ft]; src; src = src->next)
+        needed[src->c] = 1;
+    }
+  }
+  int allneeded[NUM_FIELD_COMPONENTS];
+  or_to_all(needed, allneeded, NUM_FIELD_COMPONENTS);
```
oskooi, line 487: For accurate timing statistics, it would be useful to wrap this `or_to_all` statement with `am_now_working_on(MpiAllTime);` and `finished_working();`.

stevengj: Done.

PR #1527, opened 2021-03-10
fix factor in adjoint gradient from symmetry

mochen4:
Fix the factor when using symmetry in simulation

### DIFF
--- a/src/vec.cpp
+++ b/src/vec.cpp
@@ -1217,6 +1217,15 @@ int symmetry::multiplicity() const {
     return g;
 }
 
+int symmetry::multiplicity(vec &x) const {
+  int m = multiplicity();
+  for (int n=1; n<m; ++n){
+    if (transform(x,n) == x)
+      return n;
+  }
+  return m;
+}
+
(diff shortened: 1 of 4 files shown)


src/vec.cpp
```diff
@@ -1217,6 +1217,15 @@ int symmetry::multiplicity() const {
     return g;
 }
 
+int symmetry::multiplicity(vec &x) const {
+  int m = multiplicity();
+  for (int n=1; n<m; ++n){
+    if (transform(x,n) == x)
```
stevengj, line 1223: I'm worried that this test might fail due to roundoff errors.  Maybe it should apply to an `ivec` (integer coordinates) instead?

mochen4: Ok I will fix it

mochen4: I switched to ivec, and the code worked for me.

smartalecH:
Maybe add a test to python/tests/adjoint_solver.py?

We actually don't have *any* far field adjoint tests yet. 

I would also make sure you're based off master.

PR #1528, opened 2021-03-12
user-specified chunk layouts

### LINKED ISSUE #1510: user-specified chunk layouts (for parallelization)
Perhaps the easiest way to specify this would just be as a binary tree of "cuts".   i.e. you'd say "cut the cell in half at x=3", and then for first have "cut it in half at y=7", etcetera.  (Meep would then convert this into grid points given your resolution and cell size.)   You could also optionally specify MPI ranks at the leaves of the tree to control parallelization.

oskooi:
Closes #1510.

This PR adds a new `BinaryPartition` class to the Python API following the outline proposed by @stevengj in [#1510 (comment)](https://github.com/NanoComp/meep/issues/1510#issuecomment-791976424) which can be used to specify a cell partition comprised of arbitrary sized chunks. A `BinaryPartition` class object is then passed as the `chunk_layouts` parameter of the `Simulation` class object.

A new `binary_partition` class is added in C++ which is designed to be used with a new `split_by_binarytree` routine to divide the cell into chunks similar to the existing `split_by_cost` routine.

This PR is missing the SWIG typemaps (or a glue routine) to convert the `BinaryPartition` object from Python into a `binary_partition` object in C++. It is therefore not yet ready to be merged. This PR, when ready, will also include a unit test.


stevengj:
For a SWIG wrapper you need something like the following in `meep.i`:
```cxx
%typemap(in) binary_partition * {
    $1 = py_bp_to_bp($input);
    if(!$1) {
        SWIG_fail;
    }
}

%typemap(arginit) binary_partition * {
    $1 = NULL;
}

%typemap(freearg) binary_partition * {
    delete $1;
}
```
where you have defined a function:
```cxx
binary_partition *py_bp_to_bp(PyObject *bp) {
    ....
}
```
that converts the Python object to the C++ object.

stevengj:
For `py_bp_to_bp` you have to use the Python C API, something like:
```cxx
binary_partition *py_bp_to_bp(PyObject *bp) {
    binary_partition *bp = NULL;
    if (bp == Py_None) return bp;

    PyObject *id = PyObject_GetAttrString(bp, "id");
    PyObject *split_dir = PyObject_GetAttrString(bp, "split_dir");
    PyObject *split_pos = PyObject_GetAttrString(bp, "split_pos");
    PyObject *left = PyObject_GetAttrString(bp, "left");
    PyObject *right = PyObject_GetAttrString(bp, "right");

    if (!id || !split_dir || !split_pos || !left || !right) {
        // error....
    }

    if (PyLong_Check(id)) {
         bp = new binary_partition(PyLong_AsLong(id));
    else {
         bp = new binary_partition(direction(PyLong_AsLong(split_dir)), PyFloat_AsDouble(split_pos));
         bp->left = py_bp_to_bp(left);
         bp->right = py_bp_to_bp(right);
    }

    Py_XDECREF(id);
    Py_XDECREF(split_dir);
    Py_XDECREF(split_pos);
    Py_XDECREF(left);
    Py_XDECREF(right);
    return bp; 
}
```

stevengj:
We could also add a 
```cxx
%typecheck (SWIG_TYPECHECK_POINTER) binary_partition * {
    $1 = PyObject_IsInstance($input, py_binary_partition_type());
}
```
where `py_binary_partition_type` is defined in `typemap_utils.cpp` as:
```cxx
static PyObject *get_meep_mod() {
  // Return value: Borrowed reference
  static PyObject *meep_mod = NULL;
  if (meep_mod == NULL) { meep_mod = PyImport_ImportModule("meep"); }
  return meep_mod;
}

static PyObject *py_binary_partition_type() {
  // Return value: Borrowed reference
  static PyObject *bp_type = NULL;
  if (bp_type == NULL) {
    bp_type = PyObject_GetAttrString(get_meep_mod(), "BinaryPartition");
  }
  return bp_type;
}
```

oskooi:
The following error appears during `make` with the above changes applied to `meep.i` and `typemap_utils.cpp`:
```
meep-python.cxx: In function ‘PyObject* _wrap_py_bp_to_bp(PyObject*, PyObject*)’:
meep-python.cxx:130336:42: error: ‘py_bp_to_bp’ was not declared in this scope
       result = (meep::binary_partition *)py_bp_to_bp(arg1);
```

oskooi:
Moving `py_bp_to_bp` from `meep.i` into `typemap_utils.cpp` and running `make` produces this error:
```
typemap_utils.cpp: In function ‘meep::binary_partition* py_bp_to_bp(PyObject*)’:
typemap_utils.cpp:1034:29: error: declaration of ‘meep::binary_partition* bp’ shadows a parameter
     meep::binary_partition *bp = NULL;
```

python/typemap_utils.cpp
```diff
@@ -1029,3 +1029,49 @@ static PyObject *gobj_list_to_py_list(geometric_object_list *objs) {
 
   return py_res;
 }
+
+static meep::binary_partition *py_bp_to_bp(PyObject *bp) {
```
stevengj, line 1033: Call the parameter `pybp`

python/simulation.py
```diff
@@ -5171,3 +5172,44 @@ def merge_subgroup_data(data):
...
+                if isinstance(data[0],tuple) and len(data[0]) == 2:
+                    self.split_dir = data[0][0]
+                    self.split_pos = data[0][1]
+                else:
+                    raise ValueError("expecting 2-tuple (split_dir,split_pos) but got {}".format(data[0]))
+                self.left = BinaryPartition(data=data[1])
+                self.right = BinaryPartition(data=data[2])
+            elif isinstance(data,int):
+                self.id = data
+            else:
```
stevengj, line 5207: Alternatively, you could just write a function
```py
def binarypartition(list):
    ....
```
that just constructs the C++ `meep::binary_partition` object directly (by calling its SWIG-wrapped constructors), which you can then pass directly to C++ (with no additional typemaps or conversion functions).

oskooi:
This feature seems to be working now. The current setup involves creating a `BinaryPartition` class object in Python and passing it as the `chunk_layout` parameter of the `Simulation` constructor.

A unit test has been added which verifies that the chunk ids and volumes specified in a given `BinaryPartition` class object (based on the example from #1510) are equivalent to the actual chunks created during the `structure` class initialization.

As another verification, the chunk layout for the 2d test case can be visualized using the `visualize_chunks` routine:

```py
import meep as mp
import matplotlib
matplotlib.use('agg')

chunk_layout = mp.BinaryPartition(data=[ (mp.X,-2.0), 0, [ (mp.Y,1.5), [ (mp.X,3.0), 1, [ (mp.Y,-0.5), 4,3 ] ], 2 ] ])

cell_size = mp.Vector3(10.0,5.0,0)

		    resolution=10,
                    chunk_layout=chunk_layout)

sim.init_sim()
sim.visualize_chunks()
plt.savefig('bp_chunk_layout.png',dpi=150,bbox_inches='tight')
```

bp_chunk_layout

oskooi:
An example demonstrating the use of this feature (based on the results shown above) has been added as a new section to the [Parallel Meep](https://meep.readthedocs.io/en/latest/Parallel_Meep/) page of the user manual.

doc/docs/Parallel_Meep.md
```diff
@@ -63,6 +63,38 @@ For comparison, consider the scenario where the optimization runs on just a sing
 
 Note: for optimization studies involving *random* initial conditions, the seed of the random number generator must be specified otherwise each process will have a different initial condition which will cause a crash. For example, if you are initializing the design variables with `numpy.random.rand`, then you should call `numpy.random.seed(...)` to set the same `numpy.random` seed on every process.
 
+### User-Specified Cell Partition
+
+An alternative to having Meep automatically partition the cell at runtime into chunks based on the number of MPI processes is to manually specify the cell partition via the `chunk_layout` parameter of the `Simulation` constructor as a [`BinaryPartition`](Python_User_Interface.md#binarypartition) class object. This is based on representing an arbitrary cell partition as a binary tree for which the nodes define "cuts" at a given point (e.g., -4.5, 6.3) along a given cell direction and the leaves define a unique integer-valued chunk ID (equivalent to the rank of the MPI process for that chunk).
```
stevengj, line 68: Key point: this is not a "chunk ID", it is a "process ID" which *must* be between 0 and #processes–1 (inclusive).  Note also that the same process ID can be assigned to as many chunks as you want, which just means that process timesteps multiple chunks.

python/simulation.py
```diff
@@ -5171,3 +5175,45 @@ def merge_subgroup_data(data):
...
+        and `split_pos` define the splitting direction (i.e., `mp.X`, `mp.Y`, `mp.Z`) and position (e.g., `3.5`,
+        `-4.2`, etc.) and `left` and `right` are the two branches (themselves `BinaryPartition` objects)
+        or (b) a leaf with integer value `id` in the range [0, `num_chunks`-1] for the chunk id, (2) a node
+        defined using `split_dir`, `split_pos`, `left`, and `right`, or (3) a leaf with `id`. This input format
+        enables specifying the binary tree using either a single list for the entire tree or defining the
+        nodes and leaves individually.
+        """
+        self.split_dir = None
+        self.split_pos = None
+        self.id = None
```
stevengj, line 5197: Maybe rename this to `proc` to be consistent with the rest of the source code's name for a process ID.

doc/docs/Parallel_Meep.md
```diff
@@ -63,6 +63,38 @@ For comparison, consider the scenario where the optimization runs on just a sing
...
+
+<center>
+![](images/chunk_division_binary_tree.png)
+</center>
+
+This binary tree can be described as a list of lists where each list entry is `[ (split_dir,split_pos), left, right ]` for which `split_dir` and `split_pos` define the splitting direction and position, and `left` and `right` are the left and right branches which can be either another list defining a new node or a chunk ID. There are two items to note when converting the binary tree shown above into a `BinaryPartition` class object: (1) by default, Meep defines the cell origin to be at the geometric center of the cell (i.e., at location (5.0,2.5) in the coordinate system of the figure) and thus all splitting coordinates must be shifted accordingly, and (2) MPI ranks must be defined in the range [0,`num_chunks`-1] which requires that all chunk IDs be shifted down by one. Based on these specifications, the cell partition from above can be set up as follows:
```
stevengj, line 76: I think the figure should should the correct (Meep) coordinate system, rather than showing the wrong coordinate system and warning the reader in the text.

src/structure_dump.cpp
```diff
@@ -414,6 +414,55 @@ void structure::set_chiP_from_file(h5file *file, const char *dataset, field_type
...
+  split_pos = _split_pos;
+  id = -1;
+  left = NULL;
+  right = NULL;
+}
+
+static void split_by_binarytree(grid_volume gvol,
+                                std::vector<grid_volume> &result_gvs,
+                                std::vector<int> &result_ids,
+                                const binary_partition *bp) {
```
stevengj, line 436: One option would be to change `split_by_const` etcetera so that they actually return a `binary_partition` and then we call `split_by_binarytree`.    That would make it easy to export the tree if we want, and would ensure that the `binary_partition` code gets further testing (because now all splitting would pass through it).

doc/docs/Parallel_Meep.md
```diff
@@ -63,6 +63,38 @@ For comparison, consider the scenario where the optimization runs on just a sing
...
+cell_size = mp.Vector3(10.0,5.0,0)
+
+sim = mp.Simulation(cell_size=cell_size,
+                    resolution=10,
+                    chunk_layout=chunk_layout)
+
+sim.visualize_chunks()
+plt.savefig('chunk_layout.png',dpi=150,bbox_inches='tight')
+```
+This example can be run using two different approaches: (1) with the number of MPI processes exactly equivalent to the number of user-specified chunks (i.e., `mpirun -np 5 python chunk_layout_example.py`) or (2) using *any* number of MPI processes and letting Meep automatically distribute the MPI ranks among the five chunks by additionally specifying `num_chunks=5` in the `Simulation` constructor.
```
stevengj, line 94: This is wrong: You have to use ≥ 6 processes, but then only 6 of them will be used and the rest will be idle.

src/structure_dump.cpp
```diff
@@ -457,20 +507,22 @@ void structure::load_chunk_layout(const char *filename, boundary_region &br) {
...
-void structure::load_chunk_layout(const std::vector<grid_volume> &gvs, boundary_region &br) {
+void structure::load_chunk_layout(const std::vector<grid_volume> &gvs,
+                                  const std::vector<int> &ids,
+                                  boundary_region &br) {
   // Recreate the chunks with the new grid_volumes
   for (int i = 0; i < num_chunks; ++i) {
     if (chunks[i]->refcount-- <= 1) delete chunks[i];
-    int proc = i * count_processors() / num_chunks;
-    chunks[i] = new structure_chunk(gvs[i], v, Courant, proc);
+    chunks[i] = new structure_chunk(gvs[i], v, Courant, ids[i]);
```
stevengj, line 525: Change to `ids[i] % count_processors()` and then we can use it with any number of processes.

python/simulation.py
```diff
@@ -5171,3 +5175,47 @@ def merge_subgroup_data(data):
...
+    """
+    def __init__(self, data=None, split_dir=None, split_pos=None, left=None, right=None, proc_id=None):
+        """
+        The constructor accepts three separate groups of arguments: (1) `data`: a list of lists where each
+        list entry is either (a) a node defined as `[ (split_dir,split_pos), left, right ]` for which `split_dir`
+        and `split_pos` define the splitting direction (i.e., `mp.X`, `mp.Y`, `mp.Z`) and position (e.g., `3.5`,
+        `-4.2`, etc.) and `left` and `right` are the two branches (themselves `BinaryPartition` objects)
+        or (b) a leaf with integer value `proc_id` in the range [0, `num_chunks`-1] for the process ID, (2) a node
+        defined using `split_dir`, `split_pos`, `left`, and `right`, or (3) a leaf with `proc_id`. Two items to note:
+        (1) the process ID *must* be between 0 and the number of processes and (2) the same process ID can be assigned
```
stevengj, line 5192: processes-1

doc/docs/Parallel_Meep.md
```diff
@@ -63,6 +63,38 @@ For comparison, consider the scenario where the optimization runs on just a sing
...
+cell_size = mp.Vector3(10.0,5.0,0)
+
+sim = mp.Simulation(cell_size=cell_size,
+                    resolution=10,
+                    chunk_layout=chunk_layout)
+
+sim.visualize_chunks()
+plt.savefig('chunk_layout.png',dpi=150,bbox_inches='tight')
+```
+This example can be run by specifying the number of MPI processes exactly equivalent to the number of user-specified chunks (i.e., `mpirun -np 5 python chunk_layout_example.py`) or (2) using *any* number of MPI processes and letting Meep automatically distribute the MPI ranks among the five chunks by additionally specifying `num_chunks=5` in the `Simulation` constructor.
```
stevengj, line 94: No need for `num_chunks`, but it won't actually use more than 5 processes.

stevengj:
I think the binary_partition needs to be passed to the structure *constructor*, rather than choosing the chunk division and then overwriting it with the new chunk division in `load_chunk_layout` as you are doing now.   (e.g. you shouldn't have to pass `num_chunks` as an additional parameter — it should be determined from the partition.)

oskooi:
Modifying the `structure` class constructor to accept a `binary_partition` parameter and revamping `split_by_cost` to return a `binary_partition` object would require a lot of changes which should probably be a separate PR. For now, I just added the following comment to the doc strings for `BinaryPartition` and the tutorial example: 

```
The `num_chunks` parameter of the `Simulation` constructor must be set to the
number of chunks specified by the binary tree.
```
As long as the `num_chunks` parameter is set to the number of chunks in the binary tree, the simulation will run with *any* number of MPI processes (i.e., run using not just calling `visualize_chunks` but actually time stepping the fields).

src/structure_dump.cpp
```diff
@@ -457,20 +507,22 @@ void structure::load_chunk_layout(const char *filename, boundary_region &br) {
...
 
   delete[] origins;
   delete[] nums;
 }
 
-void structure::load_chunk_layout(const std::vector<grid_volume> &gvs, boundary_region &br) {
+void structure::load_chunk_layout(const std::vector<grid_volume> &gvs,
+                                  const std::vector<int> &ids,
+                                  boundary_region &br) {
   // Recreate the chunks with the new grid_volumes
```
stevengj, line 522: Should `abort` if length of `ids` doesn't match `num_chunks`

doc/docs/Parallel_Meep.md
```diff
@@ -63,6 +63,38 @@ For comparison, consider the scenario where the optimization runs on just a sing
...
+
+chunk_layout = mp.BinaryPartition(data=[ (mp.X,-2.0), 0, [ (mp.Y,1.5),
+                                         [ (mp.X,3.0), 1, [ (mp.Y,-0.5), 4, 3 ] ], 2 ] ])
+
+cell_size = mp.Vector3(10.0,5.0,0)
+
+sim = mp.Simulation(cell_size=cell_size,
+                    resolution=10,
+                    num_chunks=5,
+                    chunk_layout=chunk_layout)
```
stevengj, line 90: Can't it get `num_chunks` from the `chunk_layout`?

stevengj: In particular:
```
def numchunks(bp=None):
    if bp is None:
        return 0
    return min(numchunks(bp.left) + numchunks(bp.right), 1)
```

doc/docs/Parallel_Meep.md
```diff
@@ -63,6 +63,38 @@ For comparison, consider the scenario where the optimization runs on just a sing
...
+
+sim.visualize_chunks()
+plt.savefig('chunk_layout.png',dpi=150,bbox_inches='tight')
+```
+
+For improved performance, it is important to order the chunks such that adjacent chunks are numbered consecutively. This is equivalent to assigning MPI ranks in [depth-first order](https://en.wikipedia.org/wiki/Depth-first_search) which ensures that adjacent chunks on the same MPI node communicate using shared memory rather than the network (which tends to be slower).
```
stevengj, line 96: it is important to order -> consider ordering

stevengj: ensures that ... -> tends to place adjacent chunks in the same MPI node (improving communication speed).

stevengj:
```suggestion
For improved performance, we recommend ordering the process IDs in [depth-first order](https://en.wikipedia.org/wiki/Depth-first_search) of the tree, which will tend to give spatially adjacent chunks nearby process IDs.  This increases the chance that adjacent chunks are on the same MPI node, improving communication speeds.
```

doc/docs/Parallel_Meep.md
```diff
@@ -63,6 +63,37 @@ For comparison, consider the scenario where the optimization runs on just a sing
...
+
+sim = mp.Simulation(cell_size=cell_size,
+                    resolution=10,
+                    chunk_layout=chunk_layout)
+
+sim.visualize_chunks()
+plt.savefig('chunk_layout.png',dpi=150,bbox_inches='tight')
+```
+
+For improved performance, consider ordering the chunks such that adjacent chunks are numbered consecutively. This is equivalent to assigning MPI ranks in [depth-first order](https://en.wikipedia.org/wiki/Depth-first_search) which ensures that adjacent chunks on the same MPI node communicate using shared memory rather than the network (which tends to be slower).
```
stevengj, line 95:
```suggestion
For improved performance, we recommend ordering the process IDs in [depth-first order](https://en.wikipedia.org/wiki/Depth-first_search) of the tree, which will tend to give spatially adjacent chunks nearby process IDs.  This increases the chance that adjacent chunks are on the same MPI node, improving communication speeds.
```

Issue #1131, opened 2020-02-19
rename omega parameters to frequency


oskooi:
#919 introduced an `omega` parameter for the frequency for various Python and C functions such as `get_epsilon_point`, `get_array`, `output_epsilon`, etc. Since `omega` does *not* refer to the angular frequency (which contains a 2π factor), it should be renamed to `frequency`. This is consistent with other Meep functions which have a frequency argument. However, to maintain backwards compatibility, `omega` should continue to be supported for those functions which currently have an `omega` parameter (but still removed from the documentation).

stevengj:
(I think the only way to implement this is to support both arguments, and give a deprecation warning if `omega` is passed.)

PR #1529, opened 2021-03-14
 remove deprecated features and minor tweaks to docs

oskooi:
Removes the deprecated `quiet` variable from the docs as well as the `omega` argument of several output functions including warnings that they are deprecated.

Also includes some minor tweaks to the docs.

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -1937,10 +1937,6 @@ class Simulation(object):
 
         self._is_initialized = True
 
-    def init_fields(self):
-        warnings.warn('init_fields is deprecated. Please use init_sim instead', DeprecationWarning)
-        self.init_sim()
-
     def initialize_field(self, cmpnt, amp_func):
         """
         Initialize the component `c` fields using the function `func` which has a single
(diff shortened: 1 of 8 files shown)


stevengj:
The `omega` parameter was renamed in #1131.

stevengj:
`quiet` became redundant in #1349?

python/simulation.py
```diff
@@ -5088,18 +5062,6 @@ def complexarray(re, im):
...
-def quiet(quietval=True):
-    """
-    Meep ordinarily prints various diagnostic and progress information to standard output.
-    This output can be suppressed by calling this function with `True` (the default). The
-    output can be enabled again by passing `False`. This sets a global variable, so the
-    value will persist across runs within the same script.
-
-    This function is deprecated, please use the [Verbosity](#verbosity) class instead.
-    """
-    verbosity(int(not quietval))
```
stevengj, line 5100: Maybe just add a deprecation warning?

stevengj:
Since it's been less than a year for `quiet`, maybe remove them from the documentation but keep them in the code with a deprecation warning message?

python/simulation.py
oskooi: Done.

oskooi:
The `quiet` function has been reinstated with a deprecation warning and its documentation has been removed from the [Python User Interface](https://meep.readthedocs.io/en/latest/Python_User_Interface) page of the user manual.

PR #1534, opened 2021-03-23
copy constructor for dft_chunk class

oskooi:
In preparation for a new feature to provide checkpointing using [protocol buffers](https://developers.google.com/protocol-buffers), this PR adds a new copy constructor for the `dft_chunk` class (which was missing) similar to what already exists for the `fields_chunk` and `structure_chunk` classes.

### DIFF
--- a/src/dft.cpp
+++ b/src/dft.cpp
@@ -112,6 +112,43 @@ dft_chunk::dft_chunk(fields_chunk *fc_, ivec is_, ivec ie_, vec s0_, vec s1_, ve
   next_in_dft = data->dft_chunks;
 }
 
+dft_chunk::dft_chunk(const dft_chunk &thed, dft_chunk *_next_in_chunk, dft_chunk *_next_in_dft) {
+  omega = thed.omega;
+  c = thed.c;
+  N = thed.N;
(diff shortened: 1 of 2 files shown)


src/dft.cpp
```diff
@@ -112,6 +112,43 @@ dft_chunk::dft_chunk(fields_chunk *fc_, ivec is_, ivec ie_, vec s0_, vec s1_, ve
...
+dft_chunk::dft_chunk(const dft_chunk &thed, dft_chunk *_next_in_chunk, dft_chunk *_next_in_dft) {
+  omega = thed.omega;
+  c = thed.c;
+  N = thed.N;
+  const int Nomega = thed.omega.size();
+  dft = new complex<realnum>[N * Nomega];
+  for (size_t i = 0; i < N * Nomega; ++i)
+    dft[i] = thed.dft[i];
+  next_in_chunk = _next_in_chunk;
+  next_in_dft = _next_in_dft;
```
stevengj, line 124: I would just set these to NULL and not pass them.

stevengj:
It's not clear that we need a copy constructor here — to migrate or checkpoint a dft_chunk, you'd just send/save it directly, not make a copy first.

PR #1539, opened 2021-03-29
computing the normal vector to an interface for a MaterialGrid using analytic gradient

### LINKED ISSUE #1537: computing the normal vector to an interface for a MaterialGrid
The following schematic shows the analytic expressions necessary to compute the gradient of the `weights` array `u` for a `MaterialGrid` in 2d from which the normal vector to an interface can be obtained at any Yee grid point using bilinear interpolation. In the figure, the `MaterialGrid` are the green points and the Yee grid points for E<sub>x</sub> and E<sub>y</sub> are the blue and red points, respectively.

Meep_Adjoint_Meeting_Notes

To set this up, we will need to modify the existing `material_function::normal_vector` routine which uses spherical quadrature such that for the case of a `MaterialGrid` it uses the analytic formulas shown above:

https://github.com/NanoComp/meep/blob/272bd0caa1cd88665389c7edfb14f3194db7bcbb/src/anisotropic_averaging.cpp#L40-L64

We can then verify that the new approach to computing the normal vector is working correctly by comparing with the results obtained using spherical quadrature for e.g. a test case involving a cylinder (similar to [#1522 (comment)](https://github.com/NanoComp/meep/pull/1522#issue-586470739)).

oskooi:
Closes #1537.

Initial attempt to add a new function `matgrid_grad` for computing the normal vector to an interface for a `MaterialGrid` using the analytic gradient via the formulas for bilinear interpolation shown in #1537. For now, it is assumed there is only one `MaterialGrid` object in 2d. The generalization to multiple overlapping `MaterialGrid`s in 3d will be added later once this base case is working.

This PR compiles and the normal vector for a test case involving a circle in 2d seems reasonably correct but for some reason the fields are diverging:

```py
import numpy as np
import meep as mp

res = 20 ## pixels/um    

design_region_res = 50  ## MaterialGrid resolution   

sigma = 3.0; ## standard deviation for Gaussian filter kernel    

cell_size = mp.Vector3(1,1,0)

rad = 0.301943

design_region_resolution = design_region_res
xv, yv = np.meshgrid(x,y)
filtered_design_params = gaussian_filter(design_params, sigma=sigma, output=np.double)

                          mp.air,
                          mp.Medium(index=3.5),
                          weights=filtered_design_params,
                          do_averaging=True,
                          beta=0,
                          eta=0.5)

                     material=matgrid)]

fcen = 0.3
df = 0.2*fcen
                     component=mp.Hz,
                     center=mp.Vector3(-0.1057,0.2094,0))]

k_point = mp.Vector3(0.3892,0.1597,0)

                    cell_size=cell_size,
                    geometry=geometry,
                    sources=sources,
                    k_point=k_point)

h = mp.Harminv(mp.Hz, mp.Vector3(0.3718,-0.2076), fcen, df)
sim.run(mp.after_sources(h),
        until_after_sources=200)

for m in h.modes:
    print("harminv:, {}, {}, {}".format(res,m.freq,m.Q))
```

*output from running script*
```
-----------
Initializing structure...
time for choose_chunkdivision = 7.2417e-05 s
Working in 2D dimensions.
Computational cell is 1 x 1 x 0 with resolution 20
     block, center = (0,0,0)
          size (1,1,0)
time for set_epsilon = 0.202574 s
-----------
Meep: using complex fields.
  File "circle2d_matgrid_test.py", line 56, in <module>
    until_after_sources=200)
  File "simulation.py", line 3645, in run
    self._run_sources_until(until_after_sources, step_funcs)
  File "simulation.py", line 2232, in _run_sources_until
    self._run_until(new_conds, step_funcs)
  File "simulation.py", line 2196, in _run_until
    self.fields.step()
  File "__init__.py", line 3948, in step
    return _meep.fields_step(self)
RuntimeError: meep: simulation fields are NaN or Inf
```

### DIFF
--- a/src/meepgeom.cpp
+++ b/src/meepgeom.cpp
@@ -308,84 +308,119 @@ bool is_metal(meep::field_type ft, const material_type *material) {
     }
 }
 
-meep::vec matgrid_grad(meep::field_type ft, const meep::volume &v, geom_box_tree tp, int oi) {
-  meep::vec gradient(zero_vec(v.dim));
-  vector3 pc = vec_to_vector3(v.center());
+meep::vec material_grid_grad(vector3 p, material_data *md) {
+  if (!is_material_grid(md)) { meep::abort("Invalid material grid detected.\n"); }
+
+  meep::vec gradient(zero_vec(dim));
+  double *data = md->weights;
+  int nx = md->grid_size.x;
+  int ny = md->grid_size.y;
+  int nz = md->grid_size.z;
+  double rx = p.x;
+  double ry = p.y;
+  double rz = p.z;
+  int stride = 1;
   int x1, y1, z1, x2, y2, z2;
   double dx, dy, dz;
 
-  if (tp) {
-    vector3 p = to_geom_box_coords(pc, &tp->objects[oi]);
-    material_data *md = (material_data *)tp->objects[oi].o->material;
+  /* mirror boundary conditions for r just beyond the boundary */
+  rx = rx < 0.0 ? -rx : (rx > 1.0 ? 1.0 - rx : rx);
+  ry = ry < 0.0 ? -ry : (ry > 1.0 ? 1.0 - ry : ry);
+  rz = rz < 0.0 ? -rz : (rz > 1.0 ? 1.0 - rz : rz);
 
-    double rx = p.x;
-    double ry = p.y;
-    double rz = p.z;
+  /* get the point corresponding to r in the epsilon array grid: */
+  x1 = meep::mirrorindex(int(rx * nx), nx);
+  y1 = meep::mirrorindex(int(ry * ny), ny);
+  z1 = meep::mirrorindex(int(rz * nz), nz);
 
-    double *data = md->weights;
+  /* get the difference between (x1,y1,z1) and the actual point */
+  dx = rx * nx - x1 - 0.5;
+  dy = ry * ny - y1 - 0.5;
+  dz = rz * nz - z1 - 0.5;
 
-    int nx = md->grid_size.x;
-    int ny = md->grid_size.y;
-    int nz = md->grid_size.z;
-    int stride = 1;
+  /* get the other closest point in the grid, with mirror boundaries: */
+  x2 = meep::mirrorindex(dx >= 0.0 ? x1 + 1 : x1 - 1, nx);
+  y2 = meep::mirrorindex(dy >= 0.0 ? y1 + 1 : y1 - 1, ny);
+  z2 = meep::mirrorindex(dz >= 0.0 ? z1 + 1 : z1 - 1, nz);
 
-    /* mirror boundary conditions for r just beyond the boundary */
-    rx = rx < 0.0 ? -rx : (rx > 1.0 ? 1.0 - rx : rx);
-    ry = ry < 0.0 ? -ry : (ry > 1.0 ? 1.0 - ry : ry);
-    rz = rz < 0.0 ? -rz : (rz > 1.0 ? 1.0 - rz : rz);
-
-    /* get the point corresponding to r in the epsilon array grid: */
-    x1 = meep::mirrorindex(int(rx * nx), nx);
-    y1 = meep::mirrorindex(int(ry * ny), ny);
-    z1 = meep::mirrorindex(int(rz * nz), nz);
-
-    /* get the difference between (x1,y1,z1) and the actual point */
-    dx = rx * nx - x1 - 0.5;
-    dy = ry * ny - y1 - 0.5;
-    dz = rz * nz - z1 - 0.5;
-
-    /* get the other closest point in the grid, with mirror boundaries: */
-    x2 = meep::mirrorindex(dx >= 0.0 ? x1 + 1 : x1 - 1, nx);
-    y2 = meep::mirrorindex(dy >= 0.0 ? y1 + 1 : y1 - 1, ny);
-    z2 = meep::mirrorindex(dz >= 0.0 ? z1 + 1 : z1 - 1, nz);
-
-    /* take abs(d{xyz}) to get weights for {xyz} and {xyz}2: */
-    bool signflip_dx = false, signflip_dy = false, signflip_dz = false;
-    if (dx != fabs(dx)) {
-      dx = fabs(dx);
-      signflip_dx = true;
-    }
-    if (dy != fabs(dy)) {
-      dy = fabs(dy);
-      signflip_dy = true;
-    }
-    if (dz != fabs(dz)) {
-      dz = fabs(dz);
-      signflip_dz = true;
-    }
-/* define a macro to give us data(x,y,z) on the grid, in row-major order: */
-#define D(x, y, z) (data[(((x)*ny + (y)) * nz + (z)) * stride])
-    double du_dx = (signflip_dx ? -1.0 : 1.0) *
-      (((-D(x1, y1, z1) + D(x2, y1, z1)) * (1.0 - dy) +
-        (-D(x1, y2, z1) + D(x2, y2, z1)) * dy) * (1.0 - dz) +
-       ((-D(x1, y1, z2) + D(x2, y1, z2)) * (1.0 - dy) +
-        (-D(x1, y2, z2) + D(x2, y2, z2)) * dy) * dz);
-    double du_dy = (signflip_dy ? -1.0 : 1.0) *
-      ((-(D(x1, y1, z1) * (1.0 - dx) + D(x2, y1, z1) * dx) +
-        (D(x1, y2, z1) * (1.0 - dx) + D(x2, y2, z1) * dx)) * (1.0 - dz) +
-       (-(D(x1, y1, z2) * (1.0 - dx) + D(x2, y1, z2) * dx) +
-        (D(x1, y2, z2) * (1.0 - dx) + D(x2, y2, z2) * dx)) * dz);
-    double du_dz = (signflip_dz ? -1.0 : 1.0) *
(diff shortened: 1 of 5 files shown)


src/meepgeom.cpp
```diff
@@ -307,6 +307,70 @@ bool is_metal(meep::field_type ft, const material_type *material) {
     }
 }
 
+/* implement mirror boundary conditions for i outside 0..n-1: */
+static int mirrorindex(int i, int n) { return i >= n ? 2 * n - 1 - i : (i < 0 ? -1 - i : i); }
```
smartalecH, line 311: Looks like this is already defined in

https://github.com/NanoComp/meep/blob/899f6d039346dd9837d1142f48c5cceed595cb56/src/meepgeom.cpp#L2377

src/meepgeom.cpp
```diff
@@ -307,6 +307,70 @@ bool is_metal(meep::field_type ft, const material_type *material) {
...
+    dz = fabs(dz);
+
+/* define a macro to give us data(x,y,z) on the grid, in row-major order: */
+#define D(x, y, z) (data[(((x)*ny + (y)) * nz + (z)) * stride])
+
+    meep::realnum du_dx = (1-dy)*(-D(x1,y2,0)+D(x2,y2,0))+dy*(-D(x1,y1,0)+D(x2,y1,0));
+    gradient.set_direction(meep::X, du_dx);
+    meep::realnum du_dy = (1-dx)*(-D(x2,y1,0)+D(x2,y2,0))+dx*(-D(x1,y1,0)+D(x1,y2,0));
+    gradient.set_direction(meep::Y, du_dy);
+
```
smartalecH, line 367: We still need the gradient in the Z direction... but it makes sense to start with 2D since most use cases will project a 2D design.

src/meepgeom.cpp
smartalecH: Ah I see you moved it. Maybe locate all the gradient codes next to each other?

stevengj:
Just take the derivative with respect to (dx,dy,dz) of the `linear_interpolate` function.  Then, because of the `dx = fabs(dx)` etcetera lines, you will need to flip the sign of that component of the gradient if dx etcetera had its sign flipped.

src/meepgeom.cpp
```diff
@@ -307,6 +307,70 @@ bool is_metal(meep::field_type ft, const material_type *material) {
...
+#undef D
+  }
+
+  return gradient/abs(gradient);
```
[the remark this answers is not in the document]

oskooi:
Following comments from @stevengj, two fixes have been made:
1. the normal vector is only computed in regions in which ε is non uniform
2. the sign of the gradient component is flipped depending on the sign of `dx` and `dy` in the interpolation scheme

With these changes, the fields no longer blow up and the convergence results for a 2d ring resonator comparing the normal vector computed using two different methods (analytic gradient vs. spherical quadrature (from the master branch)) as part of subpixel smoothing applied to the `MaterialGrid` (the test case shown at the top of this PR) now mostly agree:

relerr_vs_resolution_analytic_gradient

The discrepancy in these results is likely due to the fact that spherical quadrature is using a small finite number of quadrature points:

https://github.com/NanoComp/meep/blob/2dcfa9fcc226e07bf3d177ba8318c867a79567a9/src/sphere-quad.cpp#L207

In the limit of the number of quadrature points approaching infinity, the two sets of results should become identical.

The only thing left to do in this PR is to generalize the gradient calculation to 3d. (Unfortunately, it will become harder to validate the gradient normal implementation in 3d by comparison with spherical quadrature since the latter is much less accurate in 3d than 2d.)

oskooi:
Support for computing the gradient in 3d has also been added.

Also, the check to return a non-zero gradient for regions in which ε is non-uniform has been modified slightly: rather than requiring the weights array `u(x,y)` to have different values at each of the four (in 2d) nearest `MaterialGrid` points, the norm of the gradient is required instead to be non-zero (i.e., > 1e-8). With this change, the test results now are in better agreement with those obtained using spherical quadrature:

relerr_vs_resolution_analytic_gradient

In this plot, the results for the analytic gradient also show less irregularities than those shown in the previous comment.

This PR is now ready to be merged.

src/meepgeom.cpp
oskooi: Fixed. The gradient computation now supports 3d.

src/meepgeom.cpp
```diff
@@ -307,6 +307,70 @@ bool is_metal(meep::field_type ft, const material_type *material) {
...
+/* implement mirror boundary conditions for i outside 0..n-1: */
+static int mirrorindex(int i, int n) { return i >= n ? 2 * n - 1 - i : (i < 0 ? -1 - i : i); }
+
+meep::vec matgrid_grad(meep::field_type ft, const meep::volume &v, geom_box_tree tp, int oi) {
+  meep::vec gradient(zero_vec(v.dim));
+  vector3 pc = vec_to_vector3(v.center());
+  int x1, y1, z1, x2, y2, z2;
+  meep::realnum dx, dy, dz;
+
+  // assume there is only a single MATERIAL_GRID (i.e., no overlapping grids)
```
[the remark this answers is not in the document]
stevengj: `material_grids_addgradient_point` should probably throw an error if the grid is not `U_MEAN` or `U_DEFAULT` (or whatever it is called), since it looks like those are currently the only ones supported.

In the `U_MEAN` case, we just need the same type of loop to take the average of the gradient vectors.

stevengj:
(Note that we can't use subpixel for optimization until we get the backpropagation working too, since otherwise the gradient will be inconsistent and the optimizer may get stuck.)

src/meepgeom.cpp

src/meepgeom.cpp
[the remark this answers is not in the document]
oskooi: Done.

oskooi:
I reorganized things a bit so that the two functions added as part of this PR `matgrid_grad` and `material_grid_grad` follow a similar form to the related functions `matgrid_val` and `material_grid_val`. Support for the use case of a material grid as the default material is now also provided. Also, `matgrid_grad` aborts with an error if called for overlapping material grids with either `U_MIN` or `U_PROD`. 

The results using this feature with the latest changes are the same as shown in the figure above.

src/meepgeom.cpp
```diff
@@ -307,6 +308,121 @@ bool is_metal(meep::field_type ft, const material_type *material) {
...
+  if (!tp && is_material_grid(&default_material)) {
+    p.x = geometry_lattice.size.x == 0 ? 0
+                                       : 0.5 + (p.x - geometry_center.x) / geometry_lattice.size.x;
+    p.y = geometry_lattice.size.y == 0 ? 0
+                                       : 0.5 + (p.y - geometry_center.y) / geometry_lattice.size.y;
+    p.z = geometry_lattice.size.z == 0 ? 0
+                                       : 0.5 + (p.z - geometry_center.z) / geometry_lattice.size.z;
+    gradient = material_grid_grad(p, (material_data *)default_material);
+    ++matgrid_val_count;
+  }
```
stevengj, line 418: This code to compute `matgrid_val_count` seems duplicated; can we factor it out into a single function?

src/meepgeom.cpp
```diff
@@ -307,6 +308,121 @@ bool is_metal(meep::field_type ft, const material_type *material) {
...
+    signflip_dx = true;
+  }
+  if (dy != fabs(dy)) {
+    dy = fabs(dy);
+    signflip_dy = true;
+  }
+  if (dz != fabs(dz)) {
+    dz = fabs(dz);
+    signflip_dz = true;
+  }
```
stevengj, line 359: Can we also share some code here with `linear_interpolate`?

stevengj:
Key points: this is an analytical derivative, vs a finite-difference-like approximation in the spherical quadrature, which is why it only needs a single point.  Furthermore, the subpixel smoothing algorithm only helps in the limit where the interface is "locally flat" in the pixel, which is why a single gradient suffices.

src/meepgeom.cpp
oskooi: Fixed.

src/meepgeom.cpp
oskooi: The only duplicated code from `matgrid_grad` and `matgrid_val` that could be reasonably separated out was the coordinate mapping for the case of the default material.

oskooi:
I think this should be ready to merge now.

oskooi:
I found and fixed a bug where assigning a `MaterialGrid` as the default material was actually not doing anything. The bug was in the function call `is_material_grid(&default_material)` appearing in `matgrid_grad`, `matgrid_val`, and `material_grids_addgradient_point` (the latter two originally added in #1242) which actually was *always* evaluating to `false` because `&default_material` (the address of a `material_type` object) should in fact just have been `default_material`. 

The `default_material` object is initialized in `init_libctl` as type `material_type`:

https://github.com/NanoComp/meep/blob/2dcfa9fcc226e07bf3d177ba8318c867a79567a9/src/meepgeom.cpp#L1942-L1952

The function `is_material_grid` has a single parameter which is also a `material_type`:

https://github.com/NanoComp/meep/blob/2dcfa9fcc226e07bf3d177ba8318c867a79567a9/src/meepgeom.cpp#L265-L267

By passing `&default_material`, the results were therefore undefined. By modifying the example shown above for a ring resonator, I verified that with this change assigning a `MaterialGrid` object to the `default_material` parameter (rather than the `geometry`) of the `Simulation` object now produces the correct result whereas the current master branch does not load any ring geometry:

```
                    cell_size=cell_size,
                    default_material=matgrid,
                    sources=sources,
                    k_point=k_point)
```

A unit test has also been added for the default material case as part of the existing `python/tests/material_grid.py`.

PR #1547, opened 2021-04-07
WIP: allow add_srcdata to place sources on non-owned points

stevengj:
For ongoing work in parallelizing the placement of adjoint sources for DFT fields using `add_srcdata`, we discovered that we need to place sources on non-owned points (since these are used for voxel-center averaging from the Yee grid).   After the sources are added, this unfortunately requires a "cleanup" step, implemented in this PR, which transfers sources from non-owned points to the processors where they are actually located.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -12,7 +12,7 @@ bicgstab.hpp meepgeom.hpp material_data.hpp adjust_verbosity.hpp
 libmeep_la_SOURCES = array_slice.cpp anisotropic_averaging.cpp 		\
 bands.cpp boundaries.cpp bicgstab.cpp casimir.cpp 	\
 cw_fields.cpp dft.cpp dft_ldos.cpp energy_and_flux.cpp 	\
-fields.cpp fields_dump.cpp loop_in_chunks.cpp h5fields.cpp h5file.cpp 	\
+fields.cpp fields_dump.cpp fix_boundary_sources.cpp loop_in_chunks.cpp h5fields.cpp h5file.cpp 	\
 initialize.cpp integrate.cpp integrate2.cpp material_data.cpp monitor.cpp mympi.cpp 	\
 multilevel-atom.cpp near2far.cpp output_directory.cpp random.cpp 	\
 sources.cpp step.cpp step_db.cpp stress.cpp structure.cpp structure_dump.cpp		\
--- a/src/boundaries.cpp
+++ b/src/boundaries.cpp
@@ -182,7 +182,7 @@ void fields::connect_chunks() {
   }
 }
 
-inline bool fields::on_metal_boundary(const ivec &here) {
+bool fields::on_metal_boundary(const ivec &here) {
   LOOP_OVER_DIRECTIONS(gv.dim, d) {
     if (user_volume.has_boundary(High, d) &&
         here.in_direction(d) == user_volume.big_corner().in_direction(d)) {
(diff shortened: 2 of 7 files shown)


smartalecH:
@mawc2019 any updates on this?

mawc2019:
@smartalecH , Steven pointed out some revisions:
1. Add `id = t.id;` in `src_time(const src_time &t)` below this [line](https://github.com/NanoComp/meep/blob/fix_boundary_sources/src/meep.hpp#L954) of meep.hpp;
2. To allocate an id for the source time, register the source time in `_create_time_profile` of objective.py as
```
    def _create_time_profile(self, fwidth_frac=0.1):
        srct = mp.GaussianSource(
            np.mean(self._frequencies),
            fwidth=fwidth_frac * np.mean(self._frequencies))
        self.sim.fields.register_src_time(srct.swigobj)
        return srct
```

But the id registration of source time still fails in the adjoint run of my modified `FourierFields`, which places the adjoint sources in a way similar to that in `Near2FarFields`. I also tried to register the source time of the adjoint source by adding `self.fields.register_src_time(self.sources.swigobj)` in `change_sources` of simulation.py, but got `'NoneType' object has no attribute 'register_src_time'`.

I plan to upload my modified `FourierFields` at first and then upload a revised version of this `fix_boundary_sources` branch, so that the problem may be clearer. However, I cannot upload my modified `FourierFields` at the moment, because the adjoint and finite-difference gradients are inconsistent when my modified `FourierFields` is based on the current master branch. In contrast, they were consistent when my modified `FourierFields` was based on the old master branch a few month ago. I am looking for the reason.

smartalecH:
Thanks for the update. Maybe submit a PR for what you have, and then we can all look at it.

The adjoint recombination step was recently overhauled. In theory it shouldn't cause any problems, but you never know.

mawc2019:
With the aforementioned corrections added and the [new FourierFields](https://github.com/NanoComp/meep/pull/1821/commits/01e5ac809f854a4af8c97843bcfa9e375e346857) merged, the revised version of this WIP is [here](https://github.com/NanoComp/meep/compare/master...mawc2019:fourier_adjoint_fix_boundary_sources).

When the FourierField adjoint solver needs the boundary fix (which may happen when `yee_grid=False`), the error message `bug: fix_boundary_sources called for non-registered source` is reported during the adjoint run. An example is as follows.

```
import meep as mp
import meep.adjoint as mpa
import numpy as np

Lx,Ly,pml_size = 2,4,1
design_width,design_height = Lx,0.4

sim = mp.Simulation(

nf = mpa.FourierFields(sim,mp.Volume(center=mp.Vector3(0,design_height+0.2),size=mp.Vector3(Lx,0)),mp.Ez,yee_grid=False)
ob_list = [nf]

def J(near_data):

x0 = 0.5*np.ones(Nx*Ny)
opt.update_design([x0])
f0, dJ_deps = opt()
```

It seems that `register_src_time` in sources.cpp does not fulfill its function for the FourierFields adjoint source, which is placed in a way similar to that in Near2FarFields.

>But the id registration of source time still fails in the adjoint run of my modified FourierFields, which places the adjoint sources in a way similar to that in Near2FarFields. I also tried to register the source time of the adjoint source by adding self.fields.register_src_time(self.sources.swigobj) in change_sources of simulation.py, but got 'NoneType' object has no attribute 'register_src_time'.

stevengj:
Closed favor of #1959.

Issue #1549, opened 2021-04-14
support runtime choice of precision


stevengj:
As commented in #1544, it would be good to be able to switch precision at runtime, by making `fields<T>`, `fields_chunk<T>`, `structure<T>`, and `structure_chunk<T>` template classes parameterized by the `realnum` type `T`.

oskooi:
I have started working on this by first going through `meep.hpp` and templating any class which currently includes a `realnum` type or contains a member function which calls/uses a templated class. The problem is the sheer number of changes that are necessary. Apart from templating the classes `fields`, `fields_chunk`, `structure`, and `structure_chunk`, there are also other classes `susceptibility`, `boundary_region`, `flux_vol`, etc, that are also templated. Next I would need to template the definitions of all the member functions of these classes which are scattered across numerous files. This is not a straightforward task given the dependencies among the various classes/functions. It will take some time to set this up correctly.

A more practical alternative is to keep the current compile-time option and just enable single precision by default. All we would then need to do is modify the unit tests so that everything passes.

oskooi:
Another complication to adding templates are two existing `typedef`s in `meep.hpp` which cannot be templated but which involve parameters that are templated classes `susceptibility` and `fields_chunk`:

https://github.com/NanoComp/meep/blob/2dcfa9fcc226e07bf3d177ba8318c867a79567a9/src/meep.hpp#L1318-L1323

https://github.com/NanoComp/meep/blob/2dcfa9fcc226e07bf3d177ba8318c867a79567a9/src/meep.hpp#L1476-L1479

stevengj:
I think that any competent C++ programmer should be able to do this — it doesn't require any particular knowledge of how Meep works, and is just a (tedious but straightforward) matter of changing a type in one place and then tracing how that change  propagates through the library, which is a common software-engineering task.

stevengj:
As an alternative, when you build with `--enable-single`, it could just rename the files when it installs from `meep` to `meep_single`.

oskooi:
Replaced by #1544, #1563, #1688, etc.

PR #1550, opened 2021-04-19
make single-precision floating point the default for fields and materials

oskooi:
Following recently added support for single precision (#1544) and prior to eventually providing a user option to select the floating precision at runtime (#1549), this PR makes single-precision floating point the default for the fields and material arrays. The main changes in this PR are updates to several of the unit tests which had previously been failing for single precision. These changes mainly involve reducing the threshold relative tolerance when comparing against hard-coded values. All tests are now passing.

Some additional comments:
- Two of the tests involving the CW solver (`array_metadata.py`) and the eigensolver (`eigfreq.py`) had to be essentially disabled because single precision was causing either slow or poor convergence for the iterative solvers.
- The results for the saturable absorption unit test `multilevel_atom.py` changed significantly (nearly two orders of magnitude) when switching from double to single precision. Not sure why this the case. For now, this test has been disabled.
- It seems that for certain materials involving Lorentzian susceptibilities, using single precision causes the fields to blow up at low resolutions (not the case when using double precision). This requires increasing the resolution to ensure stability. Since this is likely a feature and not a bug, perhaps this should be documented somewhere in the user manual?
- I haven't yet verified that the results for the tutorial examples are unaffected.


oskooi:
Looking into why the results for the unit test `multilevel_atom.py` are so different for single and double precision fields, it turns out the cause of the discrepancy is [roundoff error](https://en.wikipedia.org/wiki/Round-off_error)  because the simulation is run for a large number of time steps (~10<sup>7</sup>). I verified this by comparing the fields in the laser cavity (the structure used in the unit test) from the two runs. The norm of the difference in the fields gradually increases with the simulation time: ~10<sup>-8</sup> after the first 10<sup>2</sup> timesteps vs. ~10<sup>-2</sup> at the first 10<sup>5</sup> timesteps. That's roughly an increase of six orders of magnitude in the error given three orders of magnitude increase in the runtime.

The saturable absorption feature is a rare example of needing to run the simulation for a long time in order to obtain accurate results (particularly when comparing to an analytic model such as the steady-state ab-initio laser theory). Based on these results, we should just add a note to the [tutorial](https://meep.readthedocs.io/en/latest/Python_Tutorials/Multilevel_Atomic_Susceptibility/) to explicitly mention that this feature should preferably be used with double-precision fields.

stevengj:
I think it would cause too many problems for existing users to make this the default.     For the time being, until we can make it a runtime option, it will just have to be something for "power users" who are willing to recompile.

oskooi:
Replaced by #1560.

PR #1557, opened 2021-04-27
add missing test case in meepgeom::geom_boxes_intersect

oskooi:
While investigating ways to speed up [`grid_volume::get_cost()`](https://github.com/NanoComp/meep/blob/2dcfa9fcc226e07bf3d177ba8318c867a79567a9/src/vec.cpp#L952-L957) which in turn calls [`fragment_stats::compute()`](https://github.com/NanoComp/meep/blob/2dcfa9fcc226e07bf3d177ba8318c867a79567a9/src/meepgeom.cpp#L2155-L2160), I found a bug in the function `geom_boxes_intersect` (originally added in #340) involving a missing check. This means that the computation of the fragment statistics used by the cell partitioning algorithm via [`grid_volume::find_best_split`](https://github.com/NanoComp/meep/blob/2dcfa9fcc226e07bf3d177ba8318c867a79567a9/src/vec.cpp#L984-L1036)has thus far been inaccurate. Pending further testing, this could partly explain why this feature has not been able to consistently and reliably improve load balancing of large parallel simulations.

### DIFF
--- a/src/meepgeom.cpp
+++ b/src/meepgeom.cpp
@@ -1842,11 +1842,11 @@ static int geom_boxes_intersect(const geom_box *b1, const geom_box *b2) {
   /* true if the x, y, and z ranges all intersect. */
   return (
       (BETWEEN(b1->low.x, b2->low.x, b2->high.x) || BETWEEN(b1->high.x, b2->low.x, b2->high.x) ||
-       BETWEEN(b2->low.x, b1->low.x, b1->high.x)) &&
+       BETWEEN(b2->low.x, b1->low.x, b1->high.x) || BETWEEN(b2->high.x, b1->low.x, b1->high.x)) &&
       (BETWEEN(b1->low.y, b2->low.y, b2->high.y) || BETWEEN(b1->high.y, b2->low.y, b2->high.y) ||
-       BETWEEN(b2->low.y, b1->low.y, b1->high.y)) &&
+       BETWEEN(b2->low.y, b1->low.y, b1->high.y) || BETWEEN(b2->high.y, b1->low.y, b1->high.y)) &&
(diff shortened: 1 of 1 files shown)


src/meepgeom.cpp
```diff
@@ -1842,11 +1842,11 @@ static int geom_boxes_intersect(const geom_box *b1, const geom_box *b2) {
   /* true if the x, y, and z ranges all intersect. */
   return (
       (BETWEEN(b1->low.x, b2->low.x, b2->high.x) || BETWEEN(b1->high.x, b2->low.x, b2->high.x) ||
-       BETWEEN(b2->low.x, b1->low.x, b1->high.x)) &&
+       BETWEEN(b2->low.x, b1->low.x, b1->high.x) || BETWEEN(b2->high.x, b1->low.x, b1->high.x)) &&
```
stevengj, line 1845: Isn't this redundant with the other three tests?

stevengj: That is, there are only four ways two intervals `b1 = (...)` and `b2 = [...]` can overlap:

1.   `[ ...(...]...)` : `BETWEEN(b1->low.x, b2->low.x, b2->high.x) == true`
2.   `(...[...)...]`: `BETWEEN(b1->high.x, b2->low.x, b2->high.x) == true`
3. `[...(...)...]`: both of the above checks are`true`
4. `(...[...]...)`: `BETWEEN(b2->low.x, b1->low.x, b1->high.x) == true`

So, there is no need to also check `BETWEEN(b2->high.x, b1->low.x, b1->high.x)`

oskooi: Yes, you're right: the extra check is redundant and thus unnecessary.

PR #1560, opened 2021-04-29
WIP: template classes for specifying floating-point precision at runtime

### LINKED ISSUE #1549: support runtime choice of precision

oskooi:
Closes #1549.

This is an incomplete attempt (which does not yet compile) to template all the classes which contain member objects involving fields or materials arrays.


oskooi:
Replaced by #1544, #1563, #1688, etc.

PR #1571, opened 2021-05-17
expand frequency array of adjoint-solver unit test to span entire source bandwidth

oskooi:
Currently, the unit test for the adjoint solver involves verifying the accuracy of the gradient over three hard-coded frequency values which all lie entirely *within* the bandwidth of the Gaussian source. This PR replaces these three arbitrary values with eleven linearly spaced values in the range [`fcen-0.5*df`,`fcen+0.5*df`] to ensure that the entire bandwidth of the source (including the endpoints) is tested and over a finer grid. Also includes some minor fixes to the rest of the unit test.

(Separately, I verified using this unit test that the relative error of the gradient computed using the adjoint solver vs. a finite difference decreases with increasing resolution. This would be a useful additional check to include as part of this unit test but unfortunately increases the runtime to more than a couple of minutes.)

### DIFF
--- a/python/tests/adjoint_solver.py
+++ b/python/tests/adjoint_solver.py
@@ -51,14 +51,15 @@ sources = [mp.EigenModeSource(src=mp.GaussianSource(fcen,fwidth=df),
                               eig_band=1,
                               eig_parity=eig_parity)]
 
+nf = 11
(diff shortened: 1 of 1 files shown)


stevengj:
Closing this for now, as we discussed, since it's not clear what 11 frequencies tests that 3 frequencies does not.

PR #1572, opened 2021-05-17
Add ability to override MEEP_SINGLE from the compiler command line

dalexander:
Hi there!  I'd be interested to make it easier to inject "-DMEEP_SINGLE=1" from the compiler command line, without modifying source.  This change enables that.

It might also make sense to hook this up to the autogen/configure code, and if you tell me how you'd like it done, I'm happy to take a stab at it for you.

I know Ardavan is working on making this configurability available at runtime, and I don't want to step on his toes, so please just let me know if this isn't something you folks are interested in at this time.

### DIFF
--- a/src/meep.hpp
+++ b/src/meep.hpp
@@ -34,7 +34,10 @@ namespace meep {
    point rather than double precision (the default). The reduced
    precision can provide for up to a factor of 2X improvement in the
    time-stepping rate with generally negligible loss in accuracy. */
+#ifndef MEEP_SINGLE
 #define MEEP_SINGLE 0 // 1 for single precision, 0 for double
+#endif
+
 #if MEEP_SINGLE
 typedef float realnum;
 #else


stevengj:
The problem with this is that `make install` will then install a `meep.hpp` file that won't work with user code (for people using the C++ interface) unless they also know to use that `-D flag` …

We could do it as a `configure` option — this would work by actually generating the `meep.hpp` file from a `meep.hpp.in` file, which is a bit ugly.  Alternatively, we could have `meep.hpp` do `#include "meep-config.hpp"` where `meep-config.hpp` is generated from `meep-config.hpp.in` (and is installed by `make install`), which is a bit cleaner.

dalexander:
Thanks Steven.  We can definitely add a flag like `--enable_single_precision` to `configure`.  It looks like FFTW has some settings like this.

Could we use the existing `config.h` file instead of introducing a new `meep-config.h[pp]`?

stevengj:
> Could we use the existing config.h file instead of introducing a new meep-config.h[pp]?

No, because of namespace issues. 

The new config file will have to be `#included` from `meep.hpp`, which means that it will be visible in user code that links to Meep, which means that we have to make sure that the symbols defined therein begin with `MEEP_`.   This is not true for the symbols produced by other autoconf tests.

The cleanest thing is to generate a new header file that contains only the symbols we want.

dalexander:
OK that's fine, we can do a `meep_config.hpp`.

You were right that my original code was broken but I have force-pushed what I actually intended, which should be safe... how would you feel about that as a stopgap?

My goal here is to enable injection of `-DMEEP_SINGLE=1` from our build system.  Even if I do introduce this 
meep_config.hpp, we are still going to have to figure out how to switch between two different config.h files in our build system, and that might be even harder than just injecting a `-D`.

PR #1574, opened 2021-05-18
fix segmentation fault in material_grids_addgradient of adjoint solver

oskooi:
Fixes a segmentation fault in the function `material_grids_addgradient_point` of `src/meepgeom.cpp` in the following lines:

https://github.com/NanoComp/meep/blob/8e2b0a050fd7e2d6d39019f76cdd3abcb73888c5/src/meepgeom.cpp#L2600-L2603

The problem is that `tp` (a pointer to a geometry tree) is a NULL pointer when entering this if-statement block which causes a segmentation fault when `&tp->objects[oi]` is invoked.

### DIFF
--- a/src/meepgeom.cpp
+++ b/src/meepgeom.cpp
@@ -420,6 +420,14 @@ double material_grid_val(vector3 p, material_data *md) {
 
 }
 
+static double tanh_projection(double u, double beta, double eta) {
+  double u_proj = 0;
+  double tanh_beta_eta = tanh(beta*eta);
+  u_proj = (tanh_beta_eta + tanh(beta*(u-eta))) /
+    (tanh_beta_eta + tanh(beta*(1-eta)));
+  return u_proj;
+}
+
 double matgrid_val(vector3 p, geom_box_tree tp, int oi, material_data *md) {
   double uprod = 1.0, umin = 1.0, usum = 0.0, udefault = 0.0, u;
   int matgrid_val_count = 0;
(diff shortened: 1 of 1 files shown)


src/meepgeom.cpp
```diff
@@ -2599,16 +2599,16 @@ void material_grids_addgradient_point(double *v, std::complex<double> fields_a,
   }
   // no object tree -- the whole domain is the material grid
   if (!tp && is_material_grid(default_material)) {
-    vector3 pb = to_geom_box_coords(p, &tp->objects[oi]);
+    map_lattice_coordinates(p.x,p.y,p.z);
```
smartalecH, line 2602: I've never used this before, but I assume it's the same as above but assumes a box the size of the simulation cell (the lattice in this case)?

Maybe make sure there's no weird issues with the boundaries (when I read lattice I assume periodic boundaries, which obviously isn't always the case) just make sure the last edge in each dimension is mapped to "1" and not back to "0" (or vice-versa). 

(Ideally the two routines use the same mapping logic and the only difference is the default geometry in this case)

smartalecH (commented):
Seems fine after initial inspection -- would be good to add a simple test before merging.

smartalecH:
Note that this fix is only for the case where the MaterialGrid is passed as a default material, which has never been rigorously tested (obviously since a pointer was being dereferenced when it was already determined to be NULL)

This seg fault is not associated with normal use with MaterialGrids inside of Block geometry objects (not clear from above synopsis).

src/meepgeom.cpp
oskooi: The function `map_lattice_coordinates` maps elements of the `vector3` parameter `p` into the range [0,1]:

https://github.com/NanoComp/meep/blob/8e2b0a050fd7e2d6d39019f76cdd3abcb73888c5/src/meepgeom.cpp#L374-L381

I think you were the one that set up the mapping this way. As long as `p` is defined to be within `geometry_lattice`, it should map points along the boundaries correctly.

oskooi:
#1539 added a new unit test for `MaterialGrid` as the default material but only as part of a "forward" simulation (i.e., no gradient calculation). We do not yet have a similar test involving the adjoint solver which is why this bug has gone undetected.

ianwilliamson:
> This seg fault is not associated with normal use with MaterialGrids inside of Block geometry objects (not clear from above synopsis).

Just wanted to jump in and say that I have observed segfaults from this function under the normal use that you describe (`MaterialGrid` in an `mp.Block`). I agree that what you're saying _should_ be the case, but since the function doesn't have a unit test it's hard to say for sure what's actually happening here.

PR #1577, opened 2021-05-22
pass binary-tree partition to structure constructor

### LINKED ISSUE #1573: user-specified cell partition with boundary layers
The user-specified cell partition feature added in #1528 currently requires that the boundary-layer regions (i.e., PMLs) be defined as separate chunks of the cell partition. This is inconsistent with the existing `num_chunks` API where the user specifies `N` chunks or MPI processes at runtime and then Meep forms *additional* chunks for the PML regions when creating the cell volume and assigns these new chunks the same process ID as the original chunk from which they were formed.

As a demonstration, suppose we want to divide a 10x5 cell into two equal-sized chunks by splitting at *x*=0 and also surround the cell on all sides with PML of thickness 1.0. This would involve specifying a [`BinaryPartition`](https://meep.readthedocs.io/en/latest/Python_User_Interface/#binarypartition) class object as the `chunk_layout` parameter of the `Simulation` constructor:

```py
import meep as mp

chunk_layout = mp.BinaryPartition(data=[ (mp.X,0.0), 0, 1 ])

cell_size = mp.Vector3(10.0,5.0,0)

pml_layers = [mp.PML(thickness=1)]

                    resolution=10,
                    boundary_layers=pml_layers,
                    chunk_layout=chunk_layout)

sim.visualize_chunks()
plt.savefig('chunk_layout.png',dpi=150,bbox_inches='tight')
```
Running this script aborts with the following error:
```
-----------
Initializing structure...
meep: load_chunk_layout: wrong number of chunks.
meep: load_chunk_layout: wrong number of chunks.
time for choose_chunkdivision = 0.000227449 s
MPI_ABORT was invoked on rank 0 in communicator MPI_COMM_WORLD
with errorcode 1.
```
The problem here is that even though we have specified only two chunks, there are in fact **12** chunks in total once the PML regions are included. A schematic of the actual chunk layout is shown below. Meep is expecting to be supplied 12 chunks rather than 2 and thus quits with an error.

chunk_layout_pml

In this schematic, chunks 1, 3, 6, and 8 are 2d PMLs (i.e., PML conductivity in the *x* and *y* directions) whereas chunks 2, 4, 5, 7, 9, 10 are 1d PMLs. The original user-defined chunks are 11 and 12 which have been reduced in size by their intersecting PML regions that have been separated out.

It would be useful for the user-specified `chunk_layout` feature to replicate the functionality of the `num_chunks` API so that the user does not have to define a cell partition that also includes the PML regions (which is a little cumbersome).

oskooi:
Fixes #1573.

Based on a [suggestion in #1528](https://github.com/NanoComp/meep/pull/1528#issuecomment-805453121) by @stevengj, this PR revamps the `structure` class constructor to accept a `binary_partition` argument which enables specifying the chunk volumes *once* when the user passes in a `BinaryPartition` object from the Python API. Correspondingly, the Python function `load_chunk_layout` has been removed since it is no longer necessary. The main changes involved the function `split_by_cost` in `structure.cpp` which now returns a `binary_partition` object rather than writing the chunk volumes into an array. 

This PR enables creating the PML regions as separate chunks using the existing code `structure::choose_chunkdivision`. As a result, the example test described in #1573 now produces the correct chunk layout:

bp_chunk_layout

All the C++ tests are passing yet several of the Python tests are failing. This is due to a SWIG-related issue that needs to be resolved before this PR can be merged. An additional argument has been added to the C++/Python function `create_structure_and_set_materials` in the SWIG typemaps file `meep.i` in order to pass in the `BinaryPartition` pointer object (which is then passed to the `structure` class constructor). The problem is that trying to pass in a `None` (for a `NULL` pointer) for this argument produces an error:

```
ERROR: test_dft_energy (__main__.TestDftEnergy)
  File "dft_energy.py", line 29, in test_dft_energy
  File "/meep/python/meep/simulation.py", line 3637, in run
    self.init_sim()
  File "/meep/python/meep/simulation.py", line 1909, in init_sim
    self._init_structure(self.k_point)
  File "/meep/python/meep/simulation.py", line 1699, in _init_structure
    self.chunk_layout if isinstance(self.chunk_layout,mp.BinaryPartition) else None
SystemError: <built-in function create_structure_and_set_materials> returned NULL without setting an error
```
The issue is that SWIG is expecting a `meep::binary_partition *my_bp` type argument and does not accept `NULL`.

### DIFF
--- a/python/meep.i
+++ b/python/meep.i
@@ -1456,9 +1456,6 @@ void _get_gradient(PyObject *grad, PyObject *fields_a, PyObject *fields_f, PyObj
 
 %typemap(in) meep::binary_partition * {
     $1 = py_bp_to_bp($input);
-    if(!$1) {
-        SWIG_fail;
-    }
 }
 
 %typemap(arginit) meep::binary_partition * {
--- a/python/meep.i
+++ b/python/meep.i
@@ -1891,7 +1891,8 @@ meep::structure *create_structure_and_set_materials(vector3 cell_size,
                                                     bool split_chunks_evenly,
                                                     bool set_materials,
                                                     meep::structure *existing_s,
-                                                    bool output_chunk_costs) {
+                                                    bool output_chunk_costs,
+                                                    meep::binary_partition *my_bp) {
     // Initialize fragment_stats static members (used for creating chunks in choose_chunkdivision)
     meep_geom::fragment_stats::geom = gobj_list;
     meep_geom::fragment_stats::dft_data_list = dft_data_list_;
@@ -1909,7 +1910,11 @@ meep::structure *create_structure_and_set_materials(vector3 cell_size,
 
     if (output_chunk_costs) {
          meep::volume thev = gv.surroundings();
-         std::vector<grid_volume> chunk_vols = meep::choose_chunkdivision(gv, thev, num_chunks, sym);
+         meep::binary_partition *bp = NULL;
+         if (!my_bp) bp = meep::choose_chunkdivision(gv, thev, num_chunks, sym);
+         std::vector<grid_volume> chunk_vols;
+         std::vector<int> ids;
+         meep::split_by_binarytree(gv, chunk_vols, ids, (!my_bp) ? bp : my_bp);
          for (size_t i = 0; i < chunk_vols.size(); ++i)
               master_printf("CHUNK:, %2zu, %f, %zu\n",i,chunk_vols[i].get_cost(),chunk_vols[i].surface_area());
          return NULL;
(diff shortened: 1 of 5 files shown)


oskooi:
To pass a `None` value as the `BinaryPartition` object simply required removing the SWIG failure check for this particular case which had been previously added in #1528 to the typemaps file and is no longer necessary:

https://github.com/NanoComp/meep/blob/ef8f6fc950fdf7c5feb34705afe375e40d12c795/python/meep.i#L1457-L1462

All the tests are now passing and this PR is ready to be merged.

PR #1578, opened 2021-05-28
add --enable-meep configure argument

stevengj:
Closes #1572.


oskooi:
This is working and producing the same results as compiling the master branch with setting the macro `MEEP_SINGLE` to `1` in `meep.hpp`.

PR #1583, opened 2021-06-01
Fix use-after-free issue with the global 'default_material' (issue 1483)

### LINKED ISSUE #1483: segmentation fault in _default_material argument of create_structure_and_set_materials
When I ran a simulation case involved gradient calculation (`calculate_gradient()`), `default_material` in C++ implementation is never assigned the type of `material_type` and pointed to a valid memory allocation in the workflow of gradient calculation. So the conversion from `void*` to `material_type` is invalid. A simple workaround is to delete the "else if" branch, and It passed test cases. Please check this issue.

kkg4theweb:
Signed-off-by: Krishna Gadepalli <kkg@google.com>

Closes #1483.

### DIFF
--- a/src/material_data.hpp
+++ b/src/material_data.hpp
@@ -298,14 +298,14 @@ struct material_data {
 
     memcpy(epsilon_dims, from.epsilon_dims, 3 * sizeof(size_t));
     if (from.epsilon_data) {
-      int N = from.epsilon_dims[0] * from.epsilon_dims[1] * from.epsilon_dims[2];
+      size_t N = from.epsilon_dims[0] * from.epsilon_dims[1] * from.epsilon_dims[2];
       epsilon_data = new double[N];
       memcpy(epsilon_data, from.epsilon_data, N * sizeof(double));
     }
 
     grid_size = from.grid_size;
     if (from.weights) {
-      int N = from.grid_size.x * from.grid_size.y * from.grid_size.z;
+      size_t N = from.grid_size.x * from.grid_size.y * from.grid_size.z;
       weights = new double[N];
       memcpy(weights, from.weights, N * sizeof(double));
     }
--- a/src/material_data.hpp
+++ b/src/material_data.hpp
@@ -54,6 +54,18 @@ struct transition {
   }
 
   bool operator!=(const transition &other) const { return !(*this == other); }
+
+  // NOTE: We could add a copy constructor but that requires a lot more
+  // code cleanup!
+  void copy_from(transition& from) {
+    from_level = from.from_level;
+    to_level = from.to_level;
+    transition_rate = from.transition_rate;
+    frequency = from.frequency;
+    sigma_diag = from.sigma_diag;
+    gamma = from.gamma;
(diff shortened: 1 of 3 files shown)


src/material_data.hpp
```diff
@@ -54,6 +54,18 @@ struct transition {
   }
 
   bool operator!=(const transition &other) const { return !(*this == other); }
+
+  // NOTE: We could add a copy constructor but that requires a lot more
+  // code cleanup!
+  void copy_from(transition& from) {
```
stevengj, line 60: Make the arguments `const`

src/material_data.hpp
```diff
@@ -227,6 +286,35 @@ struct material_data {
...
     material_grid_kinds = U_DEFAULT;
   }
+
+  void copy_from(material_data& from) {
+    which_subclass = from.which_subclass;
+    medium.copy_from(from.medium);
+
+    user_func = from.user_func;
+    // NOTE: the user_data field here opaque/void - so this is the best we can do.
+    user_data = from.user_data;
```
stevengj, line 296: This may be problematic — we may need to increment a Python reference count.

stevengj: Not a regression, however — filed as #1589

src/material_data.hpp
```diff
@@ -227,6 +286,35 @@ struct material_data {
...
+    memcpy(epsilon_dims, from.epsilon_dims, 3 * sizeof(size_t));
+    if (from.epsilon_data) {
+      int N = from.epsilon_dims[0] * from.epsilon_dims[1] * from.epsilon_dims[2];
+      epsilon_data = new double[N];
+      memcpy(epsilon_data, from.epsilon_data, N * sizeof(double));
+    }
+
+    grid_size = from.grid_size;
+    if (from.weights) {
+      int N = from.grid_size.x * from.grid_size.y * from.grid_size.z;
```
stevengj, line 308: Should this be `size_t`

stevengj: Probably this should just call `update_weights` and we should fix it there.

src/material_data.hpp
```diff
@@ -227,6 +286,35 @@ struct material_data {
...
+    medium.copy_from(from.medium);
+
+    user_func = from.user_func;
+    // NOTE: the user_data field here opaque/void - so this is the best we can do.
+    user_data = from.user_data;
+    do_averaging = from.do_averaging;
+
+    memcpy(epsilon_dims, from.epsilon_dims, 3 * sizeof(size_t));
+    if (from.epsilon_data) {
+      int N = from.epsilon_dims[0] * from.epsilon_dims[1] * from.epsilon_dims[2];
```
stevengj, line 301: should be `size_t`

src/material_data.hpp
stevengj: Separately, we should probably add a check that the user `grid_size` makes sense (e.g. is not negative).

kkg4theweb: Done. Made it into size_t

PR #1586, opened 2021-06-02
Fix circular dependency in imports and change namespace of JAX adjoint components

ianwilliamson:
Fixes a circular dependency in the JAX adjoint module when importing `meep.adjoint`.

### DIFF
--- a/python/Makefile.am
+++ b/python/Makefile.am
@@ -220,9 +220,8 @@ adjoint_PYTHON  = $(srcdir)/adjoint/__init__.py              \
                   $(srcdir)/adjoint/optimization_problem.py  \
                   $(srcdir)/adjoint/filters.py               \
                   $(srcdir)/adjoint/filter_source.py         \
-                  $(srcdir)/adjoint/jax/__init__.py          \
-                  $(srcdir)/adjoint/jax/wrapper.py           \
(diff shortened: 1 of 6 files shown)


ianwilliamson:
Looks like I may have also introduced an issue with the install location of the new JAX modules. Let me fix this as well...

ianwilliamson:
Okay, I think this should be good to go now. The summary is:

- I fixed some circular import dependencies in the newly introduced JAX adjoint components. 
- I also realized that the entries in the Makefile that I added were actually incorrect (they `jax/` subdirectory was not being preserved).
- In lieu of preserving this `jax/` subdirectory, I opted to just move the JAX adjoint components into the `mpa.*` namespace from `mpa.jax.*`
- Actually... the only JAX-specific module is `wrapper.py`, while the functions in `utils.py` can actually be used by the other adjoint components since these provide common functionality. Thus, I think the namespace change makes sense.
- Since the functions I previously added in `utils.py` all have unit tests, you might consider eventually switching the other part of the adjoint module over to these.

I have my local build working now and have verified that everything builds / runs successfully.

ianwilliamson:
FYI- I think the build may be broken until this gets merged.

Issue #1585, opened 2021-06-02
Backward propagating eigenmode adjoints are wrong


smartalecH:
Note that objective functions that depend on the backward propagating mode produce incorrect gradients right now.

We can get around this by calculating the *forward* coefficient in the *backward* direction (i.e. flip the coordinate frame of MPB) using something like this:

```python3
bw = mpa.EigenmodeCoefficient(sim, source_vol, kpoint_func=lambda a,b: mp.Vector3(x=-1))
```

Since this hack works, it's probably just some trivial sign error in the adjoint pipeline (which is also why I haven't pursued a fix yet...). That being said, we should add a unit test that verifies this (and someday actually fix this).

stevengj:
Test case:
![image](https://user-images.githubusercontent.com/2913679/120413075-a5ee5b80-c325-11eb-9085-715aca62c2f6.png)

ahoenselaar:
There is certainly an extremely dangerous coding practice in use here:
https://github.com/NanoComp/meep/blob/master/python/adjoint/objective.py#L34

Going from a boolean 'forward' to an integer 'self.forward' with the opposite boolean interpretation is asking for trouble.
`forward=True` becomes `self.forward=0` (interpreted as `False` in a Boolean context).

smartalecH:
> There is certainly an extremely dangerous coding practice in use here

Agreed but not the cause of the above issue, luckily (in case anyone was wondering).

ahoenselaar:
It would be great to have a small repro case with observed behavior and expected behavior. That could eventually be turned into a unit test.

ianwilliamson:
> It would be great to have a small repro case with observed behavior and expected behavior. That could eventually be turned into a unit test.

See the update to the JAX-Meep wrapper test in: https://github.com/NanoComp/meep/pull/1593/commits

ahoenselaar:
Great, thank you!

PR #1588, opened 2021-06-03
Adjoint cleanup

ianwilliamson:
This PR bases off of #1586 to clean up the adjoint module code by:

- ~~Applying formatting to the Python code to make it more readable~~ (Moved to #1590)
- Fixing several small bugs (equality rather than assignment, duplicated return statements, etc.)
- Adding method signatures to the abstract `ObjectiveQuantity` class
- Removing some non-standard Python things like `return` statements at the end of `__init__` methods
- Correcting spelling errors
- Removing unused imports

In a subsequent PR I will propose a fix for #1585.

### DIFF
--- a/python/adjoint/basis.py
+++ b/python/adjoint/basis.py
@@ -1,10 +1,6 @@
 import meep as mp
 import numpy as np
 from scipy import sparse
-#from autograd import grad, jacobian, vector_jacobian_product
-import autograd.numpy as npa
-from autograd import grad, jacobian
-#invoke python's 'abstract base class' formalism in a version-agnostic way
 from abc import ABCMeta, abstractmethod
 
 ABC = ABCMeta('ABC', (object, ), {'__slots__':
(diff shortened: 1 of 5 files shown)


stevengj:
Note that the whitespace changes will complicate rebasing and merging other pending changes to the Python code.  Is there something easier than the procedure suggested in https://github.com/NanoComp/meep/pull/662?

ianwilliamson:
> (What Python formatting tool did you use?)

`yapf` - https://github.com/google/yapf

stevengj:
It would be better to have a separate PR with just the `yapf` changes, both for ease of review and for ease of rebasing.

ianwilliamson:
> It would be better to have a separate PR with just the `yapf` changes, both for ease of review and for ease of rebasing.

Done in #1590.

ianwilliamson:
I think this PR is good to go. These changes are relatively minor code health related things that should (hopefully) be uncontroversial. Let me know if there is any additional feedback.

PR #1591, opened 2021-06-04
fix bug in copy assignment for material grid

oskooi:
Fixes a bug introduced in #1583 which is causing the unit test for a `MaterialGrid` as a default material (in `python/tests/material_grid.py`) to currently fail.

### DIFF
--- a/src/meepgeom.cpp
+++ b/src/meepgeom.cpp
@@ -29,13 +29,13 @@ material_data vacuum_material_data;
 material_type vacuum = &vacuum_material_data;
 
 static void set_default_material(material_type _default_material) {
-  if (default_material != NULL) {
+  if ((default_material != NULL) && (default_material != _default_material)) {
     material_free((material_type)default_material);
     delete (material_type)default_material;
     default_material = NULL;
   }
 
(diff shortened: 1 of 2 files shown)


oskooi:
Also fixes another bug that was causing two unit test failures in `python/tests/simulation.py`: `test_epsilon_input_file` and `test_numpy_epsilon`.

With these changes, all the unit tests are passing.

kkg4theweb (approved):

src/meepgeom.cpp
```diff
@@ -29,13 +29,13 @@ material_data vacuum_material_data;
 material_type vacuum = &vacuum_material_data;
 
 static void set_default_material(material_type _default_material) {
-  if (default_material != NULL) {
+  if ((default_material != NULL) && (default_material != _default_material)) {
```
kkg4theweb, line 32: It might be simpler to have an early exit like this ?

if (default_material == _default_material) return;

and then avoid those checks later ?

oskooi: Good suggestion. Fixed.

kkg4theweb (approved):

PR #1592, opened 2021-06-04
Refactor interfaces and internal state of adjoint module objective quantities

ianwilliamson:
This PR picks up where #1588 left off with some refactoring of the `mpa.ObjectiveQuantity` child classes:

 - Many attributes of the `mpa.ObjectiveQuantity` child classes are not correctly indicated as being private
 - Some attributes of the `mpa.ObjectiveQuantity` child classes (e.g. `source` in `EigenmodeCoefficient`) should apparently not even be part of the internal state (since they're unused) and are initialized in an adhoc manner throughout different methods of the classes
 - Functions such as `adj_src_scale()` and `create_time_profile()` should actually be methods of `mpa.ObjectiveQuantity`

In an effort to streamline the process of debugging tricky issues such as #1585, the interfaces, attributes, and data marshaling in the adjoint module should be standardized in the `mpa.ObjectiveQuantity`.

### DIFF
--- a/python/adjoint/objective.py
+++ b/python/adjoint/objective.py
@@ -47,7 +47,7 @@ class ObjectiveQuantity(abc.ABC):
             return self._eval
         else:
             raise RuntimeError(
-                'You must first run a forward simulation before resquesting the evaluation of an `ObjectiveQuantity`.'
+                'You must first run a forward simulation before requesting the evaluation of an objective quantity.'
             )
 
     def _adj_src_scale(self, include_resolution=True):
@@ -57,9 +57,7 @@ class ObjectiveQuantity(abc.ABC):
         src = self._create_time_profile()
 
         if include_resolution:
-            num_dims = ((self.sim.cell_size.x != 0) +
-                        (self.sim.cell_size.y != 0) +
-                        (self.sim.cell_size.z != 0))
+            num_dims = self.sim._infer_dimensions(self.sim.k_point)
             dV = 1 / self.sim.resolution**num_dims
         else:
             dV = 1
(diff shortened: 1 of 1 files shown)


ianwilliamson:
Okay, I think this should be good to go.

smartalecH:
> Many attributes of the mpa.ObjectiveQuantity child classes are not correctly indicated as being private

Many of these quantities don't *need* to be private, do they? From a flexibility perspective, this seems restricting.

ianwilliamson:
> Many of these quantities don't _need_ to be private, do they?

These objects use a lot of internal state to pass information between various steps (registering monitors, performing fwd and adjoint sims, etc) and if that state isn't supposed to be accessed/modified by external code then, yes, they need to be marked as private.

Part of the goal here is to indicate to tools, such as pylint, what the intended data access patterns and APIs are. Pylint can then help us to identify bugs through static analysis.

> From a flexibility perspective, this seems restricting.

There is nothing preventing you from querying these variables in environments such as colab. If there's something missing from the public API that's needed by some other piece of unchecked in code, we can incorporate that. Let me know if there's something specific.

python/adjoint/objective.py
```diff
@@ -21,9 +41,75 @@ def register_monitors(self, frequencies):
...
+    def _adj_src_scale(self, include_resolution=True):
+        """Calculates the scale for the adjoint sources."""
+        T = self.sim.meep_time()
+        dt = self.sim.fields.dt
+        src = self._create_time_profile()
+
+        if include_resolution:
+            num_dims = ((self.sim.cell_size.x != 0) +
+                        (self.sim.cell_size.y != 0) +
+                        (self.sim.cell_size.z != 0))
```
stevengj, line 62: Meep only supports 2d simulations in the xy plane, for example — if you want to do an xz plane, you do it as a 3d simulation that is 1 pixel thick and periodic in x. 

I feel like this should use `_infer_dimensions`.   (Note that cylindrical coordinates are special, but Mo has code to handle that.)

PR #1593, opened 2021-06-04
Fix incorrect backward `EigenmodeCoefficient` gradients and update test case

ianwilliamson:
This PR continues from #1592 to update the test for the Meep-JAX wrapper to check the gradients from `mpa.EigenmodeCoefficient`s with `forward=False`. In particular, this PR extends the straight waveguide test case with an additional source on the RHS of the waveguide and with two additional `mpa.EigenmodeCoefficient`s such that both the forward and backward traveling waves on each side of the waveguide are monitored and can be included in the loss function.

Currently, the cases that check the gradient of the transmission from right to left are failing (selected via `excite_port_idx` param of the test case), presumably because of #1585. Note that another option for exposing the issue described in #1585 is to check the gradient of reflection from the straight waveguide against finite differences (although this quantity should be close to zero).

### DIFF
--- a/python/adjoint/objective.py
+++ b/python/adjoint/objective.py
@@ -172,7 +172,11 @@ class EigenmodeCoefficient(ObjectiveQuantity):
         if self.kpoint_func:
             eig_kpoint = -1 * self.kpoint_func(time_src.frequency, self.mode)
         else:
-            direction = mp.Vector3(*[float(v == 0) for v in self.volume.size])
+            center_frequency = 0.5 * np.min(self.frequencies) + np.max(
+                self.frequencies)
+            direction = mp.Vector3(
+                *(np.eye(3)[self._monitor.normal_direction] *
+                  np.abs(center_frequency)))
             eig_kpoint = -1 * direction if self.forward else direction
 
         if self._frequencies.size == 1:
@@ -190,7 +194,7 @@ class EigenmodeCoefficient(ObjectiveQuantity):
         source = mp.EigenModeSource(
             src,
             eig_band=self.mode,
-            direction=mp.AUTOMATIC,
+            direction=mp.NO_DIRECTION,
             eig_kpoint=eig_kpoint,
             amplitude=amp,
             eig_match_freq=True,
@@ -205,13 +209,16 @@ class EigenmodeCoefficient(ObjectiveQuantity):
             kpoint_func = self.kpoint_func
             overlap_idx = self.kpoint_func_overlap_idx
         else:
-            direction = mp.Vector3(*[float(v == 0) for v in self.volume.size])
-            kpoint_func = lambda *not_used: direction if self.forward else -1 * direction
+            center_frequency = 0.5 * np.min(self.frequencies) + np.max(
+                self.frequencies)
+            kpoint = mp.Vector3(*(np.eye(3)[self._monitor.normal_direction] *
+                                  np.abs(center_frequency)))
+            kpoint_func = lambda *not_used: kpoint if self.forward else -1 * kpoint
             overlap_idx = 0
         ob = self.sim.get_eigenmode_coefficients(
             self._monitor,
             [self.mode],
-            direction=mp.AUTOMATIC,
+            direction=mp.NO_DIRECTION,
             kpoint_func=kpoint_func,
             **self.eigenmode_kwargs,
         )
--- a/python/adjoint/objective.py
+++ b/python/adjoint/objective.py
@@ -172,8 +172,8 @@ class EigenmodeCoefficient(ObjectiveQuantity):
         if self.kpoint_func:
             eig_kpoint = -1 * self.kpoint_func(time_src.frequency, self.mode)
         else:
-            center_frequency = 0.5 * np.min(self.frequencies) + np.max(
-                self.frequencies)
+            center_frequency = 0.5 * (np.min(self.frequencies) + np.max(
+                self.frequencies))
             direction = mp.Vector3(
                 *(np.eye(3)[self._monitor.normal_direction] *
                   np.abs(center_frequency)))
@@ -209,8 +209,8 @@ class EigenmodeCoefficient(ObjectiveQuantity):
             kpoint_func = self.kpoint_func
             overlap_idx = self.kpoint_func_overlap_idx
         else:
-            center_frequency = 0.5 * np.min(self.frequencies) + np.max(
-                self.frequencies)
+            center_frequency = 0.5 * (np.min(self.frequencies) + np.max(
+                self.frequencies))
             kpoint = mp.Vector3(*(np.eye(3)[self._monitor.normal_direction] *
                                   np.abs(center_frequency)))
             kpoint_func = lambda *not_used: kpoint if self.forward else -1 * kpoint
--- a/python/adjoint/objective.py
+++ b/python/adjoint/objective.py
@@ -115,22 +115,39 @@ class ObjectiveQuantity(abc.ABC):
 
 
 class EigenmodeCoefficient(ObjectiveQuantity):
+    """A frequency-dependent eigenmode coefficient.
+    Attributes:
+        volume: the volume over which the eigenmode coefficient is calculated.
+        mode: the eigenmode number.
+        forward: whether the forward or backward mode coefficient is returned as
+          the result of the evaluation.
+        kpoint_func: an optional k-point function to use when evaluating the eigenmode
+          coefficient. When specified, this overrides the effect of `forward`.
+        kpoint_func_overlap_idx: the index of the mode coefficient to return when
+          specifying `kpoint_func`. When specified, this overrides the effect of
+          `forward` and should have a value of either 0 or 1.
+    """
     def __init__(self,
                  sim,
                  volume,
                  mode,
                  forward=True,
                  kpoint_func=None,
+                 kpoint_func_overlap_idx=0,
                  decimation_factor=0,
                  **kwargs):
         super().__init__(sim)
+        if kpoint_func_overlap_idx not in [0, 1]:
+            raise ValueError(
+                '`kpoint_func_overlap_idx` should be either 0 or 1, but got %d'
+                % (kpoint_func_overlap_idx, ))
         self.volume = volume
         self.mode = mode
         self.forward = forward
         self.kpoint_func = kpoint_func
+        self.kpoint_func_overlap_idx = kpoint_func_overlap_idx
         self.eigenmode_kwargs = kwargs
         self._monitor = None
-        self._normal_direction = None
         self._cscale = None
         self.decimation_factor = decimation_factor
 
@@ -142,27 +159,21 @@ class EigenmodeCoefficient(ObjectiveQuantity):
             yee_grid=True,
             decimation_factor=self.decimation_factor,
         )
-        self._normal_direction = self._monitor.normal_direction
         return self._monitor
 
     def place_adjoint_source(self, dJ):
         dJ = np.atleast_1d(dJ)
-        direction_scalar = -1 if self.forward else 1
-        time_src = self._create_time_profile()
-        if self.kpoint_func is None:
-            if self._normal_direction == 0:
-                k0 = direction_scalar * mp.Vector3(x=1)
-            elif self._normal_direction == 1:
-                k0 = direction_scalar * mp.Vector3(y=1)
-            elif self._normal_direction == 2:
-                k0 = direction_scalar * mp.Vector3(z=1)
-        else:
-            k0 = direction_scalar * self.kpoint_func(time_src.frequency, 1)
         if dJ.ndim == 2:
             dJ = np.sum(dJ, axis=1)
-        da_dE = 0.5 * self._cscale  # scalar popping out of derivative
-
+        time_src = self._create_time_profile()
+        da_dE = 0.5 * self._cscale
         scale = self._adj_src_scale()
+
+        if self.kpoint_func:
+            eig_kpoint = -1 * self.kpoint_func(time_src.frequency, self.mode)
+        else:
+            direction = mp.Vector3(*[float(v == 0) for v in self.volume.size])
+            eig_kpoint = -1 * direction if self.forward else direction
 
         if self._frequencies.size == 1:
             amp = da_dE * dJ * scale
@@ -176,12 +187,11 @@ class EigenmodeCoefficient(ObjectiveQuantity):
                 self.sim.fields.dt,
             )
             amp = 1
-
         source = mp.EigenModeSource(
             src,
             eig_band=self.mode,
-            direction=mp.NO_DIRECTION,
-            eig_kpoint=k0,
+            direction=mp.AUTOMATIC,
+            eig_kpoint=eig_kpoint,
             amplitude=amp,
             eig_match_freq=True,
             size=self.volume.size,
@@ -191,17 +201,24 @@ class EigenmodeCoefficient(ObjectiveQuantity):
         return [source]
 
     def __call__(self):
-        direction = mp.NO_DIRECTION if self.kpoint_func else mp.AUTOMATIC
+        if self.kpoint_func:
+            kpoint_func = self.kpoint_func
+            overlap_idx = self.kpoint_func_overlap_idx
+        else:
+            direction = mp.Vector3(*[float(v == 0) for v in self.volume.size])
+            kpoint_func = lambda *not_used: direction if self.forward else -1 * direction
+            overlap_idx = 0
         ob = self.sim.get_eigenmode_coefficients(
             self._monitor,
             [self.mode],
-            direction=direction,
-            kpoint_func=self.kpoint_func,
+            direction=mp.AUTOMATIC,
+            kpoint_func=kpoint_func,
             **self.eigenmode_kwargs,
         )
-        # record eigenmode coefficients for scaling
-        self._eval = np.squeeze(ob.alpha[:, :, int(not self.forward)])
-        self._cscale = ob.cscale  # pull scaling factor
+        overlaps = ob.alpha.squeeze(axis=0)
+        assert overlaps.ndim == 2
+        self._eval = overlaps[:, overlap_idx]
+        self._cscale = ob.cscale
         return self._eval
 
 
--- a/python/tests/test_adjoint_jax.py
+++ b/python/tests/test_adjoint_jax.py
@@ -57,7 +57,15 @@ def build_straight_wg_simulation(
       eig_kpoint=mp.Vector3(1, 0, 0),
       size=mp.Vector3(0, wg_width + 2 * wg_padding, 0),
       center=[-sx / 2 + pml_width + source_to_pml, 0, 0],
-    )
+    ),
+    mp.EigenModeSource(
+      mp.GaussianSource(frequency=fmean, fwidth=fmean * gaussian_rel_width),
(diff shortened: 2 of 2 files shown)


ianwilliamson:
As a starting point, it seems that the fix suggested in #1585 of adding:

```python
kpoint_func=lambda a,b: mp.Vector3(x=-1)
```

to the constructor of `EigenmodeCoefficient` does not seem to have any effect.

ianwilliamson:
The issue described in #1585 should be fixed with this.

With this PR, I am proposing a slight refactoring where a private `_direction_vector` is defined upon registering the mode monitors in the forward simulation. This direction vector is then used both for pulling out the coefficients and for setting up the adjoint source (flipping the direction vector).

smartalecH:
Rather than make the hack more elegant, why not focus on fixing the actual bug?

ianwilliamson:
> Rather than make the hack more elegant, why not focus on fixing the actual bug?

Is there some reason that it's not advisable to define the forward orientation as being along the negative direction of an axis? In any case, this approach at least leads to correct gradients until someone is able to look into this more.

python/adjoint/objective.py
```diff
@@ -1,193 +1,230 @@
...
+            frequencies,
+            mp.ModeRegion(center=self.volume.center, size=self.volume.size),
+            yee_grid=True,
+        )
+        if self._monitor.normal_direction == 0:
+            direction_vector = mp.Vector3(x=1)
+        elif self._monitor.normal_direction == 1:
+            direction_vector = mp.Vector3(y=1)
+        elif self._monitor.normal_direction == 2:
+            direction_vector = mp.Vector3(z=1)
```
smartalecH, line 138: Does this preclude us from passing arbitrary `kpoint`s in for oblique waveguides?

smartalecH:
> Is there some reason that it's not advisable to define the forward orientation as being along the negative direction of an axis?

In theory, no, but if there's a bug causing the issue we should probably fix the bug. Especially since it might lead to fixing other issues.

I'm a little worried this approach actually *limits* the user (i.e. me) from defining arbitrary modes (e.g. diffraction modes, oblique waveguide modes, etc.)

ianwilliamson:
> In theory, no, but if there's a bug causing the issue we should probably fix the bug.

Sure, I agree, but the concern here is that incorrect results are being produced... so we should fix that first.

> I'm a little worried this approach actually limits the user (i.e. me) from defining arbitrary modes (e.g. diffraction modes, oblique waveguide modes, etc.)

Are you saying that you want the ability override the `forward` parameter by passing in a `kpoint_func`? That's fine. I think we could support that while still fixing the issue with the gradients.

python/adjoint/objective.py
ianwilliamson: I've reintroduced the `kpoint_func` param. WDYT?

ianwilliamson:
> I think we could support that while still fixing the issue with the gradients.

I've reintroduced the `kpoint_func` parameter of `EigenmodeCoefficient`. Let me know what you think!

python/adjoint/objective.py
smartalecH: Seems right from a high level -- thanks for looking into for me.

Have you created a test for it by chance? Or are all the tests just forward/backward (or equally up/down)?

ianwilliamson: The test case updated in this PR checks forward and backward gradients via the `forward` parameter of `EigenmodeCoefficient`, but does not have a case where `kpoint_func` is specified. I've personally not used the oblique waveguide capability but if you have something in mind let me know.

python/adjoint/objective.py
```diff
@@ -21,152 +41,231 @@ def register_monitors(self, frequencies):
...
+        return [source]
 
     def __call__(self):
-        self.time_src = create_time_profile(self)
-        direction = mp.NO_DIRECTION if self.kpoint_func else mp.AUTOMATIC
+        if self.kpoint_func:
+            kpoint_func = self.kpoint_func
+            overlap_idx = self.kpoint_func_overlap_idx
+        else:
+            kpoint_func = lambda *not_used: self._direction_vector
```
stevengj, line 213: This doesn't seem right — the k point is dimensionful, so the default guess is scaled by `fcen*(refractive index)`.

For comparison, this is the default guess in `mpb.cpp`: https://github.com/NanoComp/meep/blob/83181144fdfdd76d295139da2f4aa0ea6b192aaf/src/mpb.cpp#L487

python/adjoint/objective.py
```diff
@@ -21,152 +41,231 @@ def register_monitors(self, frequencies):
...
-        da_dE = 0.5 * self.cscale  # scalar popping out of derivative
+        time_src = self._create_time_profile()
+        da_dE = 0.5 * self._cscale
+        scale = self._adj_src_scale()
 
-        scale = adj_src_scale(self, dt)
+        if self.kpoint_func:
+            eig_kpoint = self.kpoint_func(time_src.frequency, self.mode)
+        else:
+            eig_kpoint = self._direction_vector
```
stevengj, line 181: This isn't quite the right guess, since it has the wrong units.  It would be better to use mpb.cpp's guess for this, which is obtained by passing `(0.0,0.0,0.0)` for the k-point.  To tell it to flip the sign, we could pass `(-0.0,-0.0,-0.0)` and use the [sign of zero](https://people.freebsd.org/~das/kahan86branch.pdf):
```diff
diff --git a/src/mpb.cpp b/src/mpb.cpp
index c41e3c2a..5a68ca7b 100644
--- a/src/mpb.cpp
+++ b/src/mpb.cpp
@@ -484,7 +484,7 @@ void *fields::get_eigenmode(double frequency, direction d, const volume where, c
   // which we automatically pick if kmatch == 0.
   if (match_frequency && kmatch == 0) {
     vec cen = eig_vol.center();
-    kmatch = frequency * sqrt(real(get_eps(cen, frequency)) * real(get_mu(cen, frequency)));
+    kmatch = copysign(frequency * sqrt(real(get_eps(cen, frequency)) * real(get_mu(cen, frequency))), kmatch);
     if (d == NO_DIRECTION) {
       for (int i = 0; i < 3; ++i)
         k[i] = dot_product(R[i], kdir) * kmatch; // kdir*kmatch in reciprocal basis
```

ianwilliamson:
Okay, following up on this... it seems that the `copysign()` approach while passing `mp.Vector3(0.0, 0.0, 0.0)` and `mp.Vector3(-0.0, -0.0, -0.0)` for `forward=True` and `forward=False` doesn't work. I added a print statement in `mpb.cpp` to print out `kmatch` and it seem to never change sign. Is that expected? Maybe we need to copy the sign from another quantity in that function?

stevengj:
Maybe it is easier to just pass a boolean flag.

python/adjoint/objective.py
stevengj: Looks like the sign is being lost in kmatch, so you'd have to look at `_kpoint.in_direction(d)` as the second argument of `copysign`

ianwilliamson:
> Maybe it is easier to just pass a boolean flag.

I agree. In the future it would probably be ideal to add a boolean flag, rather than having the user pass +/- zero, to specify the direction in MPB.

> Looks like the sign is being lost in kmatch, so you'd have to look at _kpoint.in_direction(d) as the second argument of copysign

I can confirm that the suggestion above in https://github.com/NanoComp/meep/pull/1593#discussion_r649455293 successfully sets the sign of `kmatch` from the input `_kpoint`. However, I am still unable to get the correct gradients.

I noticed that when `|kpoint| == 0`, MPB requires you to use `direction=mp.AUTOMATIC`, rather than `direction=mp.NO_DIRECTION`. In the approach that I used before, where I obtained correct gradients in _both_ directions, I was using `direction=mp.NO_DIRECTION`. In that case, I was also passing a non-zero k-vector. This seems to suggest that there is more going on down in that MPB function, beyond selecting the sign of `kmatch`, that we need to obtain correct gradients along both directions.

Any other thoughts? I would be in favor of using my original approach (passing a vector scaled by the frequency to specify the direction) in the near term if this is going to require a bunch of hacking in the lower level MPB code. I realize that there are some edge cases around users picking potentially weird materials or different units, but right now the adjoint module is producing incorrect gradients and I think that is probably a bigger issue.

stevengj:
> I can confirm that the suggestion above in #1593 (comment) successfully sets the sign of kmatch from the input _kpoint. However, I am still unable to get the correct gradients.

Hmm, looking at the code it looks like the reason might be that it is then computing the group velocity in the wrong direction, since `kdir` is now opposite to the direction of `k = kmatch * kdir`: https://github.com/NanoComp/meep/blob/f528873eb5cf680976102f5b307b505ab25c4b78/src/mpb.cpp#L564

Maybe it would be better to flip the sign of `kdir`, which corresponds more closely to what happens if you pass an initial-guess `kpoint` in the opposite direction.   That is, *instead* of the `copysign` line I suggested above, after [this line](https://github.com/NanoComp/meep/blob/f528873eb5cf680976102f5b307b505ab25c4b78/src/mpb.cpp#L487) do:
```
if (std::signbit(_kpoint.in_direction(d)))
    kdir[d - X] = -1; // -0.0 was passed, flip k direction
```

ianwilliamson:
> Maybe it would be better to flip the sign of kdir, which corresponds more closely to what happens if you pass an initial-guess kpoint in the opposite direction.

```c
kdir[d - X] = -1;
```
This (assignment of a negative value) seems to cause a SIGSEGV error.

oskooi:
Using `gdb` on the latest commit in this branch and running the unit test `python/tests/test_adjoint_solver.py` produces this backtrace:
```
Thread 1 "python3" received signal SIGSEGV, Segmentation fault.
0x00007ffff624cf78 in meep::vec::operator-= (this=0x80, a=...) at ../src/meep/vec.hpp:544
544	    LOOP_OVER_DIRECTIONS(dim, d) t[d] -= a.t[d];
(gdb) bt
#0  0x00007ffff624cf78 in meep::vec::operator-= (this=0x80, a=...) at ../src/meep/vec.hpp:544
#1  0x00007ffff5d30f7a in meep::fields::add_eigenmode_source (this=0x20b0d30, c0=meep::Dielectric, src=..., 
    d=meep::X, where=..., eig_vol=..., band_num=1, kpoint=..., match_frequency=true, parity=0, resolution=0, 
    eigensolver_tol=9.9999999999999998e-13, amp=..., A=0x0, dp=0x0) at mpb.cpp:816
```
The line causing the segmentation fault is `mpb.cpp:816` which in the master branch is:

https://github.com/NanoComp/meep/blob/f528873eb5cf680976102f5b307b505ab25c4b78/src/mpb.cpp#L814

But what is really causing the segfault is this line in `python/adjoint/objective.py`:
```py
_KPOINT_NEG_DIR = mp.Vector3(-0.0, 0.0, 0.0)
```
Changing this manually to `mp.Vector3(-1.0, 0.0, 0.0)` and everything runs fine. I think the problem is that a negative zero floating point is not being recognized as signed but rather unsigned which is causing the error.

src/mpb.cpp
```diff
@@ -485,6 +485,8 @@ void *fields::get_eigenmode(double frequency, direction d, const volume where, c
   if (match_frequency && kmatch == 0) {
     vec cen = eig_vol.center();
     kmatch = frequency * sqrt(real(get_eps(cen, frequency)) * real(get_mu(cen, frequency)));
+    if (std::signbit(kpoint.in_direction(d)))
+      kdir[d - X] = -1; // -0.0 was passed, flip k direction
```
stevengj, line 489: Instead, try putting this before the `kdir[d - X] = 1;` line

stevengj:
@oskooi, before that line in `get_eigenmode_source`, we should probably check whether `global_eigenmode_data` is `NULL`, which can happen if the Newton iteration failed to converge.

stevengj:
Actually, I take it back — flipping the sign of `kmatch` (and leaving `kdir` alone) seems like it should be the right thing.

ianwilliamson:
This is still failing on the check of the gradients in the backward direction using the suggested approach of flipping the sign of `kmatch` and passing `direction=mp.AUTOMATIC` from Python.

I was previously able to get the gradients to be correct in both directions **only** when `direction=mp.NO_DIRECTION` and a unit vector in the propagation direction was passed for `kpoint`. This seems to indicate that there is some additional stuff going on in the `direction=mp.NO_DIRECTION` branch of `mpb.cpp` that results in the correct fields.

stevengj:
Actually, there is a more fundamental problem, it seems.   If you look at the principle of equivalence, which is described in [this line](https://github.com/NanoComp/meep/blob/8be41f37b6fd8c99aa5fc0f304301c80797809be/src/mpb.cpp#L826-L831) (which comes from equation 4.4 in [this paper](https://arxiv.org/pdf/1301.5366.pdf), then if you're trying to make a backwards mode we should really flip the sign of nhat.

Maybe, before [this line](https://github.com/NanoComp/meep/blob/8be41f37b6fd8c99aa5fc0f304301c80797809be/src/mpb.cpp#L832), add a line
```
if (global_eigenmode_data->group_velocity < 0)
    amp *= -1; // equivalent to flipping the direction of nhat.
```

ianwilliamson:
This is still failing (currently along both directions) with the most recently suggested approach of changing the sign of `amp` based on the group velocity direction. To summarize, I am currently:

 - Using `direction = mp.AUTOMATIC` from Python for both the source construction and getting the eigenmode coeffs
 - Passing +/- `mp.Vector3(0.0, 0.0, 0.0)` to specify the direction of the mode
 - Flipping the sign of `kmatch` based on `_kpoint.in_direction(d)`
 - Flipping the sign of `amp` based on `global_eigenmode_data->group_velocity < 0`

These changes are all pushed if you want to take a look and confirm that I'm not doing something unintended. As I mentioned above, I was only able to obtain correct gradients when using `direction = mp.NO_DIRECTION` from Python.

stevengj:
Would be good to visualize the mode that is being launched here, to verify that it is going in the backward direction, or if fundamentally the mode that is being launched is incorrect.

ianwilliamson:
I've reverted this PR to use my original approach, which was to pass an appropriate signed unit vector as the k-point to MPB when defining the source and when evaluating the eigenmode coefficients. As @stevengj pointed out above, this approach has the disadvantage of not being scaled correctly / not having the appropriate units. However, I personally believe this is an acceptable trade off in order to fix the more immediate issue with the gradients being incorrect. In practice, MPB usually finds the correct mode. Perhaps we can fork off a separate issue to focus on the work of figuring out a more long term solution with what's going on down in the lower level MPB code.

WDYT?

python/adjoint/objective.py
```diff
@@ -142,28 +159,22 @@ def register_monitors(self, frequencies):
...
-
+        time_src = self._create_time_profile()
+        da_dE = 0.5 * self._cscale
         scale = self._adj_src_scale()
 
+        if self.kpoint_func:
+            eig_kpoint = -1 * self.kpoint_func(time_src.frequency, self.mode)
+        else:
+            direction = mp.Vector3(*[float(v == 0) for v in self.volume.size])
+            eig_kpoint = -1 * direction if self.forward else direction
```
stevengj, line 176: At least scale this by `abs(fcen)` to fix the units.

ianwilliamson: Done.

stevengj:
Test failure looks relevant:
```
FAIL: test_adjoint_solver_eigenmode (__main__.TestAdjointSolver)
  File "/home/runner/work/meep/meep/build/meep-1.21.0-beta/_build/sub/python/../../../python/tests/test_adjoint_solver.py", line 234, in test_adjoint_solver_eigenmode
  File "/home/runner/work/meep/meep/build/meep-1.21.0-beta/python/tests/utils.py", line 28, in assertClose
AssertionError: 0.00012112395228314199 not less than or equal to 0.00012045757274762158 : 
```

oskooi:
The failing test occurs for only the single-precision floating point case. We can just probably lower the tolerance (`epsilon=1e-6`) slightly and it should be fine.

python/tests/test_adjoint_jax.py
```diff
@@ -22,228 +22,251 @@
...
+            direction=mp.NO_DIRECTION,
+            eig_kpoint=mp.Vector3(-1, 0, 0),
+            size=mp.Vector3(0, wg_width + 2 * wg_padding, 0),
+            center=[sx / 2 - pml_width - source_to_pml, 0, 0],
         ),
-      ),
+    ]
+
+    nx = int(design_region_resolution * design_region_shape[0])
+    ny = int(design_region_resolution * design_region_shape[1])
```
oskooi, line 74: The `nx` and `ny` dimensions of the `MaterialGrid` are missing a `+ 1` (see #1760).

python/adjoint/objective.py
```diff
@@ -191,17 +205,27 @@ def place_adjoint_source(self, dJ):
         return [source]
 
     def __call__(self):
-        direction = mp.NO_DIRECTION if self.kpoint_func else mp.AUTOMATIC
+        if self.kpoint_func:
+            kpoint_func = self.kpoint_func
+            overlap_idx = self.kpoint_func_overlap_idx
+        else:
+            center_frequency = 0.5 * np.min(self.frequencies) + np.max(
+                self.frequencies)
```
oskooi, line 213: Parentheses missing around the sum involving the `np.min` and `np.max` terms?

python/adjoint/objective.py
```diff
@@ -142,28 +159,26 @@ def register_monitors(self, frequencies):
...
-
+        time_src = self._create_time_profile()
+        da_dE = 0.5 * self._cscale
         scale = self._adj_src_scale()
 
+        if self.kpoint_func:
+            eig_kpoint = -1 * self.kpoint_func(time_src.frequency, self.mode)
+        else:
+            center_frequency = 0.5 * np.min(self.frequencies) + np.max(
+                self.frequencies)
```
oskooi, line 176: Parentheses missing around the sum involving the `np.min` and `np.max` terms?

python/tests/test_adjoint_jax.py
```diff
@@ -22,228 +22,251 @@
...
 
-    for value in design_region_fields[0]:
-      self.assertIsInstance(value, onp.ndarray)
-      self.assertEqual(value.ndim, 4)  # dims: freq, x, y, pad
-      self.assertEqual(value.dtype, onp.complex128)
+    monitors = [
+        mpa.EigenmodeCoefficient(simulation,
+                                 mp.Volume(center=center, size=monitor_size),
+                                 mode=1,
+                                 forward=True,
```
oskooi, line 130: Looks like this branch was not rebased from the master branch properly because the `forward` parameter on this line rather than being a constant is swept over `True` and `False`. Also, the line below is `decimation_factor=5` in the master branch.

python/tests/test_adjoint_jax.py
ianwilliamson: Fixed.

python/adjoint/objective.py
ianwilliamson: Fixed.

python/tests/test_adjoint_jax.py
ianwilliamson: Good catch. Fixed!

ianwilliamson:
> The failing test occurs for only the single-precision floating point case. We can just probably lower the tolerance (`epsilon=1e-6`) slightly and it should be fine.

That test seems to be flaky.

PR #1594, opened 2021-06-07
Fix invalid escape sequences

ahoenselaar:


oskooi (approved):

oskooi:
Note that if you edit the docstrings in the Python source files, you will need to manually run the script `doc/generate_py_api.py` in order to generate the markdown pages for the user manual. The file [`Python_User_Interface.md.in`](https://github.com/NanoComp/meep/blob/master/doc/docs/Python_User_Interface.md.in) is used as input to the script to generate [`Python_User_Interface.md`](https://github.com/NanoComp/meep/blob/master/doc/docs/Python_User_Interface.md) and the latter file is not edited directly.

The instructions for revising the documentation for the user manual are described in https://github.com/NanoComp/meep/blob/master/doc/README.md. To verify the correctness of the proposed changes, you can also launch a local `mkdocs` server and view the user manual from your browser which is also described in that link.

ahoenselaar:
Thank you for the tip! This PR does not cause any diff in the generated documentation. 
Its sole purpose is to remove the DeprecationWarning.

stevengj:
Thanks!

PR #1596, opened 2021-06-08
GitHub workflow for CI

ahoenselaar:
Scope:
* C++ tests with MPI
* C++ tests without MPI
* Python v3.6 tests (MPI disabled) via pytest
* Python v3.9 tests (MPI disabled) via pytest

This is intended as an initial set up step.

The failing Python tests are going to be addressed in follow-up PRs. Issues identified so far include:
* Non-hermetic tests that break subsequent tests (`verbosity_mgr`)
* Memory corruption in pympb that causes later tests to fail

Other improvements for follow-up PRs:
* Cleaner presentation of test results
* Automatic inclusion of a "unit test badge" in PRs

### DIFF
--- a/python/Makefile.am
+++ b/python/Makefile.am
@@ -232,8 +232,9 @@ adjoint_PYTHON  = $(srcdir)/adjoint/__init__.py              \
 PY_PKG_FILES = $(INIT_PY) $(HL_IFACE) .libs/_meep.so
 
 meep: _meep.la $(MPB_LA) __init__.py $(HL_IFACE)
-	mkdir -p meep
+	mkdir -p meep/adjoint
 	cp $(PY_PKG_FILES) meep
+	cp $(adjoint_PYTHON) meep/adjoint
 if WITH_MPB
 	mkdir -p meep/mpb
 	cp .libs/_mpb.so meep/mpb
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -1,0 +1,71 @@
+name: CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+  workflow_dispatch:
+
+jobs:
+  build:
+    name: "Test ${{ matrix.name-prefix }}"
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name-prefix: "C++ without MPI"
+            enable-mpi: false
+            configure-options: "--without-python"
+            python-version: "none"
+          - name-prefix: "C++ with MPI"
+            enable-mpi: true
+            configure-options: "--without-python --with-mpi"
+            python-version: "none"
+          - name-prefix: "Python 3.6"
+            enable-mpi: false
+            configure-options: "--with-python --enable-maintainer-mode"
+            python-version: "3.6"
+          - name-prefix: "Python 3.9"
+            enable-mpi: false
+            configure-options: "--with-python --enable-maintainer-mode"
+            python-version: "3.9"
+
+    steps:
+    - name: Install Dependencies
+      run: |
+        sudo apt-get update -y
+        sudo apt-get install -y autoconf automake libaec-dev libctl-dev libfftw3-dev libgdsii-dev libgsl-dev libharminv-dev libhdf5-dev libtool mpb mpb-dev swig
+    - name: Install MPI
+      if: matrix.enable-mpi
+      run: |
+        sudo apt-get install libhdf5-openmpi-dev libopenmpi-dev
+        sudo update-alternatives --set hdf5.pc /usr/lib/x86_64-linux-gnu/pkgconfig/hdf5-openmpi.pc
+    - name: Check out repository
+      uses: actions/checkout@v2
+    - name: Set up Python (version ${{ matrix.python-version }})
+      if: ${{ matrix.python-version != 'none' }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install Python dependencies
+      if: ${{ matrix.python-version != 'none' }}
+      run: pip install autograd h5py jax jaxlib matplotlib mpi4py numpy parameterized pytest scipy
+    - name: Autoreconf
+      run: autoreconf -if
+    - name: configure
+      run: ./configure --with-hdf5 --without-scheme --enable-shared ${{ matrix.configure-options }}
+    - name: make
+      run: make
+    - name: Run Python tests
+      if: ${{ matrix.python-version != 'none' }}
+      env:
+        MEEP_SKIP_LARGE_TESTS: 1
+      run: |
+        export PYTHONPATH="${PWD}/python:${PYTHONPATH}"
+        pytest python/tests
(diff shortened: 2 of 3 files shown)


oskooi:
When running `make check` on my local machine for this branch, `python/tests/test_simulation.py` is failing with the following error:
```
FAIL: test_at_time (__main__.TestSimulation)
  File "./tests/test_simulation.py", line 142, in test_at_time
    self.assertTrue(os.path.exists(fname))
AssertionError: False is not true

FAIL: test_get_filename_prefix (__main__.TestSimulation)
  File "./tests/test_simulation.py", line 532, in test_get_filename_prefix
    self.assertEqual(sim.get_filename_prefix(), "simulation")
AssertionError: 'test_simulation' != 'simulation'
- test_simulation
? -----
+ simulation


FAIL: test_use_output_directory_default (__main__.TestSimulation)
  File "./tests/test_simulation.py", line 134, in test_use_output_directory_default
    self.assertTrue(os.path.exists(os.path.join(output_dir, self.fname)))
AssertionError: False is not true

FAIL: test_with_prefix (__main__.TestSimulation)
  File "./tests/test_simulation.py", line 162, in test_with_prefix
    self.assertTrue(os.path.exists(fname))
AssertionError: False is not true

Ran 32 tests in 40.139s

FAILED (failures=4)

```

configure.ac
```diff
@@ -219,24 +219,38 @@ AC_CHECK_LIB([gsl],[gsl_sf_bessel_Jn], [],
...
+    # Attempt to locate HDF5 via pkg-config and fall back to a default.
+    PKG_CHECK_MODULES(HDF5, hdf5, [
+        CPPFLAGS="$CPPFLAGS $HDF5_CFLAGS"
+        LIBS="$HDF5_LIBS $LIBS"
+    ], [
+        LIBS="-lhdf5 $LIBS"
+    ])
+
+    if test "x$with_mpi" = "xyes"; then
+        CC=$MPICPP
```
stevengj, line 237: This rename doesn't seem right

stevengj: (This is a user-visible name, people use it in their scripts to specify the MPI C++ compiler, and it's also used by the AX_MPI macro that we use.)

configure.ac
```diff
@@ -219,24 +219,38 @@ AC_CHECK_LIB([gsl],[gsl_sf_bessel_Jn], [],
...
+    save_CC="$CC"
+    save_CPPFLAGS="$CPPFLAGS"
+    save_LIBS="$LIBS"
+
+    # Attempt to locate HDF5 via pkg-config and fall back to a default.
+    PKG_CHECK_MODULES(HDF5, hdf5, [
+        CPPFLAGS="$CPPFLAGS $HDF5_CFLAGS"
+        LIBS="$HDF5_LIBS $LIBS"
+    ], [
+        LIBS="-lhdf5 $LIBS"
```
stevengj, line 233: This isn't right — we can't add `-lhdf5` by default because Meep still compiles even if HDF5 is not present at all.  You have to fall back to the `AC_CHECK_LIB(` test below.

Furthermore, the HDF5 test needs to come *after* you set the MPI compiler, because the parallel version of HDF5 must be linked with MPI in order for linking to succeed.

configure.ac
```diff
@@ -219,24 +219,38 @@ AC_CHECK_LIB([gsl],[gsl_sf_bessel_Jn], [],
 # Check for HDF5 library
 AC_ARG_WITH(hdf5, [AC_HELP_STRING([--without-hdf5],[do not allow HDF5 output])], ok=$withval, ok=yes)
 if test "$ok" = "yes"; then
```
stevengj, line 221: I wouldn't recommend changing the configure.ac script at all in this PR if possible.  For example, if you want to use HDF5 from pkgconfig to find HDF5, you can just configure with
```sh
./configure LDFLAGS=`pkg-config --libs-only-L hdf5` CPPFLAGS=`pkg-config --cflags-only-I hdf5`
```
or something like that.

stevengj: We can update the configure script to use `pkg-config` in a separate PR if we want.

configure.ac
stevengj: If github uses `$MPICPP`, then just pass `MPICXX=$MPICPP` to the `configure` script.

python/Makefile.am
```diff
@@ -8,22 +8,22 @@ if WITH_MPI
   MDPYTEST=
   MPBPYTEST=
 else
-  MDPYTEST=$(TEST_DIR)/material_dispersion.py
+  MDPYTEST=$(TEST_DIR)/test_material_dispersion.py
```
stevengj, line 11: It shouldn't be necessary to start every file with `test_*.py` — they are in a "tests" directory after all.

stevengj: In any case, it just makes it harder to review the PR if it mixes multiple unrelated changes.

python/Makefile.am
```diff
@@ -232,8 +232,9 @@ adjoint_PYTHON  = $(srcdir)/adjoint/__init__.py              \
 PY_PKG_FILES = $(INIT_PY) $(HL_IFACE) .libs/_meep.so
 
 meep: _meep.la $(MPB_LA) __init__.py $(HL_IFACE)
-	mkdir -p meep
+	mkdir -p meep/adjoint
 	cp $(PY_PKG_FILES) meep
+	cp $(adjoint_PYTHON) meep/adjoint
```
stevengj, line 237: What is going on with this change?

configure.ac
ahoenselaar: This section performs set up work for the actual check of HDF5 functionality below (via AC_CHECK_LIB and AC_CHECK_HEADER) by setting appropriate LIBS and CPPFLAGS values.

If the actual HDF5 check fails, then LIBS, CPPFLAGS and CC are all reset to their saved values.
If the actual HDF5 check succeeds, then only CC is reset to its saved non-MPI value.

configure.ac
ahoenselaar: Yes, this should be CC=$MPICXX, just like in the previous implementation.
I will push a commit to fix this.

python/Makefile.am
ahoenselaar: The use of the `test_` prefix for files and test functions is a pytest convention.
It makes it easy to distinguish actual test cases and functions from helpers such as those in `utils.py`.
Pytest is so widely used across Python projects, that most developers are likely confused by the lack of the `test_` prefix.

I am going to split this PR to facilitate reviews.

python/Makefile.am
ahoenselaar: Before this change, only the `$PY_PKG_FILES` were copied to the `meep` folder.
In a second step, the MPB interface was added below.
However, the `adjoint` subpackage was never copied to `meep`.

Do you have any insight why that is the case?

stevengj: Our Python Makefile rules seem kind of hacky here; it would be better to rewrite them to cooperate with automake if possible.

ahoenselaar:
Example output: https://github.com/ahoenselaar/meep/runs/2796065822

Shows a few broken tests and a segfault due to memory corruption.
Tests after the verbosity manager test fail but that will be addressed by #1602.

PR #1598, opened 2021-06-09
Improve autoconf handling for HDF5 library.

ahoenselaar:
*  Use pkg-config to detect HDF5 when available.
*  Add CPPFLAGS to account for HDF5 library headers.

These steps improve the chances of locating the library of Ubuntu systems and other modern distributions.

Implementation:
* Optimistically update LIBS and CPPFLAGS from pkg-config or by assuming a default
* Perform AC_CHECKs
* Revert LIBS and CPPFLAGS if any of the checks fail.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -219,24 +219,34 @@ AC_CHECK_LIB([gsl],[gsl_sf_bessel_Jn], [],
 # Check for HDF5 library
 AC_ARG_WITH(hdf5, [AC_HELP_STRING([--without-hdf5],[do not allow HDF5 output])], ok=$withval, ok=yes)
 if test "$ok" = "yes"; then
-	AC_CHECK_LIB(z,deflate, [],[AC_MSG_WARN([zlib is required for HDF5!])])
+    AC_CHECK_LIB(z,deflate, [],[AC_MSG_WARN([zlib is required for HDF5!])])
 
-	save_CC=$CC
-	if test "x$with_mpi" = "xyes"; then
-	   CC=$MPICXX
-	fi
-	save_LIBS_0="$LIBS" # save, to check later if we found any library
-	AC_CHECK_LIB(hdf5, H5Pcreate, [
-		AC_CHECK_HEADERS(hdf5.h, [LIBS="-lhdf5 $LIBS"
-                     AC_DEFINE(HAVE_HDF5,1,[Define if we have & link HDF5])])])
-	if test x"$save_LIBS_0" = x"$LIBS"; then
-		AC_MSG_WARN([Couldn't find the HDF5 library!!  Switching to --without-hdf5.])
-	fi
-	CC=$save_CC
+    save_CPPFLAGS="$CPPFLAGS"
+    save_LIBS="$LIBS"
 
-        if test "x$with_mpi" = "xyes"; then
-		AC_CHECK_FUNCS(H5Pset_mpi H5Pset_fapl_mpio)
-	fi
+    # Check whether linking with "-lhdf5" just works (installation in default
+    # location or due to user-provided flags). If not, try to locate HDF5 via
+    # pkg-config.  Optimistically update LIBS and CPPFLAGS. Revert if any
+    # AC_CHECKs fail.
+    AC_CHECK_LIB(hdf5, H5Pcreate, [], [
+        PKG_CHECK_MODULES(HDF5, hdf5, [
+            CPPFLAGS="$CPPFLAGS $HDF5_CFLAGS"
+            LIBS="$HDF5_LIBS $LIBS"
+        ], [])
+    ])
+
+    AC_CHECK_FUNC(H5Pcreate, [
+        AC_CHECK_HEADER(hdf5.h, [
+            AC_DEFINE(HAVE_HDF5,1,[Define if we have & link HDF5])
+            AC_CHECK_FUNCS(H5Pset_mpi H5Pset_fapl_mpio)
+        ])
+    ])
+
+    if test "x$ac_cv_header_hdf5_h" != "xyes"; then
+        AC_MSG_WARN([Couldn't find the HDF5 header or library!!  Switching to --without-hdf5.])
+        CPPFLAGS="$save_CPPFLAGS"
+        LIBS="$save_LIBS"
+    fi
 fi
 
 ##############################################################################


configure.ac
```diff
@@ -219,24 +219,39 @@ AC_CHECK_LIB([gsl],[gsl_sf_bessel_Jn], [],
...
+    ], [
+        LIBS="-lhdf5 $LIBS"
+    ])
+
+    if test "x$with_mpi" = "xyes"; then
+        CC=$MPICXX
+        AC_CHECK_FUNCS(H5Pset_mpi H5Pset_fapl_mpio)
+    fi
+
+    AC_CHECK_LIB(hdf5, H5Pcreate, [
```
stevengj, line 242: `AC_CHECK_LIB` adds `-lhdf5`, which you don't want to do if you've already added the library above.  You want `AC_CHECK_FUNC` instead.

configure.ac
```diff
@@ -219,24 +219,39 @@ AC_CHECK_LIB([gsl],[gsl_sf_bessel_Jn], [],
...
+    AC_CHECK_LIB(z,deflate, [],[AC_MSG_WARN([zlib is required for HDF5!])])
+
+    save_CC="$CC"
+    save_CPPFLAGS="$CPPFLAGS"
+    save_LIBS="$LIBS"
+
+    # Attempt to locate HDF5 via pkg-config and fall back to a default.
+    # Optimistically update LIBS and CPPFLAGS. Revert if any AC_CHECKs fail.
+    PKG_CHECK_MODULES(HDF5, hdf5, [
+        CPPFLAGS="$CPPFLAGS $HDF5_CFLAGS"
```
stevengj, line 231: This is theoretically problematic because `HDF5_CFLAGS` includes compiler flags as well as preprocessor flags, but `CPPFLAGS` is supposed to only have preprocessor flags.  In practice we are probably fine.

configure.ac
```diff
@@ -219,24 +219,39 @@ AC_CHECK_LIB([gsl],[gsl_sf_bessel_Jn], [],
...
-	fi
+    AC_CHECK_LIB(z,deflate, [],[AC_MSG_WARN([zlib is required for HDF5!])])
+
+    save_CC="$CC"
+    save_CPPFLAGS="$CPPFLAGS"
+    save_LIBS="$LIBS"
+
+    # Attempt to locate HDF5 via pkg-config and fall back to a default.
+    # Optimistically update LIBS and CPPFLAGS. Revert if any AC_CHECKs fail.
+    PKG_CHECK_MODULES(HDF5, hdf5, [
```
stevengj, line 230: One issue with using `PKG_CHECK_MODULES` like this is that if the user explicitly set `-L` and `-I` flags to find a particular version of HDF5 on their system, using `pkg-config` unconditionally will override that choice.  It is probably better to do `AC_CHECK_LIB` first.   See also https://stackoverflow.com/questions/10220946/pkg-check-modules-considered-harmful

In particular, it seems like we should so something like
```sh
# HDF5 libraries may need to be linked with mpicxx if they were compiled with MPI
if test "x$with_mpi" = "xyes"; then
    CC=$MPICXX
fi

# see if -lhdf5 works, then fall back to pkg-config flags (checked with AC_CHECK_FUNC below)
AC_CHECK_LIB(hdf5, H5Pcreate, [], [
    PKG_CHECK_MODULES(HDF5, hdf5, [
        CPPFLAGS="$CPPFLAGS $HDF5_CFLAGS" # assumes that CFLAGS actually contains only -I flags
        LIBS="$HDF5_LIBS $LIBS"
    ], [ac_cv_header_hdf5_h=no])
])
```
and then change the test below to `AC_CHECK_FUNC` instead of `AC_CHECK_LIBS`.   We can also just run `AC_CHECK_FUNCS(H5Pset_mpi H5Pset_fapl_mpio)` unconditionally below.

configure.ac
ahoenselaar: The default action of adding `-lhdf5` is only taken if *action-if-found* is not specified. Here, we are providing an *action-if-found*.

https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Libraries.html

configure.ac
ahoenselaar: I updated the PR based on your suggestions:
Check whether `-lhdf5` just works, otherwise try `pkg-config`.
Use AC_CHECK_FUNC and AC_CHECK_HEADER to determine whether the derived configuration works correctly.
Unconditionally check for `H5Pset_mpi` and `H5Pset_fapl_mpio`.

configure.ac
```diff
@@ -219,24 +219,41 @@ AC_CHECK_LIB([gsl],[gsl_sf_bessel_Jn], [],
...
+    AC_CHECK_LIB(hdf5, H5Pcreate, [], [
+        PKG_CHECK_MODULES(HDF5, hdf5, [
+            CPPFLAGS="$CPPFLAGS $HDF5_CFLAGS"
+            LIBS="$HDF5_LIBS $LIBS"
+        ], [])
+    ])
+
+    if test "x$with_mpi" = "xyes"; then
+        CC=$MPICXX
+    fi
```
stevengj, line 241: This needs to go before the `AC_CHECK_LIB`.    Maybe add a comment `# HDF5 may require the MPI compiler for linking to succeed`

configure.ac
```diff
@@ -219,24 +219,41 @@ AC_CHECK_LIB([gsl],[gsl_sf_bessel_Jn], [],
...
+
+    AC_CHECK_FUNC(H5Pcreate, [
+        AC_CHECK_HEADER(hdf5.h, [
+            AC_DEFINE(HAVE_HDF5,1,[Define if we have & link HDF5])
+        ])
+    ])
+    AC_CHECK_FUNCS(H5Pset_mpi H5Pset_fapl_mpio)
```
stevengj, line 248: I guess we might as well put this `AC_CHECK_FUNCS` line inside the ACTION-IF-TRUE of `AC_CHECK_HEADER`

configure.ac
ahoenselaar: I removed this part and verified that  MPICXX is used for the check.

configure.ac
ahoenselaar: Done.

PR #1601, opened 2021-06-10
Add unit test for backward-propagating mode in oblique waveguide

oskooi:
Adds a unit test to verify that for a backward-propagating mode launched within an oblique (single-mode) waveguide, the Poynting flux and the square of the mode coefficient computed downstream from the source are the same.

The figure shows the simulation setup and a snapshot of the E<sub>z</sub> fields: an eigenmode source (red line) launches a *leftward*-propagating mode. The DFT fields monitor (blue line) computes the Poynting flux exiting the waveguide as well as the mode coefficient of the backward-propagating mode. (In this figure, the time-profile of the source is a continuous wave whereas in the actual unit test it is a pulse.)

oblique_waveguide_ang35

### DIFF
--- a/python/tests/test_mode_decomposition.py
+++ b/python/tests/test_mode_decomposition.py
@@ -1,6 +1,6 @@
 import unittest
+import numpy as np
 import meep as mp
-
 
 class TestModeDecomposition(unittest.TestCase):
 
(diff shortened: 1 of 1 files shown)


python/tests/test_mode_decomposition.py
```diff
@@ -83,6 +83,50 @@ def test_mode_decomposition(self):
...
+
+        sim.run(until_after_sources=30)
+
+        flux = mp.get_fluxes(mode)[0]
+        coeff = sim.get_eigenmode_coefficients(mode,[1],
+                                               direction=mp.NO_DIRECTION,
+                                               kpoint_func=lambda f,n: kpoint).alpha[0,0,0]
+
+        print("oblique-waveguide-flux:, {:.6f}, {:.6f}".format(-flux, abs(coeff)**2))
+        self.assertAlmostEqual(-flux, abs(coeff)**2, places=1)
```
stevengj, line 129: `places=1` seems like a low tolerance.   Do they converge?

stevengj: Actually, since this is an *absolute* error, I have no idea whether it is a small or a large tolerance?  

What is the magnitude of `|flux|` here?

oskooi: This is the output of the print statement:
```
oblique-waveguide-flux:, 100.008731, 99.977897

```

oskooi: Fixed.

stevengj: Okay, so `places=1` is actually a relative tolerance of `1e-4`, which is pretty good!

stevengj: We could do
```
self.assertAlmostEqual(-1, abs(coeff)**2/flux, places=2)
```
and then maybe we could use a lower resolution.

oskooi:
```
    self.assertAlmostEqual(-1,abs(coeff)**2/flux,places=3)
AssertionError: -1 != -0.9994960396371381 within 3 places
```

Issue #1605, opened 2021-06-10
Make memory management more automatic


ahoenselaar:
Transition gradually towards less manual memory management by taking advantage of C++ features (up to C++11)

PR #1606, opened 2021-06-10
Remove unneeded `copy_from` methods

ahoenselaar:
Rely on auto-generated copy constructors and assignment operators for
`susceptibility` and `transition`.

#1605


ahoenselaar:
This PR unblocks the replacement of `susceptibility_list` with `std::vector<susceptibility>` in commit https://github.com/ahoenselaar/meep/commit/92f0d228bdb8d8ad343c35595c71e8619c3e43c6.

PR #1607, opened 2021-06-11
Replace susceptibility_list with std::vector<susceptibility>

ahoenselaar:
#1605 

To be followed by https://github.com/ahoenselaar/meep/commit/4f51bba534bb4dfd54a8a4ee58126bae9acc5ac2

### DIFF
--- a/python/meep.i
+++ b/python/meep.i
@@ -728,12 +728,6 @@ meep::volume_list *make_volume_list(const meep::volume &v, int c,
 
 %typemap(freearg) GEOMETRIC_OBJECT {
     if ($1.material) {
-        if (((material_data *)$1.material)->medium.E_susceptibilities.items) {
-            delete[] ((material_data *)$1.material)->medium.E_susceptibilities.items;
-        }
-        if (((material_data *)$1.material)->medium.H_susceptibilities.items) {
-            delete[] ((material_data *)$1.material)->medium.H_susceptibilities.items;
-        }
         delete[] ((material_data *)$1.material)->weights;
         delete[] ((material_data *)$1.material)->epsilon_data;
         delete (material_data *)$1.material;
@@ -774,12 +768,6 @@ meep::volume_list *make_volume_list(const meep::volume &v, int c,
 
 %typemap(freearg) geometric_object_list {
     for(int i = 0; i < $1.num_items; i++) {
-        if (((material_data *)$1.items[i].material)->medium.E_susceptibilities.items) {
-            delete[] ((material_data *)$1.items[i].material)->medium.E_susceptibilities.items;
-        }
-        if (((material_data *)$1.items[i].material)->medium.H_susceptibilities.items) {
-            delete[] ((material_data *)$1.items[i].material)->medium.H_susceptibilities.items;
-        }
         delete[] ((material_data *)$1.items[i].material)->epsilon_data;
         delete[] ((material_data *)$1.items[i].material)->weights;
         delete (material_data *)$1.items[i].material;
@@ -808,14 +796,6 @@ meep::volume_list *make_volume_list(const meep::volume &v, int c,
     }
 }
 
-%typemap(arginit) susceptibility_list {
-    $1.num_items = 0;
-    $1.items = NULL;
-}
-
-%typemap(freearg) susceptibility_list {
-    delete[] $1.items;
-}
 
 //--------------------------------------------------
 // typemaps needed for material grid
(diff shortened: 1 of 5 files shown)


stevengj:
LGTM.

(Historically, these `*_list` types were copied from code that was generated by libctl, which generates C code and hence was constrained by that.)

oskooi:
After trying to run `make` on the latest master branch which includes this commit on my local machine (gcc 7.5.0, SWIG 3.0.12, Python 3.5), I am getting a bunch of errors:
```
mpb-python.cxx: In function ‘PyObject* _wrap_new_mode_solver(PyObject*, PyObject*)’:
mpb-python.cxx:7941:43: error: ‘using susceptibility_list = class std::vector<meep_geom::susceptibility_struct> {aka class std::vector<meep_geom::susceptibility_struct>}’ has no member named ‘items’
       if (arg6->medium.E_susceptibilities.items) {
                                           ^~~~~
mpb-python.cxx:7942:50: error: ‘using susceptibility_list = class std::vector<meep_geom::susceptibility_struct> {aka class std::vector<meep_geom::susceptibility_struct>}’ has no member named ‘items’
         delete[] arg6->medium.E_susceptibilities.items;
                                                  ^~~~~
mpb-python.cxx:7944:43: error: ‘using susceptibility_list = class std::vector<meep_geom::susceptibility_struct> {aka class std::vector<meep_geom::susceptibility_struct>}’ has no member named ‘items’
       if (arg6->medium.H_susceptibilities.items) {
                                           ^~~~~
mpb-python.cxx:7945:50: error: ‘using susceptibility_list = class std::vector<meep_geom::susceptibility_struct> {aka class std::vector<meep_geom::susceptibility_struct>}’ has no member named ‘items’
         delete[] arg6->medium.H_susceptibilities.items;
                                                  ^~~~~
mpb-python.cxx:7956:43: error: ‘using susceptibility_list = class std::vector<meep_geom::susceptibility_struct> {aka class std::vector<meep_geom::susceptibility_struct>}’ has no member named ‘items’
       if (arg6->medium.E_susceptibilities.items) {
                                           ^~~~~
mpb-python.cxx:7957:50: error: ‘using susceptibility_list = class std::vector<meep_geom::susceptibility_struct> {aka class std::vector<meep_geom::susceptibility_struct>}’ has no member named ‘items’
         delete[] arg6->medium.E_susceptibilities.items;
```

ahoenselaar:
Your SWIG wrapper has not been regenerated. Try to delete the files
matching the pattern python/*.cxx
I think that the `clean` operation is not set up correctly.

oskooi:
Good catch. That fixed it. Note that I ran `make distclean` before building which means that indeed `make distclean` is not working properly.

ahoenselaar:
It appears there is a separate operation: `maintainer-clean`

https://github.com/NanoComp/meep/blob/2dcfa9fcc226e07bf3d177ba8318c867a79567a9/.travis.yml#L181
🤷

stevengj:
`make distclean` restores the tree to the state of a fresh unpacking of the "distributed" `.tar.gz` file — hence, it does not remove any files that are included in the `.tar.gz` file.

The swig-generated files are included in the `.tar.gz` file, because users aren't required to have swig on their machines to compile Meep.   Hence, this is a "maintainer"-generated file and requires `maintainer-clean` to remove it.  See also https://www.gnu.org/prep/standards/html_node/Standard-Targets.html

However, you shouldn't need to manually "clean" at all — it should automatically rebuild as needed.   What this means is that there is apparently a missing dependency in the Makefile rule for `mpb-python.cxx`.

stevengj:
Yup, looks like the Makefile is missing a dependency.  This fixes it:
```diff
diff --git a/python/Makefile.am b/python/Makefile.am
index 1f7d88ca..75f6f2cc 100644
--- a/python/Makefile.am
+++ b/python/Makefile.am
@@ -167,7 +167,7 @@ meep-python.cxx: $(MEEP_SWIG_SRC) $(HPPFILES)
 
 if WITH_MPB
 MPB_SWIG_SRC = mpb.i
-mpb-python.cxx: $(MPB_SWIG_SRC) $(top_srcdir)/libpympb/pympb.hpp $(top_srcdir)/src/meepgeom.hpp
+mpb-python.cxx: $(MPB_SWIG_SRC) $(top_srcdir)/libpympb/pympb.hpp $(top_srcdir)/src/meepgeom.hpp $(top_srcdir)/src/material_data.hpp
        $(SWIG) -Wextra $(AM_CPPFLAGS) $(PYMPBINCLUDE) -outdir $(builddir) -c++ -nofastunpack -python -o $@ $(srcdir)/mpb.i
 mpb.py: mpb-python.cxx
 MPB_PY = mpb.py
```
I'll push a patch.

ahoenselaar:
I cannot say that the use case is clear to me.

This maintainer mode is for somebody who:
* wants to build from source
* wants to build the Python bindings (and thus needs the `py*.cxx` files
* but does not have SWIG installed?

ahoenselaar:
> ```diff
> MPB_SWIG_SRC
> ```

What about `meep.hpp`?

stevengj:
Maintainer-mode is for omeone who is building from `git clone`, and who therefore *must* have `swig` installed (along with`autoconf` etc).

As opposed to someone who is building from source using a released `.tar.gz` file — they don't need `swig` etcetera because the `.tar.gz` file (produced by `make dist`) includes the generated source files.

stevengj:
> What about `meep.hpp`?

MPB doesn't use it.

PR #1608, opened 2021-06-12
fix failing tests for GitHub Actions CI

oskooi:
Initial attempt to eventually get all unit tests of the `make check` suite to pass on GitHub Actions CI (which was recently set up in #1596).

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -34,19 +34,6 @@ jobs:
             python-version: "3.9"
 
     steps:
-    - name: Install common dependencies
-      run: |
-        sudo apt-get -y update
-        sudo apt-get -y install autoconf automake libfftw3-dev libgsl-dev liblapack-dev guile-3.0-dev libpng-dev libtool swig
-        git clone https://github.com/NanoComp/libctl.git libctl-src
-        (cd libctl-src && git checkout master && sh autogen.sh --prefix=${HOME}/local --enable-shared && make -j 2 && make install)
-        git clone https://github.com/NanoComp/harminv.git
-        (cd harminv && git checkout master && sh autogen.sh --prefix=${HOME}/local --enable-shared && make -j 2 && make install)
-        git clone https://github.com/NanoComp/mpb.git
-        (cd mpb && git checkout master && sh autogen.sh --prefix=${HOME}/local --enable-shared LIBS=-ldl --with-libctl=${HOME}/local/share/libctl --with-hermitian-eps && make -j 2 && make install)
-        git clone https://github.com/HomerReid/libGDSII.git
-        (cd libGDSII && git checkout master && sh autogen.sh --prefix=${HOME}/local && make install)
-
     - name: Define common environment variables
       run: |
         echo "CPPFLAGS=-I${HOME}/local/include" >> $GITHUB_ENV
@@ -61,6 +48,19 @@ jobs:
         echo "HDF5_PARALLEL_CPPFLAGS=${HDF5_BASE_CPPFLAGS}/openmpi" >> $GITHUB_ENV
         echo "HDF5_SERIAL_LDFLAGS=${HDF5_BASE_LDFLAGS}/serial" >> $GITHUB_ENV
         echo "HDF5_PARALLEL_LDFLAGS=${HDF5_BASE_LDFLAGS}/openmpi" >> $GITHUB_ENV
+
+    - name: Install common dependencies
+      run: |
+        sudo apt-get -y update
+        sudo apt-get -y install autoconf automake libfftw3-dev libgsl-dev liblapack-dev guile-3.0-dev libpng-dev libtool swig
+        git clone https://github.com/NanoComp/libctl.git libctl-src
+        (cd libctl-src && git checkout master && sh autogen.sh --prefix=${HOME}/local --enable-shared && make -j 2 && make install)
+        git clone https://github.com/NanoComp/harminv.git
+        (cd harminv && git checkout master && sh autogen.sh --prefix=${HOME}/local --enable-shared && make -j 2 && make install)
+        git clone https://github.com/NanoComp/mpb.git
+        (cd mpb && git checkout master && sh autogen.sh --prefix=${HOME}/local --enable-shared LIBS=-ldl --with-libctl=${HOME}/local/share/libctl --with-hermitian-eps && make -j 2 && make install)
+        git clone https://github.com/HomerReid/libGDSII.git
+        (cd libGDSII && git checkout master && sh autogen.sh --prefix=${HOME}/local && make install)
 
     - name: Define environment variables for serial build
       if: ${{ matrix.enable-mpi == false }}
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -52,18 +52,18 @@ jobs:
     - name: Define environment variables for serial build
       if: ${{ matrix.enable-mpi == false }}
       run: |
-        echo "MPICONF=--without-mpi" >> $GITHUB_ENV
-        echo "MKCHECKFLAGS=-j 2" >> $GITHUB_ENV
-        echo "CPPFLAGS=${HDF5_SERIAL_CPPFLAGS} ${CPPFLAGS}" >> $GITHUB_ENV
-        echo "LDFLAGS="${HDF5_SERIAL_LDFLAGS} ${LDFLAGS}" >> $GITHUB_ENV
+        echo "MPICONF=\"--without-mpi\"" >> $GITHUB_ENV
+        echo "MKCHECKFLAGS=\"-j 2\"" >> $GITHUB_ENV
+        echo "CPPFLAGS=\"${HDF5_SERIAL_CPPFLAGS} ${CPPFLAGS}\"" >> $GITHUB_ENV
+        echo "LDFLAGS=\"${HDF5_SERIAL_LDFLAGS} ${LDFLAGS}\"" >> $GITHUB_ENV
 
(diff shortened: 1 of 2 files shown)


oskooi:
According to the [CI logs](https://github.com/NanoComp/meep/pull/1608/checks?check_run_id=2808376340), there is a segmentation fault occurring for `python/tests/test_mode_coeffs.py`. I suspect this is probably related to the MPB library (v1.9 from June 2019) which is installed using the [precompiled Ubuntu 20.04 package](https://packages.ubuntu.com/focal/mpb):

https://github.com/NanoComp/meep/blob/f528873eb5cf680976102f5b307b505ab25c4b78/.github/workflows/build-ci.yml#L40

The fix likely involves installing MPB from source following a similar procedure to the one outlined in the [user manual](https://meep.readthedocs.io/en/latest/Build_From_Source/#ubuntu-1604-and-1804) for Ubuntu 18.04.

ahoenselaar:
To keep the worfklow readable, we could set up a separate build process for MPB.
Ideally, it would build a package, e.g. in DEB format, that we can install during the setup for Meep's CI worklow.
https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows

.github/workflows/build-ci.yml
```diff
@@ -18,31 +18,72 @@ jobs:
...
+        export MY_CPPFLAGS="-I/usr/local/include -I/usr/include/hdf5/openmpi"
+        export MY_CC="mpicc"
+        sudo apt-get -y install libhdf5-openmpi-dev libopenmpi-dev hdf5-tools
+    - name: Install Other Dependencies
+      run: |
+        cd ~/install
+        git clone https://github.com/NanoComp/h5utils.git
+        cd h5utils/
+        sh autogen.sh CC=${MY_CC} LDFLAGS="${MY_LDFLAGS}" CPPFLAGS="${MY_CPPFLAGS}"
+        make && sudo make install
```
stevengj, line 78: Do the tests actually need h5utils?

.github/workflows/build-ci.yml
```diff
@@ -53,10 +94,9 @@ jobs:
     - name: Install Python dependencies
       if: ${{ matrix.python-version != 'none' }}
       run: pip install autograd h5py jax jaxlib matplotlib mpi4py numpy parameterized pytest scipy
-    - name: Autoreconf
-      run: autoreconf -if
     - name: configure
-      run: ./configure --with-hdf5 --without-scheme --enable-shared ${{ matrix.configure-options }}
+      run: |
+        sh autogen.sh --without-scheme --enable-shared CC=${MY_CC} LDFLAGS="${MY_LDFLAGS}" CPPFLAGS="${MY_CPPFLAGS}" ${{ matrix.configure-options }}
```
stevengj, line 99: Shouldn't this be setting `CXX=${MY_CXX}` to build Meep?   We don't really use the C compiler (`CC`).

.github/workflows/build-ci.yml
oskooi: Yes, h5utils is in fact unnecessary and has been removed.

.github/workflows/build-ci.yml
oskooi: Fixed.

python/tests/test_mode_coeffs.py
```diff
@@ -76,12 +76,12 @@ def run_mode_coeffs(self, mode_num, kpoint_func, nf=1, resolution=15):
...
-        self.assertTrue(res.kdom[0].close(mp.Vector3(0.604301, 0, 0)))
-        self.assertTrue(res.kdom[1].close(mp.Vector3(0.494353, 0, 0), tol=1e-2))
-        self.assertAlmostEqual(res.cscale[0],0.50000977,places=5)
-        self.assertAlmostEqual(res.cscale[1],0.50096888,places=5)
+        self.assertTrue(res.kpoints[0].close(mp.Vector3(0.6056758, 0, 0), tol=1e-2))
+        self.assertTrue(res.kpoints[1].close(mp.Vector3(0.4995093, 0, 0), tol=1e-2))
+        self.assertTrue(res.kdom[0].close(mp.Vector3(0.6056758, 0, 0), tol=1e-2))
+        self.assertTrue(res.kdom[1].close(mp.Vector3(0.4995093, 0, 0), tol=1e-2))
+        self.assertAlmostEqual(res.cscale[0],0.5000020,places=2)
+        self.assertAlmostEqual(res.cscale[1],0.5003438,places=2)
```
stevengj, line 84: Why did these numbers change recently?

oskooi: These tests were failing likely because of the older precompiled MPB library for Ubuntu 20.04 ([version 1.9.0](https://packages.ubuntu.com/focal/mpb) from June 2019) that (serial) Meep was linking to using the current GitHub Actions CI setup:

https://github.com/NanoComp/meep/blob/a31309269bb212f192213143816cd40bf31bf053/.github/workflows/build-ci.yml#L40

Hopefully, once the latest MPB is built from source in this PR, the original unmodified `test_mode_coeffs.py` tests will pass and these changes can then be reverted.

oskooi: Changes have been reverted since they now seem to be unnecessary.

oskooi:
Seems to be finally working. All unit tests for C++ and Python (3.6 and 3.9) for serial and MPI builds are passing. The unit tests are all executed as part of `make distcheck` which in turn calls `make check` from the root directory. We will probably want to modify this such that the Python unit tests are executed separately using `pytest` similar to what was originally set up in #1596.  For now, this PR provides the minimum functionality necessary for CI.

.github/workflows/build-ci.yml
```diff
@@ -16,56 +16,102 @@ jobs:
...
+        echo "HDF5_BASE_CPPFLAGS=-I/usr/include/hdf5" >> $GITHUB_ENV
+        echo "HDF5_BASE_LDFLAGS=-L/usr/lib/x86_64-linux-gnu/hdf5" >> $GITHUB_ENV
+        echo "GEN_CTL_IO=${HOME}/local/bin/gen-ctl-io" >> $GITHUB_ENV
+
+    - name: Define more common environment variables
+      run: |
+        echo "HDF5_SERIAL_CPPFLAGS=${HDF5_BASE_CPPFLAGS}/serial" >> $GITHUB_ENV
+        echo "HDF5_PARALLEL_CPPFLAGS=${HDF5_BASE_CPPFLAGS}/openmpi" >> $GITHUB_ENV
+        echo "HDF5_SERIAL_LDFLAGS=${HDF5_BASE_LDFLAGS}/serial" >> $GITHUB_ENV
+        echo "HDF5_PARALLEL_LDFLAGS=${HDF5_BASE_LDFLAGS}/openmpi" >> $GITHUB_ENV
```
ahoenselaar, line 50: Can `update-alternatives` in conjunction with `pkg-config` not be used here?

.github/workflows/build-ci.yml
```diff
@@ -16,56 +16,102 @@ jobs:
...
+
+    - name: Install common dependencies
+      run: |
+        sudo apt-get -y update
+        sudo apt-get -y install autoconf automake libfftw3-dev libgsl-dev liblapack-dev guile-3.0-dev libpng-dev libtool swig
+        git clone https://github.com/NanoComp/libctl.git libctl-src
+        (cd libctl-src && git checkout master && sh autogen.sh --prefix=${HOME}/local --enable-shared && make -j 2 && make install)
+        git clone https://github.com/NanoComp/harminv.git
+        (cd harminv && git checkout master && sh autogen.sh --prefix=${HOME}/local --enable-shared && make -j 2 && make install)
+        git clone https://github.com/NanoComp/mpb.git
```
ahoenselaar, line 60: Consider the checkout action instead of cloning the full repo:
https://github.com/actions/checkout

.github/workflows/build-ci.yml
```diff
@@ -16,56 +16,102 @@ jobs:
...
+        (cd mpb && git checkout master && sh autogen.sh --prefix=${HOME}/local --enable-shared LIBS=-ldl --with-libctl=${HOME}/local/share/libctl --with-hermitian-eps && make -j 2 && make install)
+        git clone https://github.com/HomerReid/libGDSII.git
+        (cd libGDSII && git checkout master && sh autogen.sh --prefix=${HOME}/local && make install)
+
+    - name: Define environment variables for serial build
```
ahoenselaar, line 65: -j$(nproc) seems better than hardcoding

.github/workflows/build-ci.yml
```diff
@@ -16,56 +16,102 @@ jobs:
...
+            configure-options: "--without-mpi"
+            python-version: "3.9"
+          - name-prefix: "Python 3.6 with MPI"
+            enable-mpi: true
+            configure-options: "--with-mpi"
+            python-version: "3.6"
+          - name-prefix: "Python 3.9 with MPI"
+            enable-mpi: true
+            configure-options: "--with-mpi"
             python-version: "3.9"
```
ahoenselaar, line 34: If this is a true Cartesian product of MPI x Python version, you don't have to rely on `include` to build the four cases manually: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix

.github/workflows/build-ci.yml
stevengj: Could also just do a `git clone --depth 1` (shallow clone).

stevengj:
Merging now since it is urgent to get CI working on other PRs, and this is in working state.  Please open a separate PR with additional cleanups.

PR #1609, opened 2021-06-12
Make C++11 support mandatory

ahoenselaar:


stevengj:
My thinking in making it optional is that this tests actually checks for a whole array of C++11 features, whereas we only need a few.  So, it's still possible that Meep will compile if C++11 is not fully supported, as long as the compiler has the features that we need — the user will find out when `make` runs in any case.   (The main virtue of the macro is that it turns on compiler flags like `-std=c++11` if needed.)

Maybe just give a warning (with `AC_WARN`) if C++11 support is not detected by the `configure` script.

PR #1610, opened 2021-06-13
Run a sanitizer workflow every 12 hours

ahoenselaar:
The sanitizers have a significantly longer runtime, so I am proposing to run them twice a day.
The workflow publishes logs that can be downloaded for further analysis.

### DIFF
--- a/.github/workflows/build-san.yml
+++ b/.github/workflows/build-san.yml
@@ -1,0 +1,58 @@
+name: Sanitizers
+
+on:
+  schedule:
+    - cron:  "24 10,22 * * *"
+  workflow_dispatch:
+
+jobs:
+  build:
+    name: "Test C++ with ${{ matrix.name-suffix }}"
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name-suffix: "ASAN"
+            sanitizer: "address"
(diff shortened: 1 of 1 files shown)


.github/workflows/build-san.yml
```diff
@@ -0,0 +1,58 @@
...
+    - name: Check out repository
+      uses: actions/checkout@v2
+    - name: Autoreconf
+      run: autoreconf -if
+    - name: configure
+      run: |
+        ./configure \
+          --with-hdf5 --without-mpi --without-python --without-scheme \
+          --enable-shared --enable-maintainer-mode \
+          CPPFLAGS="-fsanitize=${{ matrix.sanitizer }} ${{ matrix.debug-options }}" \
```
stevengj, line 46: Isn't this a `CXXFLAGS` not a `CPPFLAGS`?

stevengj:
Looks fine to me other than the tweak above.  (Once a day seems sufficient to me?)

.github/workflows/build-san.yml
ahoenselaar: Updated

ahoenselaar:
> Looks fine to me other than the tweak above. (Once a day seems sufficient to me?)

Reduced the frequency to once per day.

PR #1612, opened 2021-06-13
Reduce the number of compiler warnings

ahoenselaar:
* Removes an unused functions
* Fixes a few logic errors caused by ignored warnings

### DIFF
--- a/python/meep.i
+++ b/python/meep.i
@@ -832,7 +832,7 @@ void _get_gradient(PyObject *grad, PyObject *fields_a, PyObject *fields_f, PyObj
     void* where;
 
     PyObject *swigobj = PyObject_GetAttrString(grid_volume, "swigobj");
-    SWIG_ConvertPtr(swigobj,&where,NULL,NULL);
+    SWIG_ConvertPtr(swigobj,&where,0,NULL);
     const meep::volume* where_vol = (const meep::volume*)where;
 
     // clean the frequencies array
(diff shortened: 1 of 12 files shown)


python/meep.i
```diff
@@ -841,7 +841,7 @@ void _get_gradient(PyObject *grad, PyObject *fields_a, PyObject *fields_f, PyObj
     if (!PyArray_ISCARRAY(pao_freqs)) meep::abort("Numpy fields array must be C-style contiguous.");
     double *frequencies_c = (double *)PyArray_DATA(pao_freqs);
     int nf = PyArray_DIMS(pao_freqs)[0];
-    if (PyArray_DIMS(pao_grad)[0] != nf) meep::abort("Numpy grad array is allocated for %d frequencies; it should be allocated for %d.",PyArray_DIMS(pao_grad)[0],nf);
+    if (PyArray_DIMS(pao_grad)[0] != nf) meep::abort("Numpy grad array is allocated for %ld frequencies; it should be allocated for %ld.",PyArray_DIMS(pao_grad)[0],nf);
```
stevengj, line 844: This still doesn't seem quite correct.  `PyArray_DIMS` returns a `npy_intp`, which is an "integer the same size as a pointer", similar to `ptrdiff_t`.  On a 64-bit Windows machine, however `%ld` (`long int`) is only 4 bytes, I think?  The standard C formatting code for this is `%td`, I think, but I'm not sure what is standard in C++?

ahoenselaar: Updated to %td

stevengj:
Thanks!

PR #1614, opened 2021-06-14
update autoconf

stevengj:
Update some of the `.m4` files from ancient versions, run `autoupdate` to update the `configure.ac` file for more recent autoconf versions, and remove some other ancient cruft.


stevengj:
Python failures seem to be the same as those on master, so not a regression.

PR #1615, opened 2021-06-14
minor cleanups in meepgeom.hpp

stevengj:
rm a superfluous dependency on `meep.hpp`, and turn a few `#defines` into `consts` so that they are namespaced.


stevengj:
python error is unrelated.

PR #1580, opened 2021-05-28
WIP: Connectivity Constraint for Adjoint Solver

mochen4:
Adding support for 2D connectivity constraint

PR #1624, opened 2021-06-21
WIP: Connectivity Constraint for Adjoint Solver

mochen4:
Connectivity constraint with 3D support. Replaces https://github.com/NanoComp/meep/pull/1580


stevengj:
(In general you can do a forced-push to replace a branch in an existing PR, rather than opening a new PR.  `git push origin +connectivity` where the `+` does a forced push.)

mochen4:
I don't know why test_source.py has segfault and fails with python3.6 and MPI=false. The connectivity feature is basically independent and shouldn't affect anything.

The weird thing is if I comment out this line:
https://github.com/NanoComp/meep/blob/815f7adbe03ed652126b1222cfc2cdc24ff54144/python/adjoint/connectivity.py#L9
then the three adjoint tests will fail as expected because I passed in the default argument for sp_solver here:
https://github.com/NanoComp/meep/blob/815f7adbe03ed652126b1222cfc2cdc24ff54144/python/adjoint/connectivity.py#L14 but test_source.py will pass in this case. If I don't pass the default argument, the adjoint tests will pass but the source test will fail again.

PR #1590, opened 2021-06-03
Apply yapf to `meep.adjoint.*`

ianwilliamson:
This PR applies [YAPF](https://github.com/google/yapf) to `meep.adjoint.*` via the command:

```
yapf -p -i *.py
```

which uses the default PEP8 style.

PR #1625, opened 2021-06-21
WIP: revamped adjoint filters

smartalecH:
Updates all the adjoint filters/projections to:
* use `jax` rather than `autograd`.
* filter/project along any arbitrary dimension (i.e. support for 1D, 2D, and 3D designs)
* simplifies the interface for various filters/projections
* various bug fixes

TODO
- [ ] Update adjoint test with new filter function API
- [ ] Update tutorials
- [ ] Allow for different filter radii in each dimension (for nonuniform lengthscale constraints)

@mochen4 to pull into an existing dev branch, make sure you rebase on top of `master`, as described [here](https://github.com/NanoComp/meep/pull/1590#issuecomment-854951694).

### DIFF
--- a/python/adjoint/filters.py
+++ b/python/adjoint/filters.py
@@ -9,6 +9,7 @@ import meep as mp
 from scipy import special
 import jax.numpy.fft as fft
 
+
 def centered(arr, newshape):
     '''Helper function that reformats the padded array of the fft filter operation.
     Borrowed from scipy:
@@ -22,11 +23,12 @@ def centered(arr, newshape):
     myslice = [slice(startind[k], endind[k]) for k in range(len(endind))]
     return arr[tuple(myslice)]
 
+
 def atleast_3d(ary):
     if ary.ndim == 0:
-        result = npj.expand_dims(ary, axis=(0,1,2))#ary.reshape(1, 1, 1)
+        result = npj.expand_dims(ary, axis=(0, 1, 2))  #ary.reshape(1, 1, 1)
     elif ary.ndim == 1:
-        result = npj.expand_dims(ary, axis=(1,2))
+        result = npj.expand_dims(ary, axis=(1, 2))
     elif ary.ndim == 2:
         result = npj.expand_dims(ary, axis=(2))
     else:
@@ -48,7 +50,8 @@ def _centered(arr, newshape):
     myslice = [slice(startind[k], endind[k]) for k in range(len(endind))]
     return arr[tuple(myslice)]
 
-def meep_filter(x,kernel):
+
+def meep_filter(x, kernel):
     '''
     General convolution for 1D, 2D, and 3D design spaces.
 
@@ -78,12 +81,13 @@ def meep_filter(x,kernel):
         The output of the N-D convolution.
     '''
     # store shapes
-    x_shape = x.shape; k_shape = kernel.shape
-    
+    x_shape = x.shape
+    k_shape = kernel.shape
+
     # edge pad
-    npad = *((s,s) for s in x_shape),
-    kernelp = npj.pad(kernel,npad,mode='edge')
-    xp = npj.pad(x,npad,mode='edge')
+    npad = *((s, s) for s in x_shape),
+    kernelp = npj.pad(kernel, npad, mode='edge')
+    xp = npj.pad(x, npad, mode='edge')
 
     # convolve
     '''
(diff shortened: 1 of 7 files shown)


mochen4:
I have been trying to use the new code. I don't know if I missed something, but currently conic_filter(x,radius) gives me nan.

Specifically,
https://github.com/NanoComp/meep/blob/8903b50a2b999c3282070f99cb9e79482228722f/python/adjoint/filters.py#L236-L239 now becomes 
```
    xl = npj.linspace(-x.shape[0] / 2, x.shape[0] / 2, x.shape[0])
    yl = npj.linspace(-x.shape[1] / 2, x.shape[1] / 2, x.shape[1])
    zl = npj.linspace(-x.shape[2] / 2, x.shape[2] / 2, x.shape[2])
    X, Y, Z = npj.meshgrid(xl, yl, zl, sparse=True, indexing='ij')
```
Thus,  ``kernel = np.where(np.abs(X**2 + Y**2 + Z**2) <= radius**2, (1 - np.sqrt(npj.abs(X**2 + Y**2 + Z**2)) / radius), 0)``
is now a zero array. Shouldn't we have ``x.shape[0] / (2*resolution)``or  ``Lx/2`` or instead of ``x.shape[0] / 2`` etc. in the linspace?

smartalecH:
> Shouldn't we have x.shape[0] / (2*resolution)or Lx/2 or instead of x.shape[0] / 2 etc. in the linspace?

No,  the `radius` is in units of pixels, so there is no need to factor in the resolution.

The input array `x` needs to be three-dimensional (in your case) and `radius` needs to be >= 1.

mochen4:
Thanks! So for ``get_conic_radius_from_eta_e(b,eta_e)``,  the minimum length b will be in units of pixels instead of meep unit as before? @smartalecH

smartalecH:
> the minimum length b will be in units of pixels instead of meep unit as before

No, that function remains unchanged. So the user must manually multiply the `radius` in meep units by the corresponding design region resolution and pass the result of that value into the new filtering routines.

The idea behind the new filtering API is to keep things as simple as possible.

mochen4:
Thanks!

python/adjoint/filters.py
```diff
@@ -3,14 +3,15 @@
 """
 
 import numpy as np
-from autograd import numpy as npa
+import jax
+from jax import numpy as npj
```
ianwilliamson, line 7: Just FYI- convention is `import jax.numpy as jnp` rather than `npj`: https://jax.readthedocs.io/en/latest/notebooks/quickstart.html

stevengj:
> So the user must manually multiply the radius in meep units by the corresponding design region resolution and pass the result of that value into the new filtering routines.

Can we avoid that?  As much as possible, I'd like distances to be specified in "real" units.

smartalecH:
> I'd like distances to be specified in "real" units.

I agree, but it comes at the cost of a more complicated API with lots more parameters to pass (and generally less flexibility).

> I guess the problem with filters is that the same design variables (material grid) can be applied to multiple objects, and so effectively have different resolutions in the same simulation.

Exactly. I think it's better for the user to handle things on their side so that any required bookkeeping doesn't restrict any applications.

python/adjoint/filters.py
```diff
@@ -23,282 +24,139 @@ def _centered(arr, newshape):
...
-                            (pad[0][1], pad[1][1]))  # bottom right
-
-    out = npa.concatenate((npa.concatenate(
-        (top_left, top, top_right)), npa.concatenate((left, arr, right)),
-                           npa.concatenate(
-                               (bottom_left, bottom, bottom_right))),
-                          axis=1)
-
-    return out
+def atleast_3d(ary):
```
ianwilliamson, line 27: I think there are already two other versions of a similar function for this (one that I added in utils and one that was already in another module). It might be worth consolidating these into a single function?

In some places it will be used for standard numpy arrays and here you're using it for jax arrays. I assume you'll want to track gradients through this, so I think to support both numpy and jax array the `var.reshape(...)` approach could be used, which also creates a view. This could work with a slight modification to:

https://github.com/NanoComp/meep/blob/4d27231ae97a733b83fc1ffbff1ac4f23dccef1b/python/adjoint/utils.py#L15-L17

mochen4:
It seems that the Near2FarFields now fails. Specifically, the adjoint source amplitudes (``scale``) from https://github.com/NanoComp/meep/blob/74c97520ccf68403cf5ef64824ed70cf78902985/python/adjoint/objective.py#L333-L338 becomes negligible (10^-313) in my test case.
However, in ``optimization_problem.py``, if I import ``jacobian`` from ``autograd`` instead of ``jax``, it would work as usual. The values of the jacobian ``dJ`` are the same, and I tried to convert the jax jacobian to a numpy array with ``jnp.asarray``, but it doesn't make a difference.

smartalecH:
> However, in optimization_problem.py, if I import jacobian from autograd instead of jax, it would work as usual. 

Looks like this is a result of earlier PRs that transitioned the solver to jax.

What I typically do is use autograd for the main adjoint solver, but use jax for the filter functions used here. It's not ideal, but easier than fiddling with jax.

Although more recently I had to abandon jax altogether to get it to work on multi-node systems.

In an ideal world, we would have a package consisting of jax's autodiff function library, without losing autograd's simplicity.

mochen4:
Thanks Alec!

smartalecH:
> Resolve decision to filtfilt (shouldn't technically be needed with our even filters)

I bet this issue is due to the off-by-one bug we've been fixing in various PRs (#1769, #1760).

Also, all our filters are *separable*, which means we can actually just perform a simple 1D filter with the corresponding 1D kernel in each dimension. This not only speeds things up significantly, but should clean up the code quite a bit too (we can keep the full `fftfilt` routine on the off chance we discover a non-separable filter that's useful).

PR #1628, opened 2021-06-23
updated multi-threading branch

### LINKED ISSUE #1627: Hybrid OpenMP/MPI
It would be good to revive the OpenMP work (see #228 and #1234), and in particular to support hybrid parallelism — use shared-memory threads (via OpenMP) *within* a node, and MPI *across* nodes, in a parallel cluster.

One big motivation is to save memory for the case of 3d design grids.  We replicate the material grid (as well as several similar-size data structures for gradients and other vectors internal to NLopt) for every MPI process — this would allow us to have only one copy per *node* of a cluster.

In particular @mochen4 is now doing 3d topology optimization for 3d-printed structure, and we are finding that we are starting to run out of memory on machines with 40 cpus per node, since the 3d material grid (which can be gigabytes in size) gets replicated 40 times.

@smartalecH, did you have a branch that could be rebased here?

stevengj:
WIP to rebase @smartalecH's multithreading branch to fix #1627.


src/meep/vec.hpp
```diff
@@ -186,6 +186,70 @@ component first_field_component(field_type ft);
...
+without having to change any other code
+in the main codebase. We can proceed to do
+a collapse over all three main loops. */
+
+#define CHUNK_OPENMP _Pragma("omp parallel for")
+
+// the most generic use case where the user
+// can specify a custom clause
+#define PLOOP_OVER_IVECS_C(gv, is, ie, idx, clause)			                                                     \
+_Pragma("unroll(1)") \
```
stevengj, line 207: @smartalecH, note that these `#pragma unroll` directives affect the compiler even without openmp … did they actually improve performance or were they just an experiment?

It seems that you actually weren't using them, since in your branch they were written as `_Pragma("unroll(1")` (missing a close paren), which is an error and presumably was ignored by the compiler.

Maybe we should just delete them?

src/meep/vec.hpp
```diff
@@ -186,6 +186,70 @@ component first_field_component(field_type ft);
...
+      for (ptrdiff_t loop_i3 = 0; loop_i3 < loop_n3; loop_i3++)                                    \
+      _Pragma("unroll(1)") \
+        for (ptrdiff_t idx = idx0 + loop_i1*loop_s1 + loop_i2*loop_s2 +                            \
+           loop_i3*loop_s3, dummy_last=0;dummy_last<1;dummy_last++)
+
+// For the main timestepping events, we know
+// we want to do a simple collapse
+#define PLOOP_OVER_IVECS(gv, is, ie, idx)                                                          \
+  /*master_printf("Entered ploop\n");*/ \
+  PLOOP_OVER_IVECS_C(gv, is, ie, idx, "omp parallel for collapse(3)")
```
stevengj, line 233: @smartalecH, the branch by @HomerReid in #565 used a different looping construct instead of `#pragma parallel for collapse` to combine the multiple loops into a single parallel loop — did you ever compare your approach to his?

src/meep/vec.hpp
smartalecH: The idea was to unroll the overhead of the dummy loop (just 1 iteration) that I used to wrap everything (see comment above code). 

I didn't realize that I was missing a paren, so you can probably safely omit it. The compiler hopefully optimizes out loops of 1 iteration anyway...

src/meep/vec.hpp
smartalecH: He manually specified a scheduler and thread count. But you can do that using env variables, which is much more flexible (especially when profiling) that way you don't have to recompile each time. 

I also have a 'collapse' clause because I'm collapsing 2 loops, he just had 1 loop to parallelize. 

Everything else is the same.

src/meep/vec.hpp
stevengj: I just did a quick check in Julia with LLVM (since Julia lets me interactively examine compiler output) and it seems that it can indeed optimize out loops with a small constant number of iterations.   It should be the same in C++, and hopefully also with gcc.

stevengj:
The tests don't currently exercise the openmp code (we need to configure with `--enable-openmp` and set `OMP_NUM_THREADS` to enable this functionality), but at least they are green, so this hopefully isn't harming anything.

It would be good to do a quick performance benchmark *without* openmp on this branch, to make sure that @smartalecH's loop rewriting didn't hurt the serial performance.

stevengj:
We probably want to parallelize the `LOOP_OVER_IVECS` in `dft_chunk::update_dft` as well, since that is often a significant fraction of the runtime.

src/anisotropic_averaging.cpp
```diff
@@ -226,7 +237,8 @@ void structure_chunk::set_chi1inv(component c, material_function &medium,
   double trivial_val[3] = {0, 0, 0};
   trivial_val[idiag] = 1.0;
   ivec shift1(unit_ivec(gv.dim, component_direction(c)) * (ft == E_stuff ? 1 : -1));
-  LOOP_OVER_VOL(gv, c, i) {
+  PLOOP_OVER_IVECS_C(gv, (gv).little_corner() + (gv).iyee_shift(c),                                   \
```
stevengj, line 240: Currently, this loop doesn't seem thread-safe.  It should be changed back to `LOOP_OVER_VOL` until it is fixed.

src/fields.cpp
```diff
@@ -457,14 +457,14 @@ bool fields_chunk::alloc_f(component c) {
           component bc = direction_component(Bx, component_direction(c));
           if (!f[bc][cmp]) {
             f[bc][cmp] = new realnum[gv.ntot()];
-            for (size_t i = 0; i < gv.ntot(); i++)
+            PS1LOOP_OVER_VOL(gv, c, i)
               f[bc][cmp][i] = 0.0;
```
stevengj, line 461: It seems like overkill to thread these loops.

src/structure.cpp
```diff
@@ -613,34 +613,34 @@ void structure_chunk::mix_with(const structure_chunk *n, double f) {
       chi1inv[c][d] = new realnum[gv.ntot()];
       trivial_chi1inv[c][d] = n->trivial_chi1inv[c][d];
       if (component_direction(c) == d) // diagonal components = 1 by default
-        for (size_t i = 0; i < gv.ntot(); i++)
+        PS1LOOP_OVER_VOL(gv, c, i)
           chi1inv[c][d][i] = 1.0;
```
stevengj, line 617: Also overkill…

src/structure.cpp
```diff
@@ -762,7 +762,7 @@ structure_chunk::structure_chunk(const structure_chunk *o) : v(o->v) {
     if (is_mine() && o->chi3[c]) {
       chi3[c] = new realnum[gv.ntot()];
       if (chi3[c] == NULL) abort("Out of memory!\n");
-      for (size_t i = 0; i < gv.ntot(); i++)
+      PS1LOOP_OVER_VOL(gv, c, i)
         chi3[c][i] = o->chi3[c][i];
```
stevengj, line 766: This could just be a `memcpy`

src/fields.cpp
smartalecH: I think the idea was to allocate arrays using the same threads that would loop over them while stepping (to avoid things similar to false sharing). I didn't notice any real speedup from this, though.

tests/test_loop.cpp
```diff
@@ -0,0 +1,80 @@
+/*
```
stevengj, line 1: @smartalecH, should this file be included in `make check`?

smartalecH: No, this was just for my personal debugging. Didn't realize it was added.

oskooi:
This branch should be rebased after #1737 is merged first.

smartalecH:
All of the `LOOP` macros that involve python callbacks do not currently use the multithreading implementation.

There's a nifty swig feature, [`-threads`](https://github.com/swig/swig/issues/927), that allows us to temporarily release the python GIL, call whatever python function we need (in parallel), and then reenable the GIL. Many of the callbacks aren't required to be sequential -- rather they are just evaluating some user-defined function (e.g. a permittivity distribution). So this would be a perfect use case (and consequently, speed up the simulation init step quite a bit).

PR #1635, opened 2021-06-24
refactor sync_chunk_connections to remove barrier during fields::step

### LINKED ISSUE #1632: remove barrier from fields::connect_chunks
We should be able to remove the barrier on this line: https://github.com/NanoComp/meep/blob/8be41f37b6fd8c99aa5fc0f304301c80797809be/src/boundaries.cpp#L124-L127

by carefully putting the synchronization at any point that might set `chunk_connections_valid = false`.    The trick here is that the synchronization must be done at points where we can do collective operations (i.e. that all processors are required to be doing the same thing), which requires a little care e.g. for `fields::add_srcdata`, but I think it can be done without too much trouble.

stevengj:
Fixes #1632.

### DIFF
--- a/src/boundaries.cpp
+++ b/src/boundaries.cpp
@@ -30,6 +30,8 @@ namespace meep {
 void fields::set_boundary(boundary_side b, direction d, boundary_condition cond) {
   if (boundaries[b][d] != cond) {
     boundaries[b][d] = cond;
+    // we don't need to call sync_chunk_connections() since set_boundary()
+    // should always be called on every process
     chunk_connections_valid = false;
   }
 }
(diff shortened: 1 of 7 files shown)


oskooi:
I ran some benchmarks comparing this branch and master for a test problem based on a 3d OLED simulation using a single-machine with Intel Xeon 2.0 GHz processors with 10, 20, and 40 MPI processes. The benchmark involved time stepping the fields for some number of timesteps after the pulsed source had turned off to ensure that the fields arrays contain non-zero values. The timers were reset after the first time step to remove the time for the array initialization, etc. from the timing statistics. The results shown below are for the test case involving 10 processes and the results for 20 and 40 processes are similar.

There are two things to note from these results:
1. first figure: there is almost no change in the time stepping rate and also the total runtime of the simulation.
2. second figure: the timing statistics for this branch (labelled "PR \# 1635") shows that the time for all-to-all communication for all processes are *zero* but the time for the point-to-point communication has increased. This is probably just an error in the timing statistics but the key point is that the total MPI time does not change relative to master.

As a result, there does not seem to be any noticeable performance gains using this branch which is unexpected.

np10_sec_per_timestep_vs_time

barplot_np10

stevengj:
It could be that to see a significant timing impact from a barrier you need to be on a large distributed-memory cluster?

(I thought you guys did some profiling which was why you were complaining about this barrier?)

ahoenselaar:
We do see an improvement in our profiles. With the changes in this PR, cycles spent on connecting chunks effectively disappear.

oskooi (approved):

stevengj:
> We do see an improvement in our profiles. With the changes in this PR, cycles spent on connecting chunks effectively disappear.

Are you sure that those cycles aren't simply shifting to someplace else, as in Ardavan's test?

ahoenselaar:
Given the noise in these benchmarks and the small fraction of <1% of cycles before the PR, I cannot determine that with a reasonable investment of effort.

stevengj:
I'll go ahead and merge this, since it can't hurt to eliminate an `Allreduce` on every timestep.   I'm not convinced that we're actually seeing a net performance benefit (as opposed to an accounting change), however.

PR #1640, opened 2021-06-26
Prevent memory leaks in binary_partition

ahoenselaar:
* Use unique_ptrs to take ownership of the left and right subtrees.
* Introduce the struct `split_plane` to combine split direction and
position. This struct could be used in the future to simplify the
signatures of several volume related functions.
* Adjust SWIG typemaps
* Hide `choose_chunkdivision` from Python.

### DIFF
--- a/src/structure_dump.cpp
+++ b/src/structure_dump.cpp
@@ -18,6 +18,7 @@
 // Dump/load raw structure data to/from an HDF5 file.  Only
 // works if the number of processors/chunks is the same.
 
+#include <cassert>
 #include <stdio.h>
 #include <stdlib.h>
 #include <math.h>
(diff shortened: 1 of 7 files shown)


src/structure_dump.cpp
```diff
@@ -414,46 +414,49 @@ void structure::set_chiP_from_file(h5file *file, const char *dataset, field_type
...
-  split_pos = _split_pos;
-  proc_id = -1;
-  left = NULL;
-  right = NULL;
-}
+bool binary_partition::is_leaf() const { return !left && !right; }
+
+int binary_partition::get_proc_id() const { return proc_id; }
+
+const split_plane &binary_partition::get_plane() const { return plane; }
```
stevengj, line 431: Should these have `is_leaf()` and `!is_leaf()` assertions, respectively?

src/structure_dump.cpp
```diff
@@ -414,46 +414,49 @@ void structure::set_chiP_from_file(h5file *file, const char *dataset, field_type
...
-    result_ids.push_back(bp->proc_id);
+    result_ids.push_back(bp->get_proc_id());
     return;
   }
 
-  int split_point = (size_t)((bp->split_pos - gvol.surroundings().in_direction_min(bp->split_dir)) /
-                             gvol.surroundings().in_direction(bp->split_dir) *
-                             gvol.num_direction(bp->split_dir) + 0.5);
+  const auto &plane = bp->get_plane();
+  int split_point = static_cast<size_t>(
```
stevengj, line 449: The fact that we are casting to `size_t` but assigning to an `int` seems wrong here. (I realize it was also in the original code.)  Probably we should be casting to an `int` since that is the type required by `split_at_fraction`.

src/structure_dump.cpp
ahoenselaar: Done.

src/structure_dump.cpp
ahoenselaar: Updated.

oskooi:
I tested this branch using a 3d OLED simulation with 40 MPI processes which is currently failing with an error for the master branch as reported in #1639. While this branch does remove the error, it does not seem to produce exactly the same results (time-stepping the fields for some number of timesteps and outputting the fields at the cell origin) as the branch in #1639 even though I verified that both branches are producing the same chunk layout (just a visual inspection using the `Simulation.visualize_chunks()` routine). I would have expected the results from the two branches to be nearly identical.

oskooi:
For reference, the benchmarking test I have been using can be found [here](https://gist.github.com/oskooi/47b59852bb5965ecf98950605fccd45a). Running this script with 40 MPI processes using the default parameters produces these results:

**PR 1639**
```
field:, 1200, (-9.093591067621496e-05+0.0002530577177735094j)
```

**PR 1640**
```
field:, 1200, (-5.1244023822798436e-05+0.00047882959494642855j)
```

oskooi:
Whoops, I realized that the benchmarking test was using a source with *random* amplitude. After switching to a constant amplitude (and updating the [script](https://gist.github.com/oskooi/47b59852bb5965ecf98950605fccd45a)) and rerunning the test, the results from the two branches are identical.

Perhaps we can go now ahead and close #1639 since it is replaced by this PR?

ahoenselaar:
Glad to hear that! I really could not explain to myself how this PR could have affected the result barring other sources of randomness in Meep's behavior.

ahoenselaar:
#1605

PR #1644, opened 2021-06-28
Fix memory leaks in `src_vol`

ahoenselaar:
The old implementation's weak point was the `add_to` method, which
sometimes transferred memory ownership correctly and sometimes did not.

* Rewrite `src_vol` using STL containers.
* Replace the linked list of source volumes in `fields_chunk` with a
vector. Adjust code that iterates over source volumes accordingly.
* Exclude `fields_chunk` from SWIG wrappers.

### DIFF
--- a/src/meep_internals.hpp
+++ b/src/meep_internals.hpp
@@ -15,6 +15,7 @@
 %  Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
 */
 
+#include <vector>
 #include <string.h>
 #include "meep.hpp"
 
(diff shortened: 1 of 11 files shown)


ahoenselaar:
#1605

src/meep_internals.hpp
```diff
@@ -53,35 +54,36 @@ inline int rmin_bulk(int m) {
...
-
-  src_vol *add_to(src_vol *others);
-  src_vol *next;
+  // Constructs a new source volume. Takes ownership of `ind` and `amps`.
+  // Requirement: ind.size() == amps.size()
+  src_vol(component cc, src_time *st, std::vector<ptrdiff_t> &&ind,
+          std::vector<std::complex<double> > &&amps);
+
+  // Compares `a` and `b` for equality while ignoring the actual amplitudes.
+  static bool equal_except_for_amplitude(const src_vol &a, const src_vol &b);
```
stevengj, line 67: We could probably call this something like `coincide` or `combinable` since what it is really checking is that they coincide in space and field component, and it is used to check whether they can be combined into a single `src_vol` by adding amplitudes.

src/sources.cpp
```diff
@@ -299,26 +274,22 @@ static void src_vol_chunkloop(fields_chunk *fc, int ichunk, component c, ivec is
 
   if (idx_vol > npts) meep::abort("add_volume_source: computed wrong npts (%zd vs. %zd)", npts, idx_vol);
 
-  src_vol *tmp = new src_vol(c, data->src, idx_vol, index_array, amps_array);
   field_type ft = is_magnetic(c) ? B_stuff : D_stuff;
-  fc->sources[ft] = tmp->add_to(fc->sources[ft]);
+  fc->add_source(ft, src_vol(c, data->src, std::move(index_array), std::move(amps_array)));
```
stevengj, line 278: Isn't `std::move` already called in the `src_vol` constructor?

ahoenselaar: It is needed in both places to turn lvalues into xvalues.
If `std::move` were removed here, compilation would fail because a `vector<...>` cannot be converted to the rvalue reference `vector<...> &&`.
In the `src_vol` constructor, `std::move` is used to select the correct overload of the the `vector` constructor. If removed, compilation would succeed but the vector data would be copied instead of moved.

See the "Notes" section here: https://en.cppreference.com/w/cpp/utility/move

src/meep_internals.hpp
ahoenselaar: Updated.

PR #1646, opened 2021-06-28
Fix memory leaks in C++ tests

ahoenselaar:
Along with #1644, this shuld get us into a place where the AddressSanitizer workflow succeeds.


ahoenselaar:
#1605

PR #1653, opened 2021-07-02
preliminaries for tiled timestepping

### LINKED ISSUE #1633: support cache-tiled execution of step_curl and step_update_EDHB
Currently, `step_curl` and `step_update_EDHB` use `LOOP_OVER_VOL_OWNED`.   It would be nice to change these to use `LOOP_OVER_IVECS`, where the caller passes an `(is,ie)` pair of ivecs to loop over.  This allows the *caller* to break up the the loop over the whole domain into loops over subdomains.

The main advantage of this is that it allows you to tile the loops for greater locality.  A particularly nice way to do this is a **cache-oblivious** recursive algorithm: recursively divide the chunk into halves, with a base case large enough to amortize the function-call overhead but much smaller than any cache size.

Furthermore, you could put the loop over components *inside* the loop over tiles.  That is, for a single tile, you could update (Hx,Hy,Hz) (in three separate calls to `step_*`), to maximize the cache locality in accessing the `E` array (for the curls), while maintaining the simplicity of updating one component at a time (avoid a combinatorial explosion in the number of loop variations in the `step_*` functions).

The first step, changing to `LOOP_OVER_IVECS`, should be do-able quickly and easily.

stevengj:
Closes #1633 — adds `is,ie` arguments to `step_curl` etcetera so that we can potentially break these up into cache-efficient "tiles" in a future PR.

### DIFF
--- a/src/meep_internals.hpp
+++ b/src/meep_internals.hpp
@@ -95,18 +95,19 @@ symmetry r_to_minus_r_symmetry(int m);
 
 void step_curl(realnum *f, component c, const realnum *g1, const realnum *g2,
                ptrdiff_t s1, ptrdiff_t s2, // strides for g1/g2 shift
-               const grid_volume &gv, realnum dtdx, direction dsig, const realnum *sig,
+               const grid_volume &gv, const ivec is, const ivec ie,
+               realnum dtdx, direction dsig, const realnum *sig,
                const realnum *kap, const realnum *siginv, realnum *fu, direction dsigu,
                const realnum *sigu, const realnum *kapu, const realnum *siginvu, realnum dt,
                const realnum *cnd, const realnum *cndinv, realnum *fcnd);
 
-void step_update_EDHB(realnum *f, component fc, const grid_volume &gv, const realnum *g,
+void step_update_EDHB(realnum *f, component fc, const grid_volume &gv, const ivec is, const ivec ie, const realnum *g,
                       const realnum *g1, const realnum *g2, const realnum *u, const realnum *u1,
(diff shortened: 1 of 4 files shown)


oskooi (approved):
LGTM.

stevengj:
Some comments:

1. To see a benefit from tiling a *single* component, you need a *planar cross-section* of the chunk to exceed a cache size, at least the L1 cache for a given 3-component vector field.  i.e. `(diameter*resolution)^2 * 3 * 4bytes/float > cache`, roughly. 
2. Since you are only optimizing `step_curl`, you can strip out any extraneous computations in the benchmark.  No DFTs, no Lorentzian dispersive materials (just vacuum is fine). Just run for 1 timestep to initialize things and then benchmark a few timesteps after that — it shouldn't matter for speed whether the fields are nonzero.
3. A bigger benefit should occur if you swap the loops so that the loop over tiles goes *outside* the loop over components.   In this case you should see a benefit if the *whole chunk* exceeds the cache: roughly `(diameter*resolution)^3 * 3 * 4bytes/float > cache`.  (I would test swapping the loops *after* fixing the current benchmarks though, just so that we see the effects of swapping the loops separately.)

Also, I wouldn't divide the chunk into cubes.   To maximize spatial locality (consecutive memory axis), since we use row major order I would preferentially divide along x, then y, then z.  i.e. something like:
```
if (ntot < threshold)
    don't divide
else if (x size > 1)
    divide along x
else if (y size > 1)
    divide along y
else
    divide along z
```

Issue #1648, opened 2021-06-29
sim.chunk_layout is not accessible in Python if not specified at class instantiation time


bencbartlett:
If a `meep.Simulation` object is created without specifying `chunk_layout`, then the [default logic](https://github.com/NanoComp/meep/blob/master/src/structure.cpp#L109) for choosing chunk divisions is invoked when `sim.init_sim()` is called. However, the `chunk_layout` attribute on the Python `Simulation` object is not updated to reflect this change and will still read as `None`.

When converting from a Python `meep.BinaryPartition` object, [`py_bp_to_bp()`](https://github.com/NanoComp/meep/blob/master/python/typemap_utils.cpp#L1045) is called to convert the Python object into a C++ object. It would be useful to have the inverse operation present with a `bp_to_py_bp()` function, perhaps as a class construction method to live on `meep.BinaryPartition`. Then in `Simulation._init_structure()` at [line 1709](https://github.com/NanoComp/meep/blob/master/python/simulation.py#L1708), this function could be called to update the `Simulation.chunk_layout` attribute:

```python
if self.chunk_layout is None:
  self.chunk_layout = mp.BinaryPartition.from_cpp_object(...)
```

stevengj:
Probably the code would go in `python/typemap_utils.cpp`.  Once you have the function `PyObject *binarypartition_cpp_to_py(...)` or whatever it is called, SWIG should handle it automatically.

You may also need some refactoring in the `structure` constructor to save the binary partition it creates.

PR #1654, opened 2021-07-02
Store the binary partition in `structure`

ahoenselaar:
* Store the binary partition - whether passed in externally or generated
via `choose_chunkdivision` in `structure` and expose it via a getter.
* Remove the unused default constructor of `structure`.
* Remove a redundant and unused copy constructor of `structure`.
* Clean up the copy constructor of `structure`.
* Make `binary_partition` copyable.

Prepares for making the BinaryPartition visible in Python (#1648).

### DIFF
--- a/src/meep.hpp
+++ b/src/meep.hpp
@@ -726,18 +726,16 @@ public:
   int num_effort_volumes;
 
   ~structure();
-  structure();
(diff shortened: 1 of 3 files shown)


stevengj:
LGTM, thanks.

PR #1655, opened 2021-07-05
loop tiling for step_curl using cache-oblivious recursive algorithm

oskooi:
Initial (and incomplete) attempt to extend the revamped API for `STEP_CURL` and `STEP_UPDATE_EDHB` introduced in #1653 to support loop tiling using a cache-oblivious recursive algorithm. The focus is to first verify the correctness of the overall approach.

This PR introduces a new member `std::vector<grid_volume> gvs` to the `fields_chunks` class which stores an array of the subdomains or "tiles" of each chunk generated using a recursive-bisection algorithm implemented in `split_into_tiles`. The nice thing is is that we can reuse the existing `grid_volume` class functions `find_best_split` and `split_at_fraction` for the cell-partitioning feature inside `split_into_tiles`. The tiles are generated *once* as part of the `fields_chunk` constructor although in general they would need to be defined every time `chunk_connections_valid = false`.  A loop over the tiles is then defined around `STEP_CURL` inside `fields_chunk::step_db` (a similar approach would apply to `STEP_UPDATE_EDHB` inside `fields_chunk::update_eh`). For now, only Cartesian coordinates are supported and we can deal with Cylindrical coordinates separately after this is working.

There is a hard-coded value `base_nowned_min` inside `split_into_tiles` which determines the size of the smallest domain in the recursion. In my preliminary benchmarking results, adjusting this value from `100` and `1000` changed the average time-stepping rate by ~20% for a test case involving an OLED simulation with 4 MPI processes on a single machine. I will post more results shortly. Unfortunately, this PR is causing some of the unit tests to fail which probably needs to be resolved first.

### DIFF
--- a/src/fields.cpp
+++ b/src/fields.cpp
@@ -202,6 +202,25 @@ fields_chunk::~fields_chunk() {
   if (new_s && new_s->refcount-- <= 1) delete new_s; // delete if not shared
 }
 
+static void split_into_tiles(grid_volume gvol, std::vector<grid_volume> &result) {
+  size_t base_nowned_min = 1000;
+  if (gvol.nowned_min() < base_nowned_min) {
+    result.push_back(gvol);
+    return;
+  }
+
+  int best_split_point;
+  direction best_split_direction;
+  double left_effort_fraction;
+  gvol.find_best_split(2, false /* fragment_cost */, best_split_point, best_split_direction,
+                       left_effort_fraction);
+  grid_volume left_gvol = gvol.split_at_fraction(false, best_split_point, best_split_direction);
+  split_into_tiles(left_gvol, result);
+  grid_volume right_gvol = gvol.split_at_fraction(true, best_split_point, best_split_direction);
+  split_into_tiles(right_gvol, result);
+  return;
+}
+
 fields_chunk::fields_chunk(structure_chunk *the_s, const char *od, double m, double beta,
                            bool zero_fields_near_cylorigin, int chunkidx)
     : gv(the_s->gv), v(the_s->v), m(m), zero_fields_near_cylorigin(zero_fields_near_cylorigin),
@@ -216,6 +235,7 @@ fields_chunk::fields_chunk(structure_chunk *the_s, const char *od, double m, dou
   Courant = s->Courant;
   dt = s->dt;
   dft_chunks = NULL;
+  split_into_tiles(gv, gvs);
   FOR_FIELD_TYPES(ft) {
     polarization_state *cur = NULL;
     pol[ft] = NULL;
@@ -277,6 +297,7 @@ fields_chunk::fields_chunk(const fields_chunk &thef, int chunkidx) : gv(thef.gv)
   Courant = thef.Courant;
   dt = thef.dt;
   dft_chunks = NULL;
+  gvs = thef.gvs;
   FOR_FIELD_TYPES(ft) {
     polarization_state *cur = NULL;
     for (polarization_state *ocur = thef.pol[ft]; ocur; ocur = ocur->next) {
--- a/src/fields.cpp
+++ b/src/fields.cpp
@@ -203,7 +203,7 @@ fields_chunk::~fields_chunk() {
 }
 
 static void split_into_tiles(grid_volume gvol, std::vector<grid_volume> &result) {
-  const size_t base_nowned_min = 1000;
+  const size_t base_nowned_min = 1000000;
   if (gvol.nowned_min() < base_nowned_min) {
     result.push_back(gvol);
     return;
@@ -221,6 +221,25 @@ static void split_into_tiles(grid_volume gvol, std::vector<grid_volume> &result)
   return;
 }
 
+// First check that the tile volumes gvs do not intersect and that they add
+// up to the chunk's total grid_volume gv
+static void check_tiles(grid_volume gv, std::vector<grid_volume> gvs) {
+  grid_volume vol_intersection;
+  for (size_t i = 0; i < gvs.size(); i++)
+    for (size_t j = i + 1; j < gvs.size(); j++)
+      if (gvs[i].intersect_with(gvs[j], &vol_intersection))
+        meep::abort("gvs[%zu] intersects with gvs[%zu]\n", i, j);
+  size_t sum = 0;
+  for (size_t i = 0; i < gvs.size(); i++) {
+    size_t grid_points = 1;
+    LOOP_OVER_DIRECTIONS(gvs[i].dim, d) { grid_points *= gvs[i].num_direction(d); }
+    sum += grid_points;
+  }
+  size_t v_grid_points = 1;
+  LOOP_OVER_DIRECTIONS(gv.dim, d) { v_grid_points *= gv.num_direction(d); }
+  if (sum != v_grid_points) meep::abort("v_grid_points = %zu, sum(chunks) = %zu\n", v_grid_points, sum);
+}
+
 fields_chunk::fields_chunk(structure_chunk *the_s, const char *od, double m, double beta,
                            bool zero_fields_near_cylorigin, int chunkidx)
     : gv(the_s->gv), v(the_s->v), m(m), zero_fields_near_cylorigin(zero_fields_near_cylorigin),
@@ -236,6 +255,7 @@ fields_chunk::fields_chunk(structure_chunk *the_s, const char *od, double m, dou
   dt = s->dt;
   dft_chunks = NULL;
   split_into_tiles(gv, gvs);
+  check_tiles(gv, gvs);
   FOR_FIELD_TYPES(ft) {
     polarization_state *cur = NULL;
     pol[ft] = NULL;
--- a/src/fields.cpp
+++ b/src/fields.cpp
@@ -203,7 +203,7 @@ fields_chunk::~fields_chunk() {
 }
 
 static void split_into_tiles(grid_volume gvol, std::vector<grid_volume> *result) {
-  const size_t base_nowned_min = 20000;
+  const size_t base_nowned_min = 12000;
   if (gvol.nowned_min() < base_nowned_min) {
     result->push_back(gvol);
     return;
--- a/src/fields.cpp
+++ b/src/fields.cpp
@@ -211,9 +211,7 @@ static void split_into_tiles(grid_volume gvol, std::vector<grid_volume> *result)
 
   int best_split_point;
   direction best_split_direction;
-  double left_effort_fraction;
-  gvol.find_best_split(2, false /* fragment_cost */, best_split_point, best_split_direction,
-                       left_effort_fraction);
+  gvol.tile_split(best_split_point, best_split_direction);
   grid_volume left_gvol = gvol.split_at_fraction(false, best_split_point, best_split_direction);
   split_into_tiles(left_gvol, result);
   grid_volume right_gvol = gvol.split_at_fraction(true, best_split_point, best_split_direction);
--- a/src/fields.cpp
+++ b/src/fields.cpp
@@ -29,7 +29,8 @@ using namespace std;
 
 namespace meep {
 
-fields::fields(structure *s, double m, double beta, bool zero_fields_near_cylorigin)
+fields::fields(structure *s, double m, double beta, bool zero_fields_near_cylorigin,
+               int loop_tile_base)
     : S(s->S), gv(s->gv), user_volume(s->user_volume), v(s->v), m(m), beta(beta),
       working_on(&times_spent) {
   shared_chunks = s->shared_chunks;
@@ -57,7 +58,8 @@ fields::fields(structure *s, double m, double beta, bool zero_fields_near_cylori
   typedef fields_chunk *fields_chunk_ptr;
   chunks = new fields_chunk_ptr[num_chunks];
   for (int i = 0; i < num_chunks; i++)
-    chunks[i] = new fields_chunk(s->chunks[i], outdir, m, beta, zero_fields_near_cylorigin, i);
+    chunks[i] = new fields_chunk(s->chunks[i], outdir, m, beta,
+                                 zero_fields_near_cylorigin, i, loop_tile_base);
   FOR_FIELD_TYPES(ft) {
     typedef realnum *realnum_ptr;
     comm_blocks[ft] = new realnum_ptr[num_chunks * num_chunks];
@@ -202,9 +204,9 @@ fields_chunk::~fields_chunk() {
   if (new_s && new_s->refcount-- <= 1) delete new_s; // delete if not shared
 }
 
-static void split_into_tiles(grid_volume gvol, std::vector<grid_volume> *result) {
-  const size_t base_nowned_min = 12000;
-  if (gvol.nowned_min() < base_nowned_min) {
+static void split_into_tiles(grid_volume gvol, std::vector<grid_volume> *result,
+                             const size_t loop_tile_base) {
+  if (gvol.nowned_min() < loop_tile_base) {
     result->push_back(gvol);
     return;
   }
@@ -213,9 +215,9 @@ static void split_into_tiles(grid_volume gvol, std::vector<grid_volume> *result)
   direction best_split_direction;
   gvol.tile_split(best_split_point, best_split_direction);
   grid_volume left_gvol = gvol.split_at_fraction(false, best_split_point, best_split_direction);
-  split_into_tiles(left_gvol, result);
+  split_into_tiles(left_gvol, result, loop_tile_base);
   grid_volume right_gvol = gvol.split_at_fraction(true, best_split_point, best_split_direction);
-  split_into_tiles(right_gvol, result);
+  split_into_tiles(right_gvol, result, loop_tile_base);
   return;
 }
 
@@ -235,7 +237,7 @@ static void check_tiles(grid_volume gv, const std::vector<grid_volume> &gvs) {
 }
 
 fields_chunk::fields_chunk(structure_chunk *the_s, const char *od, double m, double beta,
-                           bool zero_fields_near_cylorigin, int chunkidx)
+                           bool zero_fields_near_cylorigin, int chunkidx, int loop_tile_base)
     : gv(the_s->gv), v(the_s->v), m(m), zero_fields_near_cylorigin(zero_fields_near_cylorigin),
       beta(beta) {
   s = the_s;
@@ -248,8 +250,12 @@ fields_chunk::fields_chunk(structure_chunk *the_s, const char *od, double m, dou
   Courant = s->Courant;
   dt = s->dt;
   dft_chunks = NULL;
-  split_into_tiles(gv, &gvs);
-  check_tiles(gv, gvs);
+  if (loop_tile_base > 0) {
+    split_into_tiles(gv, &gvs, loop_tile_base);
+    check_tiles(gv, gvs);
+  } else {
+    gvs.push_back(gv);
+  }
   FOR_FIELD_TYPES(ft) {
     polarization_state *cur = NULL;
     pol[ft] = NULL;
--- a/src/fields.cpp
+++ b/src/fields.cpp
@@ -230,14 +230,10 @@ static void check_tiles(grid_volume gv, const std::vector<grid_volume> &gvs) {
       if (gvs[i].intersect_with(gvs[j], &vol_intersection))
         meep::abort("gvs[%zu] intersects with gvs[%zu]\n", i, j);
   size_t sum = 0;
-  for (const auto& sub_gv : gvs) {
-    size_t grid_points = 1;
-    LOOP_OVER_DIRECTIONS(sub_gv.dim, d) { grid_points *= sub_gv.num_direction(d); }
-    sum += grid_points;
-  }
+  for (const auto& sub_gv : gvs) { sum += sub_gv.nowned_min(); }
   size_t v_grid_points = 1;
   LOOP_OVER_DIRECTIONS(gv.dim, d) { v_grid_points *= gv.num_direction(d); }
-  if (sum != v_grid_points) meep::abort("v_grid_points = %zu, sum(chunks) = %zu\n", v_grid_points, sum);
+  if (sum != v_grid_points) meep::abort("v_grid_points = %zu, sum(tiles) = %zu\n", v_grid_points, sum);
 }
 
 fields_chunk::fields_chunk(structure_chunk *the_s, const char *od, double m, double beta,
--- a/src/fields.cpp
+++ b/src/fields.cpp
@@ -203,7 +203,7 @@ fields_chunk::~fields_chunk() {
 }
 
 static void split_into_tiles(grid_volume gvol, std::vector<grid_volume> &result) {
-  size_t base_nowned_min = 1000;
+  const size_t base_nowned_min = 1000;
   if (gvol.nowned_min() < base_nowned_min) {
     result.push_back(gvol);
     return;
--- a/src/fields.cpp
+++ b/src/fields.cpp
@@ -202,10 +202,10 @@ fields_chunk::~fields_chunk() {
   if (new_s && new_s->refcount-- <= 1) delete new_s; // delete if not shared
 }
 
-static void split_into_tiles(grid_volume gvol, std::vector<grid_volume> &result) {
-  const size_t base_nowned_min = 1000000;
+static void split_into_tiles(grid_volume gvol, std::vector<grid_volume> *result) {
+  const size_t base_nowned_min = 20000;
   if (gvol.nowned_min() < base_nowned_min) {
-    result.push_back(gvol);
+    result->push_back(gvol);
     return;
   }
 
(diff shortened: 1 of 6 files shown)


stevengj:
Should I merge #1653 first?  That PR should be harmless in any case.

src/step_db.cpp
```diff
@@ -44,40 +44,43 @@ bool fields_chunk::step_db(field_type ft) {
...
+        realnum *f_m = have_m ? f[c_m][cmp] : NULL;
+        realnum *the_f = f[cc][cmp];
 
-      if (ft == D_stuff) { // strides are opposite sign for H curl
-        stride_p = -stride_p;
-        stride_m = -stride_m;
-      }
+        if (dsig != NO_DIRECTION && s->conductivity[cc][d_c] && !f_cond[cc][cmp]) {
+          f_cond[cc][cmp] = new realnum[gv_sub.ntot()];
+          memset(f_cond[cc][cmp], 0, sizeof(realnum) * gv_sub.ntot());
```
stevengj, line 70: This isn't right.  The conductivity etcetera should be allocated once for the whole chunk.

oskooi: That's not something that was added in this PR. The only thing that change for those lines was the indentation.

src/step_db.cpp
```diff
@@ -115,12 +115,14 @@ bool fields_chunk::step_db(field_type ft) {
...
 
-      STEP_CURL(the_f, cc, f_p, f_m, stride_p, stride_m,
-                gv, gv.little_owned_corner0(cc), gv.big_corner(),
-                Courant, dsig, s->sig[dsig],
-                s->kap[dsig], s->siginv[dsig], f_u[cc][cmp], dsigu, s->sig[dsigu], s->kap[dsigu],
-                s->siginv[dsigu], dt, s->conductivity[cc][d_c], s->condinv[cc][d_c],
-                f_cond[cc][cmp]);
+      for (size_t i = 0; i < gvs.size(); ++i) {
+        STEP_CURL(the_f, cc, f_p, f_m, stride_p, stride_m,
+                  gvs[i], gvs[i].little_owned_corner0(cc), gvs[i].big_corner(),
```
stevengj, line 120: The first argument on this line should be the original grid volume, `gv`, *not* `gvs[i]` — `STEP_CURL` needs to know "where" you are relative to the whole chunk volume in order to correctly index into the sigma arrays etcetera.

src/fields.cpp
```diff
@@ -213,6 +213,44 @@ fields_chunk::~fields_chunk() {
...
+  grid_volume left_gvol = gvol.split_at_fraction(false, best_split_point, best_split_direction);
+  split_into_tiles(left_gvol, result);
+  grid_volume right_gvol = gvol.split_at_fraction(true, best_split_point, best_split_direction);
+  split_into_tiles(right_gvol, result);
+  return;
+}
+
+// First check that the tile volumes gvs do not intersect and that they add
+// up to the chunk's total grid_volume gv
+static void check_tiles(grid_volume gv, std::vector<grid_volume> gvs) {
```
stevengj, line 237: As another check, the `little_corner()` of each grid volume should consist purely of *even* numbers.  (This should be enforced by `split_at_fraction`.)

src/step_db.cpp
```diff
@@ -115,12 +115,14 @@ bool fields_chunk::step_db(field_type ft) {
           default: meep::abort("bug - non-cylindrical field component in Dcyl");
         }
 
-      STEP_CURL(the_f, cc, f_p, f_m, stride_p, stride_m,
-                gv, gv.little_owned_corner0(cc), gv.big_corner(),
-                Courant, dsig, s->sig[dsig],
-                s->kap[dsig], s->siginv[dsig], f_u[cc][cmp], dsigu, s->sig[dsigu], s->kap[dsigu],
-                s->siginv[dsigu], dt, s->conductivity[cc][d_c], s->condinv[cc][d_c],
-                f_cond[cc][cmp]);
+      for (size_t i = 0; i < gvs.size(); ++i) {
```
stevengj, line 118: Ultimately this loop should go *outside* the loop over components, in order to get the full cache benefit.

stevengj:
See lectures 14 and 15 in [these OCW lectures](https://ocw.mit.edu/courses/electrical-engineering-and-computer-science/6-172-performance-engineering-of-software-systems-fall-2018/).   This is about accessing the memory in the right order (improving "temporal locality"), and once you do that the cache hardware should do the rest.

(The base case of the recursion has nothing to do with cache — it is "recursion coarsening" to amortize the function-call overhead.   You want to shrink the base case as much as possible until you start seeing performance degrade, which will eventually happen due to function-call overhead.)

src/fields.cpp
```diff
@@ -213,6 +213,44 @@ fields_chunk::~fields_chunk() {
   if (new_s && new_s->refcount-- <= 1) delete new_s; // delete if not shared
 }
 
+static void split_into_tiles(grid_volume gvol, std::vector<grid_volume> &result) {
```
ahoenselaar, line 216: As a convention, it would be nice to use pointers for out arguments. Otherwise, it is much harder to reason about the code at the call site while reading it.

src/fields.cpp
ahoenselaar: Pass by const ref

src/fields.cpp
```diff
@@ -213,6 +213,44 @@ fields_chunk::~fields_chunk() {
...
+// First check that the tile volumes gvs do not intersect and that they add
+// up to the chunk's total grid_volume gv
+static void check_tiles(grid_volume gv, std::vector<grid_volume> gvs) {
+  grid_volume vol_intersection;
+  for (size_t i = 0; i < gvs.size(); i++)
+    for (size_t j = i + 1; j < gvs.size(); j++)
+      if (gvs[i].intersect_with(gvs[j], &vol_intersection))
+        meep::abort("gvs[%zu] intersects with gvs[%zu]\n", i, j);
+  size_t sum = 0;
+  for (size_t i = 0; i < gvs.size(); i++) {
```
ahoenselaar, line 244: for (const auto& sub_gv : gvs)

src/step_db.cpp
ahoenselaar: for (const auto& sub_gv : gvs)

src/fields.cpp
oskooi: Fixed.

src/fields.cpp
oskooi: Fixed.

src/step_db.cpp
oskooi: Fixed.

src/fields.cpp
oskooi: Fixed.

src/step_db.cpp
oskooi: Fixed.

oskooi:
Even with the fix inside `fields_chunk::step_db` when calling `STEP_CURL` related to the first argument, some of the tests are still failing. The strange thing is that changing the size of the smallest domain via the `base_nowned_min` parameter of `split_into_tiles` (i.e., from `1000` to `20000`) is causing *different* tests to fail. The tiling seems to span the entire chunk `grid_volume` (this is verified by the function `check_tiles`). This should mean that all the grid points are being updated as part of the call to `fields::step_db` but somehow this does not seem to be the case due to the failing tests.

src/fields.cpp
```diff
@@ -213,6 +213,44 @@ fields_chunk::~fields_chunk() {
...
+  grid_volume vol_intersection;
+  for (size_t i = 0; i < gvs.size(); i++)
+    for (size_t j = i + 1; j < gvs.size(); j++)
+      if (gvs[i].intersect_with(gvs[j], &vol_intersection))
+        meep::abort("gvs[%zu] intersects with gvs[%zu]\n", i, j);
+  size_t sum = 0;
+  for (const auto& sub_gv : gvs) {
+    size_t grid_points = 1;
+    LOOP_OVER_DIRECTIONS(sub_gv.dim, d) { grid_points *= sub_gv.num_direction(d); }
+    sum += grid_points;
```
stevengj, line 247: You can just use `sum += sub_gv.ntot();`

oskooi: Unfortunately, we have to manually compute the number of grid points in `sub_gv` and can't use `sub_gv.ntot()` because to use the latter we would first need to call its private function [`grid_volume::update_ntot()`](https://github.com/NanoComp/meep/blob/master/src/meep/vec.hpp#L1038) which is not possible (otherwise the number of grid points stored in `the_ntot` and returned by `ntot()` is *not* up to date and causes `check_tiles` to abort on line 251).

oskooi:
For some reason, the CI tests seem to all be passing even though they are still failing on my local build (as I reported previously).

src/fields.cpp
oskooi: Fixed.

oskooi:
Based on some experiments, this PR seems to only work if the minimum tile volume `base_nowned_min` (i.e., the number of grid points in the smallest subdomain) is above a value of ~12000 which is the default value in this PR. For values below 12000, two of the Python unit tests (`test_cyl_ellipsoid.py`, `test_simulation.py`) are failing due to the fields blowing up. The tiles do indeed span the entire chunk which means all the grid points are being updated. It is not clear why there is a qualitative change in behavior below this minimum tile size.

src/fields.cpp
```diff
@@ -213,6 +213,40 @@ fields_chunk::~fields_chunk() {
...
+  int best_split_point;
+  direction best_split_direction;
+  double left_effort_fraction;
+  gvol.find_best_split(2, false /* fragment_cost */, best_split_point, best_split_direction,
+                       left_effort_fraction);
+  grid_volume left_gvol = gvol.split_at_fraction(false, best_split_point, best_split_direction);
+  split_into_tiles(left_gvol, result);
+  grid_volume right_gvol = gvol.split_at_fraction(true, best_split_point, best_split_direction);
+  split_into_tiles(right_gvol, result);
+  return;
```
stevengj, line 232: This is tricky, because `split_at_fraction` is designed to split a domain into *overlapping* chunks, of which the "owned" volumes are non-overlapping.  Also, the whole `find_best_split` machinery is overkill here because we just want to split into half along the longest dimension.

It might be clearer for tiling if we split the whole volume gv into non-overlapping tile (in halves, recurssively), where tile is described by an `<is,ie>` pair of corners, with `is` always an *even* grid point.

Then, when you are doing `step_curl` for a component `c` and a tile `<is,ie>`, you will pass it the something like `max(gv.little_owned_corner0(c), is + iyee_shift(c))` and `ie` for the corners.  This way it will cover the owned volume for `c` intersected with `<is,ie>`

oskooi:
I have tried just splitting the chunk in two along its longest direction but it still doesn't seem to work (i.e., some of the unit tests are failing).

**master**
```cpp
      STEP_CURL(the_f, cc, f_p, f_m, stride_p, stride_m,
                gv, gv.little_owned_corner0(cc), gv.big_corner(),
                Courant, dsig, s->sig[dsig],
                s->kap[dsig], s->siginv[dsig], f_u[cc][cmp], dsigu, s->sig[dsigu], s->kap[dsigu],
                s->siginv[dsigu], dt, s->conductivity[cc][d_c], s->condinv[cc][d_c],
                f_cond[cc][cmp]);
```

**proposed change**
```cpp
      direction longest_axis = NO_DIRECTION;
      int num_in_longest_axis = 0;
	if (gv.num_direction(d) > num_in_longest_axis) {
          longest_axis = d;
          num_in_longest_axis = gv.num_direction(d);
        }
      }
      int split_pt = gv.num_direction(longest_axis) / 2;

      grid_volume side_low;
      side_low.dim = gv.dim;
      side_low.a = gv.a;
        side_low.set_num_direction(d, gv.num_direction(d));
      }
      side_low.set_origin(gv.little_owned_corner0(cc));
      side_low.set_num_direction(longest_axis, split_pt);

      grid_volume side_high;
      side_high.dim = gv.dim;
      side_high.a = gv.a;
        side_high.set_num_direction(d, gv.num_direction(d));
      }
      side_high.set_origin(gv.little_owned_corner0(cc));
      side_high.shift_origin(longest_axis, split_pt * 2);
      side_high.set_num_direction(longest_axis, side_high.num_direction(longest_axis) - split_pt);

      ivec is = gv.little_owned_corner0(cc);
      ivec ie = min(side_low.big_corner() + gv.iyee_shift(cc), side_high.little_owned_corner0(cc));
      STEP_CURL(the_f, cc, f_p, f_m, stride_p, stride_m,
                gv, is, ie,
                Courant, dsig, s->sig[dsig],
                s->kap[dsig], s->siginv[dsig], f_u[cc][cmp], dsigu, s->sig[dsigu], s->kap[dsigu],
                s->siginv[dsigu], dt, s->conductivity[cc][d_c], s->condinv[cc][d_c],
                f_cond[cc][cmp]);

      is = max(side_low.big_corner() + gv.iyee_shift(cc), side_high.little_owned_corner0(cc));
      ie = gv.big_corner();
      STEP_CURL(the_f, cc, f_p, f_m, stride_p, stride_m,
                gv, is, ie,
                Courant, dsig, s->sig[dsig],
                s->kap[dsig], s->siginv[dsig], f_u[cc][cmp], dsigu, s->sig[dsigu], s->kap[dsigu],
                s->siginv[dsigu], dt, s->conductivity[cc][d_c], s->condinv[cc][d_c],
                f_cond[cc][cmp]);
```

stevengj:
We really should have a copy constructor for `grid_volume`…

stevengj:
Would be good to check that the `side_low`'s `ie` is `<` the `side_high`'s `is` along `longest_axis` (they should differ by 1 or 2, I think, depending on the component `cc`).

oskooi:
Outputting the `is` and `ie` for `side_low` and `side_high` shows that they are consistent but still the tests are failing.

```cpp
      master_printf("split direction: %s, %d\n",direction_name(longest_axis),split_pt);
      ivec is = gv.little_owned_corner0(cc);
      ivec ie = side_low.big_corner();
      master_printf("side_low (%s):, is=(%d,%d,%d), ie=(%d,%d,%d)\n",
                    component_name(cc),
                    is.x(),is.y(),is.z(),ie.x(),ie.y(),ie.z());
      STEP_CURL(the_f, cc, f_p, f_m, stride_p, stride_m,
                gv, is, ie,
                Courant, dsig, s->sig[dsig],
                s->kap[dsig], s->siginv[dsig], f_u[cc][cmp], dsigu, s->sig[dsigu], s->kap[dsigu],
                s->siginv[dsigu], dt, s->conductivity[cc][d_c], s->condinv[cc][d_c],
                f_cond[cc][cmp]);

      is = side_high.little_owned_corner0(cc);
      ie = gv.big_corner();
      master_printf("side_high (%s):, is=(%d,%d,%d), ie=(%d,%d,%d)\n",
                    component_name(cc),
                    is.x(),is.y(),is.z(),ie.x(),ie.y(),ie.z());
      STEP_CURL(the_f, cc, f_p, f_m, stride_p, stride_m,
                gv, is, ie,
                Courant, dsig, s->sig[dsig],
                s->kap[dsig], s->siginv[dsig], f_u[cc][cmp], dsigu, s->sig[dsigu], s->kap[dsigu],
                s->siginv[dsigu], dt, s->conductivity[cc][d_c], s->condinv[cc][d_c],
                f_cond[cc][cmp]);
```

**output for a 3d test problem**
```
side_low (bx):, is=(-108,-109,-121), ie=(2,111,-1)
side_high (bx):, is=(4,-108,-120), ie=(110,110,-2)
side_low (by):, is=(-109,-108,-121), ie=(1,112,-1)
side_high (by):, is=(2,-106,-120), ie=(110,110,-2)
side_low (bz):, is=(-109,-109,-120), ie=(1,111,0)
side_high (bz):, is=(2,-108,-118), ie=(110,110,-2)
side_low (dx):, is=(-109,-108,-120), ie=(1,112,0)
side_high (dx):, is=(2,-106,-118), ie=(110,110,-2)
side_low (dy):, is=(-108,-109,-120), ie=(2,111,0)
side_high (dy):, is=(4,-108,-118), ie=(110,110,-2)
side_low (dz):, is=(-108,-108,-121), ie=(2,112,-1)
side_high (dz):, is=(4,-106,-120), ie=(110,110,-2)
```

oskooi:
After rebasing from the master branch following #1703, this feature seems to be working now: the chunks are divided into tiles and produce the *same* results as no tiles. Some experimenting is required to determine a reasonable value for the base case of the recursive bisection on [this line](https://github.com/NanoComp/meep/pull/1655/files#diff-166204e3ad737768924818e8b08d96b092fe246a7b2045e4f30510664a61b310R206) which has been chosen somewhat arbitrarily to test that the feature is working but not necessarily to provide any kind of performance enhancement.

oskooi:
Here are preliminary benchmarking results for a single-threaded 3d OLED simulation consisting of two chunks (PML and non-PML regions). The plot shows the time spent timestepping vs. number of tiles (total for the two chunks). It seems that the no-tiled case is *always* faster than any of the tiled cases. Meep is compiled with GCC and single-precision fields.

oled_tiling

stevengj:
Some comments:

1. To see a benefit from tiling a *single* component, you need a *planar cross-section* of the chunk to exceed a cache size, at least the L1 cache for a given 3-component vector field.  i.e. `(diameter*resolution)^2 * 3 * 4bytes/float > cache`, roughly. 
2. Since you are only optimizing `step_curl`, you can strip out any extraneous computations in the benchmark.  No DFTs, no Lorentzian dispersive materials (just vacuum is fine). Just run for 1 timestep to initialize things and then benchmark a few timesteps after that — it shouldn't matter for speed whether the fields are nonzero.
3. A bigger benefit should occur if you swap the loops so that the loop over tiles goes *outside* the loop over components.   In this case you should see a benefit if the *whole chunk* exceeds the cache: roughly `(diameter*resolution)^3 * 3 * 4bytes/float > cache`.  (I would test swapping the loops *after* fixing the current benchmarks though, just so that we see the effects of swapping the loops separately.)

Also, I wouldn't divide the chunk into cubes.   To maximize spatial locality (consecutive memory axis), since we use row major order I would preferentially divide along x, then y, then z.  i.e. something like:
```
if (ntot < threshold)
    don't divide
else if (x size > 1)
    divide along x
else if (y size > 1)
    divide along y
else
    divide along z
```

src/vec.cpp
```diff
@@ -978,6 +978,24 @@ static double cost_diff(int desired_chunks, std::complex<double> costs) {
...
+  } else if (nx() > 1) {
+    best_split_point = nx() / 2;
+    best_split_direction = X;
+  } else if (ny() > 1) {
+    best_split_point = ny() / 2;
+    best_split_direction = Y;
+  } else {
+    best_split_point = nz() / 2;
+    best_split_direction = Z;
+  }
```
stevengj, line 996: It would be worth also trying to split along the longest axis.

oskooi:
I reran the benchmarking test on my local i7-7700K machine which has  much smaller cache than the machine used [previously](https://github.com/NanoComp/meep/pull/1655#issuecomment-887532642). I simplified the [benchmarking test](https://gist.github.com/oskooi/f44a8e26113470032e574957be12a210) by removing extraneous computations following the suggestions above. I ran this benchmark five separate times for each of six tile configurations. I timed just the portion spent on `step_db(B_stuff)` and `step_db(B_stuff)` of `fields::step()`. The results demonstrate that the min/mean time decreases until reaching a minimum before increasing monotonically as would be expected. However, the relative improvement in the runtime using the optimal tiling configuration is insignificant: just ~2-3% of the no-tiling configuration. These results were generated for the case in which the loop over tiles is outside the loop over components.

To verify whether it is possible to demonstrate a more significant improvement, we can try (1) increasing the problem size and (2) explore other tiling configurations beyond recursive bisection along *x*, *y*,  and *z* (i.e., splitting along the longest axis).

```
$ lscpu
Architecture:         x86_64
Thread(s) per core:   1
Model name:           Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz
L1d cache:            32K
L1i cache:            32K
L2 cache:             256K
L3 cache:             8192K
```

oled_tiles_benchmark

oskooi:
Update: I forgot to include the results for the *no-tiling* case in the previous figure. (The previous figure shows the smallest tiling configuration with 3520 tiles.) With these updated results, the relative improvement is actually **18.2%** (minimum runtime of 1.51403 s for no tiling vs.  1.23869 s for 7040 tiles) which is obviously significant. It seems therefore that this cache-oblivious loop tiling approach is indeed working to improve the runtime performance. 

oled_tiles_benchmark

stevengj:
Remember @matteo-frigo's suggestion — if you run on only a single core of your machine, you might be getting an unusually large memory bandwidth and hence not see benefits from cache optimization.  Try running on *all* of your cores at once — simply do
```py
meep.divide_parallel_processes(meep.count_processors())
```
to run the identical (serial) simulation on all processors, and use `mpirun` to launch on all cores.

oskooi:
I tried running the same benchmark on all four single-threaded cores simultaneously using `divide_parallel_processes` and the results are actually not that different than running on just a single core. Processor 1 and 3 have slightly larger (minimum) times for the no-tiling case and also show less relative improvement at the optimal tiling configuration whereas processor 2 and 4 show results similar to the single-core case from above:

```
processor 1 relative improvement: 10.41%
processor 2 relative improvement: 18.23%
processor 3 relative improvement:  1.53%
processor 4 relative improvement: 16.86%
```

The error bars of the averaged data are noticeably larger than the single-core case but nevertheless in general it still seems that there is a benefit from using loop tiling.

oled_tiles_benchmark_procs

oskooi:
**Update:** there were some inaccuracies in the results posted earlier which should therefore be ignored.

I reran the benchmarking test for 1, 2, 3, and 4 processors. The results show that the relative improvement in the time for `step_db` *increases* with the number of processors. (The noisiness of the data also seems to increase with the number of processors.) I think this validates @matteo-frigo's suggestion that the benefits of cache optimization increase with reduced memory bandwidth.

**1 processor**

```
processor 1: relative improvement of 17.26% using 7040 tiles
```

oled_tiles_benchmark_procs1

**2 Processors**
```
processor 1: relative improvement of 43.43% using 56320 tiles
processor 2: relative improvement of 41.87% using 14080 tiles
```

oled_tiles_benchmark_procs2

**3 processors**
```
processor 1: relative improvement of 62.51% using 14080 tiles
processor 2: relative improvement of 39.81% using 3520 tiles
processor 3: relative improvement of 50.10% using 3520 tiles
```

oled_tiles_benchmark_procs3

**4 processors**
```
processor 1: relative improvement of 68.03% using 3520 tiles
processor 2: relative improvement of 63.72% using 14080 tiles
processor 3: relative improvement of 62.57% using 14080 tiles
processor 4: relative improvement of 53.73% using 3520 tiles
```
oled_tiles_benchmark_procs4b

oskooi:
Added an undocumented parameter `loop_tile_base` for specifying the base case of the recursive-bisection algorithm in `split_into_tiles` to the `Simulation` constructor. The default is `0` which indicates no tiling. This API enables the user to experiment with different values at runtime rather than having to every time rebuild from source after updating a hard-coded value.

matteo-frigo:
@oskooi You are still using the i7-7700, right?  And this is a 2D or (2+epsilon)D problem, right?

oskooi:
> You are still using the i7-7700, right? And this is a 2D or (2+epsilon)D problem, right?

Yes, these results are generated using i7-7700 and a 3d simulation. The cell size is 11 x 11 x 2.2 with a resolution of 80 pixels per unit length. The total size of the **E** (*E*<sub>x</sub>,*E*<sub>y</sub>,*E*<sub>z</sub>) and **H** (*H*<sub>x</sub>,*H*<sub>y</sub>,*H*<sub>z</sub>) fields used for updating **B** and **D**, respectively, during the step-curl operation (shown below) is ~1.64 Gb = [(3 fields x 4 bytes/field) x (11 x 11 x 2.2) x 80<sup>3</sup>] / 1e9.

loop_tiling_step_curl

stevengj:
@matteo-frigo, you mentioned that at one point (prior to your [later stencil work](https://dl.acm.org/doi/10.1145/1088149.1088197)), you tried a simple recursive tiling strategy similar to the one we are using here, and got some speedup.   Do you have a reference for that?

matteo-frigo:
@stevengj No reference, sorry.

I actually tried as part of the ICS05 paper that you link, not prior to it.  In the ICS05 paper we glossed over the base case because ICS05 is mostly a theory paper and performing divide and conquer in the DeltaT=1 case only buys you a constant factor independent of the cache size, so it's kind of an "uninteresting" "detail".  For the same reason, in Figure 6, we didn't complicate the code by attempting to cut the largest dimension first.  (The cache oblivious matrix multiplication in FOCS99 cuts the largest dimension first, so that's kind of the "obvious" thing to try.)

If you are thinking of writing up the idea and giving proper credit, I don't really need or care about the credit, and I doubt that Volker cares either.  I am happy that the idea works.  Analyzing the exact conditions under which cutting the largest dimension is a good idea sounds like an interesting undergraduate/master project (how much are you saving depending on the shape of the stencil, how many rows/planes have to fit in cache for this to work, etc.).  We never went that far.

PR #1656, opened 2021-07-06
Streamline communications

ahoenselaar:
* Calculate the sequence of sends and receives once after `connect_the_chunks`
* Initiate asynchronous receives as early as possible
* Initiate asynchronous sends immediately after copying data into the communication buffers
* Prioritize chunks based on the data transfer size
* Introduce an RAII-based communications manager class that waits for requests to finish upon destruction
* Store nonzero communication sizes in a hash map instead of an array of size `#chunks x #chunks`

Performance increase by about 15% in smaller sims with light communications (small transfer sizes), e.g. a 2D sim with 12 processes on a single physical node.
Not as noticeable for larger 3D sims (e.g. 24 processes across 2 physical nodes), likely because the increased communication efficiency is overshadowed by the effects of uneven chunk balancing.

Future improvements:
* Start processing received messages immediately by using `MPI_Testsome` instead of `MPI_Waitall`. This should overlap network communications more with the copying of data into the field array.
* Combine the sequential boundary updates of {PH, H}, {PE, E} into a single [Set up Rx, copy data, trigger Tx, process Rx data] cycle
* Prioritize send operations by target process instead of source chunk. This requires more careful management of the subranges in the comms buffer array.

### DIFF
--- a/src/meep.hpp
+++ b/src/meep.hpp
@@ -17,7 +17,10 @@
 #ifndef MEEP_H
 #define MEEP_H
 
+#include <functional>
+#include <limits>
 #include <memory>
+#include <unordered_map>
 #include <stdio.h>
 #include <stddef.h>
 #include <math.h>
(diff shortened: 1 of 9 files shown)


ahoenselaar:
Are there any review comments?

src/meep.hpp
```diff
@@ -711,6 +713,80 @@ boundary_region pml(double thickness, double Rasymptotic = 1e-15, double mean_st
...
+    // Unroll hash combination to promote the generatiion of efficient code.
+    std::size_t ret = ft_hasher(key.ft);
+    ret ^= connect_phase_hasher(key.phase) + kHashAddConst + (ret << 6) + (ret >> 2);
+    ret ^= int_hasher(key.pair.first) + kHashAddConst + (ret << 6) + (ret >> 2);
+    ret ^= int_hasher(key.pair.second) + kHashAddConst + (ret << 6) + (ret >> 2);
+    return ret;
+  }
+
+private:
+  static constexpr size_t kHashAddConst = 0x9e3779b9;
```
stevengj, line 750: Should this be a 64-bit value (i.e. not have 32 zero bits) on 64-bit machines?

I guess it doesn't matter for us since our hash maps shouldn't be getting anywhere near 2^31 entries.

src/meep.hpp
```diff
@@ -17,7 +17,9 @@
 #ifndef MEEP_H
 #define MEEP_H
 
+#include <functional>
```
stevengj, line 20: Out of curiosity, what are we using from this header?

src/mympi.cpp
```diff
@@ -67,9 +67,59 @@ using namespace std;
...
 namespace meep {
 
+namespace {
+
+#ifdef HAVE_MPI
+MPI_Comm mycomm = MPI_COMM_WORLD;
+#endif
+
+// comms_manager implementation that uses MPI.
+class mpi_comms_manager : public comms_manager {
```
stevengj, line 77: I guess the thinking is that you may implement other comms managers later for non-MPI backends?

src/step.cpp
```diff
@@ -15,6 +15,9 @@
 %  Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
 */
 
+#include <array>
+#include <iostream>
```
stevengj, line 19: Are we using iostream for anything here?

src/step.cpp
```diff
@@ -133,65 +136,119 @@ void fields_chunk::phase_material(int phasein_time) {
...
-     and must be kept consistent with the code in boundaries.cpp.
-     In particular, we require that boundaries.cpp set up the connections
-     array so that all of the connections for process i come before all
-     of the connections for process i' for i < i'  */
+    const auto &sequence = comms_sequence_for_field[ft];
+    for (const comms_operation &op : sequence.receive_ops) {
+      if (chunks[op.other_chunk_idx]->is_mine()) {
+        continue;
+      }
+      manager->receive_real_async(comm_blocks[ft][op.pair_idx], static_cast<int>(op.transfer_size),
```
stevengj, line 148: Maybe the `receive_real_async ` function should do the `static_cast<int>`, i.e. pass a `size_t` here, so that this code doesn't have to know that MPI is restricted to `int` sizes?

src/mympi.cpp
ahoenselaar: And it keeps all MPI code and awarenewss of MPI out of meep.hpp

src/meep.hpp
ahoenselaar: std::hash

src/step.cpp
ahoenselaar: Removed.

src/step.cpp
ahoenselaar: Updated.

stevengj (approved):

PR #1664, opened 2021-07-09
Fix the BinaryPartition initializer

ahoenselaar:
According to the docstring `left` and `right` have to be `BinaryPartition` objects.


stevengj:
Maybe better to just fix the docstring? The current initializer seems a lot more compact.

PR #1673, opened 2021-07-14
Propagate chunk_layout to Python meep.Simulation object

### LINKED ISSUE #1648: sim.chunk_layout is not accessible in Python if not specified at class instantiation time

bencbartlett:
If a `meep.Simulation` object is created without specifying `chunk_layout`, then the default logic for choosing chunk partitionings is invoked when `sim.init_sim()` is called. However, the `chunk_layout` attribute on the Python `meep.Simulation` object previously was not updated and would still read as `None`. This pull request adds a `bp_to_py_bp()` function to convert the C++ object into a Python `meep.BinaryPartition` object and updates the `sim.chunk_layout` attribute (and for consistency, `sim.num_chunks`), in `sim._init_structure()`. (Fixes #1648)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@
 *.pyc
 _build
 .vscode
+.idea
 
 # autotools stuff
 Makefile
--- a/src/structure.cpp
+++ b/src/structure.cpp
@@ -15,7 +15,6 @@
 %  Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
 */
 
-#include <iostream>
 #include <memory>
 #include <stdio.h>
 #include <stdlib.h>
@@ -961,7 +960,6 @@ double structure_chunk::max_eps() const {
 }
 
 bool structure::equal_layout(const structure &s) const {
-  std::cout << "equal_layout num_chunks " << num_chunks << " s.num_chunks " << s.num_chunks << "\n";
   if (a != s.a || num_chunks != s.num_chunks || v != s.v || S != s.S) return false;
   for (int i = 0; i < num_chunks; ++i)
     if (chunks[i]->a != s.chunks[i]->a || chunks[i]->v != s.chunks[i]->v) return false;
--- a/src/structure.cpp
+++ b/src/structure.cpp
@@ -15,6 +15,7 @@
 %  Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
 */
 
+#include <iostream>
 #include <memory>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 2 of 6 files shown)


oskooi:
I think the failing test `test_chunk_layout.py` is due to a bug in the way the process IDs of the leaves are *always* assigned a value of -1:

https://github.com/NanoComp/meep/blob/89c93498ead04290b42acbb35b2c09c9813e04eb/src/structure.cpp#L69

The reason this bug is a little hard to track down is because if the user does not pass in a `chunk_layout` parameter to the `Simulation` object and Meep therefore falls back on using its internal cell-partition scheme, the `proc` of each chunk is set correctly but the corresponding `proc_id` of the leaves from the binary tree is never updated to this same value:

https://github.com/NanoComp/meep/blob/89c93498ead04290b42acbb35b2c09c9813e04eb/src/structure.cpp#L133-L141

What should happen is the `proc_id` of each leaf/chunk should be set to `proc` *before* returning the `binary_partition` object `bp` back to Python.

.gitignore
```diff
@@ -14,6 +14,7 @@
 *.pyc
 _build
 .vscode
+.idea
```
oskooi, line 17: Is this necessary?

src/structure.cpp
```diff
@@ -958,6 +959,7 @@ double structure_chunk::max_eps() const {
 }
 
 bool structure::equal_layout(const structure &s) const {
+  std::cout << "equal_layout num_chunks " << num_chunks << " s.num_chunks " << s.num_chunks << "\n";
```
oskooi, line 962: Is this something that was left over from debugging and therefore should be removed from this PR? 

In general, for printing to standard output in Meep, use the built-in function `master_printf("equal_layout %d s.num_chunks %d\n",num_chunks,s.num_chunks);` which will only print from the master process in a parallel simulation.

src/structure.cpp
```diff
@@ -15,6 +15,7 @@
 %  Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
 */
 
+#include <iostream>
```
oskooi, line 18: This should probably be removed (see comment below).

src/structure.cpp
bencbartlett: Removed this line

src/structure.cpp
bencbartlett: Done

.gitignore
bencbartlett: I used CLion to make these changes locally. JetBrains IDEs (CLion, IntelliJ) are pretty popular so it's probably worth including in the .gitignore file.

oskooi (approved):

PR #1674, opened 2021-07-15
Fields dump load

kkg4theweb:
First cut of a PR to add support for dumping and (re)loading the 'fields' state.

### DIFF
--- a/python/tests/test_simulation.py
+++ b/python/tests/test_simulation.py
@@ -23,11 +23,13 @@ class TestSimulation(unittest.TestCase):
 
     @classmethod
     def setUpClass(cls):
-        cls.temp_dir = mp.make_output_directory()
+        # cls.temp_dir = mp.make_output_directory()
+        cls.temp_dir = "/tmp/test_simulation"
 
     @classmethod
     def tearDownClass(cls):
-        mp.delete_directory(cls.temp_dir)
+        # mp.delete_directory(cls.temp_dir)
+        pass
 
     def test_interpolate_numbers(self):
 
@@ -203,6 +205,7 @@ class TestSimulation(unittest.TestCase):
         fn = os.path.join(self.temp_dir, 'test_in_point-ez-000200.00.h5')
         self.assertTrue(os.path.exists(fn))
 
+    @unittest.skip('References unavailable data files')
     def test_epsilon_input_file(self):
         sim = self.init_simple_simulation()
         eps_input_fname = 'cyl-ellipsoid-eps-ref.h5'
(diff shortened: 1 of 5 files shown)


python/tests/test_simulation.py
```diff
@@ -336,6 +340,9 @@ def get_ref_field_point(sim):
         if chunk_sim:
             chunk_layout = sim1
 
+        fields_dump_fn = os.path.join(self.temp_dir, 'test_load_dump_fields.h5')
```
ahoenselaar, line 343: Revert all diff in this file except for this section.
There should probably an assertion as well.

src/fields_dump.cpp
```diff
@@ -0,0 +1,183 @@
...
+#include <string.h>
+
+#include <cassert>
+
+#include "meep.hpp"
+#include "meep_internals.hpp"
+
+namespace meep {
+
+#define DUMP_FIELD(field_name)                                              \
```
ahoenselaar, line 33: Can we avoid macros in all new code?

src/fields_dump.cpp
```diff
@@ -0,0 +1,183 @@
...
+  {                                                                         \
+    /*                                                                      \
+     * make/save a num_chunks x NUM_FIELD_COMPONENTS x 2 array counting     \
+     * the number of entries in the 'field_name' array for each chunk.      \
+     */                                                                     \
+    size_t *num_f_ = new size_t[num_chunks * NUM_FIELD_COMPONENTS * 2];     \
```
ahoenselaar, line 39: Can we please avoid C-style manual allocation in new code?

python/tests/test_simulation.py
kkg4theweb: Done.

src/fields_dump.cpp
kkg4theweb: done

src/fields_dump.cpp
kkg4theweb: done.

stevengj:
It's fine to write everything to separate files (one per process, say).  You'll want to do a couple of things:

* Modify `save_dft_hdf5` to take an additional flag `single_parallel_file`.  If that's `false`, simply skip the parallel reductions in `dft_chunks_Ntotal`.   
* In your own code, follow the same two-pass structure: first compute the total size of the data on your process, then create the dataset, then write the chunks into that dataset.    Again, if ``single_parallel_file==true` we can do an optional parallel reduction after the first pass, and otherwise the code is the same.

Basically, to use HDF5 parallel I/O with a single file you have to create the file and create the dataset collectively, and then each process can call `write_chunk` as many times as it wants to write its own data into the file (as long as it writes into non-overlapping portions of the file).  That's why the `save_dft_hdf5` and `structure::dump` functions are organized the way they are.

stevengj:
Possible Python API: to re-start a simulation, run the *same* simulation script, creating the `Simulation` object as usual with all of the geometry and sources etcetera, but before you do `sim.run`, first run a `sim.load(config)` function (that loads the fields, DFT, etcetera).

`config` would be some kind of argument that says what to load.  Maybe a JSON file, or just a `dict`, whose files give the various filenames?

stevengj:
Closed by #1738?

PR #1681, opened 2021-07-16
Assign ranks to leaves of binary partition in depth-first order

oskooi:
Fixes the bug described in [#1673 (comment)](https://github.com/NanoComp/meep/pull/1673#issuecomment-881215159).

With this change, the unit test added in #1673 for `test_chunk_layout.py` passes and thus #1673 needs only to be rebased after this is merged.


stevengj:
Given this change it seems like you should also change https://github.com/NanoComp/meep/blob/89c93498ead04290b42acbb35b2c09c9813e04eb/src/structure.cpp#L134
to simply:
```cxx
const int proc = ids[i] % count_processors();
```
both for simplicity and to ensure that the chunk process id is set consistently with the value in the binary tree.

oskooi:
Good suggestion. I have made the change.

PR #1682, opened 2021-07-17
Fine-grained timing data for the time stepping phase

ahoenselaar:


python/tests/test_simulation.py
```diff
@@ -13,6 +13,9 @@
     unicode = str
 
 
+TIMESINKS_FOR_TEST = (mp.BoundarySteppingE, mp.MPBTime, mp.Connecting)
```
oskooi, line 16: Note that because this test does not involve eigenmode sources or mode decomposition, the time for `mp.MPBTime` will be zero across all processes. This could be something to verify as part of the test along with *non-zero* times for the other two categories `mp.BoundarySteppingE` and `mp.Connecting`.

python/simulation.py
```diff
@@ -3678,10 +3678,29 @@ def print_times(self):
...
+        * meep.FieldUpdateE ("updating E field")
+        * meep.BoundarySteppingB ("boundary stepping B")
+        * meep.BoundarySteppingWH ("boundary stepping WH")
+        * meep.BoundarySteppingPH ("boundary stepping PH")
+        * meep.BoundarySteppingH ("boundary stepping H")
+        * meep.BoundarySteppingD ("boundary stepping D")
+        * meep.BoundarySteppingWE ("boundary stepping WE")
+        * meep.BoundarySteppingPE ("boundary stepping PE")
+        * meep.BoundarySteppingE ("boundary stepping E")
+        * meep.Other ("everything else")
```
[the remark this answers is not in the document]
ahoenselaar: Those are hard to measure separately, as idling effectively happens within MPI communication methods.
Waiting for E field data from other nodes will show up in the `mp.BoundarySteppingE` counter.

python/tests/test_simulation.py
```diff
@@ -579,6 +582,34 @@ def print_field(sim):
...
+        df = 0.1
+
+        sources = [mp.Source(src=mp.GaussianSource(fcen, fwidth=df), component=mp.Ez,
+                             center=mp.Vector3())]
+        geometry = [mp.Block(center=mp.Vector3(), size=mp.Vector3(mp.inf, 3, mp.inf),
+                             material=mp.Medium(epsilon=12))]
+
+        sim = mp.Simulation(resolution=resolution, cell_size=cell_size, boundary_layers=pml,
+                            sources=sources, geometry=geometry, geometry_center=center)
+        sim.run(until=50)
```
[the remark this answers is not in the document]
ahoenselaar: Updated test to inspect the actual timing data.

python/tests/test_simulation.py
ahoenselaar: Updated.

python/meep.i
```diff
@@ -1433,6 +1433,27 @@ void _get_gradient(PyObject *grad, PyObject *fields_a, PyObject *fields_f, PyObj
     $1 = NULL;
 }
 
+
+// typemaps for timing data
+
+%typemap(out) std::unordered_map<meep::time_sink, std::vector<double> > {
```
stevengj, line 1439: Apparently using an `enum` as a hash key is a C++14 or later feature (https://stackoverflow.com/questions/18837857/cant-use-enum-class-as-unordered-map-key)?

Similar to #1687, you may need to pass `std::hash<int>` to `unordered_map` here to make sure this runs in Apple clang.  (Would be nice to have CI for this — if only to make sure things compile, even if we don't run the tests.)   See e.g. Argonne-National-Laboratory/neml#67 for a similar workaround in another project.

src/meep.hpp
```diff
@@ -1558,7 +1558,37 @@ enum time_sink {
...
+  BoundarySteppingH,
+  BoundarySteppingD,
+  BoundarySteppingWE,
+  BoundarySteppingPE,
+  BoundarySteppingE
+};
+
+// RAII-based profiling timer that accumulates wall time from creation until it
+// is destroyed or the `exit` method is invoked. Not thread-safe.
+class timing_scope {
```
stevengj, line 1578: Should this be in `meep_internals.hpp`?

stevengj: I guess you'd have to make `with_timing_scope` a `friend` function then.

src/time.cpp
```diff
@@ -17,118 +17,167 @@
...
+#include <iterator>
+#include <map>
+
 using namespace std;
 
 namespace meep {
 
+namespace {
+
+constexpr size_t kMeepTimingStackSize = 10;
```
stevengj, line 30: I'm not sure it's worth bounding the size the timing stack.  If the timing stack is growing out of control, we probably have bigger problems.

src/time.cpp
```diff
@@ -17,118 +17,167 @@
...
-  double now = wall_time();
-  if (last_wall_time >= 0) times_spent[working_on] += now - last_wall_time;
-  last_wall_time = now;
-  working_on = was_working_on[0];
-  for (int i = 0; i < MEEP_TIMING_STACK_SZ - 1; ++i)
-    was_working_on[i] = was_working_on[i + 1];
-  was_working_on[MEEP_TIMING_STACK_SZ - 1] = Other;
+  if (!was_working_on.empty()) { was_working_on.pop_back(); }
+  working_on =
+      with_timing_scope(!was_working_on.empty() ? was_working_on.back() : Other);
```
stevengj, line 106: Don't you need to call `pop_back` after you call `was_working_on.back()`, rather than before?

stevengj: Oh, I see that `was_working_on` actually includes the current `sink` too, which is why you need to pop first.

src/time.cpp
```diff
@@ -17,118 +17,167 @@
...
+
 using namespace std;
 
 namespace meep {
 
+namespace {
+
+constexpr size_t kMeepTimingStackSize = 10;
+
+const std::map<time_sink, const char *> kDescriptionByTimeSink{
```
stevengj, line 32: We don't actually use the Hungarian naming style anywhere else in Meep, so the `k` here (which I guess is part of the Google Coding Standards) is a bit incongruous.  `DescriptionByTimeSink` should be fine.

src/time.cpp
```diff
@@ -137,27 +186,38 @@ void fields::output_times(const char *fname) {
...
+      master_fprintf(tf, "%s%g", sep, alltimes[i * n + j]);
+      sep = ", ";
+    }
+    master_fprintf(tf, "\n");
   }
   master_fclose(tf);
-  delete[] alltimes;
+}
+
+std::unordered_map<time_sink, std::vector<double> > fields::get_timing_data() const {
```
stevengj, line 210: see above comment about enums as `unordered_map` keys

python/meep.i
ahoenselaar: This is not an enum class but a standard enum.

src/time.cpp
ahoenselaar: Agreed. Just preserving existing behavior.

src/time.cpp
ahoenselaar: Not applicable.

src/time.cpp
ahoenselaar: Updated.

python/meep.i
stevengj: Nevertheless, Apple clang is giving me an error:
```
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1/type_traits:1848:38: error: implicit instantiation of undefined template 'std::__1::hash<meep::time_sink>'
    : public integral_constant<bool, __is_empty(_Tp)> {};
                                     ^
```

src/time.cpp
stevengj: The point is that if the stack is growing out of control then we have a bug.   It would be better to have an assertion failure in that case than silently discarding things.

The old behavior was because we were too lazy to do resizing for non-`std::vector` arrays, I think.

src/time.cpp
stevengj: Apple clang is giving me an error:
```
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1/unordered_map:427:35: error: no member named 'value' in 'std::__1::is_empty<std::__1::hash<meep::time_sink>>'
          bool = is_empty<_Hash>::value && !__libcpp_is_final<_Hash>::value>
                 ~~~~~~~~~~~~~~~~~^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1/unordered_map:862:13: note: in instantiation of default argument for '__unordered_map_hasher<meep::time_sink, std::__1::__hash_value_type<meep::time_sink, std::__1::vector<double>>, std::__1::hash<meep::time_sink>>' required here
    typedef __unordered_map_hasher<key_type, __value_type, hasher>   __hasher;
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
time.cpp:210:61: note: in instantiation of template class 'std::__1::unordered_map<meep::time_sink, std::__1::vector<double>>' requested here
std::unordered_map<time_sink, std::vector<double> > fields::get_timing_data() const {
                                                            ^
In file included from time.cpp:18:
In file included from ./meep.hpp:23:
```

python/meep.i
ahoenselaar: That is really unexpected and annoying. Updated the code with `std::hash<int>`.

src/time.cpp
ahoenselaar: Updated to assert.

src/time.cpp
ahoenselaar: Fixed.

stevengj:
LGTM.  I've verified that (after rebasing on `master`) it compiles with Apple clang.

Issue #1677, opened 2021-07-15
CI for --enable-single


stevengj:
It would be good to enable CI for `--enable-single`.  This seems to require two things:

* Expose the precision to Python.  This could be as simple as exporting a 3-line `is_single_precision()` function.
* Modify the tests as needed in single precision (e.g. increasing the tolerance for comparisons).

oskooi:
Closed by #1688.

PR #1688, opened 2021-07-20
Expose floating-point precision in Python

oskooi:
Introduces a new function `meep.is_single_precision()` in Python which exposes the floating-point precision of the underlying Meep module. This function will be used by subsequent PRs to enable CI for `--enable-single` as described in #1677.


stevengj:
LGTM.  Probably should have documentation too.

oskooi:
Documentation has been added.

Issue #1575, opened 2021-05-19
better vector comparison test for tests


stevengj:
If you have two vectors `x` and `y`, the right way to compare them to some tolerance ε is `‖x-y‖ ≤ ε * max(‖x‖, ‖y‖)` where `‖⋯‖` is some norm, e.g. the ∞ norm.   It would be good to have a function for this that we can use in the tests.

Something like
```py
def assert_vec_close(x, y, tol):
    np.testing.assert_allclose(x, y, rtol=0, atol=tol*max(np.linalg.norm(x, ord=np.inf), np.linalg.norm(x, ord=np.inf)))
```

PR #1698, opened 2021-07-22
Enable single-precision floating point unit tests for CI

oskooi:
Enables single-precision floating point unit tests as part of the CI only for Python 3.9 with MPI. I have tested that all tests (C++ and Python) are passing for the *entire* CI matrix: [serial, MPI] x [Python 3.6, Python 3.9]. The changes in this PR involve mostly reducing the relative tolerance when checking against hard-coded values for a handful of the unit tests. In some of these modified tests involving a comparison of two NumPy arrays, the comparator function `numpy.testing.assert_allclose` was replaced with `assertVectorsClose` due to #1575. (We should probably replace all uses of `numpy.testing.assert_allclose` but that can be done in a separate PR.)

Three of the unit tests are skipped for single precision: 
1. `test_multilevel_atom.py`: the change in values was larger than just a few decimal digits.
2. `test_eigfreq.py`: the CW solver fails to converge.
3. `test_epsilon_warning` of `test_simulation.py`: the dispersive material Molybdenum (Mo) from the materials library is causing the fields to blow up.

### DIFF
--- a/python/tests/test_3rd_harm_1d.py
+++ b/python/tests/test_3rd_harm_1d.py
@@ -2,10 +2,9 @@ from __future__ import division
 
 import unittest
 import meep as mp
-import numpy as np
+from utils import VectorComparisonMixin
 
-
-class Test3rdHarm1d(unittest.TestCase):
+class Test3rdHarm1d(VectorComparisonMixin, unittest.TestCase):
 
     def setUp(self):
         self.sz = 100
@@ -53,7 +52,8 @@ class Test3rdHarm1d(unittest.TestCase):
 
         harmonics = [self.k, self.amp, mp.get_fluxes(self.trans1)[0], mp.get_fluxes(self.trans3)[0]]
 
-        np.testing.assert_allclose(expected_harmonics, harmonics)
+        tol = 3e-5 if mp.is_single_precision() else 1e-7
+        self.assertVectorsClose(expected_harmonics, harmonics, epsilon=tol)
 
 
 if __name__ == '__main__':
--- a/python/tests/test_adjoint_jax.py
+++ b/python/tests/test_adjoint_jax.py
@@ -16,7 +16,7 @@ jax.config.update('jax_enable_x64', True)
 _FD_STEP = 1e-4
 
 # The tolerance for the adjoint and finite difference gradient comparison
-_TOL = 2e-2
+_TOL = 0.1 if mp.is_single_precision() else 0.02
 
 mp.verbosity(0)
 
--- a/python/tests/test_multilevel_atom.py
+++ b/python/tests/test_multilevel_atom.py
@@ -1,13 +1,11 @@
-from __future__ import division
-
+import meep as mp
 import math
 import unittest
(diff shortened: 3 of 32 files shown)


python/tests/test_adjoint_jax.py
```diff
@@ -16,7 +16,7 @@
 _FD_STEP = 1e-4
 
 # The tolerance for the adjoint and finite difference gradient comparison
-_TOL = 2e-2
+_TOL = 0.1 if mp.is_single_precision() else 0.02
```
stevengj, line 19: It makes sense that the error in finite-difference approximations would depend sensitively on the precision.   If you are computing [f(x+δ)-f(x)]/δ, the optimal choice of δ depends sensitively on the precision, and if δ is too small compared to the precision you can see a large "catastrophic cancellation" error.

python/tests/test_3rd_harm_1d.py
```diff
@@ -53,7 +52,8 @@ def test_3rd_harm_1d(self):
 
         harmonics = [self.k, self.amp, mp.get_fluxes(self.trans1)[0], mp.get_fluxes(self.trans3)[0]]
 
-        np.testing.assert_allclose(expected_harmonics, harmonics)
+        tol = 3e-5 if mp.is_single_precision() else 1e-7
```
stevengj, line 55: This also makes sense — there is little hope of getting 7 significant digits in single precision.

stevengj:
We will also want to change the default tolerance of the CW solver — it currently defaults to `1e-8`, but I would change it to be `1e-8` in double precision and `1e-5` in single precision.

python/tests/test_multilevel_atom.py
```diff
@@ -1,13 +1,11 @@
...
+import meep as mp
 import math
 import unittest
 
-import meep as mp
-
 
 class TestMultiLevelAtom(unittest.TestCase):
 
+    @unittest.skipIf(mp.is_single_precision(), "double-precision floating point specific test")
```
stevengj, line 8: Would be good to open an issue about this and ping Alex Cerjan.  I'm not sure why a Maxwell–Bloch simulation would be super sensitive to floating-point errors.

stevengj: Maybe we are entering some sort of chaotic regime, in which the precise details (phase etcetera) are extremely sensitive to initial conditions and small calculational errors.

tests/symmetry.cpp
```diff
@@ -977,7 +977,9 @@ int main(int argc, char **argv) {
 #endif
 
   if (!nonlinear_ex(vol1d(1.0, 30.0), one)) meep::abort("error in 1D nonlinear vacuum\n");
-  if (!nonlinear_ex(vol3d(1.0, 1.2, 0.8, 10.0), one)) meep::abort("error in 3D nonlinear vacuum\n");
+  if (sizeof(realnum) == sizeof(double)) {
```
stevengj, line 980: Another chaotic case??

Issue #1685, opened 2021-07-19
Missing factor of 2 in adjoint gradients for force_complex_fields=True


mochen4:
When investigating https://github.com/NanoComp/meep/issues/1667, it occurred that adjoint gradients required an additional factor of 1/2 to match finite difference gradients. That factor probably came from the use of complex fields when ``sim.k_point`` is nonzero. When ``sim.k_point`` is zero, specifically setting ``force_complex_fields=True`` in simulations will also lead to the same factor of 1/2.

Preliminary tests have been performed with the same setup as in https://github.com/NanoComp/meep/issues/1667, but with multiple far points, multiple frequencies, etc. [Unit tests](https://github.com/NanoComp/meep/blob/master/python/tests/test_adjoint_solver.py) but with ``force_complex_fields=True`` are also examined. In all cases, the factor consistently appeared to be 1/2.

stevengj:
@smartalecH, any ideas of why you'd be missing a factor of 2 in the complex case?

In the real-fields case we take the real parts of the adjoint currents, but nothing seems to explicitly take this into account in the code…

smartalecH:
Nothing comes to mind immediately... I'd have to go through everything

smartalecH:
Aha!

When we started this journey (many moons ago) we had a "magic 2" factor:

https://github.com/NanoComp/meep/blob/d662b0c4f518a4d14ad530394dd2d10d82887652/src/meepgeom.cpp#L2582-L2584

and 

https://github.com/NanoComp/meep/blob/d662b0c4f518a4d14ad530394dd2d10d82887652/src/meepgeom.cpp#L2615

that dates back to Homer's original adjoint code.

Now we know this is likely because meep takes the real part of a source, `s` such that our new source is `J = 0.5(s+s.conj())`.

In the complex case, of course, meep doesn't do this.

So the proper thing to do up would be eliminate the magic 2 in the above lines and introduce a check in the source placement code (all python) to see if the simulation is complex or not.

PR #1704, opened 2021-07-27
Fix factor of 2 in adjoint gradients when not using complex fields

mochen4:
Fixes https://github.com/NanoComp/meep/issues/1685. The [test_adjoint_solver.py](https://github.com/NanoComp/meep/blob/master/python/tests/test_adjoint_solver.py) is updated correspondingly to include when ``force_complex_fields=True``.

### DIFF
--- a/python/adjoint/objective.py
+++ b/python/adjoint/objective.py
@@ -92,7 +92,7 @@ class ObjectiveQuantity(abc.ABC):
         else:
             # multi frequency simulations
             scale = dV * iomega / adj_src_phase
-        if self.sim.force_complex_fields:
+        if not self.sim.force_complex_fields:
             scale *= 0.5
         return scale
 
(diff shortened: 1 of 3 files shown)


mochen4:
I am very confused why the "CI/Test Python 3.9 with MPI (true)"  failed while the other three passed. In fact, the test that failed is ``test_adjoint_solver.py``.

python/adjoint/objective.py
```diff
@@ -92,6 +92,8 @@ def _adj_src_scale(self, include_resolution=True):
         else:
             # multi frequency simulations
             scale = dV * iomega / adj_src_phase
+        if not self.sim.force_complex_fields:
```
stevengj, line 95:
```suggestion
        # compensate for the fact that real fields take the real part of the current,
        # which halves the Fourier amplitude at the positive frequency (Re[J] = (J + J*)/2)
        if not self.sim.force_complex_fields:
```

Issue #1667, opened 2021-07-12
Adjoint solver with oblique planewave sources


mawc2019:
Does the ``Near2FarFields`` adjoint solver support oblique planewave sources? I tried to launch an oblique planewave, but the finite-difference and adjoint gradients were inconsistent. The code is as follows, where the grazing angle is ``theta_in``.
When the planewave is obliquely incident with ``theta_in​`` taking, e.g., ``np.radians(60)``, the two types of gradients​ seem uncorrelated. Even if the planewave is normally incident with ``theta_in=np.radians(90)``, the two types of gradients are still inconsistent. But in this special case, they show a good linear relationship. At ``k=mp.Vector3()``, the two types of gradients become consistent.
The inconsistency also happens in ``FourierFields``.
```
import meep as mp
import meep.adjoint as mpa
import numpy as np
from autograd import numpy as npa
import math,cmath
Si = mp.Medium(index=3.4)
SiO2 = mp.Medium(index=1.44)

resolution,design_resolution = 20,20
Lx,Ly,pml_size = 4,4,1
cell_size = mp.Vector3(Lx,Ly+2*pml_size)
pml_layers = [mp.PML(pml_size, direction=mp.Y)]
design_width,design_height = 1,1

fcen,fwidth = 1/1.55,0.01/1.55
source_center,source_size  = mp.Vector3(0,-Ly/4),mp.Vector3(Lx,0)

theta_in = np.radians(60)
k = mp.Vector3(fcen*1.44).rotate(mp.Vector3(z=1), theta_in)

def pw_amp(k,x0):
  def _pw_amp(x):
    return cmath.exp(1j*2*math.pi*k.dot(x+x0))
  return _pw_amp

src = mp.GaussianSource(frequency=fcen,fwidth=fwidth)
source = [mp.Source(src,component=mp.Ez,center=source_center,size=source_size,amp_func=pw_amp(k,source_center))]

Nx,Ny = int(round(design_resolution*design_width)),int(round(design_resolution*design_height))

design_variables = mp.MaterialGrid(mp.Vector3(Nx,Ny),SiO2,Si,grid_type='U_MEAN')
design_region = mpa.DesignRegion(design_variables,volume=mp.Volume(center=mp.Vector3(), size=mp.Vector3(design_width, design_height)))

geometry = [mp.Block(center=design_region.center, size=design_region.size, material=design_variables)] # design region
sim = mp.Simulation(
[...]

PR #1705, opened 2021-07-27
Enable nonzero k_point in adjoint solver

mochen4:
Fixes https://github.com/NanoComp/meep/issues/1667. The [test_adjoint_solver.py](https://github.com/NanoComp/meep/blob/master/python/tests/test_adjoint_solver.py) is updated correspondingly to include when ``k_point`` is ``False``, zero, and nonzero.

### DIFF
--- a/python/adjoint/objective.py
+++ b/python/adjoint/objective.py
@@ -94,7 +94,7 @@ class ObjectiveQuantity(abc.ABC):
             scale = dV * iomega / adj_src_phase
         # compensate for the fact that real fields take the real part of the current,
         # which halves the Fourier amplitude at the positive frequency (Re[J] = (J + J*)/2)
-        if not self.sim.using_real_fields():
+        if self.sim.using_real_fields():
             scale *= 2
         return scale
 
--- a/python/adjoint/objective.py
+++ b/python/adjoint/objective.py
@@ -94,7 +94,7 @@ class ObjectiveQuantity(abc.ABC):
             scale = dV * iomega / adj_src_phase
         # compensate for the fact that real fields take the real part of the current,
         # which halves the Fourier amplitude at the positive frequency (Re[J] = (J + J*)/2)
-        if not self.sim.force_complex_fields:
+        if not self.sim.using_real_fields():
             scale *= 2
         return scale
 
--- a/python/adjoint/objective.py
+++ b/python/adjoint/objective.py
@@ -92,6 +92,10 @@ class ObjectiveQuantity(abc.ABC):
         else:
             # multi frequency simulations
             scale = dV * iomega / adj_src_phase
+        # compensate for the fact that real fields take the real part of the current,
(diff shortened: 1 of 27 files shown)


mochen4:
The original setup of the ``EigenmodeCoefficient`` test in the [test_adjoint_solver.py](https://github.com/NanoComp/meep/blob/master/python/tests/test_adjoint_solver.py) has the source and monitor coincide with the sides of the PMLs, and extends into the PMLs of the other direction.  The adjoint gradients and directional derivates won't match if ``k_point`` is nonzero; if we use the original ``calculate_fd_gradient()`` in [optimization_problem.py](https://github.com/NanoComp/meep/blob/master/python/adjoint/optimization_problem.py) that calculates the fd_gradients in unit directions, the fields become ``nan``. (The forward and adjoint run are both okay, but the fields are ``nan`` after single pixel is perturbed during ``calculate_fd_gradient()``.)

Using the setup of the [tutorial example](https://github.com/NanoComp/meep/blob/master/python/examples/adjoint_optimization/01-Introduction.ipynb), the gradients agree with nonzero ``k_point``.

If the source and monitor don't extend into the PMLs, then the adjoint gradients will match the directional derivates, but the fields are still ``nan`` when using ``calculate_fd_gradient()``. If we then moved the source and monitor a little bit away from the PMLs, everything will work fine; so I further modified the test this way.

mochen4:
In the context of the PR https://github.com/NanoComp/meep/pull/1704, the factor of 2 is needed whenever ``force_complex_fields=False``. However, when ``k_point`` is nonzero,  we may not need the factor even if we don't force complex fields.

I wonder if there are other cases where we don't need the factor even if ``force_complex_fields=False``; and what will be the best way to check whether indeed only real fields are used.

oskooi:
Looks like the failing test is `test_adjoint_solver.py` for the case in which Meep is compiled with single precision (#1698). Perhaps you need to just lower the tolerance some place when `mp.is_single_precision()` is true?

mochen4:
Thanks Ardavan! I lowered the tolerance for single precision, and it works!

python/adjoint/objective.py
```diff
@@ -92,6 +92,8 @@ def _adj_src_scale(self, include_resolution=True):
         else:
             # multi frequency simulations
             scale = dV * iomega / adj_src_phase
+        if ((not self.sim.k_point) or self.sim.k_point==mp.Vector3()) and not self.sim.force_complex_fields:
```
stevengj, line 95: We should really check `fields.is_real`, assuming the `fields` object exists here.

stevengj: Even simpler, check `self.use_real()`

stevengj: `use_real` is currently an internal function of `init_sim`, which can just be split into a separate method, and probably renamed to `using_real_fields()`

PR #1709, opened 2021-07-27
flush subnormals on x86

### LINKED ISSUE #1708: flush zero subnormals
During the beginning of a simulation, when the sources are exponentially small, depending on the cutoff parameters we might be seeing [subnormal values](https://en.wikipedia.org/wiki/Denormal_number) — these are values with magnitudes `< 2.2e-308` in double precision and `< 1.2e-38` in single precision.     Unfortunately, on many CPUs the floating-point computations are much slower for subnormal values.

Since we don't actually care about representing such small values accurately, you can change the floating-point mode to simply "flush" subnormal values to `0.0`.    It might be good to just do this unconditionally when Meep is initialized.

How to do this is CPU-specific, but on x86 processors you can call the `_mm_set_csr` intrinsic.  Here is an example of how it is done in the Julia standard library: https://github.com/JuliaLang/julia/blob/bb5b98e72a151c41471d8cc14cacb495d647fb7f/src/processor_x86.cpp#L1087-L1104
(This code is [MIT-licensed](https://github.com/JuliaLang/julia/blob/master/LICENSE.md), which is GPL-compatible, so we can simply copy it into Meep if we want.)

(This needs to be done per-thread, so for OpenMP-enabled runs we will need to call this on each thread.)

stevengj:
Closes #1708.

### DIFF
--- a/python/tests/test_array_metadata.py
+++ b/python/tests/test_array_metadata.py
@@ -1,8 +1,9 @@
 import meep as mp
 import unittest
 import numpy as np
+from utils import VectorComparisonMixin
 
-class TestArrayMetadata(unittest.TestCase):
+class TestArrayMetadata(VectorComparisonMixin):
 
     def test_array_metadata(self):
         resolution = 25
@@ -80,11 +81,8 @@ class TestArrayMetadata(unittest.TestCase):
         vec_func_sum = np.sum(W*(xm**2 + 2*ym**2))
         pulse_modal_volume = np.sum(W*EpsE2)/np.max(EpsE2) * vec_func_sum
 
-        if ((mp.count_processors() % 2 == 0) and mp.is_single_precision()):
-            ref_val = 0.94
-        else:
-            ref_val = 1.00
-        self.assertAlmostEqual(cw_modal_volume/pulse_modal_volume, ref_val, places=2)
+        tol = 5e-2 if mp.is_single_precision() else 1e-2
+        self.assertVectorsClose(cw_modal_volume/pulse_modal_volume, 1.0, epsilon=tol)
 
 if __name__ == '__main__':
     unittest.main()
--- a/python/tests/test_array_metadata.py
+++ b/python/tests/test_array_metadata.py
@@ -1,9 +1,9 @@
 import meep as mp
 import unittest
 import numpy as np
-from utils import VectorComparisonMixin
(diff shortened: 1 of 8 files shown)


ahoenselaar:
I wonder whether this could negatively affect the source cutoff code here:
https://github.com/NanoComp/meep/blob/master/src/sources.cpp#L70

stevengj:
I'm not sure how it would affect that cutoff calculation, could you elaborate?  (`1e-100` is not subnormal in double precision, which is used for that calculation.)

ahoenselaar:
Oh right, the computation is always done in double precision. Should be fine then.

stevengj:
That being said, this `1e-100` cutoff (and any similar cutoffs elsewhere in the code) should really be increased when the fields are single precision.

oskooi:
This seems to have fixed the issue with the noticeable slowdown in the timestepping due to the turn on of the pulsed source for a test problem in 3d:

PR1709

I'm not sure why there is a failing C++ unit test (`harmonics.cpp`) for single-precision fields due to a small mismatch in the relative tolerance. Perhaps we can change that in a separate PR? 
```
FAIL: harmonics
===============

harmonics(2.7e-05,0.0001,1) = 9.80387e-07, 9.99195e-07
error: 3rd harmonic mismatches known val
 --- 9.99195e-07 vs. 9.99349e-07 (0.000154525 error > 0.0001)

```

oskooi:
I tested this branch on my local machine and the same failing test on CI `harmonics.cpp` also failed. The same test (along with all the other unit tests in the `make check` suite) pass using the master branch. This means something about this branch is causing a small change in the Harminv results in `harmonics.cpp` when using single precision.

stevengj:
I think the tolerance in `harmonics.cpp` (and in some of the other tests) was simply overly aggressive for single precision, so it was vulnerable to failure from slight changes in roundoff behavior.

python/tests/test_array_metadata.py
```diff
@@ -1,8 +1,9 @@
 import meep as mp
 import unittest
 import numpy as np
+from utils import VectorComparisonMixin
 
-class TestArrayMetadata(unittest.TestCase):
+class TestArrayMetadata(VectorComparisonMixin):
```
stevengj, line 6: Note: this will have to be updated when #1712 is merged.

PR #1713, opened 2021-07-28
add --with-coverage to control usage of Python coverage tests

stevengj:
Enables `--with-coverage` during CI, but the Python `coverage` module should not be used by default simply because it is present on a user's machine.

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -144,7 +144,7 @@ jobs:
       run: |
         mkdir -p build &&
         pushd build &&
-        ../configure --enable-maintainer-mode --prefix=${HOME}/local --with-libctl=${HOME}/local/share/libctl ${MPICONF} &&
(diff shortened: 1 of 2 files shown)


.github/workflows/build-ci.yml
```diff
@@ -144,7 +144,7 @@ jobs:
       run: |
         mkdir -p build &&
         pushd build &&
-        ../configure --enable-maintainer-mode --prefix=${HOME}/local --with-libctl=${HOME}/local/share/libctl ${MPICONF} &&
+        ../configure --enable-maintainer-mode --with-coverage --prefix=${HOME}/local --with-libctl=${HOME}/local/share/libctl ${MPICONF} &&
```
oskooi, line 147: The `--with-coverage` should actually have been added to this line:

https://github.com/NanoComp/meep/blob/ef2c9cad8a148c2602fa60e065a12352a89d6c21/.github/workflows/build-ci.yml#L152

PR #1714, opened 2021-07-29
Lower tolerance of CW and eigenfrequency solver tests for single precision

oskooi:
Addresses a [comment](https://github.com/NanoComp/meep/pull/1698#issuecomment-885181576) by @stevengj in #1698.

### DIFF
--- a/tests/near2far.cpp
+++ b/tests/near2far.cpp
@@ -44,7 +44,7 @@ int check_cyl(double sr, double sz, double a) {
   continuous_src_time src(w);
   vec x0 = veccyl(0.5 * sr, 0);
   f.add_point_source(c0, src, x0);
-  f.solve_cw(1e-6);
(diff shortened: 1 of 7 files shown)


tests/near2far.cpp
```diff
@@ -137,7 +137,7 @@ int check_2d_3d(ndim dim, const double xmax, double a, component c0, component c
   continuous_src_time src(w);
   f.add_point_source(c0, src, zero_vec(dim));
   if (c1 != NO_COMPONENT) f.add_point_source(c1, src, zero_vec(dim), 0.7654321);
-  f.solve_cw(1e-6);
+  f.solve_cw(sizeof(realnum) == sizeof(float) ? 1e-5 : 1e-6);
```
stevengj, line 140: Probably not necessary to change this, but it's fine as long as the test still passes.

stevengj:
Does any of the documentation need to change?

PR #1715, opened 2021-07-29
Add support for fully local HDF5 files and shared dumping of meep::structure

kkg4theweb:
Add support for dumping meep:structure to sharded files where each process writes out the data for the chunks it owns to its own file (typically named after the process rank). This is useful when running with MPI but there is no proper parallel file support on the running system.

To support this, add a 'local' mode to the h5file wrapper to support writing to local files (with parallel or MPI I/O)


stevengj (approved):

src/structure_dump.cpp
```diff
@@ -85,36 +87,60 @@ void structure::dump_chunk_layout(const char *filename) {
...
 
-void structure::dump(const char *filename) {
-  if (verbosity > 0) master_printf("creating epsilon output file \"%s\"...\n", filename);
-
-  // make/save a num_chunks x NUM_FIELD_COMPONENTS x 5 array counting
-  // the number of entries in the chi1inv array for each chunk.
-  size_t *num_chi1inv_ = new size_t[num_chunks * NUM_FIELD_COMPONENTS * 5];
-  memset(num_chi1inv_, 0, sizeof(size_t) * size_t(num_chunks * NUM_FIELD_COMPONENTS * 5));
+void structure::dump(const char *filename, bool single_parallel_file) {
+  if (verbosity > 0) printf("creating epsilon from file \"%s\" (%d)...\n", filename, single_parallel_file);
```
oskooi, line 91: Shouldn't `printf`  be `master_printf`?

src/structure_dump.cpp
```diff
@@ -542,35 +555,49 @@ void structure::load_chunk_layout(const std::vector<grid_volume> &gvs,
   check_chunks();
 }
 
-void structure::load(const char *filename) {
-  h5file file(filename, h5file::READONLY, true);
+void structure::load(const char *filename, bool single_parallel_file) {
+  h5file file(filename, h5file::READONLY, single_parallel_file, !single_parallel_file);
 
-  if (verbosity > 0) master_printf("reading epsilon from file \"%s\"...\n", filename);
+  if (verbosity > 0) printf("reading epsilon from file \"%s\" (%d)...\n", filename, single_parallel_file);
```
oskooi, line 561: Again, it seems `printf` should be `master_printf`.

python/simulation.py
```diff
@@ -1897,6 +1894,41 @@ def load_chunk_layout(self, br, source):
...
+            # the process rank to get a unique name.
+            dump_dirname = os.path.join(dirname, 'rank%02d' % mp.my_rank())
+        return dump_dirname
+
+    def dump(self, dirname, structure=True, single_parallel_file=True):
+        """
+        Dumps simulation state.
+        """
+        dump_dirname = self._get_load_dump_dirname(dirname, single_parallel_file)
+        os.makedirs(dump_dirname, exist_ok=True)
```
oskooi, line 1914: Note that the Meep module already contains a [`make_output_directory`](https://github.com/NanoComp/meep/blob/master/src/output_directory.cpp#L86-L130) function which is used by the unit tests, e.g. for a temp directory:

https://github.com/NanoComp/meep/blob/9509547b997747d9ba331c3a60a2508c8d918c67/python/tests/test_simulation.py#L25-L27

python/simulation.py
```diff
@@ -960,7 +960,6 @@ def __init__(self,
                  filename_prefix=None,
                  output_volume=None,
                  output_single_precision=False,
-                 load_structure='',
```
oskooi, line 963: Also need to update the docstrings for the function `load_structure`:

https://github.com/NanoComp/meep/blob/9509547b997747d9ba331c3a60a2508c8d918c67/python/simulation.py#L1871-L1878

Note that changing the docstrings requires building the markdown page `Python_User_Interface.md` for the user manual as described in https://github.com/NanoComp/meep/blob/master/doc/README.md in the paragraph starting with "To (re)generate the Python API documentation..."

The documentation in `Python_User_Interface.md.in` also needs to be updated:

https://github.com/NanoComp/meep/blob/9509547b997747d9ba331c3a60a2508c8d918c67/doc/docs/Python_User_Interface.md.in#L350-L373

src/structure_dump.cpp
kkg4theweb: I wanted this to be a "printf" because I wanted all the processes to log this message since the filename would be different for each of them.

src/structure_dump.cpp
kkg4theweb: ditto. Please LMK if that is not advisable for some reason.

python/simulation.py
kkg4theweb: When writing to a sharded file, each shard is written to its own sub-directory under the specified top-level directory - and hence the makedirs (see _get_load_dump_dirname)

python/simulation.py
kkg4theweb: done.

oskooi:
LGTM.

oskooi (approved):

ahoenselaar (approved):

PR #1718, opened 2021-07-31
add fields::dft_fields_norm2

### LINKED ISSUE #1711: improved stop_when_dft_decayed
Currently the adjoint code has a `stop_when_dft_decayed` function by @smartalecH that checks whether the DTFT is converged at the center frequency, but it has one big problem — it calls `get_dft_fields`, which collects all of the DFT data on every process.  (It also seems like it looks at a single frequency, which might be a problem if there is a narrow-band resonance that decays very slowly.)

It would be good to have a parallel, performant version of this, probably implemented in C++.   I think a good approach would be to check convergence of the norm of *all* the spatial points for *each* frequency separately.  That is, for each frequency you compute the L2 norm of all of the spatial points (and maybe all components, or maybe do this per DFT chunk).   Then compare this to the norm from a previous timestep (every N timesteps for some N) to see if it is converged (similarly to `stop_when_fields_decayed` etcetera).   This way the storage requirements (to cache the previous norm) are only proportional to the number of frequencies, not the number of spatial points.

stevengj:
Fix #1711 by adding a `fields::dft_fields_norm2()` function that returns the squared L2 norm of all of the fields used in DFT monitor regions.   This can be computed in parallel, and should be a reasonable quantity to use for convergence tests, similar to `stop_when_fields_decayed` but specific to DFT monitors.

Performance should be good (comparable to updating the DFTs), and is not too important since this will not typically be called every timestep.

To do:
- [ ] Add `stop_when_dft_fields_decayed` or similar function — this should use `fields.dft_maxfreq()` to determine how often to run, e.g. it can run every `5.1/fields.dft_maxfreq()` time units.
- [ ] Switch to new convergence check in tests
- [ ] Switch to new convergence check in adjoint solver

cc @smartalecH

### DIFF
--- a/src/dft.cpp
+++ b/src/dft.cpp
@@ -264,6 +264,46 @@ void dft_chunk::update_dft(double time) {
   }
 }
 
+/* Return the L2 norm^2 of all of the fields used to update DFTs.  This is useful
+   to check whether the simulation is finished (whether all relevant fields have decayed).
+   (Collective operation.) */
+double fields::dft_fields_norm2() {
+  am_now_working_on(Other);
+  double sum = 0.0;
+  for (int i = 0; i < num_chunks; i++)
+    if (chunks[i]->is_mine()) sum += chunks[i]->dft_fields_norm2();
+  finished_working();
+  return sum_to_all(sum);
+}
+
+double fields_chunk::dft_fields_norm2() const {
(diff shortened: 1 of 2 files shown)


smartalecH:
We can probably close this PR now.

PR #1720, opened 2021-08-05
decimation option for remaining DFT functions

### LINKED ISSUE #1678: offer DFT decimation option

oskooi:
Closes #1678.

#1684 introduced a new `decimation_factor` parameter for the `add_dft_fields` function to enable decimation of the DFT time-series updates. This PR adds the `decimation_factor` option to the five remaining DFT functions: `add_flux`,  `add_near2far`, `add_mode_monitor`, `add_energy`, and `add_forces`. Another DFT function [`dft_ldos`](https://meep.readthedocs.io/en/latest/Python_User_Interface/#ldos-functions) handles the DFT fields differently than the other functions (i.e., it does not initialize the fields using `fields::add_dft`) and will be dealt with in a separate PR.

This PR still requires unit tests and documentation before it can be merged.

### DIFF
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -534,10 +534,29 @@ def mean_time_spent_on(self, time_sink):
 <div class="method_docstring" markdown="1">
 
 Return the mean time spent by all processes for a type of work `time_sink` which
-can be one of ten integer values `0`-`9`: (`0`) connecting chunks, (`1`) time stepping,
-(`2`) copying boundaries, (`3`) MPI all-to-all communication/synchronization,
-(`4`) MPI one-to-one communication, (`5`) field output, (`6`) Fourier transforming,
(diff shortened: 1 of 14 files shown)


oskooi:
Unit tests and documentation have been added. This is ready to be merged.

PR #1722, opened 2021-08-07
decimation option for adjoint solver

oskooi:
Following #1684 and #1720, this PR adds a new `decimation_factor` parameter (default value of 1, i.e. no decimation) to the `OptimizationProblem` and objective functions `EigenmodeCoefficient`, `FourierFields`, `Near2FarFields` of the adjoint solver to enable decimation of the DFT time-series updates. Since the entire design region is a DFT monitor for each of the *E*<sub>x</sub>, *E*<sub>y</sub>, *E*<sub>z</sub> fields in 3d used in both the forward and adjoint runs, decimation of the time series can significantly reduce the cost of the DFT fields updates during adjoint optimization.

### DIFF
--- a/python/tests/test_adjoint_solver.py
+++ b/python/tests/test_adjoint_solver.py
@@ -72,21 +72,23 @@ def forward_simulation(design_params,mon_type,frequencies=None, use_complex=Fals
                         sources=sources,
                         geometry=geometry,
                         force_complex_fields=use_complex,
-                        k_point = k)
+                        k_point=k)
     if not frequencies:
         frequencies = [fcen]
 
     if mon_type.name == 'EIGENMODE':
         mode = sim.add_mode_monitor(frequencies,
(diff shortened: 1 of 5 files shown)


stevengj (approved):

python/tests/test_adjoint_solver.py
```diff
@@ -158,7 +164,8 @@ def J(mode_mon):
         objective_arguments = obj_list,
         design_regions = [matgrid_region],
         frequencies=frequencies,
-        decay_fields=[mp.Ez])
+        decay_fields=[mp.Ez],
+        decimation_factor=10)
```
smartalecH, line 168: Why are some default values for `decimation_factor` 5, some 10, and others 1? Shouldn't they all have the same value (either 1 or a value determined by the sources/frequencies requested)?

smartalecH (approved):

PR #1723, opened 2021-08-07
Prepare for 1.20 release

oskooi:
Also includes some minor improvements to the docs.


doc/docs/Python_User_Interface.md
```diff
@@ -534,29 +530,14 @@ def mean_time_spent_on(self, time_sink):
...
-* meep.BoundarySteppingB ("boundary stepping B")
-* meep.BoundarySteppingWH ("boundary stepping WH")
-* meep.BoundarySteppingPH ("boundary stepping PH")
-* meep.BoundarySteppingH ("boundary stepping H")
-* meep.BoundarySteppingD ("boundary stepping D")
-* meep.BoundarySteppingWE ("boundary stepping WE")
-* meep.BoundarySteppingPE ("boundary stepping PE")
-* meep.BoundarySteppingE ("boundary stepping E")
-* meep.Other ("everything else")
+can be one of the following integer constants: `0`: "time stepping", `1`: "connecting chunks",
```
ahoenselaar, line 533: Why do we use these integer values here instead of the enum?

oskooi: Two reasons: (1) integer values for the indices of the timing stack was the way this was originally documented and (2) for convenience, since the enums sometimes have too many characters to type out.

PR #1732, opened 2021-08-14
 enable automatic DFT decimation by default

### LINKED ISSUE #1731: automatic DFT decimation for band-limited sources
Now that we have the decimation option (#1678), it would be good to make this the default.   In particular, whenever we have a known bandlimited source (e.g. a gaussian, where we set the bandwidth as e.g. a tail of 1e-7), we can apply the formula from that issue to set a default decimation factor for any DFT accumulation.

For an API, I'm thinking of a default decimation factor of `decimation=0`, which will mean "use the automatic decimation if possible, otherwise set `decimation=1`".  That way, the user can still override the DFT decimation by setting it to a positive value.

oskooi:
Closes #1731.

### DIFF
--- a/src/dft.cpp
+++ b/src/dft.cpp
@@ -181,12 +181,10 @@ dft_chunk *fields::add_dft(component c, const volume &where, const double *freq,
       if (s->frequency().real()+0.5*s->fwidth() > src_freq_max)
         src_freq_max = s->frequency().real()+0.5*s->fwidth();
     }
-    if (src_freq_max > 0)
+    if (src_freq_max > 0) {
       decimation_factor = 1/(dt*(freq[Nfreq-1] + src_freq_max));
-    if (decimation_factor > 2)
-      decimation_factor -= 1;
-    else
-      decimation_factor = 1;
+    }
+    if (decimation_factor > 1) decimation_factor -= 1;
   }
   data.decimation_factor = decimation_factor;
 
--- a/src/dft.cpp
+++ b/src/dft.cpp
@@ -174,7 +174,22 @@ dft_chunk *fields::add_dft(component c, const volume &where, const double *freq,
   dft_chunk_data data;
   data.c = c;
   data.vc = vc;
+
+  if (decimation_factor == 0) {
+    double src_freq_max = 0;
+    for (src_time *s = sources; s; s = s->next) {
+      if (s->frequency().real()+0.5*s->fwidth() > src_freq_max)
+        src_freq_max = s->frequency().real()+0.5*s->fwidth();
+    }
+    if (src_freq_max > 0)
+      decimation_factor = 1/(dt*(freq[Nfreq-1] + src_freq_max));
+    if (decimation_factor > 2)
+      decimation_factor -= 1;
+    else
+      decimation_factor = 1;
+  }
   data.decimation_factor = decimation_factor;
+
   data.omega.resize(Nfreq);
   for (size_t i = 0; i < Nfreq; ++i)
     data.omega[i] = 2 * pi * freq[i];
--- a/src/dft.cpp
+++ b/src/dft.cpp
@@ -178,8 +178,13 @@ dft_chunk *fields::add_dft(component c, const volume &where, const double *freq,
   if (decimation_factor == 0) {
     double src_freq_max = 0;
     for (src_time *s = sources; s; s = s->next) {
-      if (s->frequency().real()+0.5*s->fwidth() > src_freq_max)
-        src_freq_max = s->frequency().real()+0.5*s->fwidth();
+      if (s->get_width() == 0) {
+          decimation_factor = 1;
+        } else {
+        if (s->frequency().real()+0.5/s->get_width() > src_freq_max) {
+          src_freq_max = s->frequency().real()+0.5/s->get_width();
+        }
+      }
     }
     if (src_freq_max > 0) {
       decimation_factor = 1/(dt*(freq[Nfreq-1] + src_freq_max));
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -107,6 +107,7 @@ def __init__(self,
              progress_interval=4,
              subpixel_tol=0.0001,
              subpixel_maxeval=100000,
+             loop_tile_base=0,
              ensure_periodicity=True,
              num_chunks=0,
              Courant=0.5,
@@ -715,7 +716,7 @@ of the field at that point.
 
 ```python
 def add_dft_fields(self, *args, **kwargs):
-def add_dft_fields(cs, fcen, df, nfreq, freq, where=None, center=None, size=None, yee_grid=False, decimation_factor=1):
+def add_dft_fields(cs, fcen, df, nfreq, freq, where=None, center=None, size=None, yee_grid=False, decimation_factor=0):
 ```
 
 <div class="method_docstring" markdown="1">
@@ -729,12 +730,13 @@ The volume can also be specified via the `center` and `size` arguments. The
 default routine interpolates the Fourier-transformed fields at the center of each
 voxel within the specified volume. Alternatively, the exact Fourier-transformed
 fields evaluated at each corresponding Yee grid point is available by setting
-`yee_grid` to `True`. To reduce the memory-bandwidth burden of
-accumulating DFT fields, an integer `decimation_factor` >= 1 can be
-specified. DFT field values are updated every `decimation_factor`
-timesteps. Use this feature with care, as the decimated timeseries may be
-corrupted by [aliasing](https://en.wikipedia.org/wiki/Aliasing) of high frequencies.
-The choice of decimation factor should take into account the properties of all sources
+`yee_grid` to `True`. To reduce the memory-bandwidth burden of accumulating
+DFT fields, an integer `decimation_factor` can be specified. If `decimation_factor`
+is 0 (the default), this value is automatically determined. It can be turned off by
+setting it to 1. DFT field values are updated every `decimation_factor` timesteps.
+Use this feature with care, as the decimated timeseries may be corrupted by
+[aliasing](https://en.wikipedia.org/wiki/Aliasing) of high frequencies. The choice
+of decimation factor should take into account the properties of all sources
 in the simulation as well as the frequency range of the DFT field monitor.
 
 </div>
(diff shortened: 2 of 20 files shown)


smartalecH:
Seems a bit odd that so many of the existing tests don't support decimation. Aren't all the tests involving DFTs using band-limited sources?

oskooi:
Almost all the existing unit tests in which automatic decimation had to be disabled in this PR involved comparing results based on DFT fields to hard-coded values which were generated without decimation. I suppose we could update all unit tests which involve DFT fields (a large fraction of the overall total) to use decimation to possibly speed things up but that would be a bit of work.

src/dft.cpp
```diff
@@ -174,7 +174,29 @@ dft_chunk *fields::add_dft(component c, const volume &where, const double *freq,
...
   data.c = c;
   data.vc = vc;
+
+  if (decimation_factor == 0) {
+    double src_freq_max = 0;
+    for (src_time *s = sources; s; s = s->next) {
+      if (s->get_width() == 0) {
+          decimation_factor = 1;
+        } else {
+        if (s->frequency().real()+0.5/s->get_width() > src_freq_max) {
```
stevengj, line 184: We really need a `get_fwidth(tol)` method that returns the bandwidth below which the Fourier transform decays below `tol`.

stevengj: (For most sources, `get_fwidth` will default to `0`.  For a Gaussian source we will use a formula, or maybe an approximate formula based on the continuous FT.   For adjoint sources, we have a formula for the Nutall windows.)

oskooi: Fixed. The function `gaussian_src_time::get_fwidth()` computes the bandwidth using an analytic formula for the continuous Fourier transform. The `get_fwidth()` function for the `FilteredSource` object used by the adjoint solver which involves a `CustomSource` will be added in a separate PR. In the meantime, the adjoint simulation does *not* use any decimation.

src/dft.cpp
```diff
@@ -174,7 +174,30 @@ dft_chunk *fields::add_dft(component c, const volume &where, const double *freq,
   dft_chunk_data data;
   data.c = c;
   data.vc = vc;
+
+  if (decimation_factor == 0) {
+    double tol = 1e-7;
```
stevengj, line 179: Should this be `1e-7` in single precision and `1e-15` in double precision?

stevengj: On the other hand, that would make it harder to directly compare single and double precision…

src/dft.cpp
```diff
@@ -174,7 +174,30 @@ dft_chunk *fields::add_dft(component c, const volume &where, const double *freq,
...
+    double src_freq_max = 0;
+    for (src_time *s = sources; s; s = s->next) {
+      if (s->get_fwidth(tol) == 0) {
+          decimation_factor = 1;
+        } else {
+        if (s->frequency().real()+0.5*s->get_fwidth(tol) > src_freq_max) {
```
stevengj, line 185: Maybe just
```cxx
src_freq_max = std::max(src_freq_max, s->frequency().real()+0.5*s->get_fwidth(tol));
```

src/dft.cpp
```diff
@@ -174,7 +174,30 @@ dft_chunk *fields::add_dft(component c, const volume &where, const double *freq,
...
+          src_freq_max = s->frequency().real()+0.5*s->get_fwidth(tol);
+        }
+      }
+    }
+    double freq_max = 0;
+    for (size_t i = 0; i < Nfreq; ++i) {
+      if (freq[i] > freq_max)
```
stevengj, line 192: Similarly,
```cxx
freq_max = std::max(freq_max, std::abs(freq[i]));
```

src/dft.cpp
```diff
@@ -174,7 +174,30 @@ dft_chunk *fields::add_dft(component c, const volume &where, const double *freq,
...
+        freq_max = freq[i];
+    }
+    if ((freq_max > 0) && (src_freq_max > 0)) {
+      decimation_factor = 1/(dt*(freq_max + src_freq_max));
+    }
```
stevengj, line 197: Needs an `else decimation_factor = 1;` clause.

src/dft.cpp
```diff
@@ -174,7 +174,30 @@ dft_chunk *fields::add_dft(component c, const volume &where, const double *freq,
...
+        }
+      }
+    }
+    double freq_max = 0;
+    for (size_t i = 0; i < Nfreq; ++i) {
+      if (freq[i] > freq_max)
+        freq_max = freq[i];
+    }
+    if ((freq_max > 0) && (src_freq_max > 0)) {
+      decimation_factor = 1/(dt*(freq_max + src_freq_max));
```
stevengj, line 196: Probably should be
```cxx
decimation_factor = int(std::ceil(1/(dt*(freq_max + src_freq_max))));
```
(some compilers complain if you don't do an explicit `int` cast, and we should use `ceil` rather than the default truncation to be safe)

src/dft.cpp
oskooi: Fixed.

src/dft.cpp
oskooi: Fixed.

src/dft.cpp
oskooi: Fixed.

src/dft.cpp
oskooi: Fixed.

python/tests/test_n2f_periodic.py
```diff
@@ -49,8 +49,10 @@ def test_nea2far_periodic(self):
                                 sources=sources,
                                 symmetries=symmetries)
 
-            n2f_obj = sim.add_near2far(fcen, 0, 1, mp.Near2FarRegion(center=n2f_pt, size=mp.Vector3(y=sy)), nperiods=10)
-            dft_obj = sim.add_dft_fields([mp.Ez], fcen, 0, 1, center=dft_pt, size=mp.Vector3(y=sy))
+            n2f_obj = sim.add_near2far(fcen, 0, 1, mp.Near2FarRegion(center=n2f_pt, size=mp.Vector3(y=sy)),
+                                       nperiods=10, decimation_factor=1)
+            dft_obj = sim.add_dft_fields([mp.Ez], fcen, 0, 1, center=dft_pt, size=mp.Vector3(y=sy),
+                                         decimation_factor=1)
```
stevengj, line 55: Why does this need to be disabled?

oskooi:
This is working now. I verified that the `fwidth` returned by `gaussian_src_time::get_fwidth(double tol)` for three different values of `tol` (1e-1, 1e-2, 1e-3) matches nearly exactly the bandwidth computed directly from the DTFT spectrum using this 1d test:

```
import meep as mp
import numpy as np

sz = 10.0
dpml = 2.0

fcen = 1.0
df = 0.2
src = [mp.Source(mp.GaussianSource(fcen,fwidth=df),
                 mp.Ex,
                 mp.Vector3(z=-0.5*sz+dpml))]

sim = mp.Simulation(cell_size=mp.Vector3(z=sz),
                    dimensions=1,
                    sources=src,
                    resolution=50,

nfreq = 200
dft_mon = sim.add_dft_fields([mp.Ex], fcen, 2*df, nfreq, center=mp.Vector3(z=0.5*sz-dpml), size=mp.Vector3())

ex_dft = [ np.absolute(sim.get_dft_array(dft_mon, mp.Ex, i)) for i in range(nfreq) ]
freqs = mp.get_flux_freqs(dft_mon)

tols = [ 1e-1, 1e-2, 1e-3 ]
for tol in tols:
    max_val = np.amax(ex_dft)
    idx = np.where(ex_dft > tol*max_val)[0]
    fwidth = freqs[idx[-1]] - freqs[idx[0]]
    print('fwidth({:.2E}) = {:.5f}'.format(tol,fwidth))

plt.figure()
plt.plot(freqs,ex_dft,'bo-')
plt.xlabel('frequency')
plt.ylabel('DTFT of Ex')
plt.savefig('ex_dft.png',dpi=150,bbox_inches='tight')
```

ex_dft

**`fwidth` computed using `gaussian_src_time::get_fwidth(double tol)`**
```
get_fwidth(1e-1) = 0.13662
get_fwidth(1e-2) = 0.19320
get_fwidth(1e-3) = 0.23663
```

**`fwidth` computed from DTFT spectrum**
```
fwidth(1.00E-01) = 0.13266
fwidth(1.00E-02) = 0.19296
fwidth(1.00E-03) = 0.23317
```

python/tests/test_n2f_periodic.py
oskooi: The change that caused this one test to fail using automatic decimation was introduced in c8af01d:
```
< decimation_factor = 1/(dt*(freq_max + src_freq_max));

> decimation_factor = int(std::ceil(1/(dt*(freq_max + src_freq_max))));
```
I verified that this change caused the `decimation_factor` to *increase by one* in the failing test. (I also confirmed that among these two DFT monitors `add_near2far` and `add_dft_fields`, only the `add_near2far` monitor needs to have  automatic decimation disabled or its `decimation_factor` to be reduced by one for the test to pass. Automatic decimation can be reinstated for the `add_dft_fields` monitor and the test will pass.)

There are two possible workarounds: (1) we can replace `std::ceil` with `std:floor` to be safe or (2) just continue to disable the automatic decimation for this one unit test as we are doing now.

src/dft.cpp
```diff
@@ -174,7 +174,26 @@ dft_chunk *fields::add_dft(component c, const volume &where, const double *freq,
...
+      if (s->get_fwidth(tol) == 0)
+        decimation_factor = 1;
+      else
+        src_freq_max = std::max(src_freq_max, std::abs(s->frequency().real())+0.5*s->get_fwidth(tol));
+    }
+    double freq_max = 0;
+    for (size_t i = 0; i < Nfreq; ++i)
+      freq_max = std::max(freq_max, std::abs(freq[i]));
+    if ((freq_max > 0) && (src_freq_max > 0))
+      decimation_factor = int(std::ceil(1/(dt*(freq_max + src_freq_max))));
```
stevengj, line 191:
```suggestion
      decimation_factor = std::max(1, int(std::floor(1/(dt*(freq_max + src_freq_max)))));
```

oskooi: Fixed.

python/tests/test_n2f_periodic.py
oskooi: By switching `std::ceil` to `std::floor` when setting the `decimation_factor` in 898d4385, the automatic decimation in `test_n2f_periodic.py` no longer needs to be disabled.

PR #1733, opened 2021-08-16
loop tiling for step_update_edhb

oskooi:
Initial attempt to add loop tiling to `STEP_UPDATE_EDHB` via `fields_chunk::update_eh` based on the same approach used for `STEP_CURL` via `fields_chunks::step_db` in #1655. I ran the same benchmark as #1655 (i.e., identical 3d simulation executed simultaneously on all four single-threaded cores of i7-7700) and looked at just the time spent on `update_eh`. The time for `fields_chunk::update_eh(H_stuff)` was negligible and excluded from the results and so only `fields_chunk::update_eh(E_stuff)` matters. Subpixel smoothing was turned *on* which produces anisotropic ε tensors at the discontinuous (lossless ) dielectric interfaces. (The loop tiling for `fields_chunk::step_db` was disabled not that it should affect the timing results for `update_eh`.)

Somewhat unexpectedly, the results show that increasing the number of tiles produces *slower* performance relative to the no-tiling case.

oled_subpix_updateE3_benchmark_procs4

### DIFF
--- a/src/step_db.cpp
+++ b/src/step_db.cpp
@@ -30,8 +30,8 @@ using namespace std;
 
 namespace meep {
 
-static void split_into_tiles(grid_volume gvol, std::vector<grid_volume> *result,
-                             const size_t loop_tile_base) {
+void split_into_tiles(grid_volume gvol, std::vector<grid_volume> *result,
+                      const size_t loop_tile_base) {
   if (gvol.nowned_min() < loop_tile_base) {
     result->push_back(gvol);
     return;
@@ -49,7 +49,7 @@ static void split_into_tiles(grid_volume gvol, std::vector<grid_volume> *result,
 
 // First check that the tile volumes gvs do not intersect and that they add
 // up to the chunk's total grid_volume gv
-static void check_tiles(grid_volume gv, const std::vector<grid_volume> &gvs) {
+void check_tiles(grid_volume gv, const std::vector<grid_volume> &gvs) {
   grid_volume vol_intersection;
   for (size_t i = 0; i < gvs.size(); i++)
     for (size_t j = i + 1; j < gvs.size(); j++)
@@ -63,32 +63,17 @@ static void check_tiles(grid_volume gv, const std::vector<grid_volume> &gvs) {
 }
 
 void fields::step_db(field_type ft) {
-  // split the chunk volume into subdomains for tiled execution of step_db and update_eh loops
+  if (ft != B_stuff && ft != D_stuff) meep::abort("step_db only works with B/D");
+
+  // split the chunk volume into subdomains for tiled execution of step_db loop
   for (int i = 0; i < num_chunks; i++)
     if (chunks[i]->is_mine() && changed_materials) {
-      bool is_aniso = false;
-      FOR_FT_COMPONENTS(ft, cc) {
-        const direction d_c = component_direction(cc);
-        const direction d_1 = cycle_direction(chunks[i]->gv.dim, d_c, 1);
-        const direction d_2 = cycle_direction(chunks[i]->gv.dim, d_c, 2);
-        if (chunks[i]->s->chi1inv[cc][d_1] && chunks[i]->s->chi1inv[cc][d_2]) {
-          is_aniso = true;
-          break;
-        }
-      }
-      if (chunks[i]->gvs_db.size() > 0) chunks[i]->gvs_db.clear();
-      if (loop_tile_base_db > 0 && is_aniso) {
+      if (!chunks[i]->gvs_db.empty()) chunks[i]->gvs_db.clear();
+      if (loop_tile_base_db > 0) {
         split_into_tiles(chunks[i]->gv, &chunks[i]->gvs_db, loop_tile_base_db);
         check_tiles(chunks[i]->gv, chunks[i]->gvs_db);
       } else {
         chunks[i]->gvs_db.push_back(chunks[i]->gv);
-      }
-      if (chunks[i]->gvs_eh.size() > 0) chunks[i]->gvs_eh.clear();
-      if (loop_tile_base_eh > 0 && is_aniso) {
-        split_into_tiles(chunks[i]->gv, &chunks[i]->gvs_eh, loop_tile_base_eh);
-        check_tiles(chunks[i]->gv, chunks[i]->gvs_eh);
-      } else {
-        chunks[i]->gvs_eh.push_back(chunks[i]->gv);
       }
     }
 
@@ -102,8 +87,6 @@ void fields::step_db(field_type ft) {
 
 bool fields_chunk::step_db(field_type ft) {
   bool allocated_u = false;
-
-  if (ft != B_stuff && ft != D_stuff) meep::abort("bug - step_db should only be called for B or D");
 
   for (const auto& sub_gv : gvs_db) {
     DOCMP FOR_FT_COMPONENTS(ft, cc) {
--- a/src/step_db.cpp
+++ b/src/step_db.cpp
@@ -30,52 +30,8 @@ using namespace std;
 
 namespace meep {
 
-void split_into_tiles(grid_volume gvol, std::vector<grid_volume> *result,
-                      const size_t loop_tile_base) {
-  if (gvol.nowned_min() < loop_tile_base) {
-    result->push_back(gvol);
-    return;
-  }
-
-  int best_split_point;
-  direction best_split_direction;
-  gvol.tile_split(best_split_point, best_split_direction);
-  grid_volume left_gvol = gvol.split_at_fraction(false, best_split_point, best_split_direction);
-  split_into_tiles(left_gvol, result, loop_tile_base);
-  grid_volume right_gvol = gvol.split_at_fraction(true, best_split_point, best_split_direction);
-  split_into_tiles(right_gvol, result, loop_tile_base);
-  return;
-}
-
-// First check that the tile volumes gvs do not intersect and that they add
-// up to the chunk's total grid_volume gv
-void check_tiles(grid_volume gv, const std::vector<grid_volume> &gvs) {
-  grid_volume vol_intersection;
-  for (size_t i = 0; i < gvs.size(); i++)
-    for (size_t j = i + 1; j < gvs.size(); j++)
-      if (gvs[i].intersect_with(gvs[j], &vol_intersection))
-        meep::abort("gvs[%zu] intersects with gvs[%zu]\n", i, j);
-  size_t sum = 0;
-  for (const auto& sub_gv : gvs) { sum += sub_gv.nowned_min(); }
-  size_t v_grid_points = 1;
-  LOOP_OVER_DIRECTIONS(gv.dim, d) { v_grid_points *= gv.num_direction(d); }
-  if (sum != v_grid_points) meep::abort("v_grid_points = %zu, sum(tiles) = %zu\n", v_grid_points, sum);
-}
-
 void fields::step_db(field_type ft) {
   if (ft != B_stuff && ft != D_stuff) meep::abort("step_db only works with B/D");
-
-  // split the chunk volume into subdomains for tiled execution of step_db loop
-  for (int i = 0; i < num_chunks; i++)
-    if (chunks[i]->is_mine() && changed_materials) {
-      if (!chunks[i]->gvs_db.empty()) chunks[i]->gvs_db.clear();
-      if (loop_tile_base_db > 0) {
-        split_into_tiles(chunks[i]->gv, &chunks[i]->gvs_db, loop_tile_base_db);
-        check_tiles(chunks[i]->gv, chunks[i]->gvs_db);
-      } else {
-        chunks[i]->gvs_db.push_back(chunks[i]->gv);
-      }
-    }
 
   for (int i = 0; i < num_chunks; i++)
     if (chunks[i]->is_mine())
@@ -88,7 +44,7 @@ void fields::step_db(field_type ft) {
 bool fields_chunk::step_db(field_type ft) {
   bool allocated_u = false;
 
-  for (const auto& sub_gv : gvs_db) {
+  for (const auto& sub_gv : gvs_tiled) {
     DOCMP FOR_FT_COMPONENTS(ft, cc) {
       if (f[cc][cmp]) {
         const component c_p = plus_component[cc], c_m = minus_component[cc];
--- a/src/step_db.cpp
+++ b/src/step_db.cpp
@@ -30,12 +30,73 @@ using namespace std;
 
 namespace meep {
 
+static void split_into_tiles(grid_volume gvol, std::vector<grid_volume> *result,
+                             const size_t loop_tile_base) {
+  if (gvol.nowned_min() < loop_tile_base) {
+    result->push_back(gvol);
+    return;
+  }
+
+  int best_split_point;
+  direction best_split_direction;
+  gvol.tile_split(best_split_point, best_split_direction);
+  grid_volume left_gvol = gvol.split_at_fraction(false, best_split_point, best_split_direction);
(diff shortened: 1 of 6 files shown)


src/update_eh.cpp
```diff
@@ -123,53 +123,55 @@ bool fields_chunk::update_eh(field_type ft, bool skip_w_components) {
...
-        if (!f_w_prev[ec][cmp]) f_w_prev[ec][cmp] = new realnum[gv.ntot()];
-        memcpy(f_w_prev[ec][cmp], f_w[ec][cmp] ? f_w[ec][cmp] : f[ec][cmp],
-               sizeof(realnum) * gv.ntot());
-      }
+        // save W field from this timestep in f_w_prev if needed by pols
+        if (needs_W_prev(ec)) {
+          if (!f_w_prev[ec][cmp]) f_w_prev[ec][cmp] = new realnum[gv.ntot()];
+          memcpy(f_w_prev[ec][cmp], f_w[ec][cmp] ? f_w[ec][cmp] : f[ec][cmp],
+                 sizeof(realnum) * gv.ntot());
+        }
```
stevengj, line 166: Note that you are copying the *whole* field here for every tile, which is probably not what you want!

oskooi: True but the fields being copied here are the auxiliary fields for PML and my benchmark test involves a single chunk with no PMLs. Thus, I don't think this is affecting my results. Nevertheless, I can try putting the loop over tiles around the just the call to `STEP_UPDATE_EDHB` rather than the `FOR_FT_COMPNENTS()`.

stevengj:
Maybe try a case that's more of a cubic unit cell, and try chopping along the longest axis.

oskooi:
I ran the benchmarking test for a *cubic* cell on AWS using a single c5.2xlarge instance (output of `lscpu` is below) with hyperthreading disabled (four single-threaded cores) using commit e14a6f9 of this branch. As before, I ran the same test simultaneously on all four threads in order to saturate the memory bandwidth. I also varied the resolution in order to investigate the effect of the total number of pixels (or memory) on the results. 

The results are shown below and are similar to those obtained using my desktop i7-7700k machine. The time spent on `fields_chunk::update_eh(E_stuff)` increases (nearly monotonically) with the number of tiles. For comparison, the time spent on `fields_chunk::step_db(D_stuff)` in the same run decreases with the number of tiles until reaching a minimum similar to what was demonstrated in #1655. Thus, it seems there is still no benefit to using loop tiling for `fields_chunk::updater_eh`.

```
$ lscpu
Architecture:         x86_64
CPU op-mode(s):       32-bit, 64-bit
Byte Order:           Little Endian
CPU(s):               8
On-line CPU(s) list:  0-3
Off-line CPU(s) list: 4-7
Thread(s) per core:   1
Core(s) per socket:   4
Socket(s):            1
NUMA node(s):         1
Vendor ID:            GenuineIntel
CPU family:           6
Model:                85
Model name:           Intel(R) Xeon(R) Platinum 8275CL CPU @ 3.00GHz
Stepping:             7
CPU MHz:              3615.747
BogoMIPS:             6000.00
Hypervisor vendor:    KVM
Virtualization type:  full
L1d cache:            32K
L1i cache:            32K
L2 cache:             1024K
L3 cache:             36608K
NUMA node0 CPU(s):    0-3
```

**1. update_eh**
oled_benchmark_procs4_cube_res140_updateEH

```
processor 1: relative improvement of  0.00% using 1 tiles
processor 2: relative improvement of  0.00% using 1 tiles
processor 3: relative improvement of  0.17% using 3520 tiles
processor 4: relative improvement of  2.04% using 3520 tiles
```

**2. step_db**
oled_benchmark_procs4_cube_res140_stepDB

```
processor 1: relative improvement of 20.95% using 14080 tiles
processor 2: relative improvement of 19.97% using 14080 tiles
processor 3: relative improvement of 21.66% using 28160 tiles
processor 4: relative improvement of 22.37% using 14080 tiles
```

stevengj:
I think this is because your ε tensor is diagonal in this test problem, in which case there is *no* temporal locality to exploit by tiling — you update Ex from Dx, Ey from Dy, etcetera, with each D component value being used exactly once.

To get a benefit, I think you need anistropic non-diagonal ε.   This will happen if you have interfaces along non-cartesian directions, or you can simply add some anistropic material with an arbitrarily oriented ε tensor.

src/update_eh.cpp
```diff
@@ -164,15 +165,19 @@ bool fields_chunk::update_eh(field_type ft, bool skip_w_components) {
...
       }
 
-      if (f[ec][cmp] != f[dc][cmp])
-        STEP_UPDATE_EDHB(f[ec][cmp], ec, gv, gv.little_owned_corner(ec), gv.big_corner(),
-                         dmp[dc][cmp], dmp[dc_1][cmp], dmp[dc_2][cmp],
-                         s->chi1inv[ec][d_ec], dmp[dc_1][cmp] ? s->chi1inv[ec][d_1] : NULL,
-                         dmp[dc_2][cmp] ? s->chi1inv[ec][d_2] : NULL, s_ec, s_1, s_2, s->chi2[ec],
-                         s->chi3[ec], f_w[ec][cmp], dsigw, s->sig[dsigw], s->kap[dsigw]);
+      if (f[ec][cmp] != f[dc][cmp]) {
+        for (const auto& sub_gv : gvs) {
```
stevengj, line 169: 1. This loop should be outside the `FOR_FT_COMPONENTS` loop (just make sure to do the backup copies only once).
2. If the material has a diagonal `s->chi1inv` (i.e. the off-diagonal pointers are `NULL` for this `ft`), then you *don't* want to tile (because there is no locality as explained below) — in that case you want to just set `gvs` to a single tile with the whole `gv`.

oskooi:
With the changes introduced in 150e130 and a new benchmarking test involving a cubic cell with a dielectric sphere using subpixel smoothing which produces anisotropic ε tensors at boundary voxels, the loop tiling for  `fields_chunk::update_eh(E_stuff)` produces a speedup of **~12%** in the optimal tiling configuration relative to no tiling. In the same test, the speedup for `fields_chunk::step_db(D_stuff)` was **~20%** for the *same* tiling configuration as `fields_chunk::update_eh(E_stuff)` in which the speedup was maximized. 

The benchmarking tests were run on AWS EC2 using the same set up as described [previously](https://github.com/NanoComp/meep/pull/1733#issuecomment-907877492).

**1. update_eh**

sphere_benchmark_procs4_res80_updateEH

```
processor 1: relative improvement of 10.54% using 240 tiles
processor 2: relative improvement of 10.77% using 3840 tiles
processor 3: relative improvement of 10.90% using 3840 tiles
processor 4: relative improvement of 12.08% using 3840 tiles
```

**2. step_db**

sphere_benchmark_procs4_res80_stepDB

```
processor 1: relative improvement of 17.21% using 3840 tiles
processor 2: relative improvement of 17.78% using 960 tiles
processor 3: relative improvement of 19.82% using 3840 tiles
processor 4: relative improvement of 19.15% using 3840 tiles
```

**test**
```
import meep as mp

resolution = 80

s = 3.0
cell_size = mp.Vector3(s,s,s)

fcen = 1.0
sources = [mp.Source(mp.ContinuousSource(fcen),
                     center=mp.Vector3(-0.5*s),
                     size=mp.Vector3(0,s,s),
                     component=mp.Ez)]

n_sphere = 2.0
rad = 1.0
geometry = [mp.Sphere(material=mp.Medium(index=n_sphere),
                      center=mp.Vector3(),
                      radius=rad)]

nprocs = mp.count_processors()
nproc = mp.divide_parallel_processes(mp.count_processors())

ltbs = [300,700,1600,3000,5000,25000,80000,0]

for ltb in ltbs:
                        cell_size=cell_size,
                        sources=sources,
                        k_point=mp.Vector3(),
                        geometry=geometry,
                        loop_tile_base=ltb)

    sim.init_sim()

    for r in range(1,6):
        sim.fields.step()
        sim.fields.reset_timers()

        for _ in range(5):
            sim.fields.step()

        sim.output_times('timings_res{}_tb{}_np{}_proc{}_run{}.csv'.format(resolution,ltb,nprocs,nproc,r))

        sim.restart_fields()

    sim.reset_meep()
```

oskooi:
Additional results for a different test case involving anisotropic ε *materials* also show a similar speedup of **~12%** for `fields_chunk::update_eh(E_stuff)` relative to no tiling.  It's interesting that the maximum speedup of ~12% for `fields_chunk::update_eh(E_stuff)` and ~20% for `fields_chunk::step_db(D_stuff)` for the optimal tiling is consistent across different tests and grid resolutions.

Note from the results [above](https://github.com/NanoComp/meep/pull/1733#issuecomment-911883149) and below that for the case of no tiling, the time spent per timestep on `fields_chunk::update_eh(E_stuff)` is more than 2X larger than `fields_chunk::step_db(D_stuff)`. This means that in *absolute* terms, the time savings due to loop tiling is actually more significant `fields_chunk::update_eh(E_stuff)` than it is for `fields_chunk::step_db(D_stuff)`.

I think these results finally demonstrate that loop tiling can be used to speed up `update_eh`.

**1. update_eh**
aniso_sphere_benchmark_procs4_res110_updateEH

```
processor 1: relative improvement of 12.66% using 5280 tiles
processor 2: relative improvement of 11.63% using 5280 tiles
processor 3: relative improvement of 10.62% using 660 tiles
processor 4: relative improvement of 11.96% using 10560 tiles
```

**2. step_db**
aniso_sphere_benchmark_procs4_res110_stepDB

```
processor 1: relative improvement of 20.24% using 2640 tiles
processor 2: relative improvement of 21.39% using 21120 tiles
processor 3: relative improvement of 19.10% using 2640 tiles
processor 4: relative improvement of 21.97% using 10560 tiles
```

**test**
```py
import meep as mp

resolution = 110

s = 3.0
cell_size = mp.Vector3(s,s,s)

fcen = 1.0
sources = [mp.Source(mp.ContinuousSource(fcen),
                     center=mp.Vector3(-0.5*s),
                     size=mp.Vector3(0,s,s),
                     component=mp.Ez)]

aniso_mat1 = mp.Medium(epsilon_diag=mp.Vector3(8.09946,10.12833,6.34247),
                       epsilon_offdiag=mp.Vector3(0.56199,2.84232,5.05261))

aniso_mat2 = mp.Medium(epsilon_diag=mp.Vector3(14.74,23.98,22.64),
                       epsilon_offdiag=mp.Vector3(-11.54,0,-10.69))

rad = 1.0
geometry = [mp.Sphere(material=aniso_mat2,
                      center=mp.Vector3(),
                      radius=rad)]

nprocs = mp.count_processors()
nproc = mp.divide_parallel_processes(mp.count_processors())

ltbs = [300,700,1600,3000,5000,12000,25000,40000,80000,0]

for ltb in ltbs:
                        cell_size=cell_size,
                        sources=sources,
                        k_point=mp.Vector3(),
                        geometry=geometry,
			loop_tile_base=ltb,
                        default_material=aniso_mat1,
                        eps_averaging=True)

    sim.init_sim()

    for r in range(1,6):
        sim.fields.step()
	sim.fields.reset_timers()

        for _ in range(5):
            sim.fields.step()

	sim.output_times('timings_res{}_tb{}_np{}_proc{}_run{}.csv'.format(resolution,ltb,nprocs,nproc,r))

	sim.restart_fields()

    sim.reset_meep()
```

src/update_eh.cpp
```diff
@@ -123,53 +123,74 @@ bool fields_chunk::update_eh(field_type ft, bool skip_w_components) {
...
+        }
+
+        if (f[ec][cmp] != f[dc][cmp])
+          STEP_UPDATE_EDHB(f[ec][cmp], ec, gv, ts, te,
+                           dmp[dc][cmp], dmp[dc_1][cmp], dmp[dc_2][cmp],
+                           s->chi1inv[ec][d_ec], dmp[dc_1][cmp] ? s->chi1inv[ec][d_1] : NULL,
+                           dmp[dc_2][cmp] ? s->chi1inv[ec][d_2] : NULL, s_ec, s_1, s_2, s->chi2[ec],
+                           s->chi3[ec], f_w[ec][cmp], dsigw, s->sig[dsigw], s->kap[dsigw]);
+
+        if (s->chi1inv[ec][d_1] == NULL && s->chi1inv[ec][d_2] == NULL && cmp == 1 - is_real)
```
stevengj, line 191: I would add this logic when you set up the `gvs` list in the first place — if the material is isotropic, then skip the whole subdivision of `gv` and set `gvs` to be a list of a single "tile" (the whole `gv`).

stevengj: This would require you to have a separate `gvs` list for the `step_update_edhb`, which be a good idea anyway since minimum tile size might be different.

stevengj: You'll want to set up these `gvs` lists whenever `changed_materials` is true, so you might as well call the routine to set them up right in `step_*` before they are needed.   i.e.
```
if (changed_materials)
    gvs_edhb = ....;
```

src/step_db.cpp
```diff
@@ -30,21 +30,65 @@ using namespace std;
...
+  LOOP_OVER_DIRECTIONS(gv.dim, d) { v_grid_points *= gv.num_direction(d); }
+  if (sum != v_grid_points) meep::abort("v_grid_points = %zu, sum(tiles) = %zu\n", v_grid_points, sum);
+}
+
 void fields::step_db(field_type ft) {
+  if (ft != B_stuff && ft != D_stuff) meep::abort("step_db only works with B/D");
+
+  // split the chunk volume into subdomains for tiled execution of step_db loop
+  for (int i = 0; i < num_chunks; i++)
+    if (chunks[i]->is_mine() && changed_materials) {
```
stevengj, line 70: This should probably happen in the `fields_chunk` constructor, since it doesn't depend on the materials.

stevengj: And just call it `gvs_tiled`

src/update_eh.cpp
```diff
@@ -27,6 +27,28 @@ namespace meep {
...
+      FOR_FT_COMPONENTS(ft, cc) {
+        const direction d_c = component_direction(cc);
+        const direction d_1 = cycle_direction(chunks[i]->gv.dim, d_c, 1);
+        const direction d_2 = cycle_direction(chunks[i]->gv.dim, d_c, 2);
+        if (chunks[i]->s->chi1inv[cc][d_1] && chunks[i]->s->chi1inv[cc][d_2]) {
+          is_aniso = true;
+          break;
+        }
+      }
+      if (!chunks[i]->gvs_eh.empty()) chunks[i]->gvs_eh.clear();
```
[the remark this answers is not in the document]
oskooi: Fixed.

src/step_db.cpp
oskooi: Fixed.

oskooi:
I ran benchmarks on AWS EC2 with the latest commit (57b4359) and there is again a consistent speedup across all four single-threaded processors of **~12%** for `update_e_from_b` and **~20%** for `step_db` though the tile configurations are different. (Note that tiling is turned off for `update_h_from_b` because there are no anisotropic materials which also just means H=B and therefore no update is actually performed.)

I think this feature is now ready to be merged.

**update_e_from_d**

sphere_subpix_benchmark_procs4_res110_updateEH
```
processor 1: relative improvement of 12.41% using 1320 tiles
processor 2: relative improvement of 12.86% using 1320 tiles
processor 3: relative improvement of 12.09% using 1320 tiles
processor 4: relative improvement of 12.57% using 1320 tiles
```


**step_db**
sphere_subpix_benchmark_procs4_res110_stepDB

```
processor 1: relative improvement of 22.38% using 10560 tiles
processor 2: relative improvement of 20.34% using 2640 tiles
processor 3: relative improvement of 20.92% using 10560 tiles
processor 4: relative improvement of 22.97% using 10560 tiles
```

PR #1740, opened 2021-08-25
Revamped add fields::dft_fields_norm2

### LINKED ISSUE #1711: improved stop_when_dft_decayed

smartalecH:
Fixes #1711

Similar to #1718, except it does an L2 norm over the already calculated DFT fields, rather than manually recomputing (a modified version of) them. This means the user doesn't have to run this as often (before you had to "sample" fast enough to prevent aliasing). Plus there are fewer FLOPs with this approach.

I'm sure there's a reason @stevengj didn't do things this way, just not sure what that may be.

It passed the adjoint solver test suite locally. I reran the first tutorial and saw similar relative errors in the gradients as the old method.

We should probably do some profiling to see what convergence thresholds are really needed for "accurate" gradients.


python/adjoint/optimization_problem.py
```diff
@@ -567,44 +535,9 @@ def _stop(sim):
...
-                            )  # average across space and frequency
-                        else:
-                            relative_change.append(1)
-                    else:
-                        relative_change.append(1)
-
-            relative_change = np.mean(
-                relative_change)  # average across monitors
+            current_fields  = simob.fields.dft_fields_norm()
+            relative_change =  np.abs(previous_fields-current_fields) /  previous_fields
```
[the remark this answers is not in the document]
smartalecH: Fixed.

src/dft.cpp
```diff
@@ -265,6 +265,64 @@ void dft_chunk::update_dft(double time) {
...
+  const int Nomega = omega.size();
+  LOOP_OVER_IVECS(fc->gv, is, ie, idx) {
+    for (int i = 0; i < Nomega; ++i)
+        sum += sqr(dft[Nomega * idx_dft + i]);
+    idx_dft++;
+  }
+  return sum;
+}
+
+// return the minimum abs(freq) over all DFT chunks
```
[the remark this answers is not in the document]
smartalecH: Fixed.

smartalecH:
`stop_when_dft_decayed` now supports *both* methods for computing the decay rate (see the `method` flag). Turns out both methods now give very similar results with the adjoint solver test. The default is the more conservative method in #1718.

The decay rate is calculated every `1/sim.fields.dft_maxfreq()` timesteps for both methods, which means it is called *much* more frequently than before. As a result, the decay tolerance needs to b `<1e-10` for the adjoint tests to pass. This doesn't mean the simulation is running longer than before -- quite the opposite actually. Because it's updating so frequently, it can shut off as soon as that tolerance is hit. Before, the update was only called every 50 timesteps (a rather large default value).

The `OptimizationProblem` no longer uses the `decay_dt` or `decay_fields` parameters, but I left them in to not break (mostly my) existing code. I'm sure with future releases we can drop those parameters.

stevengj:
In principle, we should set how often we check this to be `max(1/(maxfreq*dt), decimation factors)` so that we never halt without updating the DFTs first.

stevengj:
Since they are performing similarly, maybe we should just do the less conservative approach (your initial approach).

smartalecH:
> In principle, we should set how often we check this to be max(1/(maxfreq*dt), decimation factors) so that we never halt without updating the DFTs first.

Fixed.  I added a function, `min_decimation()`, that pulls the minimum decimation factor from all the DFT monitors (very similar to `dft_maxfreq()`.

> Since they are performing similarly, maybe we should just do the less conservative approach (your initial approach), and just call it dft_norm2 or even just norm2.

Fixed.

smartalecH:
I readded the original adjoint solver test file, `test_adjoint_solver.py`. It seemed to have been deleted in #1749.

src/dft.cpp
```diff
@@ -283,6 +283,81 @@ void dft_chunk::update_dft(double time) {
   }
 }
 
+/* Return the L2 norm of the DFTs themselves.  This is useful
+   to check whether the simulation is finished (whether all relevant fields have decayed).
+   (Collective operation.) */
+double fields::dft_norm2() {
```
stevengj, line 289: I would just call it `dft_norm` since it is not squared (unlike `fields_chunk::dft_norm2` or `dft_chunk::norm2`)

python/simulation.py
```diff
@@ -4425,6 +4425,41 @@ def _stop(sim):
...
+    `dt` to use based on the frequency content in the DFT monitors.
+
+    Optionally the user can specify a minimum run time (`minimum_run_time`) or a maximum
+    run time (`maximum_run_time`).
+    """
+    # Record data in closure so that we can persitently edit
+    closure = {'previous_fields':1, 't0':0, 'dt':0}
+    def _stop(_sim):
+        if _sim.fields.t == 0:
+            closure['dt'] = max(int(1/_sim.fields.dft_maxfreq()/_sim.fields.dt),int(_sim.fields.min_decimation()))
```
stevengj, line 4443: Why is `int(...)` required for `min_decimation`?  Isn't that already returning an integer?

python/simulation.py
```diff
@@ -4425,6 +4425,41 @@ def _stop(sim):
...
+        if _sim.fields.t <= closure['dt'] + closure['t0']:
+            return False
+        elif maximum_run_time and _sim.round_time() > maximum_run_time:
+            return True
+        else:
+            previous_fields = closure['previous_fields']
+            current_fields  = _sim.fields.dft_norm2()
+            relative_change = np.abs(previous_fields-current_fields) /  previous_fields
```
stevengj, line 4451: I would `return false if previous_fields == 0`, and then initialize `previous_fields` to `0`.

python/simulation.py
stevengj: Anyway, you can probably remove the `int` calls here completely since there's no reason `dt` needs to be an integer for the `t ≤ dt + t0` check.

python/simulation.py
```diff
@@ -4425,6 +4425,41 @@ def _stop(sim):
...
+            current_fields  = _sim.fields.dft_norm2()
+            relative_change = np.abs(previous_fields-current_fields) /  previous_fields   
+            
+            closure['previous_fields'] = current_fields
+            closure['t0'] = _sim.fields.t
+
+            if mp.verbosity > 0:
+                fmt = "DFT decay(t = {0:1.1f}): {1:0.4e}"
+                print(fmt.format(_sim.meep_time(), np.real(relative_change)))
+            return relative_change <= decay_by and _sim.round_time(
```
stevengj, line 4459: I'm not completely happy with the fact that the `decay_by` tolerance implicitly depends on `closure['dt']`.

(In contrast, my alternate proposal of `|fields| ≤ tol * max|fields|` has a `tol` that is independent of the sampling frequency.)

One alternative would be to check `change <= tol * maxchange`, where `change = np.abs(previous_fields-current_fields)` and `maxchange` is the maximum observed change, so that `tol` becomes independent of the sampling frequency.

smartalecH: Why does `decay_by` implicitly depend on `closure['dt']`? We are only looking at a predetermined set of frequencies. So long as the frequencies are below the Nyquist limit, the sample rate shouldn't affect the decay tolerance, right?

> One alternative would be to check change <= tol * maxchange, where change = np.abs(previous_fields-current_fields) and maxchange is the maximum observed change over the simulation, so that tol becomes independent of the sampling frequency.

Aren't we already normalizing out any dependence? How different is the current implementation:
```relative_change = np.abs(previous_fields-current_fields) /  previous_fields ```

from your proposal:
```change = np.abs(previous_fields-current_fields) /  maxchange ```

The issue with `maxchange` is now you have to compute a Δ over every spatial point, which requires *caching* all of the dft fields.

File: python/simulation.py [hunk 7]
```diff
@@ -4425,6 +4425,41 @@ def _stop(sim):
...
+    """
+    # Record data in closure so that we can persitently edit
+    closure = {'previous_fields':1, 't0':0, 'dt':0}
+    def _stop(_sim):
+        if _sim.fields.t == 0:
+            closure['dt'] = max(int(1/_sim.fields.dft_maxfreq()/_sim.fields.dt),int(_sim.fields.min_decimation()))
+        if _sim.fields.t <= closure['dt'] + closure['t0']:
+            return False
+        elif maximum_run_time and _sim.round_time() > maximum_run_time:
+            return True
```
oskooi, line 4447: Shouldn't the second if-branch which checks if the `maximum_run_time` has been exceeded come *before* the first if-branch which checks that a time interval of `dt` has elapsed?

python/adjoint/optimization_problem.py
```diff
@@ -130,7 +130,7 @@ def __init__(
                     2))  # index of center frequency
 
         self.decay_by = decay_by
-        self.decay_fields = decay_fields
+        self.decay_fields = decay_fields # left for legacy, but doesn't do anything
```
oskooi, line 133: Probably better to just remove this line here and also from `test_adjoint_solver.py` where it is also used?

python/simulation.py
stevengj: No, by `maxchange` I mean the maximum value of `change` observed over all the previous sampling *times*.

python/adjoint/optimization_problem.py
smartalecH: Fixed.

python/simulation.py
smartalecH: Fixed.

python/simulation.py
smartalecH: Fixed.

python/simulation.py
smartalecH: Fixed.

python/simulation.py
smartalecH: Fixed.

src/dft.cpp
smartalecH: Fixed.

smartalecH:
Should be ready to merge once all tests pass.

python/simulation.py
```diff
@@ -4425,6 +4425,44 @@ def _stop(sim):
...
+    def _stop(_sim):
+        if _sim.fields.t == 0:
+            closure['dt'] = max(1/_sim.fields.dft_maxfreq()/_sim.fields.dt,_sim.fields.min_decimation())
+        if maximum_run_time and _sim.round_time() > maximum_run_time:
+            return True
+        elif _sim.fields.t <= closure['dt'] + closure['t0']:
+            return False
+        else:
+            previous_fields = closure['previous_fields']
+            current_fields  = _sim.fields.dft_norm2()
```
stevengj, line 4450:
```suggestion
            current_fields  = _sim.fields.dft_norm()
```

python/simulation.py
```diff
@@ -4425,6 +4425,44 @@ def _stop(sim):
...
+    or `until_after_sources` parameter, that checks all of the Simulation's dft objects
+    every `dt` timesteps, and stops the simulation once all the components and frequencies
+    of *every* dft object have decayed by at least some tolerance, `tol`. The routine 
+    calculates the optimal `dt` to use based on the frequency content in the DFT monitors.
+
+    Optionally the user can specify a minimum run time (`minimum_run_time`) or a maximum
+    run time (`maximum_run_time`).
+    """
+    # Record data in closure so that we can persitently edit
+    closure = {'previous_fields':0, 't0':0, 'dt':0, 'maxchange':1}
```
stevengj, line 4440:
```suggestion
    closure = {'previous_fields':0, 't0':0, 'dt':0, 'maxchange':0}
```

python/simulation.py
```diff
@@ -4425,6 +4425,44 @@ def _stop(sim):
...
+            if previous_fields == 0:
+                closure['previous_fields'] = current_fields
+                return False
+            
+            closure['previous_fields'] = current_fields
+            closure['t0'] = _sim.fields.t
+            if mp.verbosity > 0:
+                fmt = "DFT decay(t = {0:1.1f}): {1:0.4e}"
+                print(fmt.format(_sim.meep_time(), np.real(change/closure['maxchange'])))
+            return (change/closure['maxchange']) <= tol and _sim.round_time() >= minimum_run_time
```
stevengj, line 4463:
```suggestion
            return (change <= closure['maxchange'] * tol and _sim.round_time() >= minimum_run_time
```

stevengj: (to handle case where `maxchange == 0`)

smartalecH:
Finally passing tests and ready to merge

python/adjoint/optimization_problem.py
```diff
@@ -69,12 +69,10 @@ def __init__(
         fcen=None,
         df=None,
         nf=None,
-        decay_dt=50,
-        decay_fields=[mp.Ez],
-        decay_by=1e-6,
+        decay_by=1e-11,
```
oskooi, line 72: Why does the default tolerance need to be reduced by five orders of magnitude? Won't such a small default value potentially lead to longer runtimes than are actually necessary particularly since the discretization errors are likely several orders of magnitude larger?

stevengj: The default tolerance now has a completely different meaning than before, so I'm not sure whether this actually corresponds to an increase in runtime…

smartalecH: Exactly. This PR actually *reduces* the runtime from `master`, since it's able to check so frequently. But the tolerance corresponds to a completely different convergence check, which is why it's value is so different than before. 

Essentially the default tolerance is chosen to make the tests pass. But the runtime is still minimized (and less than `master`).

PR #1745, opened 2021-08-28
set the default number of threads to one for multi-threading

oskooi:
Following the suggestion from @stevengj in [#1628 (comment)](https://github.com/NanoComp/meep/pull/1628#issuecomment-906827214),  this PR defines the default number of threads to one in the `fields` class constructor. The documentation is also updated.


src/fields.cpp
```diff
@@ -33,6 +37,10 @@ fields::fields(structure *s, double m, double beta, bool zero_fields_near_cylori
                int loop_tile_base)
     : S(s->S), gv(s->gv), user_volume(s->user_volume), v(s->v), m(m), beta(beta),
       working_on(&times_spent) {
+#ifdef _OPENMP
+  if (getenv("OMP_NUM_THREADS") == NULL)
+    omp_set_num_threads(1);
+#endif
```
stevengj, line 43: I would rather have this in the meep initialization in mympi.cpp

Unfortunately that initialization is not called from Python, but we could refactor some stuff into a separate initialization routine that we call from the Python front-end when it is loaded.

stevengj: In particular, here:
https://github.com/NanoComp/meep/blob/e6e9f9ca621b9df0e33ae4e2ee1c65b1ac73a9cd/src/mympi.cpp#L185-L211

stevengj: e.g. add a new function called `meep::setup()` that does this threads initialization (and potentially any other one-time initialization we want to do from both C++ and Python), call it from `initialize::initialize()`, and also call it from Python when Meep is loaded.

oskooi: I added a new function `meep::setup()` which includes the threads initialization and is called from `initialize::initialize()` of mympi.cpp. However, where is the best place to call `meep::setup()` from Python; perhaps before the MPI initialization of meep.i?

https://github.com/NanoComp/meep/blob/e2739e4a44000441f58cd61c326f426d3faf63b4/python/meep.i#L1835-L1853

stevengj: I would put in the same place where we call `set_zero_subnormals`

stevengj (approved):

Issue #1570, opened 2021-05-13
Adjoint code in cylindrical coordinates


mochen4:
It seems that adjoint code currently doesn't really support the cylindrical coordinates. For example, I get the error message ``Invalid fields::loop_in_chunks grid type ex for dimensions Cylindrical`` from https://github.com/NanoComp/meep/blob/8e2b0a050fd7e2d6d39019f76cdd3abcb73888c5/python/adjoint/optimization_problem.py#L194 in the forward run with the following setup. We probably need the different set of components for cylindrical coordinates.

```
...
resolution = 30

sr = 6
sz = 3
dimensions = mp.CYLINDRICAL
cell_size = mp.Vector3(sr, 0, sz)
pml_layers = [mp.PML(1.0)]
fcen = 1/1.55
width = 0.25
fwidth = width * fcen
source_center  = [3,0,-1]
source_size    = mp.Vector3(1,0,0)
src = mp.GaussianSource(frequency=fcen,fwidth=fwidth)
source = [mp.Source(src,component=mp.Ez,
                    center=source_center)]


design_region_resolution = 10
Nx = design_region_resolution
Ny = design_region_resolution
Si = mp.Medium(index=3.4)
SiO2 = mp.Medium(index=1.44)

design_variables = mp.MaterialGrid(mp.Vector3(Nx,Ny),SiO2,Si)
design_region = mpa.DesignRegion(design_variables,volume=mp.Volume(center=mp.Vector3(3,0,0), size=mp.Vector3(1, 0,1)))
geometry = [mp.Block(center=design_region.center, size=design_region.size, material=design_variables)]

sim = mp.Simulation(cell_size=cell_size, boundary_layers=pml_layers, geometry=geometry, sources=source,
                    eps_averaging=False, resolution=resolution, dimensions=dimensions,
                    m=2)


far_x = [mp.Vector3(0,0,30)]
NearRegions = [mp.Near2FarRegion(center=mp.Vector3(3,0 ,1),size=mp.Vector3(x=2), weight=+1)]
FarFields = mpa.Near2FarFields(sim, NearRegions ,far_x)
ob_list = [FarFields]

def J(alpha):
    return npa.abs(alpha[0,0,2])**2

opt = mpa.OptimizationProblem(
    simulation=sim,
    objective_functions=J,
    objective_arguments=ob_list,
    design_regions=[design_region],
    fcen=fcen,
    df = 0,
    nf = 1,
    decay_fields=[mp.Ez],
[...]

PR #1749, opened 2021-08-31
support adjoint gradients in cylindrical coordinate

mochen4:
Support for adjoint gradients in cylindrical coordinate. Solve https://github.com/NanoComp/meep/issues/1570.


stevengj:
Would be good to have a test, otherwise looks good!

smartalecH:
It seems that *every* cylindrical test case must run to the maximum runtime. In other words, none of the DFT fields are ever "converging"...

We should probably look into that.

PR #1754, opened 2021-09-07
WIP: Meepgeom refactoring

### LINKED ISSUE #1649: subpixel smoothing gradients
Now that material grids have support for subpixel smoothing, it would be nice to enable gradient support.

Luckily, this should be rather straightforward given the current infrastructure. Since the [current recombination step ](https://github.com/NanoComp/meep/blob/76fab2b0aa0f65efb762e9944434997ab5c6d418/src/meepgeom.cpp#L2560)uses a finite difference to compute dA/dp, we just need to make sure that the finite difference routine also calls the new subpixel averaging routine (in this [function](https://github.com/NanoComp/meep/blob/76fab2b0aa0f65efb762e9944434997ab5c6d418/src/meepgeom.cpp#L2531)).

We also need to store the adjoint D fields, rather than the E fields.

smartalecH:
Closes #1649.


stevengj:
Might be good to make a PR *just* for the refactoring, and then a separate PR for the gradient.

smartalecH:
Here's a summary of the proposed changes:
* Added new functions `make_geom_epsilon()`, `init_libctl()`, and`set_materials_from_geom_epsilon()`.
*  Moved `geom_epsilon` class to `meepgeom.hpp` (previously local to the `meepgeom.cpp` only) so that I could pass around `geom_epsilon` objects in the overloaded `set_materials_from_geom_epsilon()` and  `make_geom_epsilon()`.
*  In order to move things to the `meepgeom` header file, I had to modify the name of the symmetric matrix routines so they wouldn't clash with `mpb.cpp`, which imports `meepgeom.hpp` and `mpb.h`, both of which have symmetric matrix functions (I couldn't get the namespaces to work properly).

All of *my* code compiles, but something weird is going on with the swig generated code -- particularly with the new `set_materials_from_geom_epsilon()` and `make_geom_epsilon()` functions.

Before I spend too much time debugging, can I get some feedback as to whether this is the best way to refactor things (I'm sure it's not...)?

stevengj:
This looks fine.

smartalecH:
Almost all of the tests are passing now.

There is an issue with symmetric prism objects and subpixel smoothing. For example, two of the python tests (`test_mode_coeffs.py` and `test_bend_flux.py`) use a prism object to draw a simple, straight waveguide. Ideally, the waveguide should look like this:

image

Instead, it looks like this:

image

Notice that `x>0` looks different than `x<0`. In fact, it seems that subpixel smoothing is not working for `x<0` (for this particular prism object). If I turn subpixel smoothing off, I get:

image

So we know the issue is with subpixel smoothing *and* Prism objects. But not *all* prism objects, because the other prism object in the tests (a simple bend) looks fine:

image

Any thoughts?

smartalecH:
Oof seems like the above issue was due to an outdated version of `libctl` on my machine. 

Should be good for review/merging now.

stevengj:
LGTM.

PR #1759, opened 2021-09-18
markdown documentation for stop_when_dft_decayed

oskooi:
Markdown documentation for `stop_when_dft_decayed` (which was missing from #1740).

### DIFF
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -2691,12 +2691,12 @@ def stop_when_fields_decayed(dt=None, c=None, pt=None, decay_by=None):
 <div class="function_docstring" markdown="1">
 
 Return a `condition` function, suitable for passing to `Simulation.run` as the `until`
-or `until_after_sources` parameter, that examines the component `c` (e.g. `Ex`, etc.)
+or `until_after_sources` parameter, that examines the component `c` (e.g. `meep.Ex`, etc.)
 at the point `pt` (a `Vector3`) and keeps running until its absolute value *squared*
 has decayed by at least `decay_by` from its maximum previous value. In particular, it
 keeps incrementing the run time by `dt` (in Meep units) and checks the maximum value
 over that time period &mdash; in this way, it won't be fooled just because the field
-happens to go through 0 at some instant.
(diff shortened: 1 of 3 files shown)


doc/docs/Python_User_Interface.md
```diff
@@ -2706,6 +2706,26 @@ slow group velocities and are absorbed poorly by [PML](Perfectly_Matched_Layer.m
...
+                      minimum_run_time=0,
+                      maximum_run_time=None):
+```
+
+<div class="function_docstring" markdown="1">
+
+Return a `condition` function, suitable for passing to `Simulation.run` as the `until`
+or `until_after_sources` parameter, that checks the `Simulation`'s dft objects every `dt`
+timesteps, and stops the simulation once all the field components and frequencies of *every*
+dft object have decayed by at least some tolerance `tol` (no default value). The time interval
```
stevengj, line 2722: Shouldn’t we give a default tol, e.g. the tol from the tests?

oskooi: A default value would be useful but since `tol` is no longer related to the actual precision of the fields, it's not clear what a useful value should be. #1740 added a default value of `1e-11` to the `OptimizationProblem` class of the adjoint solver but that should probably also be removed and defined instead as part of the unit tests:

https://github.com/NanoComp/meep/blob/81f2081ca81d0b6f919dfc0d1c9465445a076441/python/adjoint/optimization_problem.py#L72

oskooi: Note that `stop_when_fields_decayed` also has no default value for its `decay_by` parameter.

smartalecH: Wouldn't a "useful value" be one for which the tests pass? We can run more tests... but at the end of the day it's a user parameter.

oskooi: I ran some experiments with the adjoint solver on a couple of different test problems and it seems that a default value of 1e-11 is reasonable (i.e., the gradient can be computed with 5-6 digits of precision at modest grid resolutions). I made the change to the default value but It would be useful at some point to add more guidance on the appropriate choice of this parameter.

PR #1760, opened 2021-09-19
additional documentation for MaterialGrid

oskooi:
Additional documentation for the `MaterialGrid` to mention that nonuniform grids are supported including a new figure showing the actual location of the grid points for a 2d example.

Also includes a fix for an off-by-one bug in the unit test `test_material_grid.py` in which the resolution of the grid did not exactly match what was specified.


python/tests/test_material_grid.py
```diff
@@ -18,8 +18,8 @@ def compute_resonant_mode(res,default_mat=False):
 
         design_shape = mp.Vector3(1,1,0)
         design_region_resolution = 50
-        Nx = int(design_region_resolution*design_shape.x)
-        Ny = int(design_region_resolution*design_shape.y)
+        Nx = int(design_region_resolution*design_shape.x)+1
```
smartalecH, line 21: Why are we adding 1 here? I realize it "fixes" a failing test, but this doesn't make sense to me.

stevengj: The first and last pixels of the material grid lie on the boundaries of the grid — that means the number of "pixels" in that direction is actually N-1 where N is the number of grid points.

oskooi: A `desigion_region_resolution` of `50` and `design_shape` of `(1,1,0`) correspond to a Δx and Δx of `0.2`. Without adding one to `Nx` and `Ny` (the current setup),  Δx and Δx are `0.020408` which is slightly off. Another solution is to use `x = np.linspace(-0.5*design_shape.x,0.5*design_shape.x,Nx,endpoint=False)`, etc.

smartalecH: Ah yes that makes sense, thanks!

smartalecH (approved):

oskooi:
For clarity, I added a comment to `test_material_grid.py` to explain why the number of grid points (`Nx` and `Ny`) of the `MaterialGrid` are chosen the way they are.

oskooi:
> Maybe the figure should show "Nx=5" and "Ny=4"

Done.

PR #1763, opened 2021-09-22
Issue# 1762 - Add an option to build SWIG python bindings with threads enabled

kkg4theweb:
Add a configure option to enable SWIG python bindings with threads enabled.
- Fix all custom binding code that use the PyAPI to be GIL safe
- Make the Py3.6 + MPI version of the CI test to enable this option for testing.


python/meep.i
```diff
@@ -1410,6 +1449,9 @@ void _get_gradient(PyObject *grad, PyObject *fields_a, PyObject *fields_f, PyObj
 }
 
 %exception {
+#ifdef MEEP_SWIG_PYTHON_DEBUG
```
stevengj, line 1452: Is it worth having a `MEEP_SWIG_PYTHON_DEBUG` just for this `printf`, vs. just adding a `printf` when you need it?  What subsequent developer is likely to need a `printf` in `_get_gradient`?

oskooi (approved):

ahoenselaar (approved):

python/meep.i
kkg4theweb: I am happy to remove it if you think it is just noise.
I found this useful to track down which SWIG wrapped methods were taking a lot of time etc.

python/meep.i
```diff
@@ -1410,6 +1449,9 @@ void _get_gradient(PyObject *grad, PyObject *fields_a, PyObject *fields_f, PyObj
...
+  printf("**SWIG**: $symname\n");
```
stevengj, line 1453:
```suggestion
  master_printf("**SWIG**: $symname\n");
```

PR #1769, opened 2021-09-27
fix off-by-one bug in dimensions of fixed-resolution MaterialGrid for adjoint-solver functions

oskooi:
Fixes several instances of a bug involving the dimensions of a `MaterialGrid` with fixed resolution. This bug was originally reported in #1760.

Also, the resolution used in the unit test `test_adjoint_solver.py` is slightly increased in order to reduce the hard-coded tolerances used to compare values.

### DIFF
--- a/python/adjoint/basis.py
+++ b/python/adjoint/basis.py
@@ -50,7 +50,7 @@ class Basis(ABC):
 
 
 class BilinearInterpolationBasis(Basis):
-    ''' 
+    '''
     Simple bilinear interpolation basis set.
     '''
     def __init__(self, resolution, symmetry=None, **kwargs):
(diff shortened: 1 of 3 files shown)


python/adjoint/basis.py
```diff
@@ -65,26 +65,26 @@ def __init__(self, resolution, symmetry=None, **kwargs):
             self.symmetry = symmetry
 
         if mp.X in set(self.symmetry):
-            self.Nx = int(resolution * self.volume.size.x / 2)
+            self.Nx = int((int(resolution * self.volume.size.x) + 1) / 2)
```
stevengj, line 68: Shouldn't we add 1 after dividing by 2, not before?

oskooi: Fixed.

smartalecH: Note that `basis.py` isn't used anywhere -- we just kept it for legacy purposes I think.

smartalecH (approved):
LGTM.

PR #1775, opened 2021-09-29
Implementation of chunk balancer feature

bencbartlett:
CONTEXT: This feature uses empirical data about the timings of each MPI process to adjust the chunk sizes to achieve optimal load balancing for repeated simulations, such as iterations in an optimization run. Processes which are consistently slow and overworked will have their chunk sizes reduced for the next iteration of an optimization run, and underworked processes will handle larger chunks. This approach has the advantage that it implicitly incorporates variable load between different machines running the MPI processes.

SCOPE:
- Added abstract class for chunk balancer
- Added DefaultChunkBalancer implementation, which adjusts chunk sizes according to per-process working time (ignoring all-all comms) while maintaining previous split directions
- Wrote unit tests for DefaultChunkBalancer to check for improvement in load-balancing, convergence to a load-balanced state, and that split_pos values are adjusted correctly for each iteration.
- Wrote unit tests for the MockSimulation class used by the DefaultChunkBalancerTests
- Added a binary_partition_utils.py library which includes lots of tree traversal algorithms which are useful to have for the chunk balancer
- Wrote unit tests for binary_partition_utils.py


doc/docs/Parallel_Meep.md
```diff
@@ -169,3 +169,105 @@ See also [FAQ/Should I expect linear speedup from the parallel Meep](FAQ.md#shou
...
+
+sim = mp.Simulation(..., chunk_layout=initial_chunk_layout)
+sim.init_sim()
+
+for iteration in range(epochs):
+  sim.run(...)
+  
+  # Adjust the chunk layout if needed
+  chunk_balancer.adjust_chunk_layout(sim, sensitivity=0.4)
+```
```
stevengj, line 234: Would be good to add an example of doing separate runs (i.e. load chunk layout from a file, run simulation, adjust_chunk_layout, and save).

(Might also be useful in that context to add methods to save/load the chunk layout, i.e. a binarypartition, from a file, without requiring the user to specifically call pickle etc.)

doc/docs/Parallel_Meep.md
```diff
@@ -169,3 +169,105 @@ See also [FAQ/Should I expect linear speedup from the parallel Meep](FAQ.md#shou
...
+### Chunk balancer interface
+
+The chunk balancer interface provides four main methods:
+- `make_initial_chunk_layout()` generates the initial chunk layout for the first iteration of a simulation. By default, `None` is returned, indicating Meep should use its default chunk partitioning strategy.
+- `should_adjust_chunk_layout()` decides whether the current layout is sufficiently poorly balanced to justify the up-front cost of reallocating the field arrays when changing chunk layouts.
+- `compute_new_chunk_layout()` looks at the current chunk layout and per-process timing data to compute a new chunk layout which has chunk sizes adjusted to improve the load balance across compute nodes.
+- `adjust_chunk_layout()` is syntactic sugar which will compute a new chunk layout, apply it to the simulation object, reset the simulation, and re-initialize the simulation.
+
+```py
+class ChunkyMonkey(abc.ABC):
```
stevengj, line 192: Maybe `AbstractChunkBalancer` would be clearer.

doc/docs/Parallel_Meep.md
```diff
@@ -169,3 +169,105 @@ See also [FAQ/Should I expect linear speedup from the parallel Meep](FAQ.md#shou
...
+  def adjust_chunk_layout(self, sim, **kwargs) -> None:
+    """Computes a new chunk layout and applies it to sim."""
+```
+
+### Usage
+
+Using the chunk balancer is very straightforward, and it can typically be integrated into existing Meep simulations with only a few lines of code. Here is a simple example:
+
+```py
+from meep.chunk_balancer import DefaultChunkBalancer
```
stevengj, line 219: maybe just `ChunkBalancer`

doc/docs/Parallel_Meep.md
bencbartlett: If we are making improvements to `BinaryPartition`, should we consider moving the `BinaryPartition` class declaration to its own file, now that we also have a `binary_partition_utils` module? It has always seemed strange to to me that it is crammed in at the end of `simulation.py`

oskooi: The `BinaryPartition` class is only used to define the `chunk_layout` of the `Simulation` object which is why its class definition is inside `simulation.py` along with various other `Simulation` related classes. I don't think it makes much sense in this PR anyways to move it to a separate file because it would probably require reorganizing the other classes as well.

doc/docs/Parallel_Meep.md
```diff
@@ -169,3 +169,136 @@ See also [FAQ/Should I expect linear speedup from the parallel Meep](FAQ.md#shou
...
+
+```py
+import meep as mp
+from meep.chunk_balancer import ChunkBalancer
+from meep.timing_measurements import MeepTimingMeasurements
+import pickle
+
+# Fetch chunk layout from a previous run
+with open("path/to/chunk_layout.pkl", "rb") as f:
+    chunk_layout = pickle.load(f)
```
stevengj, line 245: Should set `chunk_layout = None` if the file doesn't exist.

bencbartlett: Done

stevengj:
Failing tests?

bencbartlett:
Sorry about that, made a minor change that broke a test on `test_timing_measurements.py`. All fixed now.

PR #1780, opened 2021-10-06
Subpixel smoothing gradients

### LINKED ISSUE #1649: subpixel smoothing gradients

smartalecH:
Closes #1649

TODO
- [x] Modify tests to include subpixel smoothing
- [x] Quantify the effect of smoothing filter on gradient accuracy
- [x] Quantify the effect of Aᵤ finite-difference step size on gradient accuracy
- [x] Sweep β and check results
- [x] Compare gradients to naive method

### DIFF
--- a/python/adjoint/optimization_problem.py
+++ b/python/adjoint/optimization_problem.py
@@ -32,7 +32,7 @@ class OptimizationProblem(object):
         decimation_factor=0,
         minimum_run_time=0,
         maximum_run_time=None,
-        finite_difference_step=1e-6
+        finite_difference_step=utils._FD_DEFAULT
     ):
 
         self.sim = simulation
--- a/python/adjoint/optimization_problem.py
+++ b/python/adjoint/optimization_problem.py
@@ -32,7 +32,7 @@ class OptimizationProblem(object):
         decimation_factor=0,
         minimum_run_time=0,
         maximum_run_time=None,
-        finite_difference_step=utils._FD_DEFAULT
+        finite_difference_step=utils.FD_DEFAULT
     ):
 
         self.sim = simulation
--- a/python/adjoint/optimization_problem.py
+++ b/python/adjoint/optimization_problem.py
@@ -31,7 +31,8 @@ class OptimizationProblem(object):
         decay_by=1e-11,
         decimation_factor=0,
         minimum_run_time=0,
-        maximum_run_time=None
+        maximum_run_time=None,
+        finite_difference_step=1e-6
     ):
 
         self.sim = simulation
@@ -87,6 +88,7 @@ class OptimizationProblem(object):
         self.decimation_factor = decimation_factor
         self.minimum_run_time = minimum_run_time
         self.maximum_run_time = maximum_run_time
+        self.finite_difference_step = finite_difference_step # step size used in Aᵤ computation
 
         # store sources for finite difference estimations
         self.forward_sources = self.sim.sources
@@ -265,6 +267,7 @@ class OptimizationProblem(object):
                 self.D_a[ar][dri],
                 self.D_f[dri],
                 self.frequencies,
+                self.finite_difference_step
             ) for dri, dr in enumerate(self.design_regions)
         ] for ar in range(len(self.objective_functions))]
 
--- a/python/adjoint/optimization_problem.py
+++ b/python/adjoint/optimization_problem.py
@@ -99,11 +99,11 @@ class OptimizationProblem(object):
 
         self.gradient = []
 
-    def __call__(self, rho_vector=None, need_value=True, need_gradient=True):
+    def __call__(self, rho_vector=None, need_value=True, need_gradient=True, beta=None):
         """Evaluate value and/or gradient of objective function.
         """
         if rho_vector:
-            self.update_design(rho_vector=rho_vector)
+            self.update_design(rho_vector=rho_vector, beta=beta)
 
         # Run forward run if requested
         if need_value and self.current_state == "INIT":
--- a/python/adjoint/utils.py
+++ b/python/adjoint/utils.py
@@ -38,7 +38,7 @@ class DesignRegion(object):
     def update_beta(self,beta):
         self.design_parameters.beta=beta
 
-    def get_gradient(self, sim, fields_a, fields_f, frequencies):
+    def get_gradient(self, sim, fields_a, fields_f, frequencies, finite_difference_step):
         num_freqs = onp.array(frequencies).size
         shapes = []
         '''We have the option to linear scale the gradients up front
@@ -82,7 +82,8 @@ class DesignRegion(object):
         vol = sim._fit_volume_to_simulation(self.volume)
         
         # compute the gradient
-        mp._get_gradient(grad,scalegrad,fields_a,fields_f,sim.gv,vol.swigobj,onp.array(frequencies),sim.geps,shapes)
+        mp._get_gradient(grad,scalegrad,fields_a,fields_f,
+        sim.gv,vol.swigobj,onp.array(frequencies),sim.geps,shapes,finite_difference_step)
         return onp.squeeze(grad).T
 
 def _check_if_cylindrical(sim):
--- a/python/adjoint/utils.py
+++ b/python/adjoint/utils.py
@@ -19,7 +19,7 @@ _PHI_AXIS = 3
 _Z_AXIS = 1
 
 # default finite difference step size when calculating Aᵤ
-_FD_DEFAULT = 1e-3
+FD_DEFAULT = 1e-3
 
 class DesignRegion(object):
     def __init__(
@@ -108,7 +108,7 @@ def calculate_vjps(
     adj_fields: List[List[onp.ndarray]],
     design_variable_shapes: List[Tuple[int, ...]],
     sum_freq_partials: bool = True,
-    finite_difference_step: float = _FD_DEFAULT
+    finite_difference_step: float = FD_DEFAULT
 ) -> List[onp.ndarray]:
     """Calculates the VJP for a given set of forward and adjoint fields."""
     vjps = [
--- a/python/adjoint/utils.py
+++ b/python/adjoint/utils.py
@@ -105,6 +105,7 @@ def calculate_vjps(
     adj_fields: List[List[onp.ndarray]],
     design_variable_shapes: List[Tuple[int, ...]],
     sum_freq_partials: bool = True,
+    finite_difference_step: float = 1e-6
 ) -> List[onp.ndarray]:
     """Calculates the VJP for a given set of forward and adjoint fields."""
     vjps = [
@@ -113,6 +114,7 @@ def calculate_vjps(
             adj_fields[i],
             fwd_fields[i],
             frequencies,
+            finite_difference_step,
         ) for i, design_region in enumerate(design_regions)
     ]
     if sum_freq_partials:
--- a/python/adjoint/utils.py
+++ b/python/adjoint/utils.py
@@ -18,6 +18,9 @@ _RHO_AXIS = 2
 _PHI_AXIS = 3
 _Z_AXIS = 1
 
+# default finite difference step size when calculating Aᵤ
+_FD_DEFAULT = 1e-3
+
 class DesignRegion(object):
     def __init__(
             self,
@@ -105,7 +108,7 @@ def calculate_vjps(
     adj_fields: List[List[onp.ndarray]],
     design_variable_shapes: List[Tuple[int, ...]],
     sum_freq_partials: bool = True,
-    finite_difference_step: float = 1e-6
+    finite_difference_step: float = _FD_DEFAULT
 ) -> List[onp.ndarray]:
     """Calculates the VJP for a given set of forward and adjoint fields."""
     vjps = [
--- a/src/meepgeom.cpp
+++ b/src/meepgeom.cpp
@@ -2695,8 +2695,8 @@ Past versions of the code only calculated dA/dε using a finite
 difference, and then multiplied by the analytic vJp (dε/du).
 With the addition of subpixel smoothing, however, the vJp became
 much more complicated and it is easier to calculate the entire gradient
-using finite differences (at the cost of slightly less gradients due to 
-roundoff).
+using finite differences (at the cost of slightly less accurate gradients 
+due to roundoff).
  */
 void add_interpolate_weights(double rx, double ry, double rz,
                              double *data, int nx, int ny, int nz, int stride,
--- a/src/meepgeom.cpp
+++ b/src/meepgeom.cpp
@@ -2620,8 +2620,10 @@ std::complex<double> get_material_gradient(
     std::complex<double> fields_f,    // forward field at current point
     double freq,                      // frequency
     geom_epsilon *geps,               // material
-    meep::grid_volume &gv,             // simulation grid volume
-    double du                         // step size
+    meep::grid_volume &gv,            // simulation grid volume
+    double du,                        // step size
+    double *u,                        // matgrid
+    int idx                           // matgrid index
 ) {
   /*Compute the Aᵤx product from the -λᵀAᵤx calculation.
   The current adjoint (λ) field component (adjoint_c)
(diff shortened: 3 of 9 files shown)


smartalecH:
After completely refactoring the way material grid gradients are computing, there is still an issue when subpixel smoothing is turned on. I plotted the relative error in the gradient `abs((adjoint_grad - finite_diff_grad)/finite_diff_grad)` as a function of β. As can be seen, the error accumulates with increasing β (for the case where subpixel smoothing is turned on):

beta_sweep

(note the non-smoothing case also sees a rise in relative error after a certain point because the actual gradient is going to zero, as expected)

smartalecH:
Okay, I *think* I see the (dominate) issue.

Before subpixel smoothing, were were sampling our forward and adjoint fields *directly* on the yee grid (bypassing the interpolation to voxel centers that typically takes place). During the recombination step, we would loop over these points directly *one by one* (e.g. each component individually) and calculate the gradient contribution of each component *independent* of any other component. This approach is nice because we only have to worry about *one* restriction step, and is perfectly valid if the material contains a perfectly diagonal permittivity tensor. This is because the field components are totally isolated by the dA/du inner product (which is just the tensor in our case).

Unfortunately, we cannot make this assumption when subpixel smoothing is enabled, as off-diagonal terms appear. We can't just use a simple weighted inner product on the relevant field components either, because the field components (and tensor components) are all stored on *different* parts of the Yee cell. 

If we use all the field components located at the *center* of the voxel, we can do our weighted inner product. This of course requires a second interpolation step. So, in the backpropagation step, we need to perform *two* restriction steps: one that accounts for the u-grid to the permittivity grid, and one that accounts for the field components interpolating onto the voxel centers (these need to occur in reverse order). We also need to refactor a lot of the adjoint code that stores the fields, but that part seems pretty straightforward.

Here's a simple illustration:

image

Since backpropagation is a *backward* pass, we want to do the field restriction first, and then the u-grid restriction after that. The algorithm is actually rather simple:
```c++
LOOP_OVER_VOXEL_CENTERS(volume_describing_material_grid)
{
    u_scale =  E_a * dA_du * E_f; // at voxel center -- this scales all subsequent gradients (chain rule)
    LOOP_OVER_E_COMPONENTS_IN_VOXEL(v)
   { 
         //use the existing restriction code to propagate gradients to each u using postion v
   }
}
```

Notice I declared fictitious loop macros `LOOP_OVER_VOXEL_CENTERS` and `LOOP_OVER_E_COMPONENTS_IN_VOXEL`. I'm hoping we already have some macros that can at least do most if not all of the intended loop. If so, it will make the implementation very easy.

src/meepgeom.cpp
```diff
@@ -2521,63 +2583,47 @@ void get_material_tensor(const medium_struct *mm, double freq,
...
-  else if (field_dir == meep::Ez)
-    dir_idx = 2;
-  else
-    meep::abort("Invalid adjoint field component");
-
-  std::complex<double> result = fields_a * dA_du[3 * dir_idx + dir_idx] * fields_f;
+  symm_matrix chi1_1, chi1_2, dA_du;
+
+  geps->u_p = du;                   // perturbation term in one direction
+  get_chi1_tensor(r, geps, &chi1_1);
```
oskooi, line 2608: There is a bit of a complication here that is being overlooked. At the Yee grid point for e.g. *E<sub>x</sub>* (where the forward and adjoint fields for this component are computed by Meep), the row from the tensor ∂ε/∂ρ containing the effective ε (in this case containing the elements [ε<sub>xx</sub>, ε<sub>xy</sub>, ε<sub>xz</sub>]) is multiplied by the column vector of the adjoint fields [*E<sup>a</sup><sub>x</sub>*, *E<sup>a</sup><sub>y</sub>*, *E<sup>a</sup><sub>z</sub>*] to obtain "one side of the inner product." The complication, of course, is that *E<sub>y</sub>* and *E<sub>z</sub>* are not defined at the Yee grid point for *E<sub>x</sub>* which means they have to be interpolated to that point using the nearest grid values.

In the way that it is currently set up, there does not seem to be any such interpolation which means that an error is being introduced into the gradient at this step of the computation.

smartalecH:
Thanks for looking into this. Note that the forward and adjoint fields are no longer calculated on the Yee grid, but rather interpolated to the center. This way we can perform the full inner product (as described above). 

So all components are on the center of the voxel. Of course, we have to restrict out to compensate for this.

oskooi:
> Unfortunately, due to the smoothing filter (which is a requirement for our approach) we will always have interfaces requiring anisotropic entries.

By planar multilayer structure in 2d, I was referring to something like this [OLED](http://www.simpetus.com/projects.html#meep_oled). After applying filtering and subpixel smoothing, the ε tensor at the interface voxels should still be diagonal.

oskooi:
For reference, posting results here for an investigation into the impact of subpixel smoothing of a `MaterialGrid` on a continuously varying geometry. The test case is identical to the ring-resonator example from the [user manual](https://meep.readthedocs.io/en/latest/Subpixel_Smoothing/#continuously-varying-shapes-and-results) which involves two overlapping `Cylinder` objects to represent the ring geometry. The study involves computing the frequency of a low-loss resonant mode while varying the inner radius of the ring. The scripts for the simulation and post-processing the results are [here](https://gist.github.com/oskooi/c637f94aecbd84bf31d202789a1b940d). The `MaterialGrid` is using β=1000 and η=0.5 for the projection operator of the level-set function.

For a `MaterialGrid` at a resolution of 1000 and Meep grid resolution of 10 (i.e., downsampling by a factor of 10), the results for subpixel smoothing are comparable to no smoothing: while the overall frequency response is continuous, there are still some discontinuities/jumps apparent in the data.

ring_matgrid_res10_desres1000_freq_vs_radius

When the Meep grid resolution is doubled to 20 , the continuous frequency response is more apparent. However, the same trend is observed for the no-smoothing case. This is perhaps not surprising since no-smoothing is using a bilinear interpolation of the grid weights onto the Yee grid.

ring_matgrid_res20_desres1000_freq_vs_radius

For this particular test case and setup, subpixel smoothing does not seem to be providing much of a benefit compared to no smoothing. Again, this is not unexpected because no-smoothing is using an interpolation scheme which mitigates stairstepping errors in the first place. 

(One thing to try would be to increase the `MaterialGrid` resolution (to e.g. 2000 instead of 1000) in order to reduce errors in the downsampling and mitigate discontinuities for the case involving a Meep grid resolution of 10. Unfortunately the memory requirements become impractical: the 2d array used to store the grid weights is more than 20 Gb.)

smartalecH:
For some reason, the python 3.9 w/MPI build is failing tests, but the python 3.6 w/MPI passes.

@ianwilliamson are you aware of any obvious reasons why this may be the case? Any major changes to round-off-sensitive operations that may be degrading the accuracy of the gradients?

smartalecH:
The failing test is the same failing test described [here](https://github.com/NanoComp/meep/pull/1780#issuecomment-967188269).

For reference, here are the values comparing the adjoint gradients with the finite differences:

```
Directional derivative -- adjoint solver: [-4.94904819e-06 -3.02577147e-04 -2.97732191e-04], FD: [-4.75876167e-07 -2.98877141e-04 -3.12129430e-04]
```
The finite-difference gradients seem to be off (again).

oskooi:
Note that the failure occurs only when compiling using *single* precision. Maybe you just need to increase the spacing you used to compute the finite difference gradients?

smartalecH:
> Maybe you just need to increase the spacing you used to compute the finite-difference gradients?

The finite differences I'm referring to are defined by the test (i.e. I shouldn't have to modify the test at all). The *adjoint* gradients don't appear to have any issues.

It's possible that even before this PR, the finite difference gradients were pretty far off, just not enough to kill the test. This PR does introduce a little bit of noise to the adjoint sensitivities, which might "knock" it out of the predefined tolerance (even though the bulk of the error was in the finite-difference calculation). I'll take a look at that possibility.

ianwilliamson:
> Clearly something is off with the finite-difference calculation here...

It seems like the largest difference between the finite difference calculations is in that first element of the list. Even under Python 3.7, the difference between the adjoint calculation and the finite difference calculation is greater than 30% in that element.

smartalecH:
I think the real issue here is the way we are referencing our finite-difference calculation as the ground truth. The adjoint gradients are consistent from single-precision, to double-precision, python 3.6 to python 3.9, etc. But the finite differences swing quite a bit.

This is somewhat expected, as our finite differences are single-sided (first-order accurate) and the step size is arbitrary. What we *should* do is some sort of Richardson extrapolation. But tests are passing now.

python/adjoint/optimization_problem.py
```diff
@@ -31,7 +31,8 @@ def __init__(
         decay_by=1e-11,
         decimation_factor=0,
         minimum_run_time=0,
-        maximum_run_time=None
+        maximum_run_time=None,
+        finite_difference_step=1e-6
```
oskooi, line 35: Shouldn't this step size used to compute the finite-difference term Aᵤ during the recombination step be dependent on the floating-point precision of the fields similar to the step size of the finite-difference gradient? A hard-coded value of `1e-6` could be susceptible to rounding errors for single-precision fields.

oskooi:
Follow up to an earlier [thread](https://github.com/NanoComp/meep/pull/1780#issuecomment-945348014) investigating the effect of subpixel smoothing on continuously shifting boundaries of a ring resonator represented as a `MaterialGrid`. Rather than use a binarized image, applying a [Gaussian filter](https://gist.github.com/oskooi/c637f94aecbd84bf31d202789a1b940d#gistcomment-3963609) (not a signed-distance function as suggested by @stevengj) to smear the boundaries, the convergence results are not much different than before.

The inset schematic below labeled "filtered" shows the actual smoothed contour of the ring resonator adjacent to an image of the binarized representation. The "exact" result (shown in the plot as the black line) is computed at a resolution of 100 using no smoothing. What is unexpected in these results is that subpixel smoothing still does not produce a frequency response which varies continuously with continuous changes in the ring geometry. 

ring_filter_vs_binarized

smartalecH:
What filter radius are you using? Can you provide more details?

python/adjoint/optimization_problem.py
smartalecH: Does the single-precision flag also force the permittivity to single precision? I thought only the fields were single precision. If my assumption is correct, then the finite difference term behaves the same in both configurations.

oskooi: Note that the `chi1inv` member of the `structure_chunk` class has type `realnum`:

https://github.com/NanoComp/meep/blob/bc58782c0d16e6d59937028cbe2ffa402f5f24b0/src/meep.hpp#L583

Hence, along with the fields of the `fields_chunk` class the permittivity is also compiled using single-precision floating point.

oskooi:
> What filter radius are you using? Can you provide more details?

The `sigma` value of [`ndimage.gaussian_filter`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.gaussian_filter.html) used to generate these results is 1.0 (i.e., comparable to the voxel dimensions). The Meep grid resolution is 20 and the `MaterialGrid` resolution is 400. The actual code is [here](https://gist.github.com/oskooi/c637f94aecbd84bf31d202789a1b940d#gistcomment-3963609).

smartalecH:
There are a few nuances I should point out.

First, I don't think we are comparing the right things. Even when `do_averaging` is off, we *still* have a levelset, because the projection is happening after the linear interpolation. What we really want to compare is the new method to the "classical" technique, where projection happens *before* interpolation. This means `beta=0` in the `MaterialGrid` object, and we use the `tanh_projection()` function as we do in the tutorials.

Second, I think your filter radius is too small. Remember, we have a first-order Taylor expansion representing our level-set. A radius of 1 pixel (which is a bit hard to quantify using a gaussian filter...) seems too small to ensure that approximation holds. I recommend you use the in-house conic filter (as the radius is very straightforward there) and try a few different lengths (e.g. 2 pixels, 4, 8, etc.). There should be a point where it reaches second-order convergence... As we discussed before, the "pixel" unit we care about is w.r.t. the `MaterialGrid`, not the meep grid.

Third, we really should be fitting the underlying design parameters (for both our new method and the old method) to the final geometry you want at the end of the day (for a given filter radius and `beta`). Depending on your loss function, this can be cast as a *convex* problem (akin to logistic regression). But by doing the fitting, we ensure we really are comparing the right things.

Also, is the label on your x-axis right? Isn't your meep unit a micron? How are you specifying resolutions between 1 and 2 pixels per meep unit?

Finally, let's make sure we are clear about what we are trying to test with each example. Are we looking for second-order convergence? Are we making sure it behaves like a level-set?

python/adjoint/optimization_problem.py
smartalecH: True, but `eff_chi1inv_row` (which is what the recombination routine uses) actually asumes that `chi1inv_row` is always a `double`:

https://github.com/NanoComp/meep/blob/bc58782c0d16e6d59937028cbe2ffa402f5f24b0/src/meepgeom.cpp#L1014-L1015

Either way, the step size doesn't need to be so small. I reduced it to be safe.

smartalecH:
The gradients seem to be accurate now. Note that the current implementation is a brute-force, inefficient approach using finite differences at each material grid point to compute Aᵤ. This introduces a bit more "noise" in the gradient calculation than before, but it's still much better than what was reported before #1801. As far as speed is concerned, the calculation is still plenty fast. Future PRs could try to implement the true, analytic vJp if needs be.

I ran some tests to check various aspects of this PR. All of the following tests use a FP-like structure that looks like this:
image

All of the tests can be found in this [repo](https://github.com/smartalecH/subpixel_smoothing/tree/main/tests).

First, we can look at the gradient accuracy as β increases when `do_averaging` is `true` or `false`:
beta_sweep_new_30

The accuracy starts to "decrease" with β, but it's still within a workable range. Plus, it could be that the finite-difference "ground truth" measurements are also inaccurate. But compared to what's shown [above](https://github.com/NanoComp/meep/pull/1780#issuecomment-936932210), these results seem to indicate a successful implementation.

Next, I swept the finite difference step size used to estimate Aᵤ. As expected, we start to saturate after the step size reaches a certain point:
step_sweep

As mentioned [here](https://github.com/NanoComp/meep/pull/1780#discussion_r750353617) I dropped the default step size to `1e-3`.

Next, I swept the filter radius (using the same underlying design parameters) at `β=1e6`: 
radius_sweep

Note that the gradient accuracy doesn't depend on the filter radius. Rather, only the subpixel smoothing feature depends on the filter radius.

Finally, to (once again) illustrate the level-set behavior of our hybrid method, I measured the L2 norm of the *gradient* when using the classical density-based scheme and our new hybrid scheme:

beta_mag_sweep

As expected, the classical gradients *go to zero* as β increases, whereas the new hybrid method's gradients converge to a *finite value*. This effectively enables discrete shape optimization.

The tests are passing, and this PR should be ready for merging after review.

src/meepgeom.cpp
```diff
@@ -2664,45 +2666,60 @@ std::complex<double> get_material_gradient(
...
-   function in fields.cpp) to data ... this has to be changed if
-   linear_interpolate is changed!! ...also multiply by scaleby
-   etc. for different gradient types */
-
+/* A brute force approach to calculating Aᵤ using finite differences.
+Past versions of the code only calculated dA/dε using a finite
+difference, and then multiplied by the analytic vJp (dε/du).
+With the addition of subpixel smoothing, however, the vJp became
+much more complicated and it is easier to calculate the entire gradient
+using finite differences (at the cost of slightly less gradients due to
```
ianwilliamson, line 2698: "slightly less accurate gradients" you mean?

ianwilliamson (commented):
Not many comments from me.

src/meepgeom.cpp
smartalecH: Fixed

python/adjoint/utils.py
```diff
@@ -18,6 +18,9 @@
 _PHI_AXIS = 3
 _Z_AXIS = 1
 
+# default finite difference step size when calculating Aᵤ
+_FD_DEFAULT = 1e-3
```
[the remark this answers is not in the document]
smartalecH: Fixed

PR #1790, opened 2021-10-15
shorten unit test for adjoint solver

oskooi:
Currently, the unit test for the adjoint solver takes more seven minutes to run on my machine with two processes which is far longer than any of the other 50+ unit tests in the Python suite. To shorten this unit test without sacrificing coverage, we can remove the tests involving `k_point` since the boundaries are irrelevant (the test structure is a 2d waveguide surrounded on all sides by PML) and complex fields (testing real fields is sufficient). This reduces the runtime to about a minute and a half.

### DIFF
--- a/python/tests/test_adjoint_solver.py
+++ b/python/tests/test_adjoint_solver.py
@@ -20,7 +20,8 @@ sxy = 5.0
 cell_size = mp.Vector3(sxy,sxy,0)
 
 dpml = 1.0
-boundary_layers = [mp.PML(thickness=dpml)]
+pml_xy = [mp.PML(thickness=dpml)]
+pml_x = [mp.PML(thickness=dpml,direction=mp.X)]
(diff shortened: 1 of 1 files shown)


stevengj:
I think we want to test at least one case with complex fields, since we previously had a bug (off-by-2) for the adjoint solver in that case.

It would be good to have a better k-point test, because the adjoint solver needs to flip the sign of k and we should test that it is doing this.

cc @mochen4

oskooi:
A k-point test has been added involving a 2d cell with PML in the *x* direction, a square `MaterialGrid` with random weights at the cell center, and an objective function consisting of a DFT fields monitor at a single point adjacent to the `MaterialGrid` on the side opposite of a point-dipole source. There are separate tests for the single- and multi-frequency case. 

This is ready to be merged.

PR #1791, opened 2021-10-15
Increase BUFSIZE when gathering DFT field data

ahoenselaar:
A 16MB allocation is negligible and reduces the number of all_to_all communication iterations required to perform the consolidation.


stevengj:
I'm not opposed to a larger buffer here, but I'm not completely clear on why it matters — isn't `process_dft_component` only something you are doing when you fetch a DFT fields array in Python, which is typically only done at the end of the simulation?  Is that ever performance-critical?

ahoenselaar:
Users who want more control and define their own convergence criteria may
fetch the DFT fields during the simulation as well.

PR #1796, opened 2021-10-21
use maximum rather than minimum decimation factor for DFT convergence criteria

oskooi:
The `stop_when_dft_decayed` convergence-criteria function introduced in #1740 uses a measurement interval (in number of timesteps) which is the maximum of two quantities: (1) the period of the largest frequency or (2) the smallest DFT decimation factor. However, (2) is incorrect because it could mean that not all the DFT fields have been updated before checking for convergence. The correct approach would be to use the largest DFT decimation factor.

### DIFF
--- a/src/dft.cpp
+++ b/src/dft.cpp
@@ -319,7 +319,7 @@ double dft_chunk::norm2() const {
 // return the maximum decimation factor across
 // all dft regions
 int fields::max_decimation() const {
-  int maxdec = std::numeric_limits<int>::min();
+  int maxdec = 1;
   for (int i = 0; i < num_chunks; i++)
     if (chunks[i]->is_mine())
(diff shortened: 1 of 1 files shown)


src/dft.cpp
```diff
@@ -316,21 +316,21 @@ double dft_chunk::norm2() const {
   return sum;
 }
 
-// return the minimum decimation factor across
+// return the maximum decimation factor across
 // all dft regions
-int fields::min_decimation() const {
-  int mindec = std::numeric_limits<int>::max();
+int fields::max_decimation() const {
+  int maxdec = std::numeric_limits<int>::min();
```
stevengj, line 322:
```suggestion
  int maxdec = 1;
```
We never want to return a number less than 1 here.

oskooi: Fixed.

PR #1801, opened 2021-10-26
Anisotropic material gradient support

### LINKED ISSUE #1794: Proper `update_fields` interpolation scheme
The subpixel smoothing paper describes the update equation using the following figure:

image

which is interpolation method 26(c) in the original [Werner and Cary paper](https://doi.org/10.1016/j.jcp.2007.05.008).

This method performs an interpolation, then the `chi1inv` multiplication, followed by another interpolation step. What's important is that this particular `chi1inv` component does *not* seem to be located on a typical grid point that it's originally stored (i.e. in the figure, the fields are interpolated to a "node" that doesn't store ε⁻¹ₓy).

The code, however, seems to just grab two ε points, one from a "leading corner" for each side that is being interpolated:

https://github.com/NanoComp/meep/blob/b6166b1b976eea03b9763505280f650f4aed20d1/src/step_generic.cpp#L381-L382

Note that `u[i]` aligns with `g[i]` and `u[i + s]` aligns with `u[i + s]`, leaving `g[i-sx]` and `g[i-isx+s]` "unbalanced". (Note that multiplying each point by it's corresponding `chi1inv` component is a different interpolation scheme altogether and not always stable).

These two operations don't seem to be the same to me, but maybe I'm missing something with how `chi1inv` is stored (I thought an entire row of the tensor was stored at each corresponding yee point).

smartalecH:
Closes #1794 

This needs to work before #1780 (since subpixel smoothing induces anisotropic tensor elements).

Essentially, this PR *properly* recombines the forward and adjoint fields when calculating the gradients, regardless of what the material looks like (e.g. with or without off-diagonal elements). To do so, we store the `D` fields rather than the `E` fields, which means we look at `chi1inv` rather than `chi1`. 

The process is straightforward:
1. the adjoint fields are *restricted* to the respective epsilon nodes
2. the forward fields are *interpolated* to the same nodes
3. the two are "recombined" using the gradient of the epsilon inverse tensor at that point (i.e. a weighted inner product).

This is the *proper* interpolation/restriction scheme, as this is the gradient of what's actually implemented in the code.

As a nice side effect, this makes the gradients *significantly* more accurate (a few orders of magnitude when comparing relative errors for some initial tests).

TODO:
- [x] Fix merge conflicts
- [x] Fix failing tests
- [x] Renable support for dispersive/~conductive~ media (I turned this off in order to simplify the code while debugging)
- [x] Add test with material containing off-diagonal components (e.g. rotated sapphire)
- [x] Cleanup

NOTE: conductive media will require a separate PR (see #1811)


smartalecH:
Should now be ready for review/merging.

@ianwilliamson, I ended up refactoring the two APIs quite a bit, such that all of the major calls are within `utils.py` (per your suggestion). I also removed a few redundant routines that have since moved to `simulation.py` (specifically the DFT convergence test, which is now much more efficient thanks to #1740). There's still more that could be done in the future (see #1805) but that's better suited for another PR -- I only refactored the relevant pieces to this PR in order to simplify reviewing. Let me know if you have any other suggestions.

As described in #1811, the gradients are still incorrect for materials containing conductivities. Resolving that is best suited for another PR since the issue existed before this PR.

src/meepgeom.cpp
```diff
@@ -2707,99 +2799,181 @@ void material_grids_addgradient_point(double *v, std::complex<double> fields_a,
...
+            //operate on the second eps node
+            fwd1_idx = get_idx_from_ivec(gv.dim, start_ivec, the_stride, fwd_pa);
+            fwd1 = 0.5 * GET_FIELDS(fields_f,ci_forward,f_i,fwd1_idx);
+            fwd2_idx = get_idx_from_ivec(gv.dim, start_ivec, the_stride, fwd_paf);
+            fwd2 = 0.5 * GET_FIELDS(fields_f,ci_forward,f_i,fwd2_idx);
+            fwd_avg = fwd1 + fwd2;
+            meep::vec eps2 = gv[ieps2];
+            cyl_scale = (gv.dim == meep::Dcyl) ? eps2.r() : 1;
+            prod = 0.5*adj*get_material_gradient(eps2, adjoint_c, forward_c, fwd_avg, frequencies[f_i], geps, 1e-6);
+            material_grids_addgradient_point(v+ng*f_i, vec_to_vector3(eps2), scalegrad*prod.real()*cyl_scale, geps);
```
oskooi, line 2959: There is redundant code here that can be separated into a function and called twice with different arguments `(fwd_p, fwd_pf, eps1)` and `(fwd_pa, fwd_paf, eps2)`.

python/adjoint/utils.py
```diff
@@ -3,14 +3,87 @@
...
+            returns a singleton element for the forward and adjoint fields.
+            This only occurs when we are in 2D and only working with a particular
+            polarization (as the other fields are never stored). Our get_gradient
+            algorithm, however, requires we pass an array of zeros with the
+            proper shape as the design_region.            
+            '''
+            s = sim.get_array_slice_dimensions(c, vol=self.volume)[0]
+            if (onp.squeeze(fields_a[ci][0,...]).size == 1):
+                fields_a[ci] = onp.zeros(onp.insert(s,0,num_freqs))
+                fields_f[ci] = onp.zeros(onp.insert(s,0,num_freqs))
```
ianwilliamson, line 57: It would be more readable to do `onp.zeros((num_freqs,) + s)`. Also, would help to use a more descriptive name than `s`... maybe `spatial_shape` or something similar?

python/adjoint/utils.py
```diff
@@ -3,14 +3,87 @@
...
+    
+    def update_beta(self,beta):
+        self.design_parameters.beta=beta
+
+    def get_gradient(self, sim, fields_a, fields_f, frequencies):
+        num_freqs = onp.array(frequencies).size
+        shapes = []
+        com = _compute_components(sim)
+        scalegrad = 1
+        for ci, c in enumerate(com):
```
ianwilliamson, line 46: I suggest `for component_idx, component` for better readability

python/adjoint/utils.py
```diff
@@ -3,14 +3,87 @@
...
+        self.design_parameters.update_weights(design_parameters)
+    
+    def update_beta(self,beta):
+        self.design_parameters.beta=beta
+
+    def get_gradient(self, sim, fields_a, fields_f, frequencies):
+        num_freqs = onp.array(frequencies).size
+        shapes = []
+        com = _compute_components(sim)
+        scalegrad = 1
```
ianwilliamson, line 45: Can you add a comment as to why `scalegrad` is needed? If it's hardcoded to 1 here, can it just be removed?

python/adjoint/utils.py
```diff
@@ -3,14 +3,87 @@
...
+        for ci, c in enumerate(com):
+            '''We need to correct for the rare cases that get_dft_array
+            returns a singleton element for the forward and adjoint fields.
+            This only occurs when we are in 2D and only working with a particular
+            polarization (as the other fields are never stored). Our get_gradient
+            algorithm, however, requires we pass an array of zeros with the
+            proper shape as the design_region.            
+            '''
+            s = sim.get_array_slice_dimensions(c, vol=self.volume)[0]
+            if (onp.squeeze(fields_a[ci][0,...]).size == 1):
```
ianwilliamson, line 55: Why do you need the squeeze? Can't you just check the size?

python/adjoint/utils.py
```diff
@@ -3,14 +3,87 @@
...
+    def update_design_parameters(self, design_parameters):
+        self.design_parameters.update_weights(design_parameters)
+    
+    def update_beta(self,beta):
+        self.design_parameters.beta=beta
+
+    def get_gradient(self, sim, fields_a, fields_f, frequencies):
+        num_freqs = onp.array(frequencies).size
+        shapes = []
+        com = _compute_components(sim)
```
ianwilliamson, line 44: Maybe just move the call to `_compute_components()` inside `enumerate(...)`

src/meepgeom.cpp
```diff
@@ -2530,66 +2586,103 @@ void get_material_tensor(const medium_struct *mm, double freq,
...
     meep::abort("Invalid adjoint field component");
 
-  std::complex<double> result = fields_a * dA_du[3 * dir_idx + dir_idx] * fields_f;
-  return result.real();
+  if (md->trivial){
+    const double sd = 1.0; // FIXME: make user-changable?
+    meep::volume v(r);
+    LOOP_OVER_DIRECTIONS(dim, d){
+      v.set_direction_min(d, r.in_direction(d)-sd/2);
+      v.set_direction_max(d, r.in_direction(d)+sd/2);
```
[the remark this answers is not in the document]
smartalecH: I was just copying what's performed during initialization:

https://github.com/NanoComp/meep/blob/3eabcef8c90049e08449bbe39f5f876956f3875e/src/anisotropic_averaging.cpp#L204

Am I misunderstanding how this is being used?

src/meepgeom.cpp
smartalecH: I thought so too... but everything worth refactoring is already refactored in `get_material_gradient` and `material_grids_addgradient_point`.

Pulling it out further just made it more unreadable (and requires a lot more parameters to pass than just those 3).

`prod` could technically be placed within `material_grids_addgradient_point`. `cyl_scale` is repeated, but again, refactoring actually makes it harder to read.

python/adjoint/utils.py
smartalecH: > maybe spatial_shape or something similar?

Good idea

> onp.zeros((num_freqs,) + s)

unfortunately `s` is a numpy array, and can't be joined like python lists. This seems pretty standard when prepending to numpy arrays, from what I've seen (there are hundreds of possibilities of course...)

src/meepgeom.cpp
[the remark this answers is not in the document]
smartalecH: Ah I think I see. The code also uses `gv.dv()` to do some scaling under the hood. Got it.

python/adjoint/utils.py
```diff
@@ -3,14 +3,87 @@
...
+        scalegrad = 1
+        for ci, c in enumerate(com):
+            '''We need to correct for the rare cases that get_dft_array
+            returns a singleton element for the forward and adjoint fields.
+            This only occurs when we are in 2D and only working with a particular
+            polarization (as the other fields are never stored). Our get_gradient
```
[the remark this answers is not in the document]
smartalecH: Fixed.

src/meepgeom.cpp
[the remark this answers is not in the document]
smartalecH: Fixed.

python/adjoint/optimization_problem.py
```diff
@@ -254,16 +191,8 @@ def forward_run(self):
...
-            np.zeros((self.nf, c[0], c[1], c[2]), dtype=np.complex128)
-            for c in dg
-        ] for dg in self.design_grids]
-        for nb, dgm in enumerate(self.design_region_monitors):
-            for ic, c in enumerate(self.components):
-                for f in range(self.nf):
-                    self.d_E[nb][ic][f, :, :, :] = atleast_3d(
-                        self.sim.get_dft_array(dgm, c, f))
+        # Store forward fields for each set of design variables in array
+        self.d_E = utils.gather_design_region_fields(self.sim,self.design_region_monitors,self.frequencies)
```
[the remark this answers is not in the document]
smartalecH: Fixed.

python/adjoint/optimization_problem.py
```diff
@@ -318,25 +244,17 @@ def adjoint_run(self):
...
-                    for f in range(self.nf):
-                        if (self.sim.is_cylindrical or self.sim.dimensions == mp.CYLINDRICAL):
-                            # Addtional factor of 2 for cyldrical coordinate
-                            self.a_E[ar][nb][ic][f, :, :, :] = 2 * atleast_3d(self.sim.get_dft_array(dgm, c, f))
-                        else:
-                            self.a_E[ar][nb][ic][f, :, :, :] = atleast_3d(self.sim.get_dft_array(dgm, c, f))
-
-        if self.sim.is_cylindrical or self.sim.dimensions == mp.CYLINDRICAL:
+            # Store adjoint fields for each design set of design variables
+            self.a_E.append(utils.gather_design_region_fields(self.sim,self.design_region_monitors,self.frequencies))
```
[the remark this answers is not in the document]
smartalecH: Fixed.

python/adjoint/utils.py
smartalecH: Fixed.

python/adjoint/utils.py
smartalecH: Fixed.

python/adjoint/utils.py
smartalecH: Fixed.

python/adjoint/utils.py
smartalecH: Fixed.

Issue #1793, opened 2021-10-16
inconsistent DFT fields monitors of the adjoint solver


oskooi:
Currently, the `EigenmodeCoefficient` class of the adjoint solver defines the DFT field monitors **on the Yee grid**:

https://github.com/NanoComp/meep/blob/434f3fbf0ffe2334ce99cf105c6ec94ed44c95b7/python/adjoint/objective.py#L154-L162

However, the `FourierFields` and `Near2FarFields` classes define their DFT field monitors **at the voxel centers**:

https://github.com/NanoComp/meep/blob/434f3fbf0ffe2334ce99cf105c6ec94ed44c95b7/python/adjoint/objective.py#L239-L248

https://github.com/NanoComp/meep/blob/434f3fbf0ffe2334ce99cf105c6ec94ed44c95b7/python/adjoint/objective.py#L340-L347

Note: for `Near2FarFields`, no `yee_grid` option is defined but the default is the centered grid (`bool use_centered_grid = true`): 

https://github.com/NanoComp/meep/blob/434f3fbf0ffe2334ce99cf105c6ec94ed44c95b7/src/meep.hpp#L1961-L1966

When defining an objective function as part of an adjoint simulation, the user therefore needs to be consistent with these definitions. This is similar to the set up used in the unit test for the adjoint solver where the `yee_grid` is explicitly defined:

https://github.com/NanoComp/meep/blob/434f3fbf0ffe2334ce99cf105c6ec94ed44c95b7/python/tests/test_adjoint_solver.py#L79-L91

Ensuring this type of consistency (particularly for the mode coefficients which does not use the default option) can be problematic and a source of error for the average user. It would be useful if the DFT fields monitors for all objective classes were *always* defined at the voxel centers.

stevengj:
The basic idea here is that the centered grid is much easier for the user to deal with than the Yee grid.  And `FourierFields` are designed to be used directly by the user.  e.g. it would be pretty hard for the user to compute ExH if all the field components are at different points in space.

However, it would be good to have the option to get FourierFields directly on the Yee grid, e.g. the version that @mawc2019 is working on.

Whereas the DFT fields for the eigenmode coefficients and the near2far transformations are mainly for use by Meep, and Meep knows how to handle Yee-grid data.

In principle, however, we could do the DFT accumulation on the Yee grid and only convert to the centered grid when we output to the user.

oskooi:
[...]

Issue #1676, opened 2021-07-15
Simulation fields are NaN or Inf in the adjoint solver


mawc2019:
In the adjoint solver, the simulation sometimes stops in the adjoint run with the error message "RuntimeError: meep: simulation fields are NaN or Inf". I saw it happened in Near2FarFields and FourierFields, but I did not see it in EigenmodeCoefficient. An example for producing such an error in Near2FarFields is as follows, which resembles the [tutorial example](https://github.com/NanoComp/meep/blob/master/python/examples/adjoint_optimization/Near2Far-Optimization-with-Epigraph-Formulation.ipynb).

```
import meep as mp
import meep.adjoint as mpa
import numpy as np
from autograd import numpy as npa
Si = mp.Medium(index=3.4)
SiO2 = mp.Medium(index=1.44)

resolution = 30
design_region_resolution = resolution

design_region_width,design_region_height = 6,2
pml_size = 1.0
Sx,Sy = design_region_width,2*pml_size+design_region_height+5
cell_size = mp.Vector3(Sx,Sy)
pml_layers = [mp.PML(pml_size, direction=mp.Y)]

fcen,fwidth = 1/1.55,0.01/1.55
source_center,source_size  = mp.Vector3(0,-(design_region_height/2 + 1.5)),mp.Vector3(Sx,0)

src = mp.GaussianSource(frequency=fcen,fwidth=fwidth)
source = [mp.Source(src,component=mp.Ez,center=source_center,size=source_size)]
Nx,Ny = int(design_region_resolution*design_region_width),int(design_region_resolution*design_region_height)

design_variables = mp.MaterialGrid(mp.Vector3(Nx,Ny),SiO2,Si,grid_type='U_MEAN')
design_region = mpa.DesignRegion(design_variables,volume=mp.Volume(center=mp.Vector3(), size=mp.Vector3(design_region_width, design_region_height, 0)))

geometry = [mp.Block(center=design_region.center, size=design_region.size, material=design_variables)]

sim = mp.Simulation(
    cell_size=cell_size,boundary_layers=pml_layers,k_point=mp.Vector3(),geometry=geometry,
    sources=source,default_material=SiO2,resolution=resolution)

far_x = [mp.Vector3(0,40,0)]
NearRegions = [mp.Near2FarRegion(center=mp.Vector3(0,design_region_height/2+1.5), size=mp.Vector3(design_region_width+1.5,0), weight=+1)]
[...]

PR #1821, opened 2021-11-08
place fourier adjoint sources collectively for each frequency

mawc2019:
The main features of this PR are
1. The Fourier adjoint sources at each frequency are placed collectively, while in the current master branch the adjoint sources are placed point by point.
2. The Fourier adjoint solver has 'yee_grid' as a user specified argument, which can be `True` or `False`. The default option is `False`.

This PR passes all the built-in tests except test_adjoint_jax.py, test_binary_partition_utils.py, and test_chunk_balancer.py, as the current master branch does.

### DIFF
--- a/python/adjoint/objective.py
+++ b/python/adjoint/objective.py
@@ -250,12 +250,21 @@ class FourierFields(ObjectiveQuantity):
 
         mon_size = self.sim.fields.dft_monitor_size(self._monitor.swigobj, self.volume.swigobj, self.component)
         dJ = dJ.astype(np.complex128)
-        if np.prod(mon_size)*self.num_freq == dJ.size: # The objective function J is a scalar.
-            dJ = dJ.flatten()
-        elif np.prod(mon_size)*self.num_freq**2 == dJ.size: # The objective function J is a vector. Each component corresponds to a frequency.
-            dJ = np.sum(dJ,axis=1).flatten()
-        else:
+        if np.prod(mon_size)*self.num_freq != dJ.size and np.prod(mon_size)*self.num_freq**2 != dJ.size:
             raise ValueError('The format of J is incorrect!')
+
+        # The objective function J is a vector. Each component corresponds to a frequency.
+        if np.prod(mon_size)*self.num_freq**2 == dJ.size and self.num_freq > 1:
+            dJ = np.sum(dJ,axis=1)
+        '''The adjoint solver requires the objective function
+        to be scalar valued with regard to objective arguments
+        and position, but the function may be vector valued
+        with regard to frequency. In this case, the Jacobian
+        will be of the form [F,F,...] where F is the number of
+        frequencies. Because of linearity, we can sum across the
+        second frequency dimension to calculate a frequency
+        scale factor for each point (rather than a scale vector).
+        '''
 
         self.all_fouriersrcdata = self._monitor.swigobj.fourier_sourcedata(self.volume.swigobj, self.component, self.sim.fields, dJ)
 
@@ -302,7 +311,7 @@ class Near2FarFields(ObjectiveQuantity):
         sources = []
         if dJ.ndim == 4:
             dJ = np.sum(dJ, axis=0)
-        dJ = dJ.flatten()
+
         farpt_list = np.array([list(pi) for pi in self.far_pts]).flatten()
         far_pt0 = self.far_pts[0]
         far_pt_vec = py_v3_to_vec(
--- a/python/adjoint/objective.py
+++ b/python/adjoint/objective.py
@@ -99,6 +99,13 @@ class ObjectiveQuantity(abc.ABC):
             scale *= 2
         return scale
 
+    def _dft_monitor_size(self,min_max_corners):
+        """Evaluates the size of the dft monitor for FourierFields."""
+        thickness_indicator = self.sim.fields.indicate_thick_dims(self.volume.swigobj)
+        raw_size = (min_max_corners[3:6]-min_max_corners[0:3])/2+np.ones(3)
+        true_size = np.array([raw_size[i]*(thickness_indicator[i]==1)+(thickness_indicator[i]==0) for i in range(3)],dtype=int)
+        return true_size
+
     def _create_time_profile(self, fwidth_frac=0.1):
         """Creates a time domain waveform for normalizing the adjoint source(s).
 
@@ -248,32 +255,29 @@ class FourierFields(ObjectiveQuantity):
         time_src = self._create_time_profile()
         sources = []
 
-        thickness_indicator = self.sim.fields.indicate_thick_dims(self.volume.swigobj)
         min_max_corners = self.sim.fields.get_corners(self._monitor.swigobj, self.component) # ivec values of the corners
-        raw_size = (min_max_corners[3:6]-min_max_corners[0:3])/2+np.ones(3)
-        true_size = np.array([raw_size[i]*(thickness_indicator[i]==1)+(thickness_indicator[i]==0) for i in range(3)],dtype=int) # monitor size
+        mon_size = self._dft_monitor_size(min_max_corners)
 
-        if np.prod(true_size)*self.num_freq == dJ.size: # The objective function J is a scalar.
+        if np.prod(mon_size)*self.num_freq == dJ.size: # The objective function J is a scalar.
             dJ = dJ.flatten()
-        elif np.prod(true_size)*self.num_freq**2 == dJ.size: # The objective function J is a vector. Each component corresponds to a frequency.
+        elif np.prod(mon_size)*self.num_freq**2 == dJ.size: # The objective function J is a vector. Each component corresponds to a frequency.
             dJ = np.sum(dJ,axis=1).flatten()
         else:
             raise ValueError('The format of J is incorrect!')
 
         self.all_fouriersrcdata = self._monitor.swigobj.fourier_sourcedata(self.volume.swigobj, min_max_corners, dJ)
 
-        for near_data in self.all_fouriersrcdata:
-            amp_arr = np.array(near_data.amp_arr).reshape(-1, self.num_freq)
+        for fourier_data in self.all_fouriersrcdata:
+            amp_arr = np.array(fourier_data.amp_arr).reshape(-1, self.num_freq)
             scale = amp_arr * self._adj_src_scale(include_resolution=False)
             
             if self.num_freq == 1:
-                sources += [mp.IndexedSource(time_src, near_data, scale[:,0])]
+                sources += [mp.IndexedSource(time_src, fourier_data, scale[:,0])]
             else:
                 src = FilteredSource(time_src.frequency,self._frequencies,scale,self.sim.fields.dt)
                 (num_basis, num_pts) = src.nodes.shape
                 for basis_i in range(num_basis):
-                    sources += [mp.IndexedSource(src.time_src_bf[basis_i], near_data, src.nodes[basis_i])]
-
+                    sources += [mp.IndexedSource(src.time_src_bf[basis_i], fourier_data, src.nodes[basis_i])]
         return sources
 
     def __call__(self):
@@ -340,7 +344,6 @@ class Near2FarFields(ObjectiveQuantity):
                             src.nodes[basis_i],
                         )
                     ]
-
         return sources
 
     def __call__(self):
--- a/python/adjoint/objective.py
+++ b/python/adjoint/objective.py
@@ -231,98 +231,52 @@ class EigenmodeCoefficient(ObjectiveQuantity):
 
 
 class FourierFields(ObjectiveQuantity):
-    def __init__(self, sim, volume, component, decimation_factor=0):
+    def __init__(self, sim, volume, component, yee_grid=False, decimation_factor=0):
         super().__init__(sim)
-        self.volume = volume
+        self.volume = sim._fit_volume_to_simulation(volume)
         self.component = component
+        self.yee_grid = yee_grid
         self.decimation_factor = decimation_factor
 
     def register_monitors(self, frequencies):
         self._frequencies = np.asarray(frequencies)
         self._monitor = self.sim.add_dft_fields(
-            [self.component],
-            self._frequencies,
-            where=self.volume,
-            yee_grid=False,
-            decimation_factor=self.decimation_factor,
-        )
+            [self.component], self._frequencies, where=self.volume, yee_grid=self.yee_grid, decimation_factor=self.decimation_factor)
         return self._monitor
 
     def place_adjoint_source(self, dJ):
         time_src = self._create_time_profile()
         sources = []
-        scale = self._adj_src_scale()
 
-        x_dim, y_dim, z_dim = len(self._dg.x), len(self._dg.y), len(self._dg.z)
+        thickness_indicator = self.sim.fields.indicate_thick_dims(self.volume.swigobj)
+        min_max_corners = self.sim.fields.get_corners(self._monitor.swigobj, self.component) # ivec values of the corners
+        raw_size = (min_max_corners[3:6]-min_max_corners[0:3])/2+np.ones(3)
+        true_size = np.array([raw_size[i]*(thickness_indicator[i]==1)+(thickness_indicator[i]==0) for i in range(3)],dtype=int) # monitor size
 
-        if self.num_freq == 1:
-            amp = -dJ[0].copy().reshape(x_dim, y_dim, z_dim) * scale
-            if self.component in [mp.Hx, mp.Hy, mp.Hz]:
-                amp = -amp
-            for zi in range(z_dim):
-                for yi in range(y_dim):
-                    for xi in range(x_dim):
-                        if amp[xi, yi, zi] != 0:
-                            sources += [
-                                mp.Source(
-                                    time_src,
-                                    component=self.component,
-                                    amplitude=amp[xi, yi, zi],
-                                    center=mp.Vector3(self._dg.x[xi],
-                                                      self._dg.y[yi],
-                                                      self._dg.z[zi]),
-                                )
-                            ]
+        if np.prod(true_size)*self.num_freq == dJ.size: # The objective function J is a scalar.
+            dJ = dJ.flatten()
+        elif np.prod(true_size)*self.num_freq**2 == dJ.size: # The objective function J is a vector. Each component corresponds to a frequency.
+            dJ = np.sum(dJ,axis=1).flatten()
         else:
-            '''The adjoint solver requires the objective function
-            to be scalar valued with regard to objective arguments
-            and position, but the function may be vector valued
-            with regard to frequency. In this case, the Jacobian
-            will be of the form [F,F,...] where F is the number of
-            frequencies. Because of linearity, we can sum across the
-            second frequency dimension to calculate a frequency
-            scale factor for each point (rather than a scale vector).
-            '''
-            dJ = np.sum(
-                dJ,
-                axis=1)  # sum along first dimension bc Jacobian is always diag
-            dJ_4d = np.array([
-                dJ[f].copy().reshape(x_dim, y_dim, z_dim)
-                for f in range(self.num_freq)
-            ])
-            if self.component in [mp.Hx, mp.Hy, mp.Hz]:
-                dJ_4d = -dJ_4d
-            for zi in range(z_dim):
-                for yi in range(y_dim):
-                    for xi in range(x_dim):
-                        '''We only need to add a current source if the
-                        jacobian is nonzero for all frequencies at
-                        that particular point. Otherwise, the fitting
-                        algorithm is going to fail.
-                        '''
-                        if not np.all((dJ_4d[:, xi, yi, zi] == 0)):
-                            final_scale = -dJ_4d[:, xi, yi, zi] * scale
-                            src = FilteredSource(
-                                time_src.frequency,
-                                self._frequencies,
-                                final_scale,
-                                self.sim.fields.dt,
-                            )
-                            sources += [
-                                mp.Source(
-                                    src,
-                                    component=self.component,
-                                    amplitude=1,
-                                    center=mp.Vector3(self._dg.x[xi],
-                                                      self._dg.y[yi],
-                                                      self._dg.z[zi]),
-                                )
-                            ]
+            raise ValueError('The format of J is incorrect!')
+
+        self.all_fouriersrcdata = self._monitor.swigobj.fourier_sourcedata(self.volume.swigobj, min_max_corners, dJ)
+
+        for near_data in self.all_fouriersrcdata:
+            amp_arr = np.array(near_data.amp_arr).reshape(-1, self.num_freq)
+            scale = amp_arr * self._adj_src_scale(include_resolution=False)
+            
+            if self.num_freq == 1:
+                sources += [mp.IndexedSource(time_src, near_data, scale[:,0])]
+            else:
+                src = FilteredSource(time_src.frequency,self._frequencies,scale,self.sim.fields.dt)
+                (num_basis, num_pts) = src.nodes.shape
+                for basis_i in range(num_basis):
+                    sources += [mp.IndexedSource(src.time_src_bf[basis_i], near_data, src.nodes[basis_i])]
 
         return sources
 
     def __call__(self):
-        self._dg = Grid(*self.sim.get_array_metadata(dft_cell=self._monitor))
         self._eval = np.array([
             self.sim.get_dft_array(self._monitor, self.component, i)
             for i in range(self.num_freq)
--- a/python/adjoint/objective.py
+++ b/python/adjoint/objective.py
@@ -99,13 +99,6 @@ class ObjectiveQuantity(abc.ABC):
             scale *= 2
         return scale
 
-    def _dft_monitor_size(self,min_max_corners):
-        """Evaluates the size of the dft monitor for FourierFields."""
-        thickness_indicator = self.sim.fields.indicate_thick_dims(self.volume.swigobj)
-        raw_size = (min_max_corners[3:6]-min_max_corners[0:3])/2+np.ones(3)
-        true_size = np.array([raw_size[i]*(thickness_indicator[i]==1)+(thickness_indicator[i]==0) for i in range(3)],dtype=int)
-        return true_size
-
     def _create_time_profile(self, fwidth_frac=0.1):
         """Creates a time domain waveform for normalizing the adjoint source(s).
 
@@ -255,9 +248,7 @@ class FourierFields(ObjectiveQuantity):
         time_src = self._create_time_profile()
         sources = []
 
-        min_max_corners = self.sim.fields.get_corners(self._monitor.swigobj, self.component) # ivec values of the corners
-        mon_size = self._dft_monitor_size(min_max_corners)
-
+        mon_size = self.sim.fields.dft_monitor_size(self._monitor.swigobj, self.volume.swigobj, self.component)
         dJ = dJ.astype(np.complex128)
         if np.prod(mon_size)*self.num_freq == dJ.size: # The objective function J is a scalar.
             dJ = dJ.flatten()
@@ -266,7 +257,7 @@ class FourierFields(ObjectiveQuantity):
         else:
             raise ValueError('The format of J is incorrect!')
 
-        self.all_fouriersrcdata = self._monitor.swigobj.fourier_sourcedata(self.volume.swigobj, min_max_corners, dJ)
+        self.all_fouriersrcdata = self._monitor.swigobj.fourier_sourcedata(self.volume.swigobj, self.component, self.sim.fields, dJ)
 
         for fourier_data in self.all_fouriersrcdata:
             amp_arr = np.array(fourier_data.amp_arr).reshape(-1, self.num_freq)
--- a/python/adjoint/objective.py
+++ b/python/adjoint/objective.py
@@ -258,6 +258,7 @@ class FourierFields(ObjectiveQuantity):
         min_max_corners = self.sim.fields.get_corners(self._monitor.swigobj, self.component) # ivec values of the corners
         mon_size = self._dft_monitor_size(min_max_corners)
 
+        dJ = dJ.astype(np.complex128)
         if np.prod(mon_size)*self.num_freq == dJ.size: # The objective function J is a scalar.
             dJ = dJ.flatten()
         elif np.prod(mon_size)*self.num_freq**2 == dJ.size: # The objective function J is a vector. Each component corresponds to a frequency.
--- a/python/meep.i
+++ b/python/meep.i
@@ -1066,6 +1066,10 @@ void _get_gradient(PyObject *grad, double scalegrad, PyObject *fields_a, PyObjec
     $1 = PyList_Check($input);
 }
 
+%typemap(in) int* thickness_indicator {
+    $1 = (int *)thickness_indicator($input);
+}
+
 %apply int INPLACE_ARRAY1[ANY] { int [3] };
 %apply double INPLACE_ARRAY1[ANY] { double [3] };
 
@@ -1426,6 +1430,10 @@ void _get_gradient(PyObject *grad, double scalegrad, PyObject *fields_a, PyObjec
 
 %typemap(in) double* farpt_list {
     $1 = (double *)array_data($input);
+}
+
+%typemap(in) int* min_max_corners {
+    $1 = (int *)array_data($input);
 }
 
 %exception {
--- a/python/meep.i
+++ b/python/meep.i
@@ -1066,10 +1066,6 @@ void _get_gradient(PyObject *grad, double scalegrad, PyObject *fields_a, PyObjec
     $1 = PyList_Check($input);
 }
 
-%typemap(in) int* thickness_indicator {
-    $1 = (int *)thickness_indicator($input);
-}
-
 %apply int INPLACE_ARRAY1[ANY] { int [3] };
 %apply double INPLACE_ARRAY1[ANY] { double [3] };
 
--- a/python/meep.i
+++ b/python/meep.i
@@ -1432,10 +1432,6 @@ void _get_gradient(PyObject *grad, double scalegrad, PyObject *fields_a, PyObjec
     $1 = (double *)array_data($input);
 }
 
-%typemap(in) int* min_max_corners {
-    $1 = (int *)array_data($input);
-}
-
 %exception {
 #ifdef MEEP_SWIG_PYTHON_DEBUG
   // NOTE: You can do fancier things like timing the calls and using that
--- a/src/array_slice.cpp
+++ b/src/array_slice.cpp
@@ -523,35 +523,31 @@ bool increment(size_t *n, size_t *nMax, int rank) {
   return true;
 }
 
-// get the dimensions of reduced arrays
-size_t *reduce_array_dimensions(volume where, int full_rank, int &reduced_rank, size_t &reduced_grid_size, size_t dims[3], size_t stride[3], size_t reduced_stride[3], direction dirs[3], direction reduced_dirs[3]) {
+// get the dimensions of reduced arrays (i.e., arrays collapsed along empty dimensions of original volume)
+void reduce_array_dimensions(volume where, int rank, size_t dims[3], direction dirs[3], size_t stride[3], int &reduced_rank, size_t reduced_dims[3], direction reduced_dirs[3], size_t reduced_stride[3]) {
 
-  size_t *reduced_dims = new size_t[3];
   reduced_rank = 0;
-  stride[0] = stride[1] = stride[2] = reduced_stride[0] = reduced_stride[1] = reduced_stride[2] = 1;
-  for (int r = 0; r < full_rank; r++) {
+  reduced_dims[0] = reduced_dims[1] = reduced_dims[2] = stride[0] = stride[1] = stride[2] =
+      reduced_stride[0] = reduced_stride[1] = reduced_stride[2] = 1;
+  for (int r = 0; r < rank; r++) {
     if (where.in_direction(dirs[r]) == 0.0)
       reduced_stride[r] = 0; // degenerate dimension, to be collapsed
     else {
       reduced_dirs[reduced_rank] = dirs[r];
-      reduced_dims[reduced_rank++] = dims[r];
+      reduced_dims[reduced_rank++] = dims[r]; // reduced_dims is the size of the array after collapsing
     }
   }
   /*--------------------------------------------------------------*/
   /*- set up strides into full and reduced arrays                 */
   /*--------------------------------------------------------------*/
-  if (full_rank == 2)
+  if (rank == 2)
     stride[0] = dims[1]; // rstride is already all set in this case
-  else if (full_rank == 3) {
+  else if (rank == 3) {
     stride[0] = dims[1] * dims[2];
     stride[1] = dims[2];
     if (reduced_rank == 2)
       reduced_stride[reduced_stride[0] != 0 ? 0 : 1] = reduced_dims[1];
   }
-
-  /*--------------------------------------------------------------*/
-  reduced_grid_size = reduced_dims[0] * (reduced_rank == 2 ? reduced_dims[1] : 1);
-  return reduced_dims;
 }
 
 // data_size = 1,2 for real,complex-valued array
@@ -563,11 +559,11 @@ realnum *collapse_array(realnum *array, int *rank, size_t dims[3], direction dir
   /*- collapsed array                                             */
   /*--------------------------------------------------------------*/
   int full_rank = *rank;
-  size_t reduced_stride[3], stride[3]; // the latter for non-reduced array strides
+  size_t reduced_dims[3], reduced_stride[3], stride[3]; // the latter for non-reduced array strides
   direction reduced_dirs[3];
   size_t reduced_grid_size;
   int reduced_rank;
-  size_t *reduced_dims = reduce_array_dimensions(where, full_rank, reduced_rank, reduced_grid_size, dims, stride, reduced_stride, dirs, reduced_dirs);
+  reduce_array_dimensions(where, full_rank, dims, dirs, stride, reduced_rank, reduced_dims, reduced_dirs, reduced_stride);
 
   if (full_rank == 0) return array;
   if (reduced_rank==0) {
@@ -579,6 +575,7 @@ realnum *collapse_array(realnum *array, int *rank, size_t dims[3], direction dir
   /*--------------------------------------------------------------*/
   /*- allocate reduced array and compress full array into it     -*/
   /*--------------------------------------------------------------*/
+  reduced_grid_size = reduced_dims[0]*reduced_dims[1]*reduced_dims[2];
   size_t reduced_array_size = data_size * reduced_grid_size;
   realnum *reduced_array = new realnum[reduced_array_size];
   if (!reduced_array) meep::abort("%s:%i: out of memory (%zu)", __FILE__, __LINE__, reduced_array_size);
--- a/src/array_slice.cpp
+++ b/src/array_slice.cpp
@@ -524,15 +524,17 @@ bool increment(size_t *n, size_t *nMax, int rank) {
 }
 
 // get the dimensions of reduced arrays
-size_t *reduce_array_dimensions(int full_rank, int *reduced_rank, size_t dims[3], direction dirs[3], volume where, size_t stride[3], size_t reduced_stride[3], direction reduced_dirs[3], size_t *reduced_grid_size) {
-  size_t *reduced_dims = new size_t[3];
+size_t *reduce_array_dimensions(volume where, int full_rank, int &reduced_rank, size_t &reduced_grid_size, size_t dims[3], size_t stride[3], size_t reduced_stride[3], direction dirs[3], direction reduced_dirs[3]) {
 
+  size_t *reduced_dims = new size_t[3];
+  reduced_rank = 0;
+  stride[0] = stride[1] = stride[2] = reduced_stride[0] = reduced_stride[1] = reduced_stride[2] = 1;
   for (int r = 0; r < full_rank; r++) {
     if (where.in_direction(dirs[r]) == 0.0)
       reduced_stride[r] = 0; // degenerate dimension, to be collapsed
     else {
-      reduced_dirs[*reduced_rank] = dirs[r];
-      reduced_dims[(*reduced_rank)++] = dims[r];
+      reduced_dirs[reduced_rank] = dirs[r];
+      reduced_dims[reduced_rank++] = dims[r];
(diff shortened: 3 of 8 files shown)


smartalecH:
How does this PR relate to #1547? Does that PR need to work/be merged first? Are you locally testing this PR on top of that PR?

mawc2019:
>How does this PR relate to #1547?

The PR in #1547 could improve the accuracy of this PR in some circumstances.


> Does that PR need to work/be merged first?

No. This PR can work/be merged without the PR in #1547, but the PR in #1547 needs this PR to work/be merged at first.


> Are you locally testing this PR on top of that PR?

This PR has been tested separately from that PR. I also combined the two PRs, but the [combined PR](https://github.com/NanoComp/meep/compare/master...mawc2019:fourier_adjoint_fix_boundary_sources) does not work at present as described in #1547.

python/adjoint/objective.py
```diff
@@ -231,98 +231,52 @@ def __call__(self):
 
 
 class FourierFields(ObjectiveQuantity):
-    def __init__(self, sim, volume, component, decimation_factor=0):
+    def __init__(self, sim, volume, component, yee_grid=False, decimation_factor=0):
```
smartalecH, line 234: I think we should keep the default value of `yee_grid=True`. As discussed in #1793, it's better for the user to not worry about the Yee grid. i.e. if they specify a point, they shouldn't need to make sure it's on the Yee grid (or understand that it's being *snapped* under the hood). 

To make this work we'll need to do a restriction step for the adjoint sources, of course.

smartalecH:
> This PR passes all the built-in tests except test_adjoint_jax.py, test_binary_partition_utils.py, and test_chunk_balancer.py, as the current master branch does.

It looks like it's passing all tests online? Perhaps you need to rebase your local build to ensure they pass locally too.

mawc2019:
> It looks like it's passing all tests online? Perhaps you need to rebase your local build to ensure they pass locally too.
I ran the built-in tests only locally.

python/adjoint/objective.py
```diff
@@ -231,98 +231,52 @@ def __call__(self):
...
-                                    amplitude=amp[xi, yi, zi],
-                                    center=mp.Vector3(self._dg.x[xi],
-                                                      self._dg.y[yi],
-                                                      self._dg.z[zi]),
-                                )
-                            ]
+        thickness_indicator = self.sim.fields.indicate_thick_dims(self.volume.swigobj)
+        min_max_corners = self.sim.fields.get_corners(self._monitor.swigobj, self.component) # ivec values of the corners
+        raw_size = (min_max_corners[3:6]-min_max_corners[0:3])/2+np.ones(3)
+        true_size = np.array([raw_size[i]*(thickness_indicator[i]==1)+(thickness_indicator[i]==0) for i in range(3)],dtype=int) # monitor size
```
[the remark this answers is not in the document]
mawc2019: Perhaps a name like `_dft_monitor_size` sounds more pertinent. The function would be
```
    def _dft_monitor_size(self,min_max_corners):
        thickness_indicator = self.sim.fields.indicate_thick_dims(self.volume.swigobj)
        raw_size = (min_max_corners[3:6]-min_max_corners[0:3])/2+np.ones(3)
        true_size = np.array([raw_size[i]*(thickness_indicator[i]==1)+(thickness_indicator[i]==0) for i in range(3)],dtype=int)
        return true_size
```
Line 252, which computes `min_max_corners`, is not encapsulated here, because `min_max_corners` is needed by `fourier_sourcedata` later.

python/adjoint/objective.py
```diff
@@ -231,98 +231,52 @@ def __call__(self):
...
-                                    center=mp.Vector3(self._dg.x[xi],
-                                                      self._dg.y[yi],
-                                                      self._dg.z[zi]),
-                                )
-                            ]
+            raise ValueError('The format of J is incorrect!')
+
+        self.all_fouriersrcdata = self._monitor.swigobj.fourier_sourcedata(self.volume.swigobj, min_max_corners, dJ)
+
+        for near_data in self.all_fouriersrcdata:
```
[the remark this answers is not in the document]
mawc2019: I would replace `near_data` here with `fourier_data`.

python/adjoint/objective.py
```diff
@@ -231,98 +231,52 @@ def __call__(self):
...
-                                    time_src,
-                                    component=self.component,
-                                    amplitude=amp[xi, yi, zi],
-                                    center=mp.Vector3(self._dg.x[xi],
-                                                      self._dg.y[yi],
-                                                      self._dg.z[zi]),
-                                )
-                            ]
+        thickness_indicator = self.sim.fields.indicate_thick_dims(self.volume.swigobj)
+        min_max_corners = self.sim.fields.get_corners(self._monitor.swigobj, self.component) # ivec values of the corners
```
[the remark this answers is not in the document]
mawc2019: No, because `self.yee_grid` affects `min_max_corners`. In other words, `self.yee_grid==False` or `self.yee_grid==True` is already handled by Line 250.

src/dft.cpp
```diff
@@ -1375,4 +1375,91 @@ void fields::get_mode_mode_overlap(void *mode1_data, void *mode2_data, dft_flux
   get_overlap(mode1_data, mode2_data, flux, 0, overlaps);
 }
 
+std::vector<int> fields::get_corners(dft_fields fdft, component c){ // get the minimum and maximum ivec values of the dft monitor
+  volume *where = &v;
```
smartalecH, line 1379: There's a bit of subtlety when calculating the proper `ivec` corners during the DFT chunk creation that doesn't seem to be accounted for here.

If you look inside the `loop_in_chunks` routine (which is what determines the actual `ivec` corners) you notice there is some "snapping" that occurs, and that snapping depends on whether or not we are interpolating to the center of the voxel (e.g. yee_grid=False):

https://github.com/NanoComp/meep/blob/bc58782c0d16e6d59937028cbe2ffa402f5f24b0/src/loop_in_chunks.cpp#L361-L369

There's some *additional* logic below that to account for symmetries.

It's probably best to refactor *that* part of `loop_in_chunks` as `get_corners` so that we are consistently pulling the same corners.

python/adjoint/objective.py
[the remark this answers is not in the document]
smartalecH: >In other words, self.yee_grid==False or self.yee_grid==True is already handled by min_max_corners.

Hmm, I don't think that's the case. Take a look at [the comment below](https://github.com/NanoComp/meep/pull/1821/files#r748372325).

smartalecH:
The failing test is the same failing test described [here](https://github.com/NanoComp/meep/pull/1780#issuecomment-967188269).

For reference, here are the values comparing the adjoint gradients with the finite differences:

```
Directional derivative -- adjoint solver: [-4.94904819e-06 -3.02577147e-04 -2.97732191e-04], FD: [-4.75876167e-07 -2.98877141e-04 -3.12129430e-04]
```
The finite-difference gradients seem to be off (again).

src/dft.cpp
mawc2019: Steven suggested me to refactor the code from `fields::process_dft_component` in [dft.cpp](https://github.com/NanoComp/meep/blob/bc58782c0d16e6d59937028cbe2ffa402f5f24b0/src/dft.cpp#L1028). So the function `get_corners` has some code from `fields::process_dft_component` in [dft.cpp](https://github.com/NanoComp/meep/blob/bc58782c0d16e6d59937028cbe2ffa402f5f24b0/src/dft.cpp#L1028), but not from `fields::loop_in_chunks` in [loop_in_chunks.cpp](https://github.com/NanoComp/meep/blob/bc58782c0d16e6d59937028cbe2ffa402f5f24b0/src/loop_in_chunks.cpp#L340).

In my tests, `get_corners` could distinguish `self.yee_grid==False` from `True`. I have noticed the following things as expected:
1) The output of `get_corners` varies when `self.yee_grid` is changed.
2) The output of `get_corners` varies when the field component is changed in the case of `self.yee_grid==True`.
3) The output of `get_corners` does not vary when the field component is changed in the case of `self.yee_grid==False`.

In addition, after comparing the values of these outputs, I could infer that `get_corners` works on centers of voxels. (If `get_corners` does not work on centers of voxels when `self.yee_grid==False`, `get_corners` probably does not work correctly when `self.yee_grid==True`.)

mawc2019: Do you think we can keep this function as it is?

smartalecH: If it passes tests, then I see no reason to change it. Just note that the `dft` code *does* call `loop_in_chunks` when determining the boundaries (not just when computing the actual `dtft`), so we want to make sure we are always consistent.

mawc2019:
> The finite-difference gradients seem to be off (again).

It seems that this issue has been resolved. Should I rebuild this PR on the current master branch and upload again?

smartalecH:
> It seems that this issue has been resolved. Should I rebuild this PR on the current master branch and upload again?

I retriggered the tests, but they are still failing.

It's possible that the fix implemented in #1780 wasn't sufficient to resolve the issue with the new DFT adjoint sources. In which case, try building with single-precision locally and play with the finite-difference tolerance to see if that really is the issue (as described above). Make sure you rebase onto master when building locally.

mawc2019:
I rebased this PR on the current master branch locally with `--enable-single`.
The error comes from [self.assertClose](https://github.com/NanoComp/meep/blob/90f9f5280ee0b69f260e15ff312df888232f79a4/python/tests/utils.py#L21) when it is invoke at [test_damping](https://github.com/NanoComp/meep/blob/master/python/tests/test_adjoint_solver.py#L510) of `test_adjoint_solver.py`. The adjoint and finite-difference gradients being tested are
`
Directional derivative -- adjoint solver: [1.52774155e-04 1.41824694e-04 8.20296227e-05], FD: [1.53309901e-04 1.41692487e-04 8.19702919e-05]
`

As declared at its head, `assertClose` asserts that `two values or vectors satisfy ‖x-y‖ ≤ ε * max(‖x‖, ‖y‖)`.
```
def assertClose(self, x, y, epsilon = 1e-2, msg = ''):
    """Asserts that two values or vectors satisfy ‖x-y‖ ≤ ε * max(‖x‖, ‖y‖)."""
```
However, when I printed the two values under comparision, I saw `‖x-y‖ = 5.35746249993109e-07` and `ε * max(‖x‖, ‖y‖) = 6.132396030551491e-06`, which should pass the check. Is it possible that `assertClose` does not function as expected?

smartalecH:
(@oskooi I retriggered the tests after merging #1835 -- unfortunately, the tolerance adjustments in that PR were not enough to make this PR pass).

smartalecH:
> The finite-difference gradients seem still off.

The finite-difference gradients are a bit flaky, particularly when single-precision is enabled *and* the new subpixel smoothing routine is turned on.

One issue we discussed was that the new default for `MaterialGrid` subpixel smoothing routine is `do_smoothing=True`. This is a bit problematic if the underlying densities are not filtered (which appears to be the case with the [failing test](https://github.com/NanoComp/meep/blob/f9bd75748f834a1b237da8113236e70039bff4d2/python/tests/test_adjoint_solver.py#L495)).

Maybe try *filtering* the "random" design field before running the finite difference simulations. In other words, change the line

https://github.com/NanoComp/meep/blob/f9bd75748f834a1b237da8113236e70039bff4d2/python/tests/test_adjoint_solver.py#L495

to something like
```python
S12_unperturbed = forward_simulation_damping(mpa.conic_filter(p,<<put parameters here...>>), frequencies) 
```
and the same for the second part of the finite-difference calculation:

https://github.com/NanoComp/meep/blob/f9bd75748f834a1b237da8113236e70039bff4d2/python/tests/test_adjoint_solver.py#L502

Some other things we can try:
- Do a center-difference calculation
- Do some Richardson extrapolation
- Don't do a random initial design, but an initial design that makes sense (e.g. a ring or something)

mawc2019:
The failing check at [test_damping](https://github.com/NanoComp/meep/blob/master/python/tests/test_adjoint_solver.py#L511) was not caused by this PR, because `test_damping` only involves the `EigenmodeCoefficient` adjoint solver. In fact, the [previous master branch](https://github.com/NanoComp/meep/commit/90f9f5280ee0b69f260e15ff312df888232f79a4), which was released on this Tuesday morning, did not pass this test, either, when compiled with `--enable-single`.

> Maybe try filtering the "random" design field before running the finite-difference (and corresponding adjoint) simulations.

If we filter the design field in [test_damping](https://github.com/NanoComp/meep/blob/master/python/tests/test_adjoint_solver.py#L487) as follows, the check will be successful in the [previous master branch](https://github.com/NanoComp/meep/commit/90f9f5280ee0b69f260e15ff312df888232f79a4) as well as in this PR based on that.
```
    def test_damping(self):
        print("*** TESTING CONDUCTIVITIES ***")

        for frequencies in [[1/1.58, fcen, 1/1.53]]:
            ## filter the design field
            filter_radius = 0.12
            p_filtered = mpa.conic_filter(
                np.reshape(p,(Nx,Ny)),filter_radius,design_region_size.x,design_region_size.y,design_region_resolution).flatten()
            p_plus_dp_filtered = mpa.conic_filter(
                np.reshape(p+dp,(Nx,Ny)),filter_radius,design_region_size.x,design_region_size.y,design_region_resolution).flatten()
            
            ## compute gradient using adjoint solver
            adjsol_obj, adjsol_grad = adjoint_solver_damping(p_filtered, frequencies)

            ## compute unperturbed S12
            S12_unperturbed = forward_simulation_damping(p_filtered, frequencies)

            ## compare objective results
            print("S12 -- adjoint solver: {}, traditional simulation: {}".format(adjsol_obj,S12_unperturbed))

            ## compute perturbed S12
            S12_perturbed = forward_simulation_damping(p_plus_dp_filtered, frequencies)

            ## compare gradients
            if adjsol_grad.ndim < 2:
                adjsol_grad = np.expand_dims(adjsol_grad,axis=1)
            adj_scale = (dp[None,:]@adjsol_grad).flatten()
            tol = 0.06 if mp.is_single_precision() else 0.03
```
However, a different `filter_radius` may not result in a successful check. Moreover, the current master branch fails the checks of [test_adjoint_solver.py](https://github.com/NanoComp/meep/blob/master/python/tests/test_adjoint_solver.py) and [test_get_epsilon_grid.py](https://github.com/NanoComp/meep/blob/master/python/tests/test_get_epsilon_grid.py) in different ways, which cannot be resolved by the above revision.

smartalecH:
> In fact, the previous master branch, which was released on this Tuesday morning, did not pass this test, either, when compiled with --enable-single.

Hmm, I'm not sure I follow. The current master branch appears to be passing all tests (including the recent addition you mention). Are you saying that when you build `master` on your local machine (e.g. with `--enable-single`) that test fails for you?

> Moreover, the current master branch also fails the checks in test_adjoint_solver.py, which may not be caused by the same reason.

Can you be more specific? Again, `master` seems to be passing (and is passing for me locally).

smartalecH:
Even after the force push it still seems to be failing. Maybe take a closer look at exactly what's failing, as it's most likely due to something new in this PR (and not an out of date conflicting commit)


(Note you can download the python logs from the CI to see exactly where things are failing)

mawc2019:
In the log I saw the failure happened at `test_adjoint_solver.py` during `make distcheck` after `configure with single-precision floating point and swig threads`. Can I get the details of that failure at `test_adjoint_solver.py`? There should be `test_adjoint_solver.log` online.

smartalecH:
Yes, the details are [here](https://github.com/NanoComp/meep/suites/4586617600/artifacts/123637436).

Note that the `DFT` tests are failing:

```
(line 98) Directional derivative -- adjoint solver: [-131.11312903], FD: [-0.00342687]
```

Also, the fields are blowing up during the complex `EigenmodeCoefficient` tests... Does your code somehow impact those at all?

mawc2019:
> Note that the DFT tests are failing

The version of Python seems to matter a lot. I used Python 3.7.3 locally, and there were no failing tests of the `FourierFields` adjoint solver.

> Also, the fields are blowing up during the complex EigenmodeCoefficient tests...

It seems that the diverging field in the tests of complex fields is also caused by the `FourierFields` adjoint solver, and `adjoint_solver_complex_fields` only invokes the `FourierFields` adjoint solver.

> Does your code somehow impact those at all?

I do not think this PR impacts `EigenmodeCoefficient`.

smartalecH:
> It seems that the diverging field in the tests of complex fields is also caused by the FourierFields adjoint solver, and adjoint_solver_complex_fields only invokes the FourierFields adjoint solver.

You're right, I misread the log. The eigenmode tests seem to be unaffected by this (as expected)

> The version of Python seems to matter a lot

Are you sure it's the Python version and not the single-precision that's causing the issue? It seems odd that the python version would affect anything in this PR... It's especially odd that the adjoint run for the complex case blows up...

mawc2019:
> Are you sure it's the Python version and not the single-precision that's causing the issue?

I am not sure. I guess that the Python version matters, only because those failures of `FourierFields` do not appear locally when I compile Meep with or without `--enable-single`. The failure that I have seen locally is caused by `test_damping` that invokes `EigenmodeCoefficient` only. If the Python version does not matter, perhaps there are some differences between local and online tests?

> It's especially odd that the adjoint run for the complex case blows up...

Diverging results also appeared previously, as mentioned in #1676. In an [example](https://github.com/NanoComp/meep/issues/1676#issuecomment-904236002) there, the problem also appears without using the adjoint solver. I do not know whether these problems are caused by the same reason.

smartalecH:
> I am not sure. I guess that the Python version matters, only because those failures of FourierFields do not appear locally when I compile Meep with or without --enable-single. The failure that I have seen locally is caused by test_damping that invokes EigenmodeCoefficient only. If the Python version does not matter, perhaps there are some differences between local and online tests?

Hmm, I'm guessing your local branch doesn't line up with this PR+`master`.

mawc2019:
> I'm guessing your local branch doesn't line up with this PR+`master`.

I just tested this PR locally again based on the current master branch. The check now fails at the same place as the online check. By the way, I also checked the current master branch locally with single precision again, and saw that it still fails `test_damping`.

mawc2019:
In addition to the diverging fields and the inconsistency between the gradients, the adjoint gradients may be different in different runs. These problems happen only when this PR is compiled with `--enable-single`, but the `Near2Far` adjoint solver does not have these problems even if compiled with `--enable-single`.

The precision issue has already been handled in `get_gradient`, but do any places in this PR still need to handle the precision issue? In addition, this PR does not introduce any randomness explicitly. Where could the randomness in adjoint gradients come from?

smartalecH:
Can you provide a MWE that illustrates the problem? Something simpler than the entire test suite.

I can then pull your branch and look into it locally.

mawc2019:
An example is as follows.

```
import meep as mp
import meep.adjoint as mpa
import numpy as np
print("Single precision? ",mp.is_single_precision())

Lx,Ly,pml_size = 2,4,1
design_width,design_height = 1,0.6
monitor_width,monitor_height = 1,design_height+0.1

sim = mp.Simulation(

nf = mpa.FourierFields(sim,mp.Volume(center=mp.Vector3(0,monitor_height),size=mp.Vector3(monitor_width,0)),mp.Ez,yee_grid=True)
ob_list = [nf]

def J(near_data):
    return npa.abs(near_data[0,0])**2

x0 = 0.5*np.ones(Nx*Ny)
opt.update_design([x0])
f0, dJ_deps = opt()

db, choose = 1e-4, 5
np.random.seed(20211204)

g_adjoint = dJ_deps.flatten()

```

With double precision, the results are
```
Finite-difference gradient:  [206.65506121 -82.40772321  98.45018865   5.91237534  85.70581779]
Adjoint gradient:  [206.65504516 -82.38054598  98.43878076   5.91337638  85.71809238]
```
With single precision, the adjoint gradients vary among different runs and do not match the finite-difference gradients.

smartalecH:
So if I run your example on `master`, it seems to work just fine for *both* double and single precision (of course, I have to remove the `yee_grid` flag as that's something you add in this PR).

I'll take a look at your branch next.

smartalecH:
The issue lies with the precision of the Jacobian array (which is `complex64` in single precision), and what's expected inside of `fourier_sourcedata` (which is always `complex<double>`, or `complex128` in `numpy` speak). 

That discrepancy causes the `C` routine to significantly overstep the original Jacobian array, pulling garbage from memory. This is why you are seeing exploding simulations and/or inconsistent results from run to run. Of course, this is only an issue within single-precision (which seems consistent with the tests).

There are 2 solutions:
- Always force the Jacobian array to be `complex128` by casting it before passing to `fourier_sourcedata`. (This is easy, but probably not the right way...)
- Ensure that all the necessary fields within `fourier_sourcedata` are `complex<realnum>` so that the actual sources are the proper precision. This is probably the right solution, but will require quite a bit of modification to surrounding code, as all the dependencies (e.g. the `sourcedata` struct, the `near2far` adjoint code, etc.) are also defined as `complex<double>` rather than `complex<realnum`.

Maybe try the first solution first, and then go from there.

mawc2019:
> Always force the Jacobian array to be complex128 by casting it before passing to fourier_sourcedata.

I add `dJ = dJ.astype(np.complex128)` before `dJ` is passed to `fourier_sourcedata`, but the problem with single precision still exists. In addition, if the issue lies here, `Near2Far` may also be problematic with single precision, but `Near2Far` works well.

src/dft.cpp
```diff
@@ -1375,4 +1375,91 @@ void fields::get_mode_mode_overlap(void *mode1_data, void *mode2_data, dft_flux
...
+      else{ // yee_grid = false
+          // four or two neighbouring points in the yee lattice are involved in calculating the value at the center of a voxel
+        size_t site_ind[4] = {idx,idx + f->avg1,idx + f->avg2,idx + f->avg1 + f->avg2};
+        for (size_t j = 0; j < 4; ++j){
+          temp_struct.idx_arr.push_back(site_ind[j]);
+          for (size_t i = 0; i < Nfreq; ++i) {
+            EH0 = dJ_weight*dJ[Nspace*i+space_idx]*0.25; // split the amplitude of the adjoint source into four parts
+            if (is_electric(temp_struct.near_fd_comp)) EH0 *= -1;
+            EH0 /= f->S.multiplicity(ix0);
+            temp_struct.amp_arr.push_back(EH0);
```
smartalecH, line 1455: > I add dJ = dJ.astype(np.complex128) before dJ is passed to fourier_sourcedata, but the problem with single precision still exists.

Have you tried printing the values of `EH0` at each iteration to ensure you are getting what you expect? When I tried this, I was getting garbage (which is why I was hoping recasting the Jacobian would fix things). But there may be additional issues you'll have to investigate. But essentially, this is where things are breaking down.

Make sure you use `printf()`, rather than `master-printf`, so that you get info for all the chunks.

mawc2019:
> With single precision, the adjoint gradients vary among different runs and do not match the finite-difference gradients.

Such a problem of this PR is indeed related to the [commit in the master branch on November 30th](https://github.com/NanoComp/meep/commit/f9bd75748f834a1b237da8113236e70039bff4d2), as Ardavan mentioned yesterday. When this PR is rebased on the master branch before that commit, the problem does not occur. After that commit, the problem appears.

> Always force the Jacobian array to be complex128 by casting it before passing to fourier_sourcedata.

Your solution indeed solves the problem. I am sorry that I omitted to run `make` and `make install` yesterday.

> `Near2Far` may also be problematic with single precision, but `Near2Far` works well.

In the Python part of `Near2Far`, the type of `dJ` is somehow always `complex128` without any explicit conversion. So `Near2Far` does not have the same problem with single precision.

smartalecH:
> In the Python part of Near2Far, the type of dJ is somehow always complex128 without any explicit conversion. So Near2Far does not have the same problem with single precision.

Yes I think that's because the main `Near2Far` routine uses `double` (and `complex<double>`) everywhere, whereas the `dft_fields` calculation doesn't have any postprocessing and returns whatever `realnum` corresponds to.

As such, the corresponding Jacobian for `dft_fields` will also match the `realnum` type.

Glad it's working!

python/meep.i
```diff
@@ -1066,6 +1066,10 @@ void _get_gradient(PyObject *grad, double scalegrad, PyObject *fields_a, PyObjec
     $1 = PyList_Check($input);
 }
 
+%typemap(in) int* thickness_indicator {
+    $1 = (int *)thickness_indicator($input);
```
stevengj, line 1070: What is this?

src/array_slice.cpp
```diff
@@ -812,4 +812,14 @@ std::vector<double> fields::get_array_metadata(const volume &where) {
 
 } // get_array_metadata
 
+std::vector<int> fields::indicate_thick_dims(const volume &where) { // indicate which dimensions have sizes larger than 1
```
stevengj, line 815: `std::vector<bool>`?

stevengj: Why is this function needed?

src/dft.cpp
```diff
@@ -1375,4 +1375,91 @@ void fields::get_mode_mode_overlap(void *mode1_data, void *mode2_data, dft_flux
...
+  am_now_working_on(MpiAllTime);
+  max_corner = max_to_all(max_corner);
+  min_corner = -max_to_all(-min_corner); // i.e., min_to_all
+  finished_working();
+
+  std::vector<int> corners = {min_corner.x(),min_corner.y(),min_corner.z(),max_corner.x(),max_corner.y(),max_corner.z()};
+  return corners;
+}
+
+std::vector<struct sourcedata> dft_fields::fourier_sourcedata(const volume &where, int* min_max_corners, std::complex<double>* dJ){
```
stevengj, line 1399:
```suggestion
std::vector<struct sourcedata> dft_fields::fourier_sourcedata(const volume &where, const int* min_max_corners, const std::complex<double>* dJ){
```
since these arrays aren't modified?  (and change the headers etc. accordingly)

mawc2019: They are not modified. `const` can be added here and also in `near_sourcedata` of `near2far.cpp`.

python/meep.i
mawc2019: This three-element array indicates which the zero- and nonzero-thickness dimensions of the DFT monitor for the `FourierFields` adjoint solver. If the monitor has zero thickness in a dimension, the corresponding element in `thickness_indicator` is 0. If the thickness in a dimension is nonzero, the corresponding element in `thickness_indicator` is 1. This array is produced by `indicate_thick_dims` in `array_slice.cpp` and used by `_dft_monitor_size` in `objective.py`.

src/array_slice.cpp
mawc2019: I also wanted to use bool here. I changed `meep.hpp` accordingly and changed `int` to `bool` in the corresponding variable in `meep.i`. But it did not work. Perhaps I did not change `meep.i` correctly.

```
%typemap(in) int* thickness_indicator {
    $1 = (int *)thickness_indicator($input);
}
```

> Why is this function needed?

The purpose of this function is to provide assistance to the evaluation of the DFT monitor's size, which affects the size of the Jacobian `dJ`, in the `FourierFields` adjoint solver. The monitor size in a zero-thickness dimension may not be faithfully evaluated from the ivec values of the monitor's corners. To be specific, in a dimension where the thickness of the monitor is zero, the ivec values may sound like the thickness is nonzero; but in a dimension where the thickness of the monitor is nonzero, the ivec values always give reliable indications. Therefore, in order to obtain the accurate size of the DFT monitor, the code may need to know in which dimensions the monitor has zero thickness.

python/meep.i
stevengj: My point is that this makes no sense as a SWIG input conversion from `$input` (a `PyObject*`) to `$1` (an `int*`).

stevengj: `$1 = (int *)array_data($input);` would make more sense, if the Python input is a numpy array of `int32` (`int`) values.

python/adjoint/objective.py
```diff
@@ -99,6 +99,13 @@ def _adj_src_scale(self, include_resolution=True):
             scale *= 2
         return scale
 
+    def _dft_monitor_size(self,min_max_corners):
+        """Evaluates the size of the dft monitor for FourierFields."""
+        thickness_indicator = self.sim.fields.indicate_thick_dims(self.volume.swigobj)
+        raw_size = (min_max_corners[3:6]-min_max_corners[0:3])/2+np.ones(3)
+        true_size = np.array([raw_size[i]*(thickness_indicator[i]==1)+(thickness_indicator[i]==0) for i in range(3)],dtype=int)
+        return true_size
```
stevengj, line 107: You should really be using the *same* code that is used to construct the DFT slice to begin with, not reconstructing the slice dimensions.   

This means refactoring this code:  https://github.com/NanoComp/meep/blob/1c7a1edd86a378c8dcd3dcd3e2e1ed40620fe956/src/dft.cpp#L1044-L1067

to pull out the computation of the DFT slice dimensions into a separate function, so that you can call it in *both* `process_dft_component` and in your code.

stevengj: also this code:
https://github.com/NanoComp/meep/blob/1c7a1edd86a378c8dcd3dcd3e2e1ed40620fe956/src/dft.cpp#L1069-L1096

mawc2019: > to pull out the computation of the DFT slice dimensions into a separate function, so that you can call it in both process_dft_component and in your code.

Like `process_dft_component`, this _separate function_ would be a member function of `fields`, but the function `fourier_sourcedata` in my code, which needs the output of this _separate function_, is a member function of `dft_fields`. Would it be acceptable in Meep if a `fields` function is made callable from `dft_fields`?

mawc2019:
Some portions of `complex<double> fields::process_dft_component` in `dft.cpp` and `realnum *collapse_array` in `array_slice.cpp` are extracted as two new functions, i.e., `size_t *fields::get_dims` and `size_t *reduce_array_dimensions`, respectively. These two functions need to be invoked in `std::vector<struct sourcedata> dft_fields::fourier_sourcedata`. However, `Segmentation fault` and `Address not mapped` appear when `get_dims` (a `fields` function) is invoked in `fourier_sourcedata` (a `dft_fields` function). I let `fourier_sourcedata` (a `dft_fields` function) invoke different `fields` functions, and still saw those errors if the `fields` function contains some stuff such as `gv.round_vec`.

mawc2019:
I let a `dft_fields` function invoke a simple `fields` function. Then I compiled Meep with `--enable-debug` and ran the code with GDB. The error message given by GDB is as follows.
```
Thread 1 "python" received signal SIGSEGV, Segmentation fault.
0x00007ffff68e24a2 in meep::volume::get_max_corner (this=0x255) at ../src/meep/vec.hpp:988
988       vec get_max_corner() const { return max_corner; };
```

smartalecH:
> The error message given by GDB is as follows.

Now you can use gdb to figure out what's going on. It's an interactive debugger, so you can type commands and it will run them with the current program state in memory. For example, you can print the value of a variable using `p my_variable`. You can step out of the current line of code (to see what is calling this failing line) using `up 1` (where the number 1 tells you how many lines you want to step out).

The trick is to play detective now that you can probe the code in action.

mawc2019:
> I let a `dft_fields` function invoke a simple `fields` function.

Assume the `fields` function is `std::vector<int> fields::fields_test_func(dft_fields fdft, component c)`. In a `dft_fields` function, this `fields` function is invoked like
```
  fields *flds;
  std::vector<int> test_output = flds->fields_test_func(fdft, c);
```
Would `fields *flds` be problematic here?

smartalecH:
> Would fields *flds be problematic here?

Did you initialize your `fields` object with a constructor? Or is it just a pointer pointing to garbage?

mawc2019:
> Did you initialize your fields pointer with a constructor? Or is it just pointing to garbage?

In the version uploaded last week, the `fields` pointer was not initialized with a constructor. It pointed to garbage.

> Why not just use the existing `fields` object (typically called `f`)?

The latest version adopts this convention.

> Is your latest branch pushed to GitHub (so that we can look and comment at the above piece of code)?

Yes, but this latest version has new problems. It is okay when running with 1 process, but when running with multiple processes, the adjoint run hangs at `MPI_Allreduce` in [`max_to_all`](https://github.com/NanoComp/meep/blob/master/src/mympi.cpp#L414), which is invoked in one of the new functions [`get_dims`](https://github.com/NanoComp/meep/blob/3c853d6d87e721fdffb3308b71a650b16a66ba62/src/dft.cpp#L1012), which is invoked in [`fourier_sourcedata`](https://github.com/NanoComp/meep/blob/3c853d6d87e721fdffb3308b71a650b16a66ba62/src/dft.cpp#L1420). The function `get_dims` is adapted from a portion of [`process_dft_component`](https://github.com/NanoComp/meep/blob/master/src/dft.cpp#L1047-L1092), which is also invoked in the forward run but does not incur hanging.

stevengj:
> the adjoint run hangs at MPI_Allreduce in max_to_all, which is invoked in one of the new functions get_dims

This will happen if not all of the processes are calling `max_to_all`.   Maybe add a `printf("hello from process %\n", my_rank());`  to see which processes are calling `fourier_sourcedata` and hence `get_dims`?

src/dft.cpp
```diff
@@ -1375,4 +1388,94 @@ void fields::get_mode_mode_overlap(void *mode1_data, void *mode2_data, dft_flux
...
+  size_t *reduced_dims = reduce_array_dimensions(where, rank, reduced_rank, reduced_grid_size, dims, stride, reduced_stride, dirs, reduced_dirs);
+
+  std::vector<int> monitor_array = {1, 1, 1}; // lengths of the dft monitor along the three dimensions
+  for (int nd = 0; nd < 3; ++nd) if (nd < reduced_rank) monitor_array[nd] = reduced_dims[nd];
+
+  return monitor_array;
+}
+
+std::vector<struct sourcedata> dft_fields::fourier_sourcedata(const volume &where, component c, fields &f, const std::complex<double>* dJ){
+  if (!chunks) return std::vector<struct sourcedata>();
```
stevengj, line 1410: This is maybe the problem with the hanging — it means that some processes will never execute thie `get_dims` line.  Just delete this line.

python/adjoint/objective.py
```diff
@@ -231,98 +231,48 @@ def __call__(self):
...
         else:
-            '''The adjoint solver requires the objective function
-            to be scalar valued with regard to objective arguments
-            and position, but the function may be vector valued
-            with regard to frequency. In this case, the Jacobian
-            will be of the form [F,F,...] where F is the number of
-            frequencies. Because of linearity, we can sum across the
-            second frequency dimension to calculate a frequency
-            scale factor for each point (rather than a scale vector).
-            '''
```
stevengj, line 285: Keep this comment, it's helpful!

python/adjoint/objective.py
```diff
@@ -231,98 +231,48 @@ def __call__(self):
...
-                                                      self._dg.y[yi],
-                                                      self._dg.z[zi]),
-                                )
-                            ]
+        mon_size = self.sim.fields.dft_monitor_size(self._monitor.swigobj, self.volume.swigobj, self.component)
+        dJ = dJ.astype(np.complex128)
+        if np.prod(mon_size)*self.num_freq == dJ.size: # The objective function J is a scalar.
+            dJ = dJ.flatten()
+        elif np.prod(mon_size)*self.num_freq**2 == dJ.size: # The objective function J is a vector. Each component corresponds to a frequency.
+            dJ = np.sum(dJ,axis=1).flatten()
```
stevengj, line 256: Is `flatten` necessary here?   Even if `dJ` is a multidimensional numpy array, internally it is stored as a "flat" array and so C++ should get the same pointer regardless?

python/adjoint/objective.py
stevengj: @smartalecH, could you explain what is going on here?   Why can you sum over frequency if the objective function is vector-valued?

python/meep.i
stevengj: You mentioned to day that this should be deleted?

src/array_slice.cpp
```diff
@@ -523,20 +523,12 @@ bool increment(size_t *n, size_t *nMax, int rank) {
...
-                        int data_size = 1) {
-
-  /*--------------------------------------------------------------*/
-  /*- detect empty dimensions and compute rank and strides for    */
-  /*- collapsed array                                             */
-  /*--------------------------------------------------------------*/
-  int full_rank = *rank;
-  if (full_rank == 0) return array;
+// get the dimensions of reduced arrays
+size_t *reduce_array_dimensions(volume where, int full_rank, int &reduced_rank, size_t &reduced_grid_size, size_t dims[3], size_t stride[3], size_t reduced_stride[3], direction dirs[3], direction reduced_dirs[3]) {
```
stevengj, line 527: Probably better to make `size_t reduced_dims[3]` one of the arguments

By returning a `new` array you probably have a memory leak.

stevengj: Maybe put all of the input arguments first, and then all of the output arguments for clarity?

stevengj, line 527:
```suggestion
void reduce_array_dimensions(volume where, int rank, const size_t dims[3], const size_t stride[3], const direction dirs[3],
                                                     size_t &reduced_grid_size, 
                                                     int &reduced_rank, size_t reduced_dims[3], size_t reduced_stride[3], direction reduced_dirs[3]) {
```

src/array_slice.cpp
```diff
@@ -564,10 +549,36 @@ realnum *collapse_array(realnum *array, int *rank, size_t dims[3], direction dir
       reduced_stride[reduced_stride[0] != 0 ? 0 : 1] = reduced_dims[1];
   }
 
+  /*--------------------------------------------------------------*/
+  reduced_grid_size = reduced_dims[0] * (reduced_rank == 2 ? reduced_dims[1] : 1);
```
stevengj, line 553: This doesn't seem to work if `reduced_rank == 3`?

It seems easier to just initialize `reduced_dims[0] = reduced_dims[1] = reduced_dims[2] = 1` and
then the caller can compute `reduced_grid_size = reduced_dims[0] * reduced_dims[1] * reduced_dims[2]` if they want.

Then don't have a `reduced_grid_size` argument here.

src/dft.cpp
```diff
@@ -985,6 +985,60 @@ complex<double> dft_chunk::process_dft_component(int rank, direction *ds, ivec m
   return integral;
 }
 
+// get variables that are needed by complex<double> fields::process_dft_component
+size_t *fields::get_dims(dft_chunk **chunklists, int num_chunklists, component c, ivec &min_corner, ivec &max_corner, size_t &array_size, size_t &bufsz, int &rank, direction *ds, int *array_rank, size_t *array_dims, direction *array_dirs){
```
stevengj, line 989: Maybe `get_dft_component_dims`

src/dft.cpp
```diff
@@ -985,6 +985,60 @@ complex<double> dft_chunk::process_dft_component(int rank, direction *ds, ivec m
...
+  /***************************************************************/
+  /* get statistics on the volume slice **************************/
+  /***************************************************************/
+  volume *where = &v; // use full volume of fields
+  bufsz = 0;
+  min_corner = gv.round_vec(where->get_max_corner()) + one_ivec(gv.dim);
+  max_corner = gv.round_vec(where->get_min_corner()) - one_ivec(gv.dim);
```
stevengj, line 996: You can just use `v.get_max_corner()` etcetera and eliminate `where` completely?

src/dft.cpp
stevengj: Make `size_t dims[3]` and argument and return `void` here.

src/dft.cpp
```diff
@@ -985,6 +985,60 @@ complex<double> dft_chunk::process_dft_component(int rank, direction *ds, ivec m
...
+  am_now_working_on(MpiAllTime);
+  max_corner = max_to_all(max_corner);
+  min_corner = -max_to_all(-min_corner); // i.e., min_to_all
+  finished_working();
+
+  /***************************************************************/
+  /***************************************************************/
+  /***************************************************************/
+  rank = 0;
+  size_t *dims = new size_t[3];
```
stevengj, line 1020: (this introduces a memory leak if this is never freed)

src/dft.cpp
```diff
@@ -1171,6 +1182,8 @@ complex<double> fields::process_dft_component(dft_chunk **chunklists, int num_ch
 /***************************************************************/
 complex<realnum> *collapse_array(complex<realnum> *array, int *rank, size_t dims[3], direction dirs[3], volume where);
 
+size_t *reduce_array_dimensions(volume where, int full_rank, int &reduced_rank, size_t &reduced_grid_size, size_t dims[3], size_t stride[3], size_t reduced_stride[3], direction dirs[3], direction reduced_dirs[3]);
```
stevengj, line 1185: Probably both of these should be in `meep-internal.hpp` … and make sure that header file is included both here and in `array_slice.cpp` where the functions are defined.

The reason is that this way the compiler will check that the declaration in the header matches the declaration where the function is defined.

src/dft.cpp
```diff
@@ -1375,4 +1388,94 @@ void fields::get_mode_mode_overlap(void *mode1_data, void *mode2_data, dft_flux
   get_overlap(mode1_data, mode2_data, flux, 0, overlaps);
 }
 
+// return the size of the dft monitor
+std::vector<int> fields::dft_monitor_size(dft_fields fdft, const volume &where, component c){
```
stevengj, line 1392: Does `vector<size_t>` work or does SWIG not convert `size_t` to Python correctly?

src/dft.cpp
```diff
@@ -1375,4 +1388,94 @@ void fields::get_mode_mode_overlap(void *mode1_data, void *mode2_data, dft_flux
...
+  ivec min_corner, max_corner;
+  int rank, reduced_rank;
+  direction dirs[3], reduced_dirs[3];
+  size_t array_size, bufsz, reduced_grid_size, reduced_stride[3], stride[3];
+  dft_chunk *chunklists[1];
+  chunklists[0] = chunks;
+  
+  size_t *dims = f.get_dims(chunklists, 1, c, min_corner, max_corner, array_size, bufsz, rank, dirs);
+  size_t *reduced_dims = reduce_array_dimensions(where, rank, reduced_rank, reduced_grid_size, dims, stride, reduced_stride, dirs, reduced_dirs);
+  size_t monitor_array[3] = {1, 1, 1}; // lengths of the dft monitor along the three dimensions
```
stevengj, line 1421: With the above changes, you should be able to get rid of `monitor_array` and just use `reduced_grid_size` directly.

src/dft.cpp
```diff
@@ -1375,4 +1388,94 @@ void fields::get_mode_mode_overlap(void *mode1_data, void *mode2_data, dft_flux
...
+  int rank, reduced_rank;
+  direction dirs[3], reduced_dirs[3];
+  size_t array_size, bufsz, reduced_grid_size, reduced_stride[3], stride[3];
+  dft_chunk *chunklists[1];
+  chunklists[0] = fdft.chunks;
+
+  size_t *dims = get_dims(chunklists, 1, c, min_corner, max_corner, array_size, bufsz, rank, dirs);
+  size_t *reduced_dims = reduce_array_dimensions(where, rank, reduced_rank, reduced_grid_size, dims, stride, reduced_stride, dirs, reduced_dirs);
+
+  std::vector<int> monitor_array = {1, 1, 1}; // lengths of the dft monitor along the three dimensions
```
stevengj, line 1403: Probably call this `reduced_dims_vec` to be more descriptive

src/dft.cpp
```diff
@@ -1375,4 +1388,94 @@ void fields::get_mode_mode_overlap(void *mode1_data, void *mode2_data, dft_flux
...
+  direction dirs[3], reduced_dirs[3];
+  size_t array_size, bufsz, reduced_grid_size, reduced_stride[3], stride[3];
+  dft_chunk *chunklists[1];
+  chunklists[0] = chunks;
+  
+  size_t *dims = f.get_dims(chunklists, 1, c, min_corner, max_corner, array_size, bufsz, rank, dirs);
+  size_t *reduced_dims = reduce_array_dimensions(where, rank, reduced_rank, reduced_grid_size, dims, stride, reduced_stride, dirs, reduced_dirs);
+  size_t monitor_array[3] = {1, 1, 1}; // lengths of the dft monitor along the three dimensions
+  for (int nd = 0; nd < 3; ++nd) if (nd < reduced_rank) monitor_array[nd] = reduced_dims[nd];
+  const size_t monitor_size = monitor_array[0]*monitor_array[1]*monitor_array[2]; // total number of points in the monitor
```
stevengj, line 1423: the `const` is unnecessary here

src/dft.cpp
```diff
@@ -1375,4 +1388,94 @@ void fields::get_mode_mode_overlap(void *mode1_data, void *mode2_data, dft_flux
...
+
+      double dJ_weight = 1; // weight for linear interpolation
+      int nd = 0;
+      LOOP_OVER_DIRECTIONS(f->fc->gv.dim, d){
+        if (where.in_direction(d) > 0) position_array[nd++] = int((ix0.in_direction(d)-min_corner.in_direction(d))/2);
+        else dJ_weight *= (1-abs(x0.in_direction(d)-where.in_direction_min(d))/(f->fc->gv.inva)); // based on distances
+      }
+
+      // index when dJ is flattened to a one-dimenional array
+      size_t idx_1d = position_array[0]*monitor_array[1]*monitor_array[2]+position_array[1]*monitor_array[2]+position_array[2];
```
stevengj, line 1451:
```suggestion
      size_t idx_1d = (position_array[0]*monitor_array[1] + position_array[1])*monitor_array[2] + position_array[2];
```

src/dft.cpp
```diff
@@ -1375,4 +1388,94 @@ void fields::get_mode_mode_overlap(void *mode1_data, void *mode2_data, dft_flux
...
+
+      if (f->avg1==0 && f->avg2==0){ // yee_grid = true
+        temp_struct.idx_arr.push_back(idx);
+        for (size_t i = 0; i < Nfreq; ++i) {
+          EH0 = dJ_weight*dJ[monitor_size*i+idx_1d];
+          if (is_electric(temp_struct.near_fd_comp)) EH0 *= -1;
```
stevengj, line 1457: Why flip sign here?

stevengj: cc @mochen4

mochen4: The flip of sign is consistent with [Owen Miller's thesis](http://optoelectronics.eecs.berkeley.edu/ThesisOwenMiller.pdf) Eq (5.15b). It says the change of permittivity ε in the design region induces an electric dipole, whose effects on E and H in some region χ, is equivalent to change of E field in the design region from the electric and negative magnetic dipole in the region χ, by reciprocity.

oskooi:
Looks like this PR, particularly the contents of `dft_fields::fourier_sourcedata`, is causing failures in the [address sanitizer](https://github.com/NanoComp/meep/runs/5154712153?check_suite_focus=true) related to the misuse of `size_t` (unsigned) and `ptrdiff_t` (signed).

Issue #1811, opened 2021-11-01
Incorrect gradients when materials contain conductivities


smartalecH:
As described in #1788, the adjoint solver does not produce accurate gradients when materials contain conductivities (including artificial damping terms). This is why the following lines are commented out on `master`:

https://github.com/NanoComp/meep/blob/7019c11e8b870ac8859166dd66a7f1f8df2b7c71/src/meepgeom.cpp#L603-L606

(Note that even if these lines were active, the actual conductivities themselves still need to be interpolated, as described in our paper -- but that's not why the gradients were failing, as the artificial damping terms introduced in #1788 also fail.)

When calculating Aᵤ (the derivative of the system matrix w.r.t. the design parameters), do a finite difference, performed here:

https://github.com/NanoComp/meep/blob/7019c11e8b870ac8859166dd66a7f1f8df2b7c71/src/meepgeom.cpp#L2566-L2592

The code that evaluates ε(ω,u) is found here:

https://github.com/NanoComp/meep/blob/7019c11e8b870ac8859166dd66a7f1f8df2b7c71/src/meepgeom.cpp#L2506-L2533

What's interesting is gradients for dispersive materials without any global conductivities (e.g. LN in the material database) work just fine -- meaning at least part of the above code is working well...

[Here](https://gist.github.com/smartalecH/e2e84d5f5b806a52582ba9dabc841c4f) is a test script that can be used during debugging.

([Here](https://gist.github.com/smartalecH/6511bfa9a1b002e33ebddaf6d2d22c77) is another script that uses LN for the material grid and produces accurate gradients)

(cc @mochen4)

stevengj:
Note that conductivity is handled in a very different place in the code from other dispersive materials — in the `step_curl` operation, in `update_edhb`.   Not sure if this would affect the adjoint formula?

mochen4:
It seems that when damping is on, dA_du isn't calculated from finite difference but from the trivial case here: https://github.com/NanoComp/meep/blob/7019c11e8b870ac8859166dd66a7f1f8df2b7c71/src/meepgeom.cpp#L2553-L2555
Besides check additionally ``md->damping == 0`` in the if statement above, shouldn't we also use ``2*meep::pi*freq`` instead of ``freq`` here?
https://github.com/NanoComp/meep/blob/7019c11e8b870ac8859166dd66a7f1f8df2b7c71/src/meepgeom.cpp#L2519

Making these two changes seems to fix the issue.

[...]

PR #1823, opened 2021-11-09
Fix adjoint gradient with connectivity

mochen4:
Fix https://github.com/NanoComp/meep/issues/1811 after PR https://github.com/NanoComp/meep/pull/1801


smartalecH:
Would be good to add a simple test to `test_adjoint_solver.py`.

Maybe just borrow a simple test like this...

https://github.com/NanoComp/meep/blob/d4b1e1a519037f4a63c8e81caf4711de50f5378b/python/tests/test_adjoint_solver.py#L308-L333

...and retrofit it to include a damping term.

PR #1826, opened 2021-11-12
Fix memory leaks in SWIG wrappers.

kkg4theweb:
Fixes memory leaks in the SWIG wrappers for meep::make_output_directory and the automatic freeing of material_type_list

### DIFF
--- a/python/meep.i
+++ b/python/meep.i
@@ -1405,6 +1405,7 @@ void _get_gradient(PyObject *grad, PyObject *fields_a, PyObject *fields_f, PyObj
     if ($1.num_items != 0) {
         for (int i = 0; i < $1.num_items; i++) {
             material_free($1.items[i]);
+            delete $1.items[i];
         }
     }
     delete[] $1.items;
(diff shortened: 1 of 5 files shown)


python/meep.i
```diff
@@ -1405,6 +1405,7 @@ void _get_gradient(PyObject *grad, PyObject *fields_a, PyObject *fields_f, PyObj
     if ($1.num_items != 0) {
         for (int i = 0; i < $1.num_items; i++) {
             material_free($1.items[i]);
+            delete $1.items[i];
```
stevengj, line 1408: Maybe it would be better to put this in `material_free`?

kkg4theweb: We could - but then it might break the code of anybody using the C++ api directly.
I dont know how much we worry about that.

stevengj: I'm not so worried about that.

stevengj: I pushed a patch to move `delete` into `material_free`.   It looks like there was another memory leak in `set_default_material` that this fixes.

kkg4theweb: Thanks for making the change. I tested this internally and everything looks good.
Please go ahead and merge this change.

PR #1828, opened 2021-11-16
documentation notes on the relationship between free- and bound-charge currents

oskooi:
As an additional demonstration, might be useful to include a plot of the fields vs. time for a 2d cell with PML containing a point-dipole source with `is_integrated` set to `True` and `False`. The fields generated by the `is_integrated=True` source should eventually decay away to zero after the source has turned off whereas `is_integrated=False` should converge to a small non-zero value.


doc/docs/FAQ.md
```diff
@@ -73,10 +73,16 @@ If you are worried about this, then you are probably setting up your calculation
...
+
+Note that a free-charge current in a polarizable medium such as a dielectric material (including vacuum) creates [bound charges](https://en.wikipedia.org/wiki/Maxwell%27s_equations#Bound_charge_and_current) which in turn can act as a secondary current source. In an FDTD simulation, these bound charges $\rho_b$ (which are equivalent to the divergence of the polarization field $\vec{P}$ via Gauss's law or [electric dipole moment](https://en.wikipedia.org/wiki/Electric_dipole_moment)) may make it seem that the current source never turns off because the fields do not decay away to zero. This is demonstrated in the figure below showing a plot of the electric energy over the entire cell versus simulation time for a point dipole in 3d vacuum surrounded by PML. To ensure that the net dipole moment is zero when the source turns off, the source should be the *integral* of $\mathcal{J}$ rather than just $\mathcal{J}$. This feature is enabled in Meep by setting the `is_integrated` parameter of the [`GaussianSource`](Python_User_Interface.md#gaussiansource) class object to `True` (default is `False`).
```
stevengj, line 78: The integral of free current is free polarization (hence free charge), not bound charge.  (Of course, free charges *generate* polarization of the bound charge if you have a polarization, but the *source* of the field is ultimately the free charge density).

PR #1830, opened 2021-11-18
Fix adjoint gradient with conductivities

### LINKED ISSUE #1811: Incorrect gradients when materials contain conductivities

mochen4:
Replaces https://github.com/NanoComp/meep/pull/1823. Fixes #1811.

### DIFF
--- a/python/tests/test_adjoint_solver.py
+++ b/python/tests/test_adjoint_solver.py
@@ -261,7 +261,85 @@ def adjoint_solver_complex_fields(design_params, frequencies=None):
     sim.reset_meep()
 
     return f, dJ_du
+    
+def forward_simulation_damping(design_params, frequencies=None, mat2=silicon):
+    matgrid = mp.MaterialGrid(mp.Vector3(Nx,Ny),
+                              mp.air,
+                              mat2,
+                              weights=design_params.reshape(Nx,Ny),
+                              damping = 3.14*fcen)
 
+    matgrid_geometry = [mp.Block(center=mp.Vector3(),
+                                 size=mp.Vector3(design_region_size.x,design_region_size.y,0),
+                                 material=matgrid)]
+
+    geometry = waveguide_geometry + matgrid_geometry
+
+    sim = mp.Simulation(resolution=resolution,
+                        cell_size=cell_size,
+                        boundary_layers=pml_xy,
+                        sources=wvg_source,
+                        geometry=geometry)
+
+    if not frequencies:
+        frequencies = [fcen]
+
+    mode = sim.add_mode_monitor(frequencies,
+                                    mp.ModeRegion(center=mp.Vector3(0.5*sxy-dpml-0.1),
+                                                  size=mp.Vector3(0,sxy-2*dpml,0)),
+                                    yee_grid=True,
+                                    eig_parity=eig_parity)
+
+    sim.run(until_after_sources=mp.stop_when_dft_decayed())
+
+
+    coeff = sim.get_eigenmode_coefficients(mode,[1],eig_parity).alpha[0,:,0]
+    S12 = np.power(np.abs(coeff),2)
+    sim.reset_meep()
+    return S12
+
+def adjoint_solver_damping(design_params, frequencies=None, mat2=silicon):
+    matgrid = mp.MaterialGrid(mp.Vector3(Nx,Ny),
+                              mp.air,
+                              mat2,
+                              weights=np.ones((Nx,Ny)),
+                              damping = 3.14*fcen)
+    matgrid_region = mpa.DesignRegion(matgrid,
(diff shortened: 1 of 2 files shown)


stevengj:
Test failure?

mochen4:
> Test failure?

The double precision tests past but single precision one failed. I am still looking at why.

stevengj:
Maybe try changing `FD_DEFAULT` in `python/adjoint/utils.py` to see if it is the fault of the finite-difference dε/du.

stevengj:
In discussion with @mochen4, it seems like simply increasing the runtime fixes the problem — there seems to be a problem with the `stop_when_dft_decayed` convergence criterion in single precision.

mawc2019:
When I compile the current master branch with `--enable-single` and Python 3.7.3, the check at `test_damping` still fails as described in [#1821](https://github.com/NanoComp/meep/pull/1821#issuecomment-982959565). The finite-difference gradients change a lot between double precision and single precision. I increased `minimum_run_time` to `2000` in [forward_simulation_damping](https://github.com/NanoComp/meep/blob/master/python/tests/test_adjoint_solver.py#L294) and [adjoint_solver_damping](https://github.com/NanoComp/meep/blob/master/python/tests/test_adjoint_solver.py#L339), but the check still failed. A tweak following @smartalecH 's [suggestion](https://github.com/NanoComp/meep/pull/1821#issuecomment-983757992) is as follows, with which the check passes locally. Maybe this issue is negligible because the online check does not fail.

```
    def test_damping(self):
        print("*** TESTING CONDUCTIVITIES ***")

        for frequencies in [[1/1.58, fcen, 1/1.53]]:
            ## filter the design field
            filter_radius = 0.12
            p_filtered = mpa.conic_filter(
                np.reshape(p,(Nx,Ny)),filter_radius,design_region_size.x,design_region_size.y,design_region_resolution).flatten()
            dp_filtered = mpa.conic_filter(
                np.reshape(dp,(Nx,Ny)),filter_radius,design_region_size.x,design_region_size.y,design_region_resolution).flatten()
            
            ## compute gradient using adjoint solver
            adjsol_obj, adjsol_grad = adjoint_solver_damping(p_filtered, frequencies)

            ## compute unperturbed S12
            S12_unperturbed = forward_simulation_damping(p_filtered, frequencies)

            ## compare objective results
            print("S12 -- adjoint solver: {}, traditional simulation: {}".format(adjsol_obj,S12_unperturbed))

            ## compute perturbed S12
            S12_perturbed = forward_simulation_damping(p_filtered+dp_filtered, frequencies)

            ## compare gradients
            if adjsol_grad.ndim < 2:
                adjsol_grad = np.expand_dims(adjsol_grad,axis=1)
            adj_scale = (dp_filtered[None,:]@adjsol_grad).flatten()
            tol = 0.06 if mp.is_single_precision() else 0.03
```

mochen4:
``test_damping`` is only added after this PR, so certainly you wouldn't see the failure for commits before this one. But that test is only about the added features in this PR, so I don't think it is the reason why #1821 fails locally on your machine.

> Maybe this issue is negligible because the online check does not fail.

The CI test runs the single-precision case with python3.9. I tested the master branch locally under python3.9 in single-precision, and tests passed. So I think the issue is either something with the python version or something in your local machine.

@mawc2019

Issue #1797, opened 2021-10-22
Materialization of DFT field data on the master only


ahoenselaar:
DFT fields returned to the user are gathered in full on each process. This can result in massive memory allocations, especially when multiple Meep processes are running on the same physical node.

In many scenarios, the DFT field data is not needed on each worker/process but only on the master. Being able to materialize DFT fields on the master only could alleviate many out-of-memory situations.

stevengj:
(As much as possible, it would be nice to avoid collecting the DFT field data on any single process, but rather to leave it distributed and to compute with it in that form, e.g. as we do for the near-to-far transformation.)

stevengj:
In particular, the adjoint solver should never call get_dft — it should compute a distributed dot product of the forward and adjoint fields.

stevengj:
The outline I have in mind is:

1. Distribute the computation of the "dot product" between the forward and adjoint fields that gives the gradient.  This way, you won't need to collect the DFT fields on *any* process.  However, the degrees of freedom and the gradients (just 2d arrays for 2d material grids) will still be replicated.
2. In the long run, one could have a distributed version of the CCSA algorithm, so that each process only stores a portion of the degrees of freedom (e.g. you break the material grid into "chunks" according to process boundaries, and chunks that are not needed are not stored locally) and the CCSA algorithm operates in parallel on distributed data.  This way, from the user perspective it will be the same as now — you just write one "serial" Meep script that happens to run in parallel — but it will scale better to huge problems (e.g. volumetric degrees of freedom for 3d printing).

PR #1833, opened 2021-11-20
support for single-precision floating point for fields array functions

oskooi:
#1544 and #1675 added support for the single-precision floating point type for the time-domain and DFT fields. However, the fields array *functions* such as `fields::get_array_slice` and `fields::get_dft_array` only use double precision which unfortunately exacerbates the memory bottleneck described in #1797. This PR modifies all of the array functions to support single precision.

### DIFF
--- a/src/energy_and_flux.cpp
+++ b/src/energy_and_flux.cpp
@@ -58,10 +58,10 @@ double fields::field_energy_in_box(const volume &where) {
   return electric_energy_in_box(where) + cur_step_magnetic_energy;
 }
 
-static complex<realnum> dot_integrand(const complex<realnum> *fields, const vec &loc, void *data_) {
+static complex<double> dot_integrand(const complex<realnum> *fields, const vec &loc, void *data_) {
   (void)loc;
   (void)data_; // unused;
-  return real(conj(fields[0]) * fields[1]);
+  return real(conj(cdouble(fields[0])) * cdouble(fields[1]));
 }
 
 double fields::field_energy_in_box(component c, const volume &where) {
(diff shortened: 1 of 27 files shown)


oskooi:
This is finally ready to be merged. There were lots of (little) changes necessary to make everything work. The documentation for `get_array` and `get_dft_array` has been updated to mention that the type of the array elements returned by these two functions is the same as the fields themselves.

For now, the gradient computed by the adjoint solver is still in double precision regardless of the type of the forward and adjoint DFT fields. This is not a performance limitation since the gradient is much smaller in memory consumption than the DFT fields .To support computing the gradient in single precision would require changes to `geom_epsilon::eff_chi1inv_row` and other material-related functions which currently return the ε tensor in double precision. This can be addressed in a separate PR, if necessary.

src/energy_and_flux.cpp
```diff
@@ -58,7 +58,7 @@ double fields::field_energy_in_box(const volume &where) {
   return electric_energy_in_box(where) + cur_step_magnetic_energy;
 }
 
-static complex<double> dot_integrand(const complex<double> *fields, const vec &loc, void *data_) {
+static complex<realnum> dot_integrand(const complex<realnum> *fields, const vec &loc, void *data_) {
```
stevengj, line 61: Why is the return value changed?

src/meep.hpp
```diff
@@ -2077,13 +2077,13 @@ class fields {
...
   /* flux and mode fields.)                               */
   /********************************************************/
-  std::complex<double> process_dft_component(dft_chunk **chunklists, int num_chunklists,
-                                             int num_freq, component c, const char *HDF5FileName,
-                                             std::complex<double> **field_array = 0, int *rank = 0,
-                                             size_t *dims = 0, direction *dirs = 0,
-                                             void *mode1_data = 0, void *mode2_data = 0,
-                                             component c_conjugate = Ex, bool *first_component = 0,
-                                             bool retain_interp_weights = true);
+  std::complex<realnum> process_dft_component(dft_chunk **chunklists, int num_chunklists,
```
stevengj, line 2080: Why did the return value change?

stevengj:
The basic principle is to try to keep scalar calculations, like accumulating the integral, in double precision.  Use `realnum` only for arrays like the fields.   So, when you take the fields array and integrate it, you convert the value to double precision first (since C++ doesn't support mixed-precision calculations on complex numbers).

It might be helpful to write a little utility function, e.g.
```cxx
std::complex<double> cdouble(std::complex<float> z) { return std::complex<double>(real(z), imag(z)); }
std::complex<double> cdouble(std::complex<double> z) { return z; }
```

src/energy_and_flux.cpp
oskooi: Return type reverted to `std::complex<double>`.

src/meep.hpp
oskooi: Return type reverted to `std::complex<double>`.

oskooi:
The scalar calculations have been reverted to double precision and the `realnum` type is now used only for the fields arrays.

There are two slight adjustments to the tolerances of the checks in `test_adjoint_solver.py` which are actually unrelated to this PR. These were necessary because they were causing test failures on my local i7-7700k machine when using single precision.

PR #1827, opened 2021-11-13
plot geometry for dispersive materials without initializing structure object

oskooi:
Closes #1824.

Removed the `frequency` parameter from `plot2D` since this cannot be passed to `get_epsilon_grid`. As a result, this API change is not backwards compatible. As a workaround, we can leave the `frequency` parameter in but just mention in the documentation that it is simply ignored.

cc @joamatab

PR #1835, opened 2021-11-27
unit test for `get_epsilon_grid`

oskooi:
Following the support for dispersive materials added to `get_epsilon_grid` in #1827, this PR adds a unit test which verifies that the output from `get_epsilon_grid` is equivalent to `get_epsilon_point` (which involves initializing the `structure` object) over a 2d cell of arbitrary geometries and materials (including SiN and Co from the materials library) at multiple frequencies. 

real_grid_nx101_ny101

### DIFF
--- a/python/Makefile.am
+++ b/python/Makefile.am
@@ -65,6 +65,7 @@ TESTS =                                        \
     $(TEST_DIR)/test_gaussianbeam.py           \
     $(TEST_DIR)/test_geom.py                   \
     $(TEST_DIR)/test_get_point.py              \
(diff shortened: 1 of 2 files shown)


mawc2019:
The current master branch fails at [test_adjoint_solver.py](https://github.com/NanoComp/meep/blob/master/python/tests/test_adjoint_solver.py) and [test_get_epsilon_grid.py](https://github.com/NanoComp/meep/blob/master/python/tests/test_get_epsilon_grid.py) when compiled with `--enable-single`.

PR #1839, opened 2021-12-01
Fix memory leaks

kkg4theweb:
Fix memory leaks introduced with recent changes to geom_epsilon::geom_epsilon (8dd39f7946f1ab3ec8fe59db3cf3bbfc19184375)

Also fix existing memory leaks in GDSIIgeom.cpp


stevengj:
Thanks!

Issue #1834, opened 2021-11-22
Meep + MPB interaction causes crash


stevengj:
The following crashes for me:
```py
import numpy as np
import math
import meep as mp
from meep import mpb

ms_sq = mpb.ModeSolver(
        resolution=32,
        num_bands=8,
        geometry_lattice=mp.Lattice(size=mp.Vector3(1, 1)),
        geometry=[mp.Cylinder(0.2, material=mp.Medium(epsilon=12))],
        k_points = [ mp.Vector3(0.5, 0) ]
        )
ms_sq.run_tm()

sim = mp.Simulation(cell_size=mp.Vector3(15, 1),
                    sources=[mp.Source(mp.GaussianSource(0.2, fwidth=0.6), mp.Ez,
                                        center=(-6.5,0), size=(0,1))],
                    resolution=20)
sim.init_sim()
```
giving
```
Initializing structure...
python(14647,0x110da3e00) malloc: *** error for object 0x7fbc4bbc0e00: pointer being freed was not allocated
python(14647,0x110da3e00) malloc: *** set a breakpoint in malloc_error_break to debug
```
There is no crash if I comment out the MPB calculation, so somehow there is a weird interaction here?

@kkg4theweb, maybe you could run this example through your memory sanitizer?

oskooi:
I used `git bisect` to track down the first commit which triggers this error to #1583.

Unfortunately, this is not fixed by #1839.

kkg4theweb:
This wont be fixed by #1839 since that fix is only for a memory *leak* and not a violation.
I will try running this through the sanitizer and report back

kkg4theweb:
Indeed there is a crash. The problem is here:

https://github.com/NanoComp/meep/blob/087d92c49002554f182d68b0f1aafb0188075076/libpympb/pympb.cpp#L679

  default_material = _default_material;

We should not be doing this (setting the global directly) anymore but instead use "set_material" from meepgeom.cpp (after making it public) since set_material is supposed to make a copy of "_default_material"

The way it is, the next call to set_material after this is done, ends up freeing the original "_default_material" which is again freed by the swig wrapper code causing a crash.

I will send out a PR

kkg4theweb:
Sorry, I mean "set_default_material" and not "set_material"

kkg4theweb:
https://github.com/NanoComp/meep/pull/1840

stevengj:
Closed by #1840

PR #1840, opened 2021-12-02
Fix for Issue #1834 

kkg4theweb:
Expose set_default_material and use it in libpympb/pympb.cpp instead of directly setting the 'default_material' global which can lead to crashes since set_default_material actually makes a copy and frees up the old pointer before pointing to the new one.


stevengj:
Thanks!

PR #1842, opened 2021-12-06
Define what happens when `β=∞` and `u=η`

smartalecH:
#1801 allows us to take gradients when `β=∞`.

Currently, if `β=∞` and `u=η`, `NaN` is returned (as `∞×0` is not defined). 

This PR simply returns a 1 in that case (and thereby defines the default behavior).

### DIFF
--- a/src/meepgeom.cpp
+++ b/src/meepgeom.cpp
@@ -480,6 +480,7 @@ double material_grid_val(vector3 p, material_data *md) {
 
 static double tanh_projection(double u, double beta, double eta) {
   if (beta == 0) return u;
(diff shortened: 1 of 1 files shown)


src/meepgeom.cpp
```diff
@@ -480,6 +480,7 @@ double material_grid_val(vector3 p, material_data *md) {
 
 static double tanh_projection(double u, double beta, double eta) {
   if (beta == 0) return u;
+  if ((beta == std::numeric_limits<double>::infinity()) && (u == eta)) return 1.0; // 0*∞ is undefined, so let's just assign 1 here
```
stevengj, line 483: This seems sufficient:
```suggestion
  if (u == eta) return 1.0; // 0*∞ is NaN, so let's just assign 1 here
```

PR #1847, opened 2021-12-08
Alternative approach for stop_when_dft_decayed

### LINKED ISSUE #1831: stop_when_dft_decayed in single precision
The [current `stop_when_dft_decayed` implementation](https://github.com/NanoComp/meep/blob/03ea7ad3f56a5fb40b5211d69a885572e23b2f2b/python/simulation.py#L4527-L4564) seems like it may be problematic in single precision, as @mochen4 noticed in #1830.

In particular, it checks whether `| ‖DFT‖ - ‖prev DFT‖ | < tol max(‖DFT‖)` for a default tolerance of `tol = 1e-11`, which seems to be trying to detect a fractional change less than the single precision.   On the other hand, while the DFT fields are single precision, the norm is computed in double precision.

We might want to switch to the original approach that I suggested https://github.com/NanoComp/meep/pull/1740#issuecomment-906489048, which computes the norm of the fields contributing to the DFT, rather than the change of the norm of the DFT.     That has the advantage that it avoids the ill-conditioned step of taking a difference of two nearly equal quantities.

cc @smartalecH

mochen4:
Fixes #1831.

With the alternative approach, I couldn't find a default ``tol``/``decay_by`` parameter that would pass the single precision tests. The simplest solution I found is to use the ``minimum_run_time``,  as I currently have in this PR.

Alternatively, I could set a specific ``decay_by`` for each unit test in ``test_adjoint_solver.py``

### DIFF
--- a/python/adjoint/optimization_problem.py
+++ b/python/adjoint/optimization_problem.py
@@ -28,7 +28,7 @@ class OptimizationProblem(object):
         fcen=None,
         df=None,
         nf=None,
-        decay_by=1e-11,
+        decay_by=6e-7,
         decimation_factor=0,
         minimum_run_time=0,
         maximum_run_time=None,
--- a/python/adjoint/optimization_problem.py
+++ b/python/adjoint/optimization_problem.py
@@ -28,7 +28,7 @@ class OptimizationProblem(object):
         fcen=None,
         df=None,
         nf=None,
-        decay_by=6e-7,
+        decay_by=7.5e-7,
         decimation_factor=0,
         minimum_run_time=0,
         maximum_run_time=None,
--- a/python/adjoint/wrapper.py
+++ b/python/adjoint/wrapper.py
@@ -99,7 +99,7 @@ class MeepJaxWrapper:
         monitors: List[EigenmodeCoefficient],
         design_regions: List[DesignRegion],
         frequencies: List[float],
-        dft_threshold: float = 1e-11,
+        dft_threshold: float = 6e-7,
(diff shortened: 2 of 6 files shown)


stevengj:
Would be good to visualize the field in the single-precision case when the decay stagnates — i.e. what modes are causing it to have such a high Q?  The fields at the Nyquist frequency, or maybe at zero frequency?

mochen4:
I noticed the decay stagnated for the eigenmode test in the single-precision case at the traditional forward simulation:
https://github.com/NanoComp/meep/blob/5453a98c6dcfe0b8d501f9425a823da2abfe8750/python/tests/test_adjoint_solver.py#L401 but shouldn't it be basically identical to the forward simulation in ``OptimizationProblem`` (which decays fine)? @smartalecH 
https://github.com/NanoComp/meep/blob/5453a98c6dcfe0b8d501f9425a823da2abfe8750/python/tests/test_adjoint_solver.py#L398

smartalecH:
As discussed, I think the issue is that `forward_simulation` isn't placing `DFT` monitors in the design region (where you are seeing some high-frequency speckling). The simulation itself should be identical between the two, but the `adjoint_solver` routine is explicitly waiting for the "speckle" to converge.

mochen4:
Here are a few plots of the field when it stagnates

540 4833374023438fields
579 6500244140625fields
600 0166625976562fields
620 3833618164062fields

mochen4:
The field's high-frequency pattern is probably due to the round-off at single-precision.

PR #1852, opened 2021-12-09
switch type from double to realnum for monitor-related functions

oskooi:
This PR switches the type from `double` to `realnum` of all the functions defined in `montior.cpp`. None of these functions are performance critical since they are not typically called at every time step. The benefit of switching their type is for consistency with the rest of the API. Note, however, that the fields array input to Harminv is still in double precision since this change would have required modifying the Harminv library to also support single precision which is probably not necessary.

Unfortunately, `make` is failing so this PR is not yet ready to be merged. For some reason, the modified function `matrix_invert` of `monitor.cpp` is not being SWIG-wrapped properly which is causing `make` to fail. However, if the SWIG wrappers for Scheme and Python are modified by hand to force the correct type conversion for this function then `make` and `make check` both pass.

### DIFF
--- a/python/tests/test_mode_coeffs.py
+++ b/python/tests/test_mode_coeffs.py
@@ -116,7 +116,7 @@ class TestModeCoeffs(unittest.TestCase):
         ex_at_eval_point = emdata.amplitude(eval_point, mp.Ex)
         hz_at_eval_point = emdata.amplitude(eval_point, mp.Hz)
 
-        places = 5 if mp.is_single_precision() else 7
+        places = 4 if mp.is_single_precision() else 7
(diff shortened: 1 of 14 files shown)


python/tests/test_mode_coeffs.py
```diff
@@ -116,7 +116,7 @@ def test_modes(self):
         ex_at_eval_point = emdata.amplitude(eval_point, mp.Ex)
         hz_at_eval_point = emdata.amplitude(eval_point, mp.Hz)
 
-        places = 5 if mp.is_single_precision() else 7
+        places = 4 if mp.is_single_precision() else 7
```
smartalecH, line 119: Why do you now need to drop the precision of the test?

stevengj:
As a general principle, I'd prefer to use single precision only for storing large arrays.

Issue #1832, opened 2021-11-19
Non-materialized DFT field storage for streamlined adjoint calculations


smartalecH:
As discussed in #1797, we can alleviate several memory constraints while computing gradients during adjoint optimization by not materializing the DFT fields on every process. Instead, we can distribute the computation of the gradient itself. Here's what needs to happen/change in order for this to work.

1. The DFT constructor needs to accept a flag that allows the DFT chunks to persist *after* a simulation object is garbage collected (i.e. after we run the forward run). We need to be aware of what fields in the DFT object cannot be accessed anymore (because they will point to deallocated objects). There might be some python garbage collection that happens to the `dft_fields` object too (thanks to SWIG), so we'll need to double-check that.
2. We want to store DFT chunks with different field components, but the same `grid_volume` on the same process (this might already be the default). This will reduce a lot of communication during the recombination step (and simplify the code).
3. The `OptimizationProblem` class will no longer need to materialize the DFT fields. Rather, they will just store a pointer to the DFT object itself.
4. We need to modify the `get_gradient` routine to accept `dft_fields` pointers, rather than a pointer to the fields themselves.
5. We need to modify the main gradient code to "loop in chunks". Each chunk will update a local copy of the gradients.
6. Once all the local copies are done, we perform a simple `sum_to_all`.

Some gotchas
* There's going to need to be some boundary communication along the edges of chunks in some cases. The logic of this will be a little hairy.
* Symmetries are going to pose a small problem. It might be best, in this case, to only compute the simplest part of the gradient under the hood, and then return a full copy with the proper symmetries applied to the user (just like we do with everything else). For example, If we have a simulation with mirror symmetry, only the gradients for one half are computed during the `loop_in_chunks` routine. But we should probably support the use case of returning the full, symmetric array to the user (like the code currently operates).

(cc @oskooi)

stevengj:
With regards to step 2, that is already true for the centered grid, I think.  If the DFT chunks are on the Yee grid, then different field components aren't on the same grid points so it's not clear what you mean by the "same" grid volume.   I guess you can't process the components separately because of anisotropic materials?

I think it might be possible to avoid boundary communications, but I think we need to sit down and go through exactly what is needed here.

With regard to symmetries, I think it would be cleanest to operate similarly to loop in chunks, looping over each chunk potentially multiple times depending on the symmetry operations that map it into the design region.

stevengj:
[...]

Issue #1555, opened 2021-04-24
Adjoint gradients in Near2Far change with the number of processes


mawc2019:
The adjoint gradients in the Near2Far class change with the number of mpi processes. The test was based on the example code at [Near2Far Optimization with Epigraph Formulation](https://github.com/NanoComp/meep/blob/master/python/examples/adjoint_optimization/Near2Far-Optimization-with-Epigraph-Formulation.ipynb), with the objective function replaced by
```
def J(alpha):
    return npa.sum(npa.abs(alpha[0,:,2])**2)

opt = mpa.OptimizationProblem(
    simulation = sim,
    objective_functions = J,
    objective_arguments = ob_list,
    design_regions = [design_region],
    frequencies=frequencies,
    decay_by = 1e-6,
    decay_fields=[mp.Ez],
    maximum_run_time = 2000
)
```
The adjoint gradients were calculated as
```
x0 = 0.5*np.ones(Nx*Ny)
opt.update_design([x0])
f0, dJ_deps = opt()
g_adjoint = np.sum(dJ_deps, axis=1)
```

I ran the code with 1, 16, and 32 mpi processes. Their relative differences are as follows. The horizontal axis labels different pixels in the design region, and the vertical axis represents the relative differences w.r.t. the 1-process adjoint gradients, namely, ```16-process adjoint gradients / 1-processes adjoint gradients − 1``` and ```32-processes adjoint gradients / 1-processes adjoint gradients − 1```.
![image](https://user-images.githubusercontent.com/60903466/115972801-960d7b80-a51e-11eb-8985-b1112a6a8bef.png)
![image](https://user-images.githubusercontent.com/60903466/115972811-a58cc480-a51e-11eb-9346-a2a05c89153c.png)

If we take the absolute value and plot them in the log scale, we can see that the differences are ubiquitous rather than limited to those spikes.
![image](https://user-images.githubusercontent.com/60903466/115972834-d2d97280-a51e-11eb-84af-0ee59ecc8547.png)
![image](https://user-images.githubusercontent.com/60903466/115972838-d8cf5380-a51e-11eb-89d3-149dab0ee1fa.png)

I also tested EigenmodeCoefficient, but did not see the variation of adjoint gradients among different numbers of processes.

smartalecH:
See #1547, which should hopefully fix this.

mawc2019:
[...]

PR #1855, opened 2021-12-13
Non-materialized dft fields for adjoint calculations

### LINKED ISSUE #1797: Materialization of DFT field data on the master only

smartalecH:
Closes #1797, #1832, #1861

TODO
- [x] Add attribute to `dft_fields` allowing them to persist after a simulation is run 
- [x] Store additional boundary (i.e. padding) around dft subvolume
- [x] Devise way to easily access different components from single string of chunks
- [x] Pass tests


smartalecH:
Currently, the `dft_array` object creates a ["linked list" of sorts](https://github.com/NanoComp/meep/blob/ece7a0b7db8f672e27944cad1831c35619b8bbd0/src/dft.cpp#L851-L854) for all of the chunks on the current process, *regardless* of the field component. As such there doesn't seem to be a convenient way to traverse this list in order to find the proper field component during the recombination step.

For example, assume we are looping over the *adjoint* field chunks, and we are currently on the `Dx` chunk. Eventually, we will need the `Dy` and `Dz` *forward* field points that correspond to the current adjoint points (they will be spatially offset, of course). This is a bit tricky though because *all* of the chunks are cascaded together. We'd have to perform several checks on *all* the chunks at *every* point.

I was thinking it may be worth it to create a modified version of `dft_fields` (e.g. maybe `dft_adjoint_fields`) that explicitly separates the components, making the postprocessing step much easier. The Poynting Flux object [does something like this](https://github.com/NanoComp/meep/blob/ece7a0b7db8f672e27944cad1831c35619b8bbd0/src/dft.cpp#L537-L538). Any thoughts @stevengj @oskooi ?

stevengj:
Probably easier to just call `add_dft_fields` once for each component.

src/vec.cpp
```diff
@@ -1644,13 +1644,14 @@ grid_volume grid_volume::subvolume(ivec is, ivec ie) {
...
   LOOP_OVER_DIRECTIONS(dim, d) {
-    num[(int)d] = (ie - is).in_direction(d) / 2;
+    num[(int)d] = (ie - is).in_direction(d)/2;
     origin.set_direction(d, is.in_direction(d));
   }
   num_changed();
-  center_origin();
-  shift_origin(origin);
+  //center_origin();
+  set_origin(origin);
```
smartalecH, line 1654: This is the last bit of code that needs to work in order for this PR to pass (I hope).

I am using `subvolumes` to pull the proper `dft` array index given a specified `ivec` (using the `grid_volume::index()` function). Unfortunately, the `subvolume` code was never tested or debugged.

The current issue is that the `index()` routine is not supplying the proper index...

The `strides`, and `num` arrays are set properly, which means that the issue has to be with the `io` field (the origin of the `subvolume()`.

Any thoughts?

src/meepgeom.cpp
```diff
@@ -2853,182 +2853,169 @@ void material_grids_addgradient_point(double *v, vector3 p, double scalegrad, ge
...
+        for (int ci_forward=0; ci_forward<3; ci_forward++){
+          if (forward_dft_chunks[ci_forward].size() == 0) continue;
+          meep::dft_chunk* fwd_chunk = forward_dft_chunks[ci_forward][cur_chunk];
+          meep::grid_volume gv_fwd = gv.subvolume(fwd_chunk->is,fwd_chunk->ie);
+          meep::component forward_c = fwd_chunk->c;
+
+          // loop over each point of interest
+          LOOP_OVER_IVECS(gv,adj_chunk->is_old,adj_chunk->ie_old,idx){
+            meep::ivec ip = gv.iloc(adjoint_c,idx);
+            size_t idx_adj =  gv_adj.index(adjoint_c,ip);
```
smartalecH, line 2920: For reference, this is where I'm hoping to use the `index` function within `grid_volume` (which is actually a `subvolume`).

`ip` is the `ivec` corresponding to `idx` supplied by `LOOP_OVER_IVECS`.

src/vec.cpp
stevengj: From the `grid_volume::index` method, it seems like the origin should be set to `is - iyee_shift(c)` for your use case.

stevengj:
As discussed in https://github.com/NanoComp/meep/issues/1555, we'll probably need to ensure that loop_in_chunks only loops over owned points for this to work.

src/dft.cpp
```diff
@@ -70,6 +72,21 @@ dft_chunk::dft_chunk(fields_chunk *fc_, ivec is_, ivec ie_, vec s0_, vec s1_, ve
...
+  to account for boundary effects. We will pad
+  by 1 pixel in each dimension, while ensuring
+  we don't step outside of the chunk loop itself
+  */
+  if(data->expand){
+    is_old = is_;
+    ie_old = ie_;
+    is = max(is-one_ivec(fc->gv.dim)*2,fc->gv.little_corner());
+    ie = min(ie+one_ivec(fc->gv.dim)*2,fc->gv.big_corner());
+  }
```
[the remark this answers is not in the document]
smartalecH: It seems like the issue lies with chunks next to each other within the region. There's no problem with the edge of the dft region, so long as it remains in the same chunk. But the interface *between* chunks is problematic. Here's a poor man's illustration:
image

Specifically, `index()` returns an out of bounds index when it tries to pull the off-diagonal terms need for the inner product.

Ideally, the padding should account for this. But it appears that the edge of the `fields` chunk is preventing us from padding an extra pixel, in these cases (at least that's what it looks like when I print all the relevant values).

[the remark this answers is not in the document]
smartalecH: I reverted some unintentional `iyee_shift` in `dft.cpp`. I no longer see the issue described above, but the gradients are still inaccurate when multiple chunks are used (although the relative error is still pretty low).

I'm not sure the error lies with `loop_in_chunks()` either as I'm not using that in the recombination step (although `loop_in_chunks()` is used while creating the `dft_chunk`s, so that may be the issue).

python/adjoint/wrapper.py
```diff
@@ -156,13 +156,8 @@ def _run_fwd_simulation(self, design_variables):
...
 
         monitor_values = utils.gather_monitor_values(self.monitors)
-        fwd_fields = utils.gather_design_region_fields(
-            self.simulation,
-            design_region_monitors,
-            self.frequencies,
-        )
         return (jnp.asarray(monitor_values),
-                jax.tree_map(jnp.asarray, fwd_fields))
+                fwd_design_region_monitors)
```
smartalecH, line 160: @ianwilliamson as you can see, we no longer pull the fields onto each process. Rather, we just hold onto opaque pointers which are later passed to `_get_gradient` (`fwd_design_region_monitors` and `adj_design_region_monitors` are the pointers) .

Do I need to do anything special to ensure jax plays nicely with this opaque data structures (before, you were using `tree_map`s and some other rather interesting data structures to wrap the arrays)?

smartalecH:
Status update for @oskooi 

Currently, cases with complex fields are failing (including cylindrical coordinates). Also, the jax wrapper needs some love.

But other than that, things look good. There is still some error introduced (hopefully) due to the bug within #1555. But the error is rather small. And not present when only a single chunk holds DFT fields.

python/adjoint/wrapper.py
smartalecH: I think I got it, thanks.

smartalecH:
Finally passing tests and ready for review.

@ahoenselaar, do you want to run this on some of your own internal tests to see if you have any issues?

python/visualization.py
```diff
@@ -533,7 +533,7 @@ def plot_fields(sim, ax=None, fields=None, output_plane=None, field_parameters=N
     field_parameters = default_field_parameters if field_parameters is None else dict(default_field_parameters, **field_parameters)
 
     # user specifies a field component
-    if fields in [mp.Ex, mp.Ey, mp.Ez, mp.Er, mp.Ep, mp.Hx, mp.Hy, mp.Hz]:
+    if fields in [mp.Ex, mp.Ey, mp.Ez, mp.Er, mp.Ep, mp.Dx, mp.Dy, mp.Dz, mp.Hx, mp.Hy, mp.Hz]:
```
oskooi, line 536: Why does this change in the visualization module need to be a part of this PR?

src/meepgeom.cpp
```diff
@@ -2853,188 +2853,180 @@ void material_grids_addgradient_point(double *v, vector3 p, double scalegrad, ge
...
-            fwd2 = 0.5 * meep::cdouble(GET_FIELDS(fields_f, ci_forward, f_i, fwd2_idx));
-            fwd_avg = fwd1 + fwd2;
-            meep::vec eps2 = gv[ieps2];
-            cyl_scale = (gv.dim == meep::Dcyl) ? eps2.r() : 1;
-            material_grids_addgradient_point(
-                v+ng*f_i, vec_to_vector3(eps2), scalegrad*cyl_scale, geps,
-                adjoint_c, forward_c, fwd_avg, 0.5*adj, frequencies[f_i], gv, du);
+            /* more complicated case requires interpolation/restriction */
+            } else if (
+                      md->do_averaging ||                         /* account for subpixel smoothing     */
```
oskooi, line 2947: Let's try to use consistent formatting: no new line for the first clause of the `if` statement and also subsequent lines should be aligned properly.

python/adjoint/utils.py
```diff
@@ -43,53 +43,19 @@ def update_beta(self,beta):
...
-
         grad = onp.zeros((num_freqs, self.num_design_params))  # preallocate
-        geom_list = sim.geometry
-        f = sim.fields
         vol = sim._fit_volume_to_simulation(self.volume)
-
         # compute the gradient
-        mp._get_gradient(grad,scalegrad,fields_a,fields_f,
+        mp._get_gradient(grad,scalegrad,
+                            fields_a[0].swigobj,fields_a[1].swigobj,fields_a[2].swigobj,
```
oskooi, line 55: Misalignment of adjacent lines.

python/adjoint/wrapper.py
```diff
@@ -172,9 +167,12 @@ def _run_adjoint_simulation(self, monitor_values_grad):
                 'regions are present.')
         adjoint_sources = utils.create_adjoint_sources(self.monitors,
                                                        monitor_values_grad)
-        self.simulation.reset_meep()
+        # TODO refactor with optimization_problem.py #
+        self.simulation.restart_fields()
+        self.simulation.clear_dft_monitors()
         self.simulation.change_sources(adjoint_sources)
-        design_region_monitors = utils.install_design_region_monitors(
+        #                                            #
```
oskooi, line 174: Remove blank line?

src/meep.hpp
```diff
@@ -1142,6 +1142,11 @@ class dft_chunk {
   // the frequencies to loop_in_chunks
   std::vector<double> omega;
 
+  // decide whether to "persist" after fields class
+  // is deleted (e.g. for adjoint calculations)
+  bool persist = false;
+  bool expand = false;
```
oskooi, line 1148: Is it necessary to have a separate variable for `expand` since its value is (always?) identical to `persist` (i.e., whenever `persist` is `true` then `expand` must also be `true` and vice versa.). Is there a case where they could have different values?

python/adjoint/utils.py
```diff
@@ -149,18 +115,20 @@ def install_design_region_monitors(
...
+            [comp],
             frequencies,
             where=design_region.volume,
             yee_grid=True,
-            decimation_factor=decimation_factor
-        ) for design_region in design_regions
-    ]
+            decimation_factor=decimation_factor,
+            persist=True,
+            expand=True
```
oskooi, line 129: Why do these values override the defaults where `persist` and `expand` are both `False`? Maybe just change the defaults to `True`?

smartalecH: Because we don't want the default to be true.

python/visualization.py
smartalecH: It's important for debugging.

python/geom.py
```diff
@@ -541,7 +541,7 @@ def __init__(self,
                  medium2,
                  weights=None,
                  grid_type="U_DEFAULT",
-                 do_averaging=False,
+                 do_averaging=True,
```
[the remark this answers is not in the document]
smartalecH: Yes this is intentional.

I can add an additional test.

src/meep.hpp
smartalecH: Future applications may require this. I wanted to keep it as general as possible for that reason.

ianwilliamson: This code is already pretty complex, so my suggestion would be to remove one of the duplicated parameters. Having two variables that need to be kept in sync increases the surface area for potentially tricky bugs. If a new parameter is needed in the future, it can be added at that time.

python/tests/test_adjoint_jax.py
```diff
@@ -152,32 +152,6 @@ def test_mode_monitor_helpers(self):
         self.assertEqual(monitor_values.shape,
                          (len(self.monitors), len(self.frequencies)))
 
-    def test_design_region_monitor_helpers(self):
```
ianwilliamson, line 155: Why is this test case being deleted? If something changed with the underlying function, my suggestion would be to update this test.

smartalecH: The point of the test was to check to make sure the function that pulled the dft fields works properly.

We no longer pull the dft fields onto each proc.

python/geom.py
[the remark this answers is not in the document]
ianwilliamson: I have seen a significant performance regression when `do_averaging=True`, so it may be preferred to keep this off by default.

python/adjoint/wrapper.py
smartalecH: It's not a blank line. It's indicating what block of code corresponds to the TODO.

python/geom.py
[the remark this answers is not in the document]
smartalecH: You've seen a regression with this PR applied? Can you be more specific?

python/adjoint/wrapper.py
ianwilliamson: Sorry, I somehow missed this comment. If the the JAX test passes then I think it should be okay. The typical approach is to explicitly register the object type (or some class wrapping the object)) as a pytree.

smartalecH: No worries. It works now, thanks. 

The key here is to keep it within the class object itself since we want it to remain opaque to jax.

python/geom.py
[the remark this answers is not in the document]
ianwilliamson: The time taken to calculate the gradient is significantly higher with `do_averaging=True` than with `do_averaging=False`.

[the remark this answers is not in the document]
smartalecH: Yes, on `master`. The point of this PR is to alleviate that issue. Have you applied this PR and seen similar regression?

python/tests/test_adjoint_jax.py
ianwilliamson: Understood, but I would suggest updating this case to test the new functionality, rather than deleting it.

For example, you can test that you get the expected number of pointers, or that the fields they point to have the expected attributes (shape, dtype, etc.). For complex code like this it is very valuable to have test coverage like this, even for things which might seem trivial.

python/geom.py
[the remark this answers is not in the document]
ianwilliamson: I have not.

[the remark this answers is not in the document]
smartalecH: Just for reference, my own benchmarks (using both the meep tests and my own test files) show similar computational speeds when smoothing is off or on.

I'm sure you have many more tests you use to debug new features. If you want to run this on those, that would be great.

But as stated in #1903, the primary reason for disabling smoothing was because of the speed degradation. Since this PR supposedly resolves that, let's revert the default to true.

While I'm on my soapbox, I'll point out that the point of smoothing isn't so much to gain second-order acuraccy (the benefits of that seem to be relatively minor). Rather, it allows us to compute gradients at *any* beta (that's not true when smoothing is off, or for any traditional density method for that matter). So in short, enabling smoothing is very important from a practical design perspective.

stevengj:
As discussed this may still have a problematic corner case until #1895 is fixed.

If the user's `where` volume (where they added the DFT fields) extends right to the edge of a chunk, it could be that this edge will be assigned to the chunk that does *not* own those points, and the adjacent chunk will get *no* points at all (so you will have nothing to pad).

A potential workaround for this might be to pad the user's volume `where` as well, but this will screw up the edge weights unless you dig into `loop_in_chunks`

python/tests/test_adjoint_jax.py
smartalecH: Thanks. I added a test to check that the number of Pointers is correct.

src/meep.hpp
smartalecH: Fixed.

python/adjoint/utils.py
smartalecH: Fixed.

src/meepgeom.cpp
smartalecH: Fixed.

smartalecH:
@oskooi I'm unable to reproduce the CI test failure locally (`test_adjoint_jax.py` using MPI). Can you try on your machine?

The CI logs indicate that the fields get rather large:

```
FAIL: tests/test_adjoint_jax
============================

....WARNING:absl:No GPU/TPU found, falling back to CPU. (Set TF_CPP_MIN_LOG_LEVEL=0 and rerun for more info.)
WARNING:absl:No GPU/TPU found, falling back to CPU. (Set TF_CPP_MIN_LOG_LEVEL=0 and rerun for more info.)
......FF....
FAIL: test_wrapper_gradients_3_1500_1550bw_01relative_gaussian_port2 (__main__.WrapperTest)
Tests gradient from the JAX-Meep wrapper against finite differences [with _='1500_1550bw_01relative_gaussian_port2', frequencies=[0.6666666666666666, 0.6559139784946236, 0.6451612903225806], gaussian_rel_width=0.1, design_variable_fill_value=0.5, excite_port_idx=1].
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/parameterized/parameterized.py", line 533, in standalone_func
    return func(*(a + p.args), **p.kwargs)
  File "../../../python/tests/test_adjoint_jax.py", line 253, in test_wrapper_gradients
    epsilon=_TOL,
  File "/home/runner/work/meep/meep/build/meep-1.22.0/python/tests/utils.py", line 28, in assertClose
AssertionError: 7.541084240345138e+227 not less than or equal to 1.8852710600862845e+226 : 

Ran 8 tests in 87.057s

FAILED (failures=1)
```

oskooi:
I was able to reproduce both test failures: (1) line 534 of `test_adjoint_solver.py` and (2) line 250 of `test_adjoint_jax.py`. Note that these failing tests are using *single*-precision floating point for the fields and materials arrays.

src/dft.cpp
```diff
@@ -289,30 +306,52 @@ double fields::dft_norm() {
...
+  const size_t Nomega = omega.size();
+  /* looping over chunks that have been "expanded"
+  for adjoint calculations requires some care. Namely,
+  we want to make sure we don't double count the padding
+  and can replicate results with different chunk combinations.
+  */
+  if (persist) {
+    grid_volume subgv = fgv.subvolume(is,ie,c);
+    LOOP_OVER_IVECS(fgv, is_old, ie_old, idx) {
+      idx_dft = subgv.index(c,fgv.iloc(c,idx));
```
stevengj, line 336:
```suggestion
    LOOP_OVER_IVECS(subgv, is_old, ie_old, idx) {
      idx_dft = idx;
```

src/dft.cpp
```diff
@@ -289,30 +306,52 @@ double fields::dft_norm() {
...
-    idx_dft++;
+  size_t idx_dft;
+  const size_t Nomega = omega.size();
+  /* looping over chunks that have been "expanded"
+  for adjoint calculations requires some care. Namely,
+  we want to make sure we don't double count the padding
+  and can replicate results with different chunk combinations.
+  */
+  if (persist) {
+    grid_volume subgv = fgv.subvolume(is,ie,c);
```
stevengj, line 334: check: the number of points in `subgv` should match the length of the `dft` array?

src/dft.cpp
```diff
@@ -289,30 +306,52 @@ double fields::dft_norm() {
...
+          sum += sqr(dft[Nomega * idx_dft + i]);
+    } 
+  } 
+  /* note we place the if outside of the
+  loop to avoid branching. This routine gets
+  called a lot, so let's try to stay efficient
+  (at the expense of uglier code).
+   */
+  else{
+    LOOP_OVER_IVECS(fgv, is, ie, idx) {
```
stevengj, line 348: Similar to above.

stevengj: Actually it's okay because we use `IVEC_LOOP_COUNTER` here … so you don't need `subgv` at all.

src/meepgeom.cpp
```diff
@@ -2853,188 +2853,179 @@ void material_grids_addgradient_point(double *v, vector3 p, double scalegrad, ge
...
+        for (int ci_forward=0; ci_forward<3; ci_forward++){
+          size_t num_f_chunks = forward_dft_chunks[ci_forward].size();
+          if ((num_f_chunks == 0) || (cur_chunk>=num_f_chunks)) continue;
+          meep::dft_chunk* fwd_chunk = forward_dft_chunks[ci_forward][cur_chunk];
+          meep::component forward_c = fwd_chunk->c;
+          meep::grid_volume gv_fwd = gv.subvolume(fwd_chunk->is,fwd_chunk->ie,forward_c);
+          
+          // loop over each point of interest
+          LOOP_OVER_IVECS(gv,adj_chunk->is_old,adj_chunk->ie_old,idx){
+            meep::ivec ip = gv.iloc(adjoint_c,idx);
```
stevengj, line 2922:
```suggestion
            IVEC_LOOP_ILOC(gv, ip);
```

stevengj: Probably need to change https://github.com/NanoComp/meep/blob/9c5582ed99251bec706400d3639f04a347d985fb/src/meep/vec.hpp#L352 to use `meep::ivec`

src/meepgeom.cpp
```diff
@@ -2853,188 +2853,179 @@ void material_grids_addgradient_point(double *v, vector3 p, double scalegrad, ge
...
+        // loop over forward components
+        for (int ci_forward=0; ci_forward<3; ci_forward++){
+          size_t num_f_chunks = forward_dft_chunks[ci_forward].size();
+          if ((num_f_chunks == 0) || (cur_chunk>=num_f_chunks)) continue;
+          meep::dft_chunk* fwd_chunk = forward_dft_chunks[ci_forward][cur_chunk];
+          meep::component forward_c = fwd_chunk->c;
+          meep::grid_volume gv_fwd = gv.subvolume(fwd_chunk->is,fwd_chunk->ie,forward_c);
+          
+          // loop over each point of interest
+          LOOP_OVER_IVECS(gv,adj_chunk->is_old,adj_chunk->ie_old,idx){
```
stevengj, line 2921:
```suggestion
          LOOP_OVER_IVECS(gv_adj,adj_chunk->is_old,adj_chunk->ie_old,idx_adj){
```

src/meepgeom.cpp
```diff
@@ -2853,188 +2853,179 @@ void material_grids_addgradient_point(double *v, vector3 p, double scalegrad, ge
...
+            meep::ivec ip = gv.iloc(adjoint_c,idx);
+            size_t idx_adj =  gv_adj.index(adjoint_c,ip); 
+            if (idx_adj >= adj_chunk->N) meep::abort("index %d larger than limit %d\n",idx_adj,adj_chunk->N);    
+            meep::vec p  = gv.loc(adjoint_c,idx);
```
stevengj, line 2925: Same thing: use `IVEC_LOOP_LOC(gv_adj, p);`

src/meepgeom.cpp
```diff
@@ -2853,188 +2853,179 @@ void material_grids_addgradient_point(double *v, vector3 p, double scalegrad, ge
...
+            std::complex<meep::realnum> adj = adj_chunk->dft[nf*idx_adj+f_i];
+            material_type md;
+            geps->get_material_pt(md, p);
+            if (!md->trivial) adj *= cond_cmp(adjoint_c,p,frequencies[f_i], geps);
```
stevengj, line 2929: Maybe add a comment here.

src/meepgeom.cpp
```diff
@@ -2853,188 +2853,179 @@ void material_grids_addgradient_point(double *v, vector3 p, double scalegrad, ge
...
+  /* ------------------------------------------------------------ */
+  // Broadcast results
+  /* ------------------------------------------------------------ */
+  meep::sum_to_all(v_local, v, ng*nf);
+
+  /* ------------------------------------------------------------ */
+  // cleanup
+  /* ------------------------------------------------------------ */
+  // clear the array used for local sum to all
+  delete[] v_local;
```
stevengj, line 3018: Could use `std::unique_ptr` instead.

src/meepgeom.cpp
```diff
@@ -2853,188 +2853,180 @@ void material_grids_addgradient_point(double *v, vector3 p, double scalegrad, ge
...
+          if ((num_f_chunks == 0) || (cur_chunk>=num_f_chunks)) continue;
+          meep::dft_chunk* fwd_chunk = forward_dft_chunks[ci_forward][cur_chunk];
+          meep::component forward_c = fwd_chunk->c;
+          meep::grid_volume gv_fwd = gv.subvolume(fwd_chunk->is,fwd_chunk->ie,forward_c);
+          
+          // loop over each point of interest
+          LOOP_OVER_IVECS(gv_adj,adj_chunk->is_old,adj_chunk->ie_old,idx_adj){
+            double cyl_scale;   
+            IVEC_LOOP_ILOC(gv_adj, ip);
+            IVEC_LOOP_LOC(gv, p);
```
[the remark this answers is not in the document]
smartalecH: oof. good catch.

smartalecH:
I can't replicate the test failures on my machine. @oskooi can you try one more time on your machine?

Note I've already bumped the tolerance, but am still seeing errors.

oskooi:
With the most recent commit (9aa8296), there are two separate CI test failures: 
1. `test_adjoint_solver.py` (single precision)
2. `test_adjoint_cyl.py` (double precision)

On my local machine, I was only able to reproduce (1).

mawc2019:
This branch sometimes causes a segmentation fault in the adjoint run. An example is as follows. Before this PR, there is no segmentation fault. After this PR, the segmentation fault appears.

```
import meep as mp
import meep.adjoint as mpa
import numpy as np

Si = mp.Medium(index=3.48)
Air = mp.Medium(index=1)

resolution,design_resolution = 40,40

design_width,design_height = 1.0,0.4
Lx,Ly = design_width,4
dpml = 1.5

monitor_height,monitor_width = 1.5,0.8

pml_layers = [mp.PML(dpml)]
cell_size = mp.Vector3(Lx+2*dpml,Ly+2*dpml)

fcen,fwidth = 1/1.55,0.2/1.55

source_center,source_size  = mp.Vector3(0,-Ly/4),mp.Vector3(Lx,0)
source = [mp.EigenModeSource(src,eig_band=1,size=source_size,center=source_center)]

design_variables = mp.MaterialGrid(mp.Vector3(Nx,Ny),Air,Si,grid_type='U_MEAN')
design_region = mpa.DesignRegion(design_variables,volume=mp.Volume(center=mp.Vector3(0,design_height/2),size=mp.Vector3(design_width,design_height)))

sim = mp.Simulation(cell_size=cell_size,boundary_layers=pml_layers,geometry=geometry,sources=source,eps_averaging=False,resolution=resolution)

TE0 = mpa.EigenmodeCoefficient(sim,mp.Volume(center=mp.Vector3(0,monitor_height),size=mp.Vector3(monitor_width,0)),mode=1)
ob_list = [TE0]

def J(alpha):
    return npa.abs(alpha) ** 2

    simulation = sim,objective_functions = J,objective_arguments = ob_list,
    design_regions = [design_region],fcen=fcen,df = 0,nf = 1,decay_by = 1e-5)

x0 = 0.5*np.ones(Nx*Ny)
opt.update_design([x0])
f0, dJ_deps = opt()
```

mawc2019:
Before this PR, although there is no segmentation fault, the finite-difference and adjoint gradients are inconsistent in the [above example](https://github.com/NanoComp/meep/pull/1855#issuecomment-1103429038). With the following code added to the end,
```
db = 1e-4
choose = 5
np.random.seed(1905)

g_adjoint = dJ_deps.flatten()

(m, b) = np.polyfit(np.squeeze(g_discrete), g_adjoint[idx], 1)
```
the results are printed as
```
Finite-difference gradient:  [ 0.03372703 -1.49253234 -0.21201943  1.02494487  1.02513747]
Adjoint gradient:  [2.9224977  1.75185184 3.31514926 1.11873224 1.11873224]
```

mawc2019:
After the PR, in addition to the above issue in the `EigenmodeCoefficient` adjoint solver, sometimes the finite-difference and adjoint gradients are inconsistent in the `FourierFields` adjoint solver. The example code is as follows.
```
import meep as mp
import meep.adjoint as mpa
import numpy as np

resolution,design_resolution = 40,40
Lx,Ly,pml_size = 2,4,1

source = [mp.EigenModeSource(src,eig_band=1,size=source_size,center=source_center)]

sim = mp.Simulation(

nf = mpa.FourierFields(sim,mp.Volume(center=mp.Vector3(0,design_height+0.2),size=mp.Vector3(design_width,0)),mp.Ex,yee_grid=True)
ob_list = [nf]

def J(near_data):
    return npa.abs(near_data[0,0])**2

    fcen=fcen,df = 0,nf = 1,decay_by = 1e-6)

x0 = 0.5*np.ones(Nx*Ny)
opt.update_design([x0])
f0, g_adjoint = opt()

db, choose = 1e-4, 5
np.random.seed(20211204)

g_adjoint = g_adjoint.flatten()

```
Before this PR, the results are
```
Finite-difference gradient:  [-41.65764109 -36.27053419  22.04881095  38.55970822 -29.69618643]
Adjoint gradient:  [-41.65784098 -36.27197005  22.0480579   38.55931979 -29.69604378]
```
After this PR, the results are
```
Finite-difference gradient:  [-29.95933836 -39.96501542  14.17651164  26.06499857 -21.2956607 ]
Adjoint gradient:  [-49.61541877 -62.96284535  26.39349287  45.95704617 -35.40822268]
```

PR #1857, opened 2021-12-19
unit test for conductivity

oskooi:
#1804 added a "damping" property to the `MaterialGrid` which is based on the σ<sub>D</sub> feature used for [modeling complex ε over a narrow bandwidth](https://meep.readthedocs.io/en/latest/Materials/#conductivity-and-complex). #1830 subsequently added a unit test for this feature to the adjoint-solver module.  However, currently there is no standalone unit test for the σ<sub>D</sub> feature outside of the adjoint-solver module.

This PR adds a new unit test for σ<sub>D</sub> which involves verifying that the power transmitted through a propagation distance *d* of a single-mode waveguide with attenuation coefficient ɑ is *P*<sub>0</sub>exp(-ɑd) where *P*<sub>0</sub> is the incident power from the source in a lossless waveguide. The tricky part is getting the units right when converting ɑ into a complex ε (which is then used to compute σ<sub>D</sub> using a formula). The relationship between ɑ and  ε is derived analytically in this [Wikipedia article](https://en.wikipedia.org/wiki/Mathematical_descriptions_of_opacity). This test can also be used as a reference for users wishing to import experimental material properties into Meep from [ellipsometry measurements](https://en.wikipedia.org/wiki/Ellipsometry).

There is, however, one important caveat: Meep's mode launcher (currently) only supports *lossless* waveguides which means there are some inaccuracies when trying to launch a mode of a lossy system. For now, as long as the loss is relatively small (i.e., attenuation coefficient of a few dB per cm), this error is (mostly) negligible and the simulated results can be validated against the correct analytic result.


smartalecH:
> #1804 added a "damping" property to the MaterialGrid which is based on the σD feature used for modeling complex ε over a narrow bandwidth. #1830 

Note that the damping term added in #1804 is an *artificial* damping term, not meant to model any actual loss from materials. In fact, as the material becomes binary (e.g `u=0` or `u=1`) the damping term goes to 0 (i.e. disappears).

In other words, I'm not sure a test outside the adjoint solver is needed.

If one wants to interpolate between materials that have loss, this isn't the way to do it. We already have built-in support for materials that are lossy due to conductivities.

oskooi:
> I'm not sure a test outside the adjoint solver is needed.

As I described above, this is meant to be a standalone test for *just* the `D_conductivity` feature (currently missing) which involves comparing to a known analytic result. This test is also not about interpolating between materials that have loss which we can (and probably should) add separately.

python/tests/test_conductivity.py
```diff
@@ -0,0 +1,96 @@
...
+        #       The imaginary part of ε is ignored and the launched
+        #       waveguide mode is therefore inaccurate. For small values
+        #       of imag(ε) (which is proportional to att_coeff), this
+        #       inaccuracy tends to be insignificant.
+        sources = [mp.EigenModeSource(src=mp.GaussianSource(fsrc,fwidth=0.2*fsrc),
+                                      center=mp.Vector3(-5.),
+                                      size=mp.Vector3(y=10.),
+                                      eig_parity=mp.EVEN_Y+mp.ODD_Z)]
+
+        # ref: https://en.wikipedia.org/wiki/Mathematical_descriptions_of_opacity
```
stevengj, line 35:
```suggestion
        # ref: https://en.wikipedia.org/wiki/Mathematical_descriptions_of_opacity
        # Note that this is the loss of a planewave, which is only approximately
        # the loss of a waveguide mode.  In principle, we could compute the latter
        # semi-analytically if we wanted to run this unit test to greater accuracy
        # (e.g. to test convergence with resolution).
```

PR #1869, opened 2021-12-26
Ensure sim.plot2D() contour option plots all levels

simbilod:
The default ```matplotlib.pyplot.contour``` function does not always choose levels appropriately if there are both small and large contrasts. See for instance the example from [#1827](https://github.com/NanoComp/meep/pull/1827) ([comment](https://github.com/NanoComp/meep/pull/1827#issuecomment-972430404)) :

plot2D_test

plot2D_test_contour

While the user can always provide the appropriate ```levels``` options, we can ensure the desired behaviour by hardcoding the ```levels``` parameter to look up the unique epsilon values from the geometry (which are used for plotting instead of a multitude of subpixel-averaged values as of [#1827](https://github.com/NanoComp/meep/pull/1827)):

plot2D_test_contour_levels

### DIFF
--- a/python/visualization.py
+++ b/python/visualization.py
@@ -402,7 +402,7 @@ def plot_eps(sim, ax, output_plane=None, eps_parameters=None, frequency=None):
 
     if mp.am_master():
         if eps_parameters['contour']:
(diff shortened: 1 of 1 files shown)


oskooi:
LGTM.

PR #1872, opened 2021-12-29
fix memory leaks in structure and fields load during checkpointing

oskooi:
Fixes two memory leaks introduced in #1738 and reported by the [address sanitizer](https://gist.github.com/ahoenselaar/13d0b764bb38bbab8b579ed414c49592).

cc @kkg4theweb

### DIFF
--- a/src/fields_dump.cpp
+++ b/src/fields_dump.cpp
@@ -189,7 +189,7 @@ void fields::load_fields_chunk_field(h5file *h5f, bool single_parallel_file,
           } else {
             if (n != ntot)
               meep::abort("grid size mismatch %zd vs %zd in fields::load", n, ntot);
-            *f = new realnum[ntot];
+            if (!*f) *f = new realnum[ntot];
             my_ntot += ntot;
           }
         }
--- a/src/fields_dump.cpp
+++ b/src/fields_dump.cpp
@@ -189,7 +189,12 @@ void fields::load_fields_chunk_field(h5file *h5f, bool single_parallel_file,
           } else {
             if (n != ntot)
               meep::abort("grid size mismatch %zd vs %zd in fields::load", n, ntot);
-            if (!(*f)) *f = new realnum[ntot];
+            // here we need to allocate the fields array for H in the PML region
+            // because of H = B in fields_chunk::alloc_f whereby H is lazily
+            // allocated in fields_chunk::update_eh during the first timestep
+            const direction d_c = component_direction(c);
+            if (!(*f) || (*f && is_magnetic(component(c)) && chunks[i]->s->sigsize[d_c] > 1))
+              *f = new realnum[ntot];
             my_ntot += ntot;
           }
         }
--- a/src/fields_dump.cpp
+++ b/src/fields_dump.cpp
@@ -189,7 +189,11 @@ void fields::load_fields_chunk_field(h5file *h5f, bool single_parallel_file,
           } else {
             if (n != ntot)
               meep::abort("grid size mismatch %zd vs %zd in fields::load", n, ntot);
-            if (!*f) *f = new realnum[ntot];
+            if (*f) {
+              delete[] * f;
+              *f = NULL;
+            }
+            *f = new realnum[ntot];
             my_ntot += ntot;
           }
         }
--- a/src/fields_dump.cpp
+++ b/src/fields_dump.cpp
@@ -184,16 +184,12 @@ void fields::load_fields_chunk_field(h5file *h5f, bool single_parallel_file,
           size_t n = num_f[(chunk_i * NUM_FIELD_COMPONENTS + c) * 2 + d];
           realnum **f = field_ptr_getter(chunks[i], c, d);
           if (n == 0) {
-            delete[] * f;
+            delete[] *f;
             *f = NULL;
           } else {
             if (n != ntot)
               meep::abort("grid size mismatch %zd vs %zd in fields::load", n, ntot);
-            if (*f) {
-              delete[] * f;
-              *f = NULL;
-            }
-            *f = new realnum[ntot];
+            if (!(*f)) *f = new realnum[ntot];
             my_ntot += ntot;
           }
         }
--- a/src/structure_dump.cpp
+++ b/src/structure_dump.cpp
@@ -604,7 +604,11 @@ void structure::load(const char *filename, bool single_parallel_file) {
           }
           else {
             if (n != ntot) meep::abort("grid size mismatch %zd vs %zd in structure::load", n, ntot);
-            if (!chunks[i]->chi1inv[c][d]) chunks[i]->chi1inv[c][d] = new realnum[ntot];
+            if (chunks[i]->chi1inv[c][d]) {
+              delete[] chunks[i]->chi1inv[c][d];
+              chunks[i]->chi1inv[c][d] = NULL;
+            }
+            chunks[i]->chi1inv[c][d] = new realnum[ntot];
             my_ntot += ntot;
           }
         }
--- a/src/structure_dump.cpp
+++ b/src/structure_dump.cpp
@@ -604,11 +604,7 @@ void structure::load(const char *filename, bool single_parallel_file) {
           }
           else {
             if (n != ntot) meep::abort("grid size mismatch %zd vs %zd in structure::load", n, ntot);
-            if (chunks[i]->chi1inv[c][d]) {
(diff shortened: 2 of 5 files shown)


src/fields_dump.cpp
```diff
@@ -189,6 +189,10 @@ void fields::load_fields_chunk_field(h5file *h5f, bool single_parallel_file,
           } else {
             if (n != ntot)
               meep::abort("grid size mismatch %zd vs %zd in fields::load", n, ntot);
+            if (*f) {
+              delete[] * f;
+              *f = NULL;
+            }
```
stevengj, line 195:
```suggestion
            delete[] *f;
```
According to the C++ standard, `delete[]` should be safe (do nothing) for a `NULL` pointer.  And there's no point in assigning it to `NULL` since we immediately re-assign `*f` on the next line.

Similarly for the other part of this PR.

src/structure_dump.cpp
```diff
@@ -604,6 +604,10 @@ void structure::load(const char *filename, bool single_parallel_file) {
           }
           else {
             if (n != ntot) meep::abort("grid size mismatch %zd vs %zd in structure::load", n, ntot);
+            if (chunks[i]->chi1inv[c][d]) {
+              delete[] chunks[i]->chi1inv[c][d];
+              chunks[i]->chi1inv[c][d] = NULL;
+            }
```
stevengj, line 610:
```suggestion
            delete[] chunks[i]->chi1inv[c][d];
```

oskooi:
While this PR with the latest commit (e2eae75) fixes the memory leak in the C++ unit test `dump_load.cpp`, it is causing the Python unit test `test_dump_load.py` to fail. The failing Python unit test involves time stepping the fields in a reference simulation until t=15, saving the fields at t=15, and running  until t=20. The fields at some arbitrary monitor point are saved at every 1 time units and the results are compared to a second test simulation which involves loading the fields from the reference simulation at t=15. What is strange is that both simulations produce identical results for t=15 and t=16 but then the test simulation starts to diverge.

```
Saving temp files to dir: /tmp/meepTwK9OZ
Running test_load_dump_fields_2d
-----------
Initializing structure...
time for choose_chunkdivision = 0.000271299 s
Working in 2D dimensions.
Computational cell is 5 x 5 x 0 with resolution 50
     block, center = (0,0,0)
          size (1,1,1e+20)
          dielectric constant epsilon diagonal = (10.24,10.24,10.24)
     block, center = (1,0,0)
          size (1,1,1e+20)
          dielectric constant epsilon diagonal = (13,13,13)
time for set_epsilon = 0.0707238 s
-----------
run 0 finished at t = 15.0 (1500 timesteps)
creating epsilon from file "/tmp/meepTwK9OZ/test_load_dump_fields/structure.h5" (1)...
Dumped structure to file: /tmp/meepTwK9OZ/test_load_dump_fields/structure.h5 (True)
creating fields output file "/tmp/meepTwK9OZ/test_load_dump_fields/fields.h5" (1)...
Dumped fields to file: /tmp/meepTwK9OZ/test_load_dump_fields/fields.h5 (True)
run 1 finished at t = 20.0 (2000 timesteps)
-----------
Initializing structure...
time for choose_chunkdivision = 9.4034e-05 s
Working in 2D dimensions.
Computational cell is 5 x 5 x 0 with resolution 50
time for set_epsilon = 0.0330667 s
-----------
Working in 2D dimensions.
Computational cell is 5 x 5 x 0 with resolution 50
time for set_epsilon = 0.0333062 s
-----------
reading epsilon from file "/tmp/meepTwK9OZ/test_load_dump_fields/structure.h5" (1)...
Loaded structure from file: /tmp/meepTwK9OZ/test_load_dump_fields/structure.h5 (True)
reading fields from file "/tmp/meepTwK9OZ/test_load_dump_fields/fields.h5" (1)...
Loaded fields from file: /tmp/meepTwK9OZ/test_load_dump_fields/fields.h5 (True)
on time step 1500 (time=15), 2.0487 s/step
run 0 finished at t = 20.0 (2000 timesteps)
F
FAIL: test_load_dump_fields_2d (__main__.TestLoadDump)
  File "test_dump_load.py", line 257, in test_load_dump_fields_2d
    self._load_dump_fields_2d()
  File "test_dump_load.py", line 251, in _load_dump_fields_2d
    self.assertAlmostEqual(ref_field_points[t], v)
AssertionError: 0.3447239547967911 != 0.34472334384918213 within 7 places

Ran 1 test in 0.464s

FAILED (failures=1)
```
Here are the results from the two runs.
```
ref:, 15.0, 0.42263978719711304
ref:, 16.0, 0.3860078901052475
ref:, 17.0, 0.3447239547967911
ref:, 18.0, 0.30143244564533234
ref:, 19.0, 0.25082115083932877
ref:, 20.0, 0.19293657690286636

test:, 15.0, 0.42263978719711304
test:, 16.0, 0.3860078901052475
test:, 17.0, 0.34472334384918213 ***
test:, 18.0, 0.3033006340265274 ***
test:, 19.0, 0.2514386847615242 ***
test:, 20.0, 0.18865110725164413 ***
```

src/fields_dump.cpp
```diff
@@ -184,12 +184,12 @@ void fields::load_fields_chunk_field(h5file *h5f, bool single_parallel_file,
...
           realnum **f = field_ptr_getter(chunks[i], c, d);
           if (n == 0) {
-            delete[] * f;
+            delete[] *f;
             *f = NULL;
           } else {
             if (n != ntot)
               meep::abort("grid size mismatch %zd vs %zd in fields::load", n, ntot);
-            *f = new realnum[ntot];
+            if (!(*f)) *f = new realnum[ntot];
```
stevengj, line 192:
```suggestion
            if (*f)
                memset((void*) *f, 0, sizeof(realnum)*ntot);
            else
                *f = new realnum[ntot];
```

src/structure_dump.cpp
```diff
@@ -604,7 +604,7 @@ void structure::load(const char *filename, bool single_parallel_file) {
           }
           else {
             if (n != ntot) meep::abort("grid size mismatch %zd vs %zd in structure::load", n, ntot);
-            chunks[i]->chi1inv[c][d] = new realnum[ntot];
+            if (!chunks[i]->chi1inv[c][d]) chunks[i]->chi1inv[c][d] = new realnum[ntot];
```
stevengj, line 607:
```suggestion
            if (chunks[i]->chi1inv[c][d])
                memset((void*) chunks[i]->chi1inv[c][d], 0, sizeof(realnum)*ntot);
            else
                chunks[i]->chi1inv[c][d] = new realnum[ntot];
```

oskooi:
I think I have tracked down the cause of the two failing tests. There seems to be a bug in saving/load the fields in the **PML regions**. This can be demonstrated by simply removing `pml_layers` from the `Simulation` object which causes both of the failing tests shown below to pass. It's unclear though why this is related to this PR.

https://github.com/NanoComp/meep/blob/e41a735460f41839f1bb734fecf65c6e37715cf4/python/tests/test_dump_load.py#L180-L196

https://github.com/NanoComp/meep/blob/e41a735460f41839f1bb734fecf65c6e37715cf4/python/tests/test_dump_load.py#L261-L277

oskooi:
This is finally working. The bug was a bit subtle to track down and has to do with the fact that during the fields initialization of `fields_chunk::alloc_f` *H* = *B* in the PML region:

https://github.com/NanoComp/meep/blob/e41a735460f41839f1bb734fecf65c6e37715cf4/src/fields.cpp#L475-L483

*H* is lazily allocated during the *first* timestep in `fields_chunk::update_eh`:

https://github.com/NanoComp/meep/blob/e41a735460f41839f1bb734fecf65c6e37715cf4/src/update_eh.cpp#L166-L172

The problem here is that when the saved fields are being loaded from the HDF5 file, *H* = *B* (because `fields::step()` and thus `fields_chunk::update_eh` has not yet been invoked) and thus the saved *H* fields were never actually being loaded into a separate *H* fields array. The *H* fields in the PML region were being assigned to the saved *B* fields in the PML region which is why the results were slightly different from the reference simulation. The fix involved manually allocating the fields array for *H* in the PML region within `fields::load_fields_chunk_field` (rather than in `fields_chunk::update_eh`) so that the saved *H* fields from could be then be correctly loaded from the file into the *H* fields array.

I added a comment to explain all this in `fields::load_fields_chunk_field`.

PR #1873, opened 2021-12-29
Add cylindrical coordinates support for `plot2d`

smartalecH:
Basic support for visualization of cylindrical coordinates.

Example image of (nonoptimized) metalens when plotting with `master` (incorrect)

image

Correct plot using this PR:

image

With fields

image

(cc @mochen4)

### DIFF
--- a/python/visualization.py
+++ b/python/visualization.py
@@ -189,14 +189,19 @@ def get_2D_dimensions(sim, output_plane):
     elif sim.output_volume:
         plane_center, plane_size = mp.get_center_and_size(sim.output_volume)
     else:
-        plane_center, plane_size = (sim.geometry_center, sim.cell_size)
+        if (sim.dimensions == mp.CYLINDRICAL) or sim.is_cylindrical:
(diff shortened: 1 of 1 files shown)


oskooi (commented):
It doesn't seem that this PR supports plotting ε because `plot_eps` is using `get_epsilon_grid` to obtain the ε data which currently only supports Cartesian coordinates:

https://github.com/NanoComp/meep/blob/e28f3916dd9b486f7d105430780d94861fe88460/python/visualization.py#L401

smartalecH:
Seems to work just fine based on the above example...

Why do you think it doesn't support cylindrical coordinates?

It treats the geometry definition as a simple 2d profile in the XZ domain (it thinks it's Cartesian, which is just fine for our purposes. Obviously we'd have issues with asking for a full 3D revolved rendition).

oskooi:
I tested this PR using the tutorial example for cylindrical coordinates using a [binary-phase zone plate](https://meep.readthedocs.io/en/latest/Python_Tutorials/Cylindrical_Coordinates/#focusing-properties-of-a-binary-phase-zone-plate). It appears to be working although the location of the near-field monitor (blue line) appears slightly off.

zone_plate_cyl_plot2D

PR #1881, opened 2022-01-04
restrict weights parameter of MaterialGrid to range [0, 1]

oskooi:
The `weights` array parameter of the `MaterialGrid` class is required to have values in the range [0, 1]. However, this does not seem to be actually enforced anywhere even though it is assumed to be true for the `epsilon_material_grid` function which computes the interpolated material quantity given an element `u` of the `weights` array:

https://github.com/NanoComp/meep/blob/a1884b25f141c131cb45c4ee7b1e67abe19a33c3/src/meepgeom.cpp#L570-L572 

https://github.com/NanoComp/meep/blob/a1884b25f141c131cb45c4ee7b1e67abe19a33c3/src/meepgeom.cpp#L534-L547

Not ensuring a valid user input for the `weights` array can produce unexpected results. As an example, the three unit tests for the adjoint solver (`test_adjoint_solver.py`, `test_adjoint_jax.py`, `test_adjoint_cyl.py`) all involve applying a small random perturbation to the `weights` array without checking that the results are valid.

This PR adds a check to the `MaterialGrid` class in Python to make sure that the values of the `weights` array are valid. If they are not, it prints a warning and also clips the values to lie in the range [0, 1] which may not necessarily be the correct thing to do (i.e., it may be better to abort and leave it up to the user to provide valid input). In any event, with this change, `test_adjoint_jax.py` is *failing* precisely because of a `weights` violation which requires further investigation.

### DIFF
--- a/python/geom.py
+++ b/python/geom.py
@@ -552,9 +552,10 @@ class MaterialGrid(object):
 
         Elements of the `weights` array must be in the range [0,1] where 0 is `medium1` and 1 is `medium2`.
         The `weights` array is used to define a linear interpolation from `medium1` to `medium2`.
-        Two material types are supported: (1) frequency-independent isotropic $\\varepsilon$ or $\\mu$
-        and (2) `LorentzianSusceptibility` (only the `sigma` parameter is interpolated). `medium1` and
-        `medium2` must both be the same type. The materials are [bilinearly interpolated](https://en.wikipedia.org/wiki/Bilinear_interpolation)
+        Two material types are supported: (1) frequency-independent isotropic $\\varepsilon$ (`epsilon_diag` and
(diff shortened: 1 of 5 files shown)


stevengj:
CI failing?

python/geom.py
```diff
@@ -615,7 +628,10 @@ def update_weights(self,x):
         """
         if x.size != self.num_params:
             raise ValueError("weights of shape {} do not match user specified grid dimension: {}".format(self.weights.size,self.grid_size))
-        self.weights[:]=x.flatten().astype(np.float64)
+        if np.amin(x) < 0. or np.amax(x) > 1.:
```
ianwilliamson, line 631: Rather than duplicating the code used for the check, you should create a method

oskooi: Done.

PR #1882, opened 2022-01-04
Ex field plot

mochen4:
``mp.Ex`` is 0, so ``if fields`` is ``false`` and ``mp.Ex`` fields wouldn't be plotted in plot2D


stevengj:
Good catch.  Python's "implicit boolean" feature seems to be an invitation for bugs of this sort.

PR #1883, opened 2022-01-05
connectivity constraint fix and example

mochen4:
Fix a few issues in the connectivity constraint and add an [example](https://github.com/mochen4/meep/blob/connectivity_fix/python/examples/adjoint_optimization/ConnectivityConstraint.ipynb).

It seems that for a structure like this (Dirichlet BC on the top, i.e. components should want to connect to the top):
image
the gradients look like this:
image

In other words, it inclines to bridge gaps instead of eliminating isolated islands; on the other hands, small islands could be captured and diminished by the geometric constraint.


smartalecH:
How big does that island need to be before the constraint starts building a bridge for it?

Does the island have to "touch" and edge (ie so that it's not truly an island)?

I'm just thinking there are probably plenty of island possibilities that will still satisfy the geometric constraint, but may not encourage any "bridging".

mochen4:
> How big does that island need to be before the constraint starts building a bridge for it?
> 

The island can be really small. In fact, in the pictures above, the island on the upper left is just 1 pixel by 1 pixel, and there is already some negative gradients on the top and left that try to bridge it to the top boundary and the right structure.

On the other hand, I should mention that the void has a small nonzero value 0.0001. In the actual optimization, the material density is rarely exactly zero anyway. If it is exactly zero, there won't be any bridges; the structure is completely ignorant of the island, like this:
image


> Does the island have to "touch" and edge (ie so that it's not truly an island)?
>

It seems irrelevant whether touching an edge matters.

Issue #1861, opened 2021-12-22
Out-of-bounds indexing in `material_grids_addgradient`


ahoenselaar:
Calculated indices `fwd1_idx` and `fwd2_idx` into the DFT field arrays are, at times, negative and cause out-of-bounds data to be read.

Repro:
Instrument `material_grids_addgradient` as follows:

```
...
fwd1_idx = get_idx_from_ivec(gv.dim, start_ivec, the_stride, fwd_p);
if (fwd1_idx < 0) { meep::abort("Negative index for first eps node fwd1_idx=%ld", fwd1_idx); }
fwd1 = 0.5 * meep::cdouble(GET_FIELDS(fields_f, ci_forward, f_i, fwd1_idx));
fwd2_idx = get_idx_from_ivec(gv.dim, start_ivec, the_stride, fwd_pf);
if (fwd2_idx < 0) { meep::abort("Negative index for first eps node fwd2_idx=%ld", fwd2_idx); }
fwd2 = 0.5 * meep::cdouble(GET_FIELDS(fields_f, ci_forward, f_i, fwd2_idx));
fwd_avg = fwd1 + fwd2;
meep::vec eps1 = gv[ieps1];
cyl_scale = (gv.dim == meep::Dcyl) ? eps1.r() : 1;
material_grids_addgradient_point(
    v+ng*f_i, vec_to_vector3(eps1), scalegrad*cyl_scale, geps,
    adjoint_c, forward_c, fwd_avg, 0.5*adj, frequencies[f_i], gv, du);

//operate on the second eps node
fwd1_idx = get_idx_from_ivec(gv.dim, start_ivec, the_stride, fwd_pa);
if (fwd1_idx < 0) { meep::abort("Negative index for second eps node fwd1_idx=%ld", fwd1_idx); }
fwd1 = 0.5 * meep::cdouble(GET_FIELDS(fields_f, ci_forward, f_i, fwd1_idx));
fwd2_idx = get_idx_from_ivec(gv.dim, start_ivec, the_stride, fwd_paf);
if (fwd2_idx < 0) { meep::abort("Negative index for second eps node fwd2_idx=%ld", fwd2_idx); }
...
```

then run the Python test suite and observe failures in adjoint tests.

Without the introduction of safer indexing primitives and better test coverage, bugs of this type are going to occur over and over.

smartalecH:
What's an example of a "better indexing primitive" we can use to avoid issues like this in the future?

[...]

PR #1886, opened 2022-01-07
compute adjoint gradient using restriction/interpolation only when anisotropic materials are present

oskooi:
#1780 added support for subpixel smoothing of the adjoint gradients in which ε is an anisotropic tensor. However, in the current implementation of `material_grids_addgradient` of `meepgeom.cpp`, regardless of the value of the `do_averaging` property of the `MaterialGrid` object, the adjoint gradient is *always* computed for the off-diagonal components of ε which is a bug. Also, as reported in #1861, computing these off-diagonal elements involves a memory violation in which the array for the forward fields is read outside of its bounds. The memory violation issue is exacerbated by the fact that `do_averaging=True` is the default setting in `MaterialGrid`:

https://github.com/NanoComp/meep/blob/efe3b4b1fff538cf1b1ad2e34bb47c7a3ddfb28a/python/geom.py#L538-L547

This PR provides a small fix to `material_grids_addgradient` to ensure that when `do_averaging=False` only the diagonal ε elements are computed (and the memory violation is therefore avoided).

### DIFF
--- a/src/meepgeom.cpp
+++ b/src/meepgeom.cpp
@@ -2965,8 +2965,14 @@ void material_grids_addgradient(double *v, size_t ng, std::complex<meep::realnum
             material_grids_addgradient_point(
                 v+ng*f_i, vec_to_vector3(p), scalegrad*cyl_scale, geps,
                 adjoint_c, forward_c, fwd, adj, frequencies[f_i], gv, du);
-          /* more complicated case requires interpolation/restriction */
-          } else if (md->do_averaging) {
+          /* anisotropic materials require interpolation/restriction */
+          } else if (md->do_averaging ||
+                     md->medium_1.epsilon_offdiag.x.re != 0 ||
+                     md->medium_1.epsilon_offdiag.y.re != 0 ||
+                     md->medium_1.epsilon_offdiag.z.re != 0 ||
+                     md->medium_2.epsilon_offdiag.x.re != 0 ||
+                     md->medium_2.epsilon_offdiag.y.re != 0 ||
+                     md->medium_2.epsilon_offdiag.z.re != 0) {
             /* we need to restrict the adjoint fields to
             the two nodes of interest (which requires a factor
             of 0.5 to scale), and interpolate the forward fields
--- a/src/meepgeom.cpp
+++ b/src/meepgeom.cpp
@@ -2959,7 +2959,7 @@ void material_grids_addgradient(double *v, size_t ng, std::complex<meep::realnum
           /********* compute -λᵀAᵤx *************/
 
           /* trivial case, no interpolation/restriction needed        */
-          if (forward_c == adjoint_c) {
+          if (forward_c == adjoint_c || !md->do_averaging) {
             std::complex<double> fwd = GET_FIELDS(fields_f, ci_forward, f_i, idx_fields);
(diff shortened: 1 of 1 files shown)


smartalecH:
This isn't quite right.

1. Even if subpixel smoothing is turned off (e.g. `do_averaging=false`), there are still occasions when we need to average the other components. Specifically, materials with off-diagonal materials need this (that was the whole point of #1801). In theory, there should be *no* overstepping when we just care about the diagonal components, as we are always using the same field data, regardless of the averaging condition.
2. Right now, if averaging is turned off (and there are no off-diagonal eps components) then dA/du should be 0, meaning that if there is a bug here, it will be on those routines (e.g.  `material_grids_addgradient_point()`). Perhaps a better check would be something like
```c++
else if (md->do_averaging && !(md-> medium_1->epsilon_offdiag) && !(md-> medium_2->epsilon_offdiag))
```
such that the routine is bypassed. (this needs a close look, however).

3. The *real* issue is the overstepping due to the [`GET_FIELDS`](https://github.com/NanoComp/meep/blob/efe3b4b1fff538cf1b1ad2e34bb47c7a3ddfb28a/src/meepgeom.cpp#L2925) macro (as discussed in the above-linked issue, #1861). The logic for this is horrendous. There are a lot of corner cases.
4. It is **more** work to get this patched than to get #1855 working (it's very close). The new method of accessing data directly from the chunks removes the need for `GET_FIELDS`'s crazy logic.

ahoenselaar:
> 4. It is **more** work to get this patched than to get [Non-materialized dft fields for adjoint calculations #1855](https://github.com/NanoComp/meep/pull/1855) working (it's very close). The new method of accessing data directly from the chunks removes the need for `GET_FIELDS`'s crazy logic.

Attempting to fix forward and pile more changes on top of an incorrect implementation seems very risky.

smartalecH:
> Attempting to fix forward and pile more changes on top of an incorrect implementation seems very risky.

But the implementation is changing... normally I would agree with you but in this case, it's apples and oranges.

oskooi:
It turns out this PR actually prevents the memory violations reported in #1861 even when subpixel smoothing is *enabled* (`do_averaging=True`). I tracked down the cause of the memory violations to be related to [`LOOP_OVER_IVECS(gv, is[ci_adjoint], ie[ci_adjoint], idx)`](https://github.com/NanoComp/meep/blob/efe3b4b1fff538cf1b1ad2e34bb47c7a3ddfb28a/src/meepgeom.cpp#L2932) of `material_grids_addgradient` which involves points *outside* the `grid_volume` of the `MaterialGrid`. It is only points outside of the `MaterialGrid` (for which `do_averaging=False`) which trigger the memory violation. With this PR, any grid point with `do_averaging=False` never executes the restriction/interpolate code which is why the memory violation is avoided.

I suspect there is a bug somewhere in this block which determines the starting and ending `ivec`s for the `LOOP_OVER_IVECS`:

https://github.com/NanoComp/meep/blob/efe3b4b1fff538cf1b1ad2e34bb47c7a3ddfb28a/src/meepgeom.cpp#L2903-L2910

Anyhow, at the very least, this PR fixes a bug in `material_grids_addgradient` related to disabling subpixel smoothing when the user specifies `do_averaging=False`.

PR #1901, opened 2022-01-12
resolve multiplication bug in Prism

Niccolo-Marcucci:
I wasn't being able to initiate a Prism. Turns out performing the operation Vector3*scalar returns Vector3, while scalar*Vector3 doesn't. 
I'm not sure the latter could work anyway, but in the meanwhile this change allows the Prism to work.

### DIFF
--- a/python/geom.py
+++ b/python/geom.py
@@ -1189,7 +1189,7 @@ class Prism(GeometricObject):
         """
 
         centroid = sum(vertices, Vector3(0)) * (1.0 / len(vertices)) # centroid of floor polygon
-        original_center = centroid + (0.5*height)*axis               # center as computed from vertices, height, axis
+        original_center = centroid + axis*(0.5*height)               # center as computed from vertices, height, axis
         if center is not None and len(vertices):
             center = Vector3(*center)
             # translate vertices to center prism at requested center


oskooi:
I'm not sure whether this change is necessary because the result of multiplying a `Vector3` by a scalar is always a `Vector3`, regardless of their order:
```py
>>> import meep as mp
>>> a = mp.Vector3(1.5,2.5,3.5)
>>> a * 4.3
Vector3<6.449999999999999, 10.75, 15.049999999999999>
>>> 4.3 * a
Vector3<6.449999999999999, 10.75, 15.049999999999999>
```

Niccolo-Marcucci:
Oh, this is interesting.  Turns out the problem appears when multiplying scalar of type `numpy.float64`.
```
>>> import meep as mp
>>> import numpy as np
>>> a = mp.Vector3(1.5,2.5,3.5)     
>>> b = np.float64(4.3)
>>> a * b
Vector3<6.449999999999999, 10.75, 15.049999999999999>
>>> b * a
array([ 6.45, 10.75, 15.05])
```
It just happens that i was setting the `Prism` height with an imported variable that was indeed `numpy.float64`. 
I thought of submitting a pull request because it gave me a headache solving the bug. Although I didn't notice back then that with standard `float` it was working.

I'm sorry, I think you can discard this pull request. Do you think i should submit a bug report or is this an unavoidable behavior?

ianwilliamson:
> I'm sorry, I think you can discard this pull request. Do you think i should submit a bug report or is this an unavoidable behavior?

AFAIK, this is unavoidable because Python will first call the left object's `__rmul__()` method in `a * b`. Unfortunately, Numpy's `__rmul__()` is somewhat aggressive in converting things to arrays.

oskooi:
Closing this as the reported behavior is a feature rather than a bug.

Issue #1501, opened 2021-02-15
GDS datatype import


flaport:
Hi,

I'm getting started with importing gds files into meep and found this tutorial really helpful: https://meep.readthedocs.io/en/latest/Python_Tutorials/GDSII_Import.

However, I was wondering if there is a way to specify the datatype as well during import. Right now it seems that specifying a layer number imports all polygons irrespective of their datatype. It would be useful if I could specify a (layer, datatype) tuple instead of just the layer index.

Or maybe there's a different way to distinguish the datatypes of the imported volumes after importing? If so, I can't seem to find it.

oskooi:
To import a layer as a tuple from a GDS file rather than as just a single number, you will need to use [`gdspy`](https://gdspy.readthedocs.io/en/stable/) because this feature is not supported by `libGDSII`.

Here is a demonstration of how to set this up:
```py
## load the GDS file                
gds = gdspy.GdsLibrary(infile=gds_file)

## define cell size and center                  
box = gds.top_level()[0].get_bounding_box()
cell_center = 0.5*mp.Vector3(box[1][0] + box[0][0],box[1][1] + box[0][1])

## define the geometry using all the polygons from layer (37,4)                    
polygons = gds.top_level()[0].get_polygons(True)[37,4]

design_geometry = []
for pg in polygons:
  vertices = []
  for vt in pg:
    # define vertices relative to center of cell                                           
    vertices.append(mp.Vector3(vt[0],vt[1])-cell_center)
    design_geometry.append(mp.Prism(vertices=vertices,
                                    height=0.5,
                                    axis=mp.Vector3(0,0,+1),
                                    material=mp.Medium(index=3.5)))
    design_geometry.append(mp.Prism(vertices=vertices,
                                    height=0.5,
                                    axis=mp.Vector3(0,0,-1),
                                    material=mp.Medium(index=3.5)))
```
Note that there are *two* `Prism` objects: one extending in the +z direction and the other in -z with a combined height of `1.0`.

It would probably be useful to add this example to https://meep.readthedocs.io/en/latest/Python_Tutorials/GDSII_Import.

flaport:
Thanks a lot @oskooi, this method works good enough for me. 

It might still be useful to add native support for this into libGDSII as well. [I have opened an issue there](https://github.com/HomerReid/libGDSII/issues/13).

PR #1905, opened 2022-01-14
Remove duplicate vertices in `Prism` object

### LINKED ISSUE #1520: `Prism` objects with duplicated vertices cause a segfault
Defining a Meep prism object with repeated coordinates seems to cause a segfault. It may be more user-friendly to perform some input validation here and, instead, raise an exception. Consider the example below, where the first and last vertex are the same:

```
import meep as mp

geometry = [
    mp.Prism(
        vertices=[
            mp.Vector3(51.07817888570662, -41.40026050039933, 0.0),
            mp.Vector3(26.07817888570662, -41.400436735950734, 0.0),
            mp.Vector3(16.07817888570662, -51.40027726778226, 0.0),
            mp.Vector3(11.875133679754981, -51.400343290466935, 0.0),
            mp.Vector3(-48.92182111429338, -51.400343290466935, 0.0),
            mp.Vector3(-48.92182111429338, 48.599656709533065, 0.0),
            mp.Vector3(51.07817888570662, 48.599656709533065, 0.0),
            mp.Vector3(51.07817888570662, -41.40026050039933, 0.0),
        ],
        height=100,
        sidewall_angle=0,
        material=mp.Medium(index=3.5),
    ),
]

simulation = mp.Simulation(
    cell_size=mp.Vector3(100.0, 100.0, 0.0),
    geometry=geometry,
    resolution=0.02,
)

simulation.init_sim()
```

oskooi:
Closes #1520.

### DIFF
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -5699,7 +5699,7 @@ Construct a `Prism`.
 + **`vertices` [list of `Vector3`]** — The vertices that make up the prism. They
   must lie in a plane that's perpendicular to the `axis`. Note that infinite
   lengths are not supported. To simulate infinite geometry, just extend the edge
-  of the prism beyond the cell.
+  of the prism beyond the cell. Duplicate entries will be removed.
 
 + **`height` [`number`]** — The prism thickness, extruded in the direction of
   `axis`. `mp.inf` can be used for infinite height. No default value.
--- a/python/geom.py
+++ b/python/geom.py
@@ -1168,7 +1168,7 @@ class Prism(GeometricObject):
         + **`vertices` [list of `Vector3`]** — The vertices that make up the prism. They
           must lie in a plane that's perpendicular to the `axis`. Note that infinite
           lengths are not supported. To simulate infinite geometry, just extend the edge
-          of the prism beyond the cell.
+          of the prism beyond the cell. Duplicate entries will be removed.
 
         + **`height` [`number`]** — The prism thickness, extruded in the direction of
           `axis`. `mp.inf` can be used for infinite height. No default value.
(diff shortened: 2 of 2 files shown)


ahoenselaar:
Wouldn't one want to only consider consecutive vertices?

stevengj:
Also, shouldn't this happen in libctl in the low-level constructor?

oskooi:
> Wouldn't one want to only consider consecutive vertices?

This is true in the case of a GDS file for which the shapes must be [weakly simple polygons](https://en.wikipedia.org/wiki/Simple_polygon#Weakly_simple_polygon). This means that the polygon has segments which can intersect but not cross. This also means that representing holes requires connecting their boundary to the boundary of the enclosing shape which can give rise to duplicate vertices that are not consecutive.

In the general definition of a prism which involves convexity, any duplicate vertex would need to be removed whether it is consecutive or not.

> Also, shouldn't this happen in libctl in the low-level constructor?

Sure, we can move this feature to libctl and then add a test in Meep.

ianwilliamson:
> This is true in the case of a GDS file for which the shapes must be [weakly simple polygons](https://en.wikipedia.org/wiki/Simple_polygon#Weakly_simple_polygon). This means that the polygon has segments which can intersect but not cross. This also means that representing holes requires connecting their boundary to the boundary of the enclosing shape which can give rise to duplicate vertices that are not consecutive.
> 
> In the general definition of a prism which involves convexity, any duplicate vertex would need to be removed whether it is consecutive or not.

Does the issue described in #1520 actually extend to the case of having _any_ duplicated vertices? I only showed the segfault with the duplicated first / last vertex while having subpixel smoothing enabled. Though, I could also see how there would be issues with having consecutive vertices.

Support for GDS-style polygons in the prism strikes me as being particularly useful given #1501, but if you do want to move forward with this approach, I would suggest raising an error instead of silently removing vertices. This would be quite confusing for users.

stevengj:
To me, a polygon with duplicate consecutive vertices seems like a well-defined limiting case, that seems like a polygon with one fewer vertex.  (e.g. the limit of a quadrilateral as a side length goes to zero should be a triangle, not an error)

stevengj:
e.g. matlab `polyshape` automatically deletes duplicate and even co-linear vertices. https://www.mathworks.com/matlabcentral/answers/606836-why-does-the-polyshape-function-deletes-the-vertices-in-the-input-when-there-are-no-duplicated-poi … and it seems pretty common in geometry packages to support deleting consecutive duplicate vertices, even if they are only approximately equal.

PR #1919, opened 2022-01-28
add type annotations to `Simulation` class

oskooi:
Even though Python is a dynamically typed language, type annotations (or hints) can be used for [static analysis](https://www.python.org/dev/peps/pep-0484/#rationale-and-goals) using tools such as [mypy](https://github.com/python/mypy) or [pytype](https://github.com/google/pytype). It would be useful to eventually add type annotations to *all* Python classes and functions in the API. This way, we could add a static analyzer (including a linter) as a GitHub Action as part of the CI.

One anomaly with the current setup is that using the automated script [`generate_api_doc.py`](https://github.com/NanoComp/meep/blob/master/doc/generate_py_api.py) to update the user manual markdown page `Python_User_Interface.md` which is generated from the docstrings shows the module filename as part of the class name. For example, the `Simulation` constructor includes:
```py
             boundary_layers: List[PML] = None,
             symmetries: List[Symmetry] = None,
```
which appears in the user manual as:
```py
             boundary_layers: List[meep.simulation.PML] = None,
             symmetries: List[meep.simulation.Symmetry] = None,
```
It would be cleaner if the module filename `meep.simulation` did not appear. **EDIT:** Actually, this can be fixed by adding:
```
from __future__ import annotations
```


doc/docs/Python_User_Interface.md
```diff
@@ -87,40 +87,40 @@ control various parameters of the Meep computation.
...
-             sources=None,
-             eps_averaging=True,
-             dimensions=3,
-             boundary_layers=None,
-             symmetries=None,
-             force_complex_fields=False,
-             default_material=Medium(),
-             m=0,
-             k_point=False,
+             cell_size: meep.geom.Vector3 = None,
```
stevengj, line 90: Remember that these arguments can be tuples instead of `Vector3`, and they get converted to `Vector3` internally

doc/docs/Python_User_Interface.md
```diff
@@ -87,40 +87,40 @@ control various parameters of the Meep computation.
...
+             resolution: float = None,
```
stevengj, line 91: What happens if the user then passes an integer?

stevengj:
As a general principle, this is fine, but one has to be very carefully about overly restricting the types, because then you use flexibility — Python potentially has a wide variety of types that can, for example, represent a real number.

stevengj:
Note also that this requires Python 3.5; what's the minimum version of Python that we currently support?

doc/docs/Python_User_Interface.md
stevengj: (Note that unions of types can be expressed as `Union[X, Y]` if you do `from typing import Union`).

stevengj: Note also that you presumably need to allow `NoneType` in order to accept a `None` argument.

ianwilliamson: You can wrap the type with `Optional` to allow for `None`, e.g. `Optional[Vector3] = None`

oskooi: I have wrapped the type of several of the constructor arguments which have `None` as the default with `Optional`. Also, I introduced a `Vector3Type` alias as `Union[Vector3, Tuple[float, ...], Tuple[int, ...]` which includes tuples.

doc/docs/Python_User_Interface.md
oskooi: I modified the type of `resolution` to `Union[float, int]` in order to support both number types.

oskooi:
> Note also that this requires Python 3.5; what's the minimum version of Python that we currently support?

The GitHub actions CI uses Python 3.6 and 3.9. The Conda packages for the latest version 1.22 use Python 3.7 - 3.10.

[Python 3.5](https://www.python.org/downloads/release/python-350/) was released in September 2015 which is fairly outdated.

doc/docs/Python_User_Interface.md
```diff
@@ -87,40 +87,40 @@ control various parameters of the Meep computation.
...
-             sources=None,
-             eps_averaging=True,
-             dimensions=3,
-             boundary_layers=None,
-             symmetries=None,
-             force_complex_fields=False,
-             default_material=Medium(),
-             m=0,
-             k_point=False,
+             cell_size:Union[meep.geom.Vector3, Tuple[float, ...], Tuple[int, ...]]=None,
```
stevengj, line 90: I think we still need `NoneType` here.

doc/docs/Python_User_Interface.md
```diff
@@ -87,40 +87,40 @@ control various parameters of the Meep computation.
...
+             resolution:Union[float, int]=None,
+             geometry:Union[List[meep.geom.GeometricObject], NoneType]=None,
+             sources:Union[List[meep.source.Source], NoneType]=None,
+             eps_averaging:bool=True,
+             dimensions:int=3,
+             boundary_layers:Union[List[meep.simulation.PML], NoneType]=None,
+             symmetries:Union[List[meep.simulation.Symmetry], NoneType]=None,
+             force_complex_fields:bool=False,
+             default_material:meep.geom.Medium=Medium(),
+             m:int=0,
```
stevengj, line 100: can also be `float`

doc/docs/Python_User_Interface.md
```diff
@@ -87,40 +87,40 @@ control various parameters of the Meep computation.
...
+             k_point:bool=False,
```
stevengj, line 101: `bool` or Vector3 or tuple

doc/docs/Python_User_Interface.md
```diff
@@ -87,40 +87,40 @@ control various parameters of the Meep computation.
...
              kz_2d='complex',
```
stevengj, line 102: string

doc/docs/Python_User_Interface.md
```diff
@@ -87,40 +87,40 @@ control various parameters of the Meep computation.
...
-             default_material=Medium(),
-             m=0,
-             k_point=False,
+             cell_size:Union[meep.geom.Vector3, Tuple[float, ...], Tuple[int, ...]]=None,
+             resolution:Union[float, int]=None,
+             geometry:Union[List[meep.geom.GeometricObject], NoneType]=None,
+             sources:Union[List[meep.source.Source], NoneType]=None,
+             eps_averaging:bool=True,
+             dimensions:int=3,
+             boundary_layers:Union[List[meep.simulation.PML], NoneType]=None,
```
stevengj, line 96: are absorbers subtypes of PML?

python/simulation.py
```diff
@@ -933,40 +936,40 @@ class Simulation(object):
...
+                 resolution: Union[float, int] = None,
+                 geometry: Optional[List[GeometricObject]] = None,
+                 sources: Optional[List[Source]] = None,
+                 eps_averaging: bool = True,
+                 dimensions: int = 3,
+                 boundary_layers: Optional[List[PML]] = None,
+                 symmetries: Optional[List[Symmetry]] = None,
+                 force_complex_fields: bool = False,
+                 default_material: Medium = mp.Medium(),
+                 m: int = 0,
```
stevengj, line 949: Can be a float — doesn't need to be an integer if the cell doesn't contain the origin.

doc/docs/Python_User_Interface.md

doc/docs/Python_User_Interface.md
oskooi: Fixed. Note that passing `int` when a `float` is expected is acceptable as discussed in https://stackoverflow.com/questions/50928592/mypy-type-hint-unionfloat-int-is-there-a-number-type.

doc/docs/Python_User_Interface.md

doc/docs/Python_User_Interface.md

python/simulation.py
oskooi: Fixed.

stevengj:
So int is a subtype of float in this type system?

oskooi:
> So int is a subtype of float in this type system?

Yes, that seems to be the case ever since [PEP 484](https://www.python.org/dev/peps/pep-0484/) which contains the [`typing`](https://docs.python.org/3/library/typing.html) module was introduced as part of the Python 3.5 release in September 2015.

stevengj:
Is there a way to work with earlier Python releases?   Requiring Python 3.7 for this is a bit annoying, since that is less than 4 years old.  I don't mind requiring Python 3.5.

PR #1930, opened 2022-02-07
pympb ports of of MPB's `compute_symmetries`

thchr:
This is essentially just a copy-port of https://github.com/NanoComp/mpb/issues/134, which enables MPB to compute overlap integrals of eigenfields transformed by a space group operation.

I also pulled in [**edit:** a small part of] https://github.com/NanoComp/mpb/issues/100 because https://github.com/NanoComp/mpb/issues/134 depended on parts of that. ~My understanding is that the Python interface to MPB does not currently run under MPI - so it's not clear that it's worth pulling in https://github.com/NanoComp/mpb/issues/100 which was mainly about making `get_val` run under MPI. On the other hand, I figured there wouldn't be any harm in it, in case it ever becomes a priority to make MPI work for MPB's Python interface.~[**edit:** I've left out the MPI parts]

I haven't tried connecting Python and C(++) before, so the PR here is mainly based on reading the current source. Please let me know if I missed anything.

If this gets merged, I'll make a Python-docs PR to the MPB repo as well.

### DIFF
--- a/libpympb/pympb.cpp
+++ b/libpympb/pympb.cpp
@@ -117,6 +117,15 @@ void matrix3x3_to_arr(mpb_real arr[3][3], matrix3x3 m) {
 
 cnumber cscalar2cnumber(scalar_complex cs) { return make_cnumber(CSCALAR_RE(cs), CSCALAR_IM(cs)); }
 
+cvector3 cscalar32cvector3(const scalar_complex *cs)
+{
+     cvector3 v;
+     v.x = cscalar2cnumber(cs[0]);
+     v.y = cscalar2cnumber(cs[1]);
+     v.z = cscalar2cnumber(cs[2]);
+     return v;
+}
+
 // Return a string describing the current parity, used for frequency and filename
 // prefixes
 const char *parity_string(maxwell_data *d) {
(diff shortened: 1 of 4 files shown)


libpympb/pympb.cpp
```diff
@@ -1699,22 +1708,41 @@ std::vector<mpb_real> mode_solver::get_output_k() {
...
+
+  #else /* no MPI */
+    mpb_real val = data[(((ix * ny) + iy) * nz + iz) * stride];
+  #endif
+
+  // TODO: this is in MPB's implementation, but seems to cause issues here - why?
+  // if (conjugate)
+  //   return -val;
+  // else
+  //   return val;
```
thchr, line 1738: In MPB's implementation of `get_val`, the `conjugate` keyword works like this. In the previous implementation here though, the `conjugate` keyword is ignored.
When I was testing this, I found that I needed to ignore it here as well for things to turn out right - but I don't understand why there's this difference. Any explanation would be helpful!

stevengj: In MPB's implementation, the `conjugate` keyword is modified here: https://github.com/NanoComp/mpb/blob/74fb432d31905a43b2d2c1d9b88a9f4bbfd0a8b6/mpb/fields.c#L567

That is, a `conjugate == 1` argument is ignored unless (1) MPB is compiled `--with-inv-symmetry` (`! SCALAR_COMPLEX`) and (2) the requested point is in the non-stored "other half" of the data (which is conjugated).     Basically, it corresponds to real-antisymmetric data.    

This isn't relevant in the Python interface because it doesn't support `--with-inv-symmetry` IIRC.

thchr: It seemed to have an impact here though: if I have the above lines uncommented, `get_bloch_field_point` seemed to return the conjugate of the right answer.

stevengj: That's because you are missing the logic from [this line](https://github.com/NanoComp/mpb/blob/74fb432d31905a43b2d2c1d9b88a9f4bbfd0a8b6/mpb/fields.c#L567).

thchr: Ahh, that's some bad copy-pasting on my part - thanks for figuring that one out..!
Should I port that bit over, even though the Python interface currently doesn't support `--with-inv-symmetry` (e.g., in case that becomes a desire)?

stevengj: No, I would just omit it, but add a comment to that effect.

stevengj: Similarly for the MPI stuff — since right now, it can't be tested, I'm reluctant to include it.

thchr: Okay, I've done that in [c743e25](https://github.com/NanoComp/meep/pull/1930/commits/c743e2521581a773c52d76e7f9bb6822b3fa0aef).

stevengj (approved):

thchr:
Hi @stevengj,

I have a follow-up question to this PR, if you the time:

I realized that this feature doesn't work unless I compile meep serially (i.e., without `--with-mpi`). I had naively thought that meep might take care of calling mpb serially, even if meep was compiled with MPI.
Is it correctly understood that the Python interface to MPB will only work if I compile meep serially? I.e., I cannot have co-existing parallel meep and MPB programs? Most (all? except this feature) MPB-related things seemed to work fine if I compiled meep with `--with-mpi`, but the symmetry eigenvalue computation does not.

stevengj:
If it doesn't work (albeit serially) when you compile meep with MPI, I would consider that as a bug that we should fix.

thchr:
Okay, I will try to find a fix - if you have any thoughts about what could be likely culprits, please let me know.

Just to verify: everything in pympb.cpp is run on one core, regardless of how meep is built?

thchr:
OK, I think I have an idea of the problem: in `pympp.cpp`, `HAVE_MPI` is defined if compiled with `--with-mpi`. Nothing depends on this, *except* for `XYZ_LOOP` which picks up an MPI-specific implementation here: https://github.com/NanoComp/meep/blob/31d2794979ec3ca13d2cfb618401f09283325f3a/libpympb/pympb.cpp#L28-L39

So the way we loop over `mdata` depends on `HAVE_MPI` - but, if I understand correctly, `mdata` is always created from a serial MPB, so its structure doesn't actually match that of the MPI-specific `XYZ_LOOP`.
(This would mean that e.g., `compute_field_integral` probably also does something wrong when compiled with `--with-mpi` currently)

Does this sound right? If so, I can submit a PR that removes the MPI-variant of `XYZ_LOOP`.

mawc2019:
> I realized that this feature doesn't work unless I compile meep serially (i.e., without --with-mpi).

@mochen4 and I have tried to compile the current master branch without `--with-mpi`, but it does not work, either. The error `'CASSIGN_CONJ_MULT' was not declared in this scope` is reported during `make`.

thchr:
> > I realized that this feature doesn't work unless I compile meep serially (i.e., without --with-mpi).
> 
> @mochen4 and I have tried to compile the current master branch without `--with-mpi`, but it does not work, either. The error `'CASSIGN_CONJ_MULT' was not declared in this scope` is reported during `make`.

This macro should be pulled in from the mpb header files that are included at the top of `pympp.cpp` (e.g., `mpb.h` includes `maxwell.h` which includes `scalar.h` which has the definition for [`CASSIGN_CONJ_MULT`](https://github.com/NanoComp/mpb/blob/74fb432d31905a43b2d2c1d9b88a9f4bbfd0a8b6/src/matrices/scalar.h#L67))

mawc2019:
Thanks! Steven just uploaded the fix. The updated master branch compiles now.

PR #1938, opened 2022-02-10
update near2far tutorial and docs

### LINKED ISSUE #1909: fix incorrect claim that near-to-far transform only works in the far field
This line seems wrong: https://github.com/NanoComp/meep/blob/1f5833454b2e0954f91658a5e86a60ef2ae4ac85/doc/docs/Python_User_Interface.md.in#L305

The name "near-to-far field transform" is something of a misnomer — since we use the full Green's function, this is just an application of the Equivalence Principle, and we should be able to compute fields *anywhere* outside the near2far surface.   The only limiting factor should be discretization errors, but for any given distance it should converge with resolution (once the distance >> resolution).

oskooi:
Fixes #1909.


stevengj:
Some notes:

* If you have a *closed* near-field surface, then the only error is discretization (which will cause a bigger and bigger mismatch as you get further away, due to numerical dispersion).
* If you have a *non-closed* near-field surface, you get an error from the truncation
  - a guided mode does not radiate if you have an *infinite* surface next to the guided mode, since it is under the light cone, but if you truncate the equivalent currents to a finite surface then you get radiation from the edges (you "smeared out" the Fourier transform with a sinc function so that it leaks under the light cone)
  - If you move the finite/truncated near-field surface farther away, then the spurious radiation from the guided mode decreases exponentially fast.  But if you move it too far away, then you will miss more and more of the radiation from the cavity mode and this will cause errors.

Ideally, we could let the near-field surface extend all the way into the PML, so that the fields decayed away and we could truncate it.   But then you have to use the "stretched-coordinate" PML Green's function for the near-to-far transformation, and these will be tricky to get right.   See e.g. https://ieeexplore.ieee.org/document/1265144

oskooi:
I have expanded the discussion in the tutorial to describe the errors introduced by the finite truncation of the near-field surface as described above.

PR #1942, opened 2022-02-10
allow construction of `Matrix` from `numpy.array`s

thchr:
I was trying to use the Python interface via PyCall in Julia and wanted to construct a `meep.Matrix` like so
```jl
using PyCall
mp = pyimport("meep")
mp.Matrix([1,2,3], [4,5,6], [7,8,9])
```
and noticed that it errored because the Julia vectors are converted to numpy arrays, which didn't like the chained comparison (throwing `ValueError('The truth value of an array with more than one element is ambiguous. [...]`).
This fixes that.

PS. I know I could just do `mp.Matrix(mp.Vector3(1,2,3), mp.Vector3(4,5,6), mp.Vector3(7,8,9))` - so this is just intended to make it a little bit more ergonomic.

### DIFF
--- a/python/geom.py
+++ b/python/geom.py
@@ -1247,10 +1247,10 @@ class Matrix(object):
         self.c1 = Vector3(*c1)
         self.c2 = Vector3(*c2)
         self.c3 = Vector3(*c3)
-        if c1 == c2 == c3 == Vector3():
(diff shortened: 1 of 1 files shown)


stevengj:
Thanks!

PR #1943, opened 2022-02-10
mpb: make `solve_kpoint` also set `freqs`

thchr:
I noticed that running `ms.solve_kpoint(k)` would not set the `ms.freqs` variable, unlike doing `ms.run()`.
I figured that might be nice to set in `solve_kpoint` directly.


stevengj:
LGTM.

PR #1949, opened 2022-02-13
pympp: remove `HAVE_MPI`-variation of `XYZ_LOOP`

thchr:
This is split into three commits, the first of which is the main one: 

1. Implements what I suggested in https://github.com/NanoComp/meep/pull/1930#issuecomment-1036999119: i.e., it comments out the `HAVE_MPI`-specific code for `XYZ_LOOP` in pympb.cpp (since the code doesn't match the layout of `mdata`). Fixes https://github.com/NanoComp/meep/pull/1930#issuecomment-1036360784.

2. Adds a test of the symmetry-functionality from #1930 which previously gave the wrong answer when compiled with `--with-mpi`.

3. Tries to reenable Python-MPB tests (originally removed in https://github.com/NanoComp/meep/pull/262). Locally, I saw that there was a single failure (in `test_output_efield_z`) but the margin of error was very small, so figured I'd check if it did the same here (and reenable if it did not).

### DIFF
--- a/libpympb/pympb.cpp
+++ b/libpympb/pympb.cpp
@@ -17,7 +17,7 @@ extern "C" int mpb_verbosity = 2;
 #endif
 
 // xyz_loop.h
-#ifndef HAVE_MPI
+// #ifndef HAVE_MPI
 #define LOOP_XYZ(md)                                                                               \
   {                                                                                                \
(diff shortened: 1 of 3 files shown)


thchr:
Gentle bump; any concerns with these changes?

stevengj:
Thanks, looks good!

Issue #1916, opened 2022-01-26
Analytic gradient computation Aᵤ


smartalecH:
Currently, we use a finite difference approximation to compute the gradient of the operator (A) w.r.t the design variables (u). This approach is convenient and flexible but has its limitations. Namely, it's rather hard to choose a sufficient finite-difference step size when β approaches ∞. (see #1854)

The proper implementation should involve the analytic "backpropagation" through the averaging routine, which involves another integral.

Not only would this approach improve the accuracy of the gradients, but it would also speed the gradient computation up by a factor of 2.

smartalecH:
(cc @oskooi )

We *also* may be able to circumvent the brute force quadrature altogether (at least when β=∞). The averaging integral has a nice form that can be simplified by parts.

stevengj:
One thing we might consider is to *always* do the subpixel averaging as if β = infinity, regardless of the actual value of β.

Rationale:

1. The β = infinity averaging is easy.  You just use the [analytical formula for the volume of a spherical cap](https://en.wikipedia.org/wiki/Spherical_cap) to compute the fraction of the sphere occupied by each material, and then the effective ε formula is simple (even if we do full "Kottke" averaging for the anisotropic case, which we aren't doing yet here).
2. Subpixel averaging only really matters for accuracy when the interface is discontinuous (or approximately so on the scale of a pixel), in which case the β=infinity approximation should be good.  When β is small and the interface is smooth, you don't need to do any averaging at all — you get second-order accuracy even just by sampling the center of the pixel.  So I'm hopeful that you would still get second-order accuracy from the fictitious "β=infinity" average in this case, though that would have to be checked more carefully.

PR #1951, opened 2022-02-16
Approximate material grid smoothing (β=∞)

smartalecH:
As described in [this comment](https://github.com/NanoComp/meep/issues/1916#issuecomment-1034426933), this PR attempts to perform subpixel smoothing assuming β=∞ for all values actual of β.

From a high-level (abstract) perspective, the technique seems to work (with some important exceptions). Here is the resulting geometry (plotted by pulling the trace of the ε tensor so we can explicitly see the smoothing) for various values of β:

β=2
image

β=64
image

β=512
image

β=∞
image

There are still some artifacts that need to be resolved for lower values of β, but it's clear that at least the routine is doing averaging across the boundries.

There is also a bug with the recombination step (while calculating the gradient). There are occasions when `get_front_object` does not set `mat_behind`. This never happens during the forward/adjoint runs -- which is unexpected as the same "routines" are used between these two cases, so we shouldn't see different behavior...

Of course, we still need to perform convergence tests, gradient tests, etc.

### DIFF
--- a/src/meepgeom.cpp
+++ b/src/meepgeom.cpp
@@ -1021,6 +1021,43 @@ static bool get_front_object(const meep::volume &v, geom_box_tree geometry_tree,
   return true;
 }
 
+double get_material_grid_fill(meep::ndim dim, double d, double r, double u, double eta,
+  material_type &mat, material_type &mat_behind){
+  /* Lets assume that the "background material" is void (u=0) and that the
+  "foreground material is solid (u=1)". The fill fraction then describes
+  the amount of foreground material inhabits the cell.
+
+  The analytic volume expressions for the end cap only specify the fill fraction
+  of the cap relative to the whole sphere... but we don't know what the cap 
+  actually is (solid or void). So we need a little extra logic to sort that out.
+  
+  Occasionally, our distance to the nearest interface is outside the current
+  sphere, which means we don't need to do any averaging (the fill is 1). Again,
+  we don't know if that we means we are in void or solid, however, until we look
+  at u. Then we can determine whether the fill is 0 or 1. To make things easy, we'll
+  always assume the cap is solid.
+  */
+  double rel_fill;
+  if (abs(d) > abs(r)){
+    mat_behind = mat;
+    return -1.0; // garbage fill
+  } else {
+    if (dim == meep::D1)
+      rel_fill = (r-d)/(2*r);
+    else if (dim == meep::D2 || dim == meep::Dcyl){
+      rel_fill = (1/(r*r*meep::pi)) * (r*r*std::acos(d/r)-d*std::sqrt(r*r-d*d));
+    }
+    else if (dim == meep::D3)
+      rel_fill = (((r-d)*(r-d))/(4*meep::pi*r*r*r))*(2*r+d);
+  }
+
+  if (u <= eta){
+    return rel_fill;
+  }else{
+    return 1-rel_fill;
+  }
+}
+
 void geom_epsilon::eff_chi1inv_row(meep::component c, double chi1inv_row[3], const meep::volume &v,
                                    double tol, int maxeval) {
   symm_matrix meps_inv;
@@ -1059,6 +1096,7 @@ void geom_epsilon::eff_chi1inv_matrix(meep::component c, symm_matrix *chi1inv_ma
   material_type mat, mat_behind;
   symm_matrix meps;
   vector3 p, shiftby, normal;
+  double fill;
   fallback = false;
 
   if (maxeval == 0) {
@@ -1072,15 +1110,35 @@ void geom_epsilon::eff_chi1inv_matrix(meep::component c, symm_matrix *chi1inv_ma
 
   if (!get_front_object(v, geometry_tree, p, &o, shiftby, mat, mat_behind)) {
     get_material_pt(mat, v.center());
-    if (mat && (mat->which_subclass == material_data::MATERIAL_USER ||
-                mat->which_subclass == material_data::MATERIAL_GRID)
-        && mat->do_averaging) {
+    if (mat && (mat->which_subclass == material_data::MATERIAL_USER)) {
       fallback = true;
       return;
+    }
+    else if(mat && (mat->which_subclass == material_data::MATERIAL_GRID)){
+      // debug
+      //if (mat_behind)
+      //  printf("c %d | mat %d | behind %d | (x,y) (%f,%f) \n",c,mat->which_subclass,mat_behind->which_subclass,v.center().x(),v.center().y());
+      // proceed
     }
     else {
       goto trivial;
     }
+  }
+
+  if (mat->which_subclass == material_data::MATERIAL_GRID){
+    int oi;
+    geom_box_tree tp = geom_tree_search(p, restricted_tree, &oi);
+    
+    meep::vec normal_vec(matgrid_grad(p, tp, oi, mat));
+    double nabsinv = 1.0 / meep::abs(normal_vec);
+    LOOP_OVER_DIRECTIONS(normal_vec.dim, k) { normal_vec.set_direction(k,normal_vec.in_direction(k)*nabsinv); }
+    
+    double uval = matgrid_val(p, tp, oi, mat)+this->u_p;
+    double d = (mat->eta-uval) * nabsinv;
+    double r = v.diameter()/2;
+    
+    fill = get_material_grid_fill(normal_vec.dim,d,r,uval,mat->eta,mat,mat_behind);
+    normal = vec_to_vector3(normal_vec);
   }
 
   /* check for trivial case of only one object/material */
@@ -1089,14 +1147,17 @@ void geom_epsilon::eff_chi1inv_matrix(meep::component c, symm_matrix *chi1inv_ma
   // it doesn't make sense to average metals (electric or magnetic)
   if (is_metal(meep::type(c), &mat) || is_metal(meep::type(c), &mat_behind)) goto noavg;
 
-  normal = unit_vector3(normal_to_fixed_object(vector3_minus(p, shiftby), *o));
-  if (normal.x == 0 && normal.y == 0 && normal.z == 0)
-    goto noavg; // couldn't get normal vector for this point, punt
-  geom_box pixel = gv2box(v);
-  pixel.low = vector3_minus(pixel.low, shiftby);
-  pixel.high = vector3_minus(pixel.high, shiftby);
+  if (mat->which_subclass != material_data::MATERIAL_GRID) {
+    normal = unit_vector3(normal_to_fixed_object(vector3_minus(p, shiftby), *o));
+    if (normal.x == 0 && normal.y == 0 && normal.z == 0)
+      goto noavg; // couldn't get normal vector for this point, punt
+    
+    geom_box pixel = gv2box(v);
+    pixel.low = vector3_minus(pixel.low, shiftby);
+    pixel.high = vector3_minus(pixel.high, shiftby);
 
-  double fill = box_overlap_with_object(pixel, *o, tol, maxeval);
+    fill = box_overlap_with_object(pixel, *o, tol, maxeval);
+  }
 
   material_epsmu(meep::type(c), mat, &meps, chi1inv_matrix);
   symm_matrix eps2, epsinv2;


stevengj:
See also #1399 for what to do when `mat != mat_behind`, even if one (or both) of them are material grids or user functions.   Basically, we want to treat the material as piecewise constant and do the subpixel averaging using the geometric fill factor.

src/meepgeom.cpp
```diff
@@ -977,50 +987,96 @@ static bool get_front_object(const meep::volume &v, geom_box_tree geometry_tree,
...
     shiftby_front = shiftby2;
     mat_front = mat2;
+    p_front = p2;
     mat_behind = mat1;
+    p_behind = p1;
   }
   return true;
 }
 
+double get_material_grid_fill(meep::ndim dim, double d, double r, double u, double eta,
```
smartalecH, line 1040: Analytically computes fill fraction.

src/meepgeom.cpp
```diff
@@ -1070,33 +1133,107 @@ void geom_epsilon::eff_chi1inv_matrix(meep::component c, symm_matrix *chi1inv_ma
...
+  if (is_material_grid(mat_behind)){
+    int oi;
+    geom_box_tree tp = geom_tree_search(p_mat_behind, restricted_tree, &oi);
+    meep::vec normal_vec_behind(matgrid_grad(p_mat_behind, tp, oi, mat_behind));
+    
+    double nabsinv = 1.0 / meep::abs(normal_vec_behind);
+    LOOP_OVER_DIRECTIONS(normal_vec_behind.dim, k) { normal_vec_behind.set_direction(k,normal_vec_behind.in_direction(k)*nabsinv); }
+    
+    double uval = matgrid_val(p_mat_behind, tp, oi, mat_behind)+this->u_p;
+    double d = (mat_behind->eta-uval) * nabsinv;
```
smartalecH, line 1181: distance to an interface

src/meepgeom.cpp
```diff
@@ -1070,33 +1133,107 @@ void geom_epsilon::eff_chi1inv_matrix(meep::component c, symm_matrix *chi1inv_ma
...
+  to do any averaging, or if the current pixel is smooth enough
+  to just evaluate.
+  
+  Keep in mind that if both the front and back materials are material grids,
+  we assume they are the same material grid. (all of the complicated overlapping
+  symmetry conditions are taken care of elsewhere).
+   */
+  double fill_front, fill_back;
+  vector3 normal_front, normal_back;
+  if (is_material_grid(mat)){
```
stevengj, line 1156: This should only occur if the *entire pixel* is filled by the material grid.

src/meepgeom.cpp
```diff
@@ -1070,33 +1133,107 @@ void geom_epsilon::eff_chi1inv_matrix(meep::component c, symm_matrix *chi1inv_ma
...
+    LOOP_OVER_DIRECTIONS(normal_vec_front.dim, k) { normal_vec_front.set_direction(k,normal_vec_front.in_direction(k)*nabsinv); }
+    
+    double uval = matgrid_val(p_mat, tp, oi, mat)+this->u_p;
+    double d = (mat->eta-uval) * nabsinv;
+    double r = v.diameter()/2;
+    
+    fill = get_material_grid_fill(normal_vec_front.dim,d,r,uval,mat->eta,&mg_averaging,&center_is_void);
+    normal = vec_to_vector3(normal_vec_front);
+  }
+  if (is_material_grid(mat_behind)){
```
stevengj, line 1172: Should be no need for this.  If you have two overlapping objects, both with a material grid, then you should treat the material grids as if they were each *uniform* materials and fall back on our geometric-object fill fraction, similar to variable materials.

src/meepgeom.cpp
```diff
@@ -1057,11 +1113,18 @@ void geom_epsilon::eff_chi1inv_matrix(meep::component c, symm_matrix *chi1inv_ma
...
   symm_matrix meps;
   vector3 p, shiftby, normal;
+  double fill;
   fallback = false;
+  bool mg_averaging = false;
+  bool center_is_void = true;
 
-  if (maxeval == 0) {
+  if (maxeval == 0 ||
+      (!get_front_object(v, geometry_tree, p, &o, shiftby, mat, mat_behind, p_mat, p_mat_behind)
```
stevengj, line 1125: I would change `get_front_object` to return the number of objects found (treating the default material as an "object"). 

* If it's `> 2`, then we go to the `noavg:` branch. 
*  If it `== 2`, then even if one of the objects is a material grid or a user material then we treat it as a uniform material (by evaluating at `p_mat` or `p_mat_behind`, respectively) and do the geometric-object averaging (if the resulting `mat` and `mat_behind` are still unequal).   
* If it `== 1`, then if it's a material grid we do the first-order expansion to find the fill fraction etcetera.

src/meepgeom.cpp
```diff
@@ -977,50 +987,96 @@ static bool get_front_object(const meep::volume &v, geom_box_tree geometry_tree,
...
+  }
+
+  *mg_averaging = true;
+  if (u <= eta){
+    *center_is_void = true;
+    return rel_fill;   // center is void, so cap must be sold
+  }else{
+    *center_is_void = false;
+    return 1-rel_fill; // center is solid, so cap must be void
+  }
```
stevengj, line 1077: May not be needed any more — just return the fill fraction of the u=1 material always, for example.

smartalecH:
To debug the gradient step, make sure the forward, adjoint, and recombination steps all have the same geometry tree. See libctl `display_geom_box_tree`:

https://github.com/NanoComp/libctl/blob/069e736614d86d070ce9bfc222c995c818f24191/utils/geom.c#L1749

smartalecH:
Quick update:
- The new routine seems to work well, except that the relatively high gradient error we used to just see with β=∞ now occurs at *all* values of β (about 1%-10% relative error on average). This is somewhat expected, as there are "tricky" interfaces for all β now. These interfaces don't play well with the finite difference approximation of the A operator (see #1916). To get better gradients, we need to backprop through the Kotkke algorithm too.
- The current tests that are failing are mostly due to either API fixes I still need to make (i.e. remove certain `do_smoothing` flags) or because the gradient is no longer accurate enough to pass. There is, however, an issue with the artificial conductivities. Those gradients seem very off. I'm assuming I'm missing something there...

smartalecH:
I've added support for analytic gradients using duals (as discussed), but there are a couple of issues.

First, our approximation is having trouble accurately representing the material grid at lower values of β. You can see the artifacts here:
image

This is somewhat expected (and not related at all to the duals recently implemented) as we *always* assume β=∞ near interfaces.

In addition, there are still some inaccuracies with the gradient when β≠0 (for both cases when smoothing is off or on). It's a somewhat complicated routine with lots of logic. A second (or third) pair of eyes would help a lot.

stevengj:
Possible solution: do the β=infinity averaging for pixels that cross the u=η level set, but don't use ε1 and ε2.   Instead, use an "effective" ε1 and ε2 given by ε(project(u, β)) for u = η ± u₀′Δx/2

stevengj:
Regarding the merge with #1399 — in order to retain piecewise differentiability, when you have an "edge" pixel that *also* crosses the u=η level set, instead of sampling the MG at an arbitrary point you should probably compute the fill fraction and then do a simple weighted average of the "effective" ε1 and ε2 as defined above.

stevengj:
More detail on the "effective" epsilons.   Let ε₀ denote the projected value (with the actual β) at u₀ (the center of the sphere) and let ε₊ and ε₋ denote the projected values at u₀±u₀′Δx/2 (the edges of the sphere).   d=(η-u₀)/u₀′ be the distance of the u=η level set from the center.

Now we need an effective ε̃₊ and ε̃₋ on the two sides as inputs for the Kottke averaging.   We *cannot* simply use ε₊ and ε₋ because then the material will change discontinuously to ε₀ when the interface leaves the cell.  We need to make sure things are continuous both as the interface leaves the cell (|d| > Δx/2) and when the interface crosses the center (d=0).

Case 1: d > 0.   Then let ε̃₊ = ε₊ and let ε̃₋ = [ε₀d + ε₋(Δx/2 – d)] / (Δx/2).

Case 2: d < 0.   Then let ε̃₋ = ε₋ and let ε̃₊ = [-ε₀d + ε₊(Δx/2 + d)] / (Δx/2).

smartalecH:
@stevengj your trick seems to work well!

image

Now onto fixing the gradients....

smartalecH:
cc @oskoi you can rerun the test with this PR

python/adjoint/filters.py
```diff
@@ -7,6 +7,8 @@
 import meep as mp
 from scipy import special
 from scipy import signal
+import skfmm
```
oskooi, line 10: Is it necessary to include this additional module just for its signed distance function (rather than write our own or find a similar function in e.g. SciPy)? skfmm does not seem to be widely used which could make installing it a challenge for some users building from source.

smartalecH: > could make installing it a challenge for some users building from source.

From their [docs](https://github.com/scikit-fmm/scikit-fmm):

```
pip install scikit-fmm
```

What's challenging about this particular package vs `numpy`, `scipy`, etc.?

It's convenient for testing, but we can remove it if it really is a challenge to install.

oskooi: I ran into some package conflicts during the `pip install`. In general, it is useful to keep the number of package dependencies, particularly smaller ones, to a minimum.

python/tests/test_material_grid.py
```diff
@@ -137,109 +158,70 @@ def compute_resonant_mode_2d(res,default_mat=False):
...
-                              material=Si)]
-
-    sources = [mp.Source(src=mp.GaussianSource(fcen,fwidth=df),
-                         size=mp.Vector3(),
-                         center=mp.Vector3(0.13,0.25,0.06),
-                         component=mp.Ez)]
+    ms = mpb.ModeSolver(num_bands=1,
+                        k_points=[mp.cartesian_to_reciprocal(k_point, geometry_lattice)],
+                        geometry=geometry,
+                        geometry_lattice=geometry_lattice,
```
stevengj, line 168: This is wrong Alex!!!!!

smartalecH: Who's Alex?

stevengj:
@smartalecH, I pushed a rebased version following the instructions from https://github.com/NanoComp/meep/pull/2164#issuecomment-1199944618 to fix conflicts from the reformatting.

src/meepgeom.cpp
```diff
@@ -1191,6 +1255,251 @@ void geom_epsilon::eff_chi1inv_matrix(meep::component c, symm_matrix *chi1inv_ma
...
+      uval = matgrid_val(p_mat, tp, oi, mat);
+      normal = matgrid_grad(p_mat, tp, oi, mat);
+      if (cvector3_re(normal).x == 0 && cvector3_re(normal).y == 0 && cvector3_re(normal).z == 0)
+        /* couldn't get normal vector for this point; no averaging is needed,
+        but we still need to interpolate onto our grid using duals */
+        goto mgavg;
+      duals::duald u_prime = normalize_dual(normal);
+      duals::duald d = (mat->eta - uval) / u_prime;
+      double r = v.diameter() / 2;
+      fill = get_material_grid_fill(v.dim, d, r, uval, mat->eta);
```
smartalecH, line 1415: Ok I think I have finally narrowed down the reason why some tests are failing.

Note that the `matgrid_grad` is especially sensitive to roundoff error (it calls several routines from both `meep` and `libctl`). There are many cases where the normal vector should be `(0,0,0)` (e.g. a uniform design region) but is typically off by a few $\epsilon$ in a random direction. **Especially when the design variables are uniform, which is the case for the failing tests**.

This is especially problematic when `uval`≈`mat->eta`, because then we get some weird $\frac{0}{0}$ conditions (this produces some dramatic swings in both the value for `d` and its dual derivative).

We need to think about some ways to make this part of the code more robust to roundoff error. This is a bit outside of my wheelhouse...

@stevengj, @oskooi, do you have any suggestions?

smartalecH: As an example, suppose I'm optimizing a uniform design field (such that `rho=0.5` everywhere). Of course, I'm doing TO, so I run this through a conic filter first. Ideally, my output is also `rho_tilde=0.5` everywhere, but in reality there are some perturbations of the order of $\epsilon$ scattered throughout.

Now, because of this slight variation, when I try to compute the normal vector at a particular point `r` (i.e. the *spatial* gradient of the design field) using `matgrid_grad`, I *hope* to get a vector that looks like `(0,0,0)` but most of the time in these cases it looks something like `(3.330669e-15, 6.661338e-15, 0.000000e+00)`, where again, I'm off by a few $\epsilon$ in the `x` and `y` directions.

Furthermore, when I try to compute $d=\frac{\eta-\rho}{|\nabla\rho|}$, the quantity $\eta-\rho$ is *also* sensitive to floating point roundoff, and I occasionally get values like `-2.775558e-16`. Which means I have a really small number divided by an even smaller number ($|\nabla\rho|$ is pretty small), and we get some interesting behavior.

I've tried some thresholded compare statements (i.e. create a function `equals(double var1, double var2)` that checks if `var1` and `var2` are within a few $\epsilon$ of each other). Of course, these methods are a bit ad hoc and not ideal.

I'm thinking there's another way to tackle this problem that lets us avoid the issue of checking for roundoff altogether... just not sure what that is yet...

oskooi: Why don't we try adding a check to the end of [`meep::vec material_grid_grad`](https://github.com/NanoComp/meep/blob/ea43f83c2f8c3e74aab2f8dae3e2ed96b7347777/src/meepgeom.cpp#L366) to set the gradient vector to zero if all the components happen to be smaller than some threshold value? That way, we just need to modify only one place in the code. (For consistency, we also should do the same thing for other functions which compute the normal vector i.e. [`material_function::normal_vector`](https://github.com/NanoComp/meep/blob/ea43f83c2f8c3e74aab2f8dae3e2ed96b7347777/src/anisotropic_averaging.cpp#L44-L68)).

smartalecH: I added a function, `AreEqual()`, which checks two numbers (I templated it so that it works with doubles, floats, etc.) within $2\epsilon$. This threshold (like all thresholds) is a bit arbitrary though...

This mitigates many of the errors I've seen so far. But there are still some issues slipping through the cracks, again due to roundoff error.

stevengj: In general, I would expect our final effective ε to be insensitive to errors like this, because in a region where ε is nearly constant its arithmetic and harmonic means are nearly equal so the choice of the normal vector becomes irrelevant.   So, we should have a well-conditioned function to differentiate.

This seems like a classic type of problem in numerical analysis where you have some nice well-conditioned function that you want to compute, but your computational algorithm involves some intermediate step (like the computation of `d`) that is badly conditioned, and screws up your calculation.  The solution is usually to reformulate the calculation algebraically to avoid the ill-conditioned step, e.g. to replace `d` with `d√|∇ρ|`.  Sometimes it requires some clever algebra, e.g. see the `exp1` and quadratic-formula examples in [these notes](https://nbviewer.org/github/mitmath/18335/blob/spring21/notes/Floating-Point-Intro.ipynb#Cancellation-error).

stevengj:
Another thing to do is to set `opt.set_param("verbosity", 1)` to see the internal progress of the MMA or CCSA algorithm, to see if it is getting stuck on inner iterations (e.g. if the "rho" penalty parameter is increasing without bound).   See the [internal parameters for MMA](https://nlopt.readthedocs.io/en/latest/NLopt_Algorithms/#mma-method-of-moving-asymptotes-and-ccsa) in the nlopt docs.

PR #1958, opened 2022-02-22
minor tweaks and fixes to docs

oskooi:


stevengj:
Would be good to put in the figure title that the near2far fields were "erroneously truncated" or something like that to emphasize that they aren't supposed to agree.

PR #1959, opened 2022-02-22
WIP: allow add_srcdata to place sources on non-owned points

mawc2019:
This PR continues @stevengj 's work in #1547. At present, it still fails when too many processes are involved, which can be illustrated by the [example](https://github.com/NanoComp/meep/pull/1547#issuecomment-963528760) with 7 processes. The error is `unknown src_time_id (missing registration?)`.

### DIFF
--- a/python/adjoint/objective.py
+++ b/python/adjoint/objective.py
@@ -109,10 +109,12 @@ class ObjectiveQuantity(abc.ABC):
         The user may specify a scalar valued objective function across multiple frequencies (e.g. MSE) in
         which case we should check that all the frequencies fit in the specified bandwidth.
         """
-        return mp.GaussianSource(
+        srct = mp.GaussianSource(
             np.mean(self._frequencies),
             fwidth=fwidth_frac * np.mean(self._frequencies),
         )
+        self.sim.fields.register_src_time(srct.swigobj)
+        return srct
 
 
 class EigenmodeCoefficient(ObjectiveQuantity):
@@ -273,12 +275,12 @@ class FourierFields(ObjectiveQuantity):
             scale = amp_arr * self._adj_src_scale(include_resolution=False)
             
             if self.num_freq == 1:
-                sources += [mp.IndexedSource(time_src, fourier_data, scale[:,0])]
+                sources += [mp.IndexedSource(time_src, fourier_data, scale[:,0], not self.yee_grid)]
             else:
                 src = FilteredSource(time_src.frequency,self._frequencies,scale,self.sim.fields.dt)
                 (num_basis, num_pts) = src.nodes.shape
                 for basis_i in range(num_basis):
-                    sources += [mp.IndexedSource(src.time_src_bf[basis_i], fourier_data, src.nodes[basis_i])]
+                    sources += [mp.IndexedSource(src.time_src_bf[basis_i], fourier_data, src.nodes[basis_i], not self.yee_grid)]
         return sources
 
     def __call__(self):
--- a/python/adjoint/objective.py
+++ b/python/adjoint/objective.py
@@ -109,12 +109,10 @@ class ObjectiveQuantity(abc.ABC):
         The user may specify a scalar valued objective function across multiple frequencies (e.g. MSE) in
         which case we should check that all the frequencies fit in the specified bandwidth.
         """
-        srct = mp.GaussianSource(
+        return mp.GaussianSource(
             np.mean(self._frequencies),
             fwidth=fwidth_frac * np.mean(self._frequencies),
         )
-        self.sim.fields.register_src_time(srct.swigobj)
-        return srct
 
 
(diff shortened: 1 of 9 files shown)


src/fix_boundary_sources.cpp
```diff
@@ -0,0 +1,176 @@
...
+        }
+    }
+  }
+
+  // we need each process's data to be contiguous
+  srcpt_info_compare compare = {chunks};
+  std::sort(boundarysources.begin(), boundarysources.end(), compare);
+
+  // collect 2d (row-major) arrays offsets and numcomm,
+  // where numcomm[i,j] is the number of srcpt_info items
```
stevengj, line 84: Note: `numcomm[i,j]` here is stored at `numcomm[i*P + j]` (this is "row major order").

mawc2019:
In `fix_boundary_sources.cpp`, although the value of `srcpts[0].src_time_id` in the source process immediately after `MPI_Send(srcpts, N, mpi_srcpt_info, pdest, psrc*P + pdest, mycomm)` at Line [164](https://github.com/NanoComp/meep/blob/997a1a720425f7b98c33135ca0d023037521dd01/src/fix_boundary_sources.cpp#L164) is the same as that in the destination process immediately after `MPI_Recv(srcpts, N, mpi_srcpt_info, psrc, psrc*P + pdest, mycomm, &status)` at Line [129](https://github.com/NanoComp/meep/blob/997a1a720425f7b98c33135ca0d023037521dd01/src/fix_boundary_sources.cpp#L129), `src_time *srctime = lookup_src_time(srcpts[0].src_time_id)` gives different `srctime`, which is not `NULL` in the source process but is often `NULL` in the destination process. The null `srctime` in the destination process causes the [error](https://github.com/NanoComp/meep/blob/997a1a720425f7b98c33135ca0d023037521dd01/src/fix_boundary_sources.cpp#L148) later. Occasionally the error does not appear even though the same code is run.

stevengj:
Went over the code with @mawc2019 today, and found a typo in the `srcpt_info_offsets` definition which was causing the problem noted [above](https://github.com/NanoComp/meep/pull/1959#issuecomment-1062370494).

mawc2019:
The error message of the online check is as follows, but that c++ file already has `#include <algorithm>`.
```
../../../src/fix_boundary_sources.cpp: In member function ‘void meep::fields::fix_boundary_sources()’:
../../../src/fix_boundary_sources.cpp:81:8: error: ‘sort’ is not a member of ‘std’; did you mean ‘sqrt’?
   81 |   std::sort(boundarysources.begin(), boundarysources.end(), compare);
      |        ^~~~
```
There is no error when it is checked offline.

src/fix_boundary_sources.cpp
```diff
@@ -0,0 +1,176 @@
...
+#ifdef NEED_UNDEF_SEEK_FOR_MPI
+// undef'ing SEEK_* is needed for MPICH, possibly other MPI versions
+#undef SEEK_SET
+#undef SEEK_END
+#undef SEEK_CUR
+#endif
+#include <mpi.h>
+#include <algorithm>
+#endif
+
```
stevengj, line 14: You are only doing `#include <algorithm>` on MPI builds (it is inside the `#ifdef HAVE_MPI`) — you have to do it on serial builds too.

```suggestion
#include <mpi.h>
#endif

#include <algorithm>

```

stevengj:
See my suggested patch above.

python/adjoint/objective.py
```diff
@@ -273,12 +273,12 @@ def place_adjoint_source(self, dJ):
             scale = amp_arr * self._adj_src_scale(include_resolution=False)
             
             if self.num_freq == 1:
-                sources += [mp.IndexedSource(time_src, fourier_data, scale[:,0])]
+                sources += [mp.IndexedSource(time_src, fourier_data, scale[:,0], not self.yee_grid)]
```
stevengj, line 276: Note that this only affects code that use Fourier fields `yee_grid=false`.

mawc2019:
Here is an example. Without this PR, the finite-difference and adjoint gradients are inconsistent when the number of processes is large, such as 8. With this PR, the finite-difference and adjoint gradients become consistent.
image
The code is as follows.

```
import meep as mp
import meep.adjoint as mpa
import numpy as np

Lx,Ly,pml_size = 2,4,1

sim = mp.Simulation(

nf = mpa.FourierFields(sim,mp.Volume(center=mp.Vector3(0,design_height+0.2),size=mp.Vector3(Lx,0)),mp.Ez)
ob_list = [nf]

def J(near_data):

x0 = 0.5*np.ones(Nx*Ny)
opt.update_design([x0])
f0, dJ_deps = opt()

db, choose = 1e-4, 5
np.random.seed(20211204)

g_adjoint = dJ_deps.flatten()

```

PR #1968, opened 2022-02-27
add support for out-of-plane wavevector in 2d cell to fields::get_eigenmode

### LINKED ISSUE #1957: mode decomposition for 2d cell with out-of-plane wavevector
All the examples in [Tutorial/Mode Decomposition](https://meep.readthedocs.io/en/latest/Python_Tutorials/Mode_Decomposition) involve a 2d cell with *in-plane* wavevector. Since Meep supports the special case of a [2d cell with *out-of-plane* wavevector](https://meep.readthedocs.io/en/latest/2d_Cell_Special_kz/), it would be useful to demonstrate this feature using mode decomposition. Currently, there is only one example for the `kz_2d` feature involving a flux calculation which is not that interesting.  As a practical example, we could demonstrate how to compute the diffraction orders via `get_eigenmode_coefficients` of a 2d binary-phase grating for an incident planewave with out-of-plane wavevector.

oskooi:
Closes #1957.

### DIFF
--- a/python/tests/test_binary_grating.py
+++ b/python/tests/test_binary_grating.py
@@ -1,3 +1,5 @@
+# -*- coding: utf-8 -*-
+
 import unittest
 import parameterized
 import meep as mp
@@ -9,14 +11,14 @@ import numpy as np
 class TestEigCoeffs(unittest.TestCase):
   @classmethod
   def setUpClass(cls):
-    cls.resolution = 30        # pixels/μm
+    cls.resolution = 30    # pixels/μm
 
-    cls.dpml = 1.0             # PML thickness
-    cls.dsub = 1.0             # substrate thickness
-    cls.dpad = 1.0             # padding thickness between grating and PML
-    cls.gp = 6.0               # grating period
-    cls.gh = 0.5               # grating height
-    cls.gdc = 0.5              # grating duty cycle
+    cls.dpml = 1.0         # PML thickness
+    cls.dsub = 1.0         # substrate thickness
+    cls.dpad = 1.0         # padding thickness between grating and PML
+    cls.gp = 6.0           # grating period
+    cls.gh = 0.5           # grating height
+    cls.gdc = 0.5          # grating duty cycle
 
     cls.sx = cls.dpml+cls.dsub+cls.gh+cls.dpad+cls.dpml
     cls.sy = cls.gp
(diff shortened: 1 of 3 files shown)


oskooi:
The unit test for mode decomposition and the `kz_2d` feature added in this PR to `test_binary_grating.py` as `test_binary_grating_special_kz` is failing because of a possible bug in this feature (or possibly the test itself, which is unlikely). The test attempts to verify that the total reflected and transmitted flux obtained using `get_fluxes` for an incident planewave with out-of-plane wavevector is equivalent to the sum of the flux of all the individual reflected and transmitted diffraction orders computed using `get_eigenmode_coefficients`. The test is failing using `kz_2d="3d"` and I have yet to try it with the other two options `"complex"` and `"real/imag"`. This test is identical to the existing `test_binary_grating_oblique` in `test_binary_grating.py` which involves an in-plane wavector and is passing.

stevengj:
Maybe try a longer wavelength where there is only the 0th (specular) diffracted order, just to simplify things.

stevengj:
Don't forget that kz ≠ 0 allows the grating structure to couple the s and p polarizatios.  So you have to compute both the s and p diffracted orders and add up their powers.

oskooi:
Summing up the flux in each of the diffraction orders for both the S- and P-polarizations does indeed produce the same result as the total Poynting flux. I have added a unit test in `test_binary_grating.py` which verifies this result for each of the three cases for `kz_2d` of `real/imag`, `complex`, and `3d`.

stevengj:
Notes:

* you are specifying kx and ky, but that doesn't mean ky = 0.   ky is determined by ω, kx, and ky.
* this means that the plane of incidence is oblique with respect to the xy or xz planes.
* you used an Ez line source, but this will generate an incident wave that has some S and some P component, because z is not orthogonal to or parallel to the plane of incidence.
* even if your incident wave were purely S or purely P, the scattered wave would contain both S and P, because the surface breaks the rotational symmetry.

stevengj:
Might be nice to have a test for the relative phase of the eigenmode coefficients.   You can just do a transmission calculations through a flat interface between a material ε1 and air, and another simulation through ε2 and air; then the relative phase of the two eigenmode transmission coefficients should match the prediction from the Fresnel equations.

PR #2010, opened 2022-03-21
Conic Filters off by 1 in the Dimension

danogbe:
When I was doing the tutorials for the meep adjoint solver module, every time the conic_filter function gets called, the Nx and Ny is off by 1, so I don't think that there should be an extra one added in lines 232 and 233 in filters.py.


smartalecH:
(take a look at #2012)

stevengj:
Actually, it looks to me that the [tutorial](https://nbviewer.org/github/NanoComp/meep/blob/master/python/examples/adjoint_optimization/03-Filtered_Waveguide_Bend.ipynb) is off: the lines
```py
Nx = int(design_region_resolution*design_region_width)
Ny = int(design_region_resolution*design_region_height)
```
should both have been changed to add ` + 1` when we fixed this in the material grid.

PR #2016, opened 2022-03-23
Fix off-by-one filtering errors.

### LINKED ISSUE #2012: New adjoint filtering scheme
The current filtering scheme often induces a spatial delay in each dimension. This implies the filter kernel is not a ["zero-phase" filter](https://ccrma.stanford.edu/~jos/fp/Zero_Phase_Filters_Even_Impulse.html). In theory, all of the example impulse responses currently in the codebase (top-hat, conic, and Gaussian) *should* be zero-phase, indicating our setup is wrong.

One way to get around this is to implement a "filt-filt" (essentially filter in both directions, and remove the phase delay). This is what is implemented in #1625. This, of course, is *not* what we want, as this changes the underlying kernel of the final filter operation, rendering all of our minimum length-scale constraint math invalid. It should be possible to properly implement a single-pass filter with the above kernels.

Since all of our kernels are *separable*, we can actually do a batched convolution (natively supported by `autograd`) so long as we construct our filter kernels properly and apply the right padding (to preserve the correct shape).

Right now we do an fft-based filter. We may not need to with a batched convolution (depending on the size of the kernel, an fft convolution [may be *slower*](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.fftconvolve.html)).

(cc @hammy4815)

smartalecH:
Fixes #2012.

Note this is just a patch. A future PR could really improve the performance of our filtering, as described in #2012.

For now, this PR ensures that the impulse response is always Type I (even with an odd number of taps) so that it is zero phase. It adds a new function, `compute_mg_dims`, to help with this.

Also adds a test to check for "off-by-one" errors.

### DIFF
--- a/python/adjoint/filters.py
+++ b/python/adjoint/filters.py
@@ -7,6 +7,19 @@ from autograd import numpy as npa
 import meep as mp
 from scipy import special
 
+def compute_mg_dims(Lx,Ly,resolution):
+    '''Compute the material grid dimensions from
+    the corresponding resolution, x-size, and y-size.
+    The grid dimensions must be odd.    
+    '''
+    Nx = int(Lx * resolution)
+    Ny = int(Ly * resolution)
+    if (Nx % 2) == 0:
+        Nx += 1
+    if (Ny % 2) == 0:
+        Ny += 1
+    return Nx, Ny
+
 
 def _centered(arr, newshape):
     '''Helper function that reformats the padded array of the fft filter operation.
@@ -98,8 +111,7 @@ def simple_2d_filter(x, kernel, Lx, Ly, resolution, symmetries=[]):
         The output of the 2d convolution.
     """
     # Get 2d parameter space shape
-    Nx = int(Lx * resolution) + 1
-    Ny = int(Ly * resolution) + 1
+    Nx, Ny = compute_mg_dims(Lx,Ly,resolution)
     (kx, ky) = kernel.shape
 
     # Adjust parameter space shape for symmetries
@@ -179,8 +191,7 @@ def cylindrical_filter(x, radius, Lx, Ly, resolution, symmetries=[]):
     density-based topology optimization. Archive of Applied Mechanics, 86(1-2), 189-218.
     '''
     # Get 2d parameter space shape
-    Nx = int(Lx * resolution) + 1
-    Ny = int(Ly * resolution) + 1
+    Nx, Ny = compute_mg_dims(Lx,Ly,resolution)
 
     # Formulate grid over entire design region
     xv, yv = np.meshgrid(np.linspace(-Lx / 2, Lx / 2, Nx),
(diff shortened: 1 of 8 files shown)


oskooi:
For consistency, we also need to update the following lines in `test_material_grid.py` to use `mpa.compute_mg_dims`:

https://github.com/NanoComp/meep/blob/34b62e94f9bdbc035d9c0eba1c87743c9bb32f0b/python/tests/test_material_grid.py#L17-L18

https://github.com/NanoComp/meep/blob/34b62e94f9bdbc035d9c0eba1c87743c9bb32f0b/python/tests/test_material_grid.py#L91-L94

smartalecH:
> For consistency, we also need to update the following lines in test_material_grid.py to use mpa.compute_mg_dims

Fixed, thanks.

smartalecH:
CI doesn't want to trigger for some reason...

python/adjoint/filters.py
```diff
@@ -7,6 +7,19 @@
 import meep as mp
 from scipy import special
 
+def compute_mg_dims(Lx,Ly,resolution):
+    '''Compute the material grid dimensions from
+    the corresponding resolution, x-size, and y-size.
+    The grid dimensions must be odd.
```
stevengj, line 13: Why?

smartalecH: If the number of "taps" is not odd, it's no longer a zero-phase filter.

smartalecH:
Should be ready to merge now. 

The implementation is agnostic to the number of filter taps. No need to "add 1" or check that's it's odd.

stevengj:
LGTM.

mawc2019:
It seems that the 2d filters in `filters.py`, including [`cylindrical_filter`](https://github.com/NanoComp/meep/blob/master/python/adjoint/filters.py#L89), [`conic_filter`](https://github.com/NanoComp/meep/blob/master/python/adjoint/filters.py#L135), and [`gaussian_filter`](https://github.com/NanoComp/meep/blob/master/python/adjoint/filters.py#L181), should have kernels with circular boundaries, but the code in these functions implies that the cross sections are rectangular. For example, in `cylindrical_filter`, the kernel is defined in [this way](https://github.com/NanoComp/meep/blob/master/python/adjoint/filters.py#L115-L126). This issue was also mentioned [here](https://github.com/NanoComp/meep/commit/11545a1b859222df2c404370673115666b589c22#r82552977). To avoid this problem, we may define the kernels as what was done before this PR (with some minor revisions).

In addition, as one of the steps for fixing the off-by-one error, this PR forces the size of each kernel to be an odd number along every direction, which is indicated by [`_proper_pad`](https://github.com/NanoComp/meep/blob/master/python/adjoint/filters.py#L22). This measure may cause errors in the sizes of kernels. Could the off-by-one error be fixed in a way without introducing the odd-number restriction?

PR #2017, opened 2022-03-24
revamp test_refl_angular.py to validate results against theory rather than hard-coded values

oskooi:
Currently, [`test_refl_angular.py`](https://github.com/NanoComp/meep/blob/master/python/tests/test_refl_angular.py) involves computing the reflectance of a 1d interface for the *S* polarization at *normal* incidence and validating the results against hard-coded values. The filename is therefore a misnomer since the test does not actually involve a planewave at oblique incidence. This PR modifies this test to validate the results using the Fresnel equations for two angles of incidence (0° and 20.6°). This makes for a more rigorous and complete test.

This test will eventually be expanded to include a validation of the phase of a total internal reflected mode obtained using mode decomposition via the `kz_2d` feature as described by @stevengj in [#1968 (comment)](https://github.com/NanoComp/meep/pull/1968#issuecomment-1069836738). (The Fresnel equations for the phase are provided in slide 20 of [these notes](http://www.patarnott.com/atms749/pdf/FresnelEquations.pdf).)

Also included in this PR are small fixes/tweaks to the related [Tutorial/Angular Reflectance Spectrum of a Planar Interface](https://meep.readthedocs.io/en/latest/Python_Tutorials/Basics/#angular-reflectance-spectrum-of-a-planar-interface).


stevengj:
`h5py.File(fname, 'w')` should overwrite any existing files.

stevengj:
We could run into trouble if multiple tests are run in parallel and are writing to files with the same names — we should make sure that the tests use a unique filename prefix for each test (e.g. based on the name of the test file).

oskooi:
This is good to go.

PR #2021, opened 2022-03-25
add stop_when_energy_decayed termination condition

oskooi:
For simulations involving a large computational cell in which the sources and the monitors are separated by a large distance (i.e., tens to hundreds of wavelengths), the available termination conditions `stop_when_fields_decayed` and `stop_when_dft_decayed` are inadequate because they often stop the simulation prematurely as the fields have not had sufficient time to propagate throughout the entire cell.

This PR introduces a new termination condition `stop_when_energy_decayed` which tracks the electric energy over the *entire* cell. However, the cost of calling the function `electric_energy_in_box` at every timestep to obtain the energy can be significant: in tests, ~20X slow down in the time-stepping rate compared to using a fixed runtime termination condition.

The performance penalty notwithstanding, some preliminary tests have demonstrated that this feature provides an improvement compared to using a large, fixed, and overly conservative run time chosen arbitrarily.

### DIFF
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -2852,10 +2852,9 @@ def stop_when_energy_decayed(dt=None, decay_by=None):
 <div class="function_docstring" markdown="1">
 
 Return a `condition` function, suitable for passing to `Simulation.run` as the `until`
-or `until_after_sources` parameter, that examines the electric energy over the entire
-cell volume and keeps running until its absolute value has decayed by at least `decay_by`
-from its maximum previous value. In particular, it keeps incrementing the run time by `dt`
-(in Meep units) and checks the maximum value over that time period &mdash.
+or `until_after_sources` parameter, that examines the field energy over the entire
+cell volume at every `dt` time units and keeps incrementing the run time by `dt`  until
+its absolute value has decayed by at least `decay_by` from its maximum recorded value.
 
 Note that, if you make `decay_by` very small, you may need to increase the `cutoff`
 property of your source(s), to decrease the amplitude of the small high-frequency
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -87,40 +87,40 @@ control various parameters of the Meep computation.
 
 ```python
 def __init__(self,
-             cell_size:Union[meep.geom.Vector3, Tuple[float, ...], NoneType]=None,
-             resolution:float=None,
-             geometry:Union[List[meep.geom.GeometricObject], NoneType]=None,
-             sources:Union[List[meep.source.Source], NoneType]=None,
-             eps_averaging:bool=True,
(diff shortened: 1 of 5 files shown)


stevengj:
> The cost of calling the function electric_energy_in_box at every timestep to obtain the energy is trivial and thus there is no performance penalty for using this new feature.

Is there any particular reason to call it on every timestep?

oskooi:
> Is there any particular reason to call it on every timestep?

Given the performance penalty of calling `field_energy_in_box` which requires looping over all grid points, it is now called at `dt` time units (which is a required parameter) rather than at each time step.

> Also, note that you might want to look at field_energy_in_box instead of just the electric energy.

Done.

doc/docs/Python_User_Interface.md
```diff
@@ -87,40 +87,40 @@ control various parameters of the Meep computation.
...
-             filename_prefix:Union[str, NoneType]=None,
-             output_volume:Union[meep.simulation.Volume, NoneType]=None,
-             output_single_precision:bool=False,
-             geometry_center:Union[meep.geom.Vector3, Tuple[float, ...]]=Vector3<0.0, 0.0, 0.0>,
-             force_all_components:bool=False,
-             split_chunks_evenly:bool=True,
+             cell_size: Union[meep.geom.Vector3, Tuple[float, ...], NoneType] = None,
+             resolution: float = None,
+             geometry: Optional[List[meep.geom.GeometricObject]] = None,
+             sources: Optional[List[meep.source.Source]] = None,
```
stevengj, line 93: What's going on with all of these `Union -> Optional` changes?   Seems like they belong in a separate PR.

oskooi:
Those changes were caused by running `make python_api_doc` which was necessary to update the markdown page in the documentation.

doc/docs/Python_User_Interface.md
oskooi: #1919 added the `Optional` type hint to various parameters of the `Simulation` constructor such as `geometry`, `sources`, etc. The type hint `Optional[Type]` is equivalent to `Union[Type, None]` according to [PEP 484](https://peps.python.org/pep-0484/#union-types). The default values of each type are parsed by the `ast` module in [`doc/generate_py_api.py`](https://github.com/NanoComp/meep/blob/55f881fef388d591d201fec7b5b139d39db6747b/doc/generate_py_api.py#L140-L142). Because I have been using different versions of Python (3.5, 3.8) to call `generate_py_api.py` in various PRs which involve updating the docs (i.e., #1958, #1926, etc.), the different versions of `ast` seem to have been interchanging `Optional` and `Union`. The same thing seems to be happening again in this PR. `ast` is also inserting an extra space before and after the `=` sign. We therefore probably need to edit `generate_py_api.py` in a separate PR to ensure consistent parsing of the type hints and thus avoid these unnecessary changes.

PR #2040, opened 2022-04-13
validation check in loop_in_chunks

### LINKED ISSUE #2028: add validation check to loop_in_chunks
For things like #1895, it might be good to add a validation check to the `loop_in_chunks` code: as we loop over the chunks, keep a running sum of the total number of grid points in all of the [`chunkloop` calls](https://github.com/NanoComp/meep/blob/55f881fef388d591d201fec7b5b139d39db6747b/src/loop_in_chunks.cpp#L497-L498), and then check at the end that this total equals the number of points in the [original volume](https://github.com/NanoComp/meep/blob/55f881fef388d591d201fec7b5b139d39db6747b/src/loop_in_chunks.cpp#L355-L356).  (If not, emit a warning.)

I think this should be cheap, so it wouldn't hurt to do it all the time.  But if we want we could only enable it in `--enable-debug` mode.

cc @mochen4, @mawc2019

mochen4:
Closes #2028. Added validation check to loop_in_chunks.

### DIFF
--- a/src/loop_in_chunks.cpp
+++ b/src/loop_in_chunks.cpp
@@ -354,7 +354,10 @@ void fields::loop_in_chunks(field_chunkloop chunkloop, void *chunkloop_data, con
   volume wherec(where + yee_c);
   ivec is(vec2diel_floor(wherec.get_min_corner(), gv.a, zero_ivec(gv.dim)) - iyee_c);
   ivec ie(vec2diel_ceil(wherec.get_max_corner(), gv.a, zero_ivec(gv.dim)) - iyee_c);
-  
+
+  vec s0(gv.dim), e0(gv.dim), s1(gv.dim), e1(gv.dim);
+  compute_boundary_weights(gv, where, is, ie, snap_empty_dimensions, s0, e0, s1, e1);
+
   int original_vol = 1;
   LOOP_OVER_DIRECTIONS(gv.dim, d){
     if (boundaries[High][d] == Periodic && boundaries[Low][d] == Periodic) {
@@ -363,9 +366,6 @@ void fields::loop_in_chunks(field_chunkloop chunkloop, void *chunkloop_data, con
       original_vol *= (min(user_volume.big_owned_corner(cgrid), ie).in_direction(d) - max(user_volume.little_owned_corner(cgrid),is).in_direction(d))/2+1;
     }
   }
-
-  vec s0(gv.dim), e0(gv.dim), s1(gv.dim), e1(gv.dim);
-  compute_boundary_weights(gv, where, is, ie, snap_empty_dimensions, s0, e0, s1, e1);
 
   // loop over symmetry transformations of the chunks:
   int vol_sum = 0;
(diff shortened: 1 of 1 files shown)


src/loop_in_chunks.cpp
```diff
@@ -358,7 +358,18 @@ void fields::loop_in_chunks(field_chunkloop chunkloop, void *chunkloop_data, con
   vec s0(gv.dim), e0(gv.dim), s1(gv.dim), e1(gv.dim);
   compute_boundary_weights(gv, where, is, ie, snap_empty_dimensions, s0, e0, s1, e1);
 
+  int original_vol = 1;
+  LOOP_OVER_DIRECTIONS(gv.dim, d){
+    if (boundaries[High][d] == Periodic && boundaries[Low][d] == Periodic) {
+      original_vol *= (ie.in_direction(d) - is.in_direction(d))/2+1;
+    } else {
+      int vol_size_d = (min(user_volume.big_owned_corner(cgrid), ie).in_direction(d) - max(user_volume.little_owned_corner(cgrid),is).in_direction(d))/2+1;
```
mochen4, line 366: Sometimes, ``where`` doesn't intersect with the ``user_volume``, so the size on that direction is negative.

src/loop_in_chunks.cpp
```diff
@@ -508,6 +524,8 @@ void fields::loop_in_chunks(field_chunkloop chunkloop, void *chunkloop_data, con
       }
     } while (ishift != min_ishift);
   }
+  int vol_sum_all = sum_to_all(vol_sum);
+  if (use_symmetry && vol_sum_all != original_vol) master_printf("WARNING vol mismatch:, original_vol %i, looped vol_sum %i \n", original_vol, vol_sum_all);
```
mochen4, line 528: When ``use_symmetry=false``,  e.g. 
https://github.com/NanoComp/meep/blob/75ced5296a91d8e49ad61b1bb07093e7abbb95e9/src/sources.cpp#L460
the symmetry operations won't be applied to chunks and thus don't intersect the entire original volume.

src/loop_in_chunks.cpp
```diff
@@ -406,7 +417,7 @@ void fields::loop_in_chunks(field_chunkloop chunkloop, void *chunkloop_data, con
         grid_volume gvu(chunks[i]->gv);
         ivec _iscoS(S.transform(gvu.little_owned_corner(cS), sn));
         ivec _iecoS(S.transform(gvu.big_owned_corner(cS), sn));
-        ivec iscoS(max(user_volume.little_owned_corner(cgrid), min(_iscoS, _iecoS))), iecoS(max(_iscoS, _iecoS)); // fix ordering due to to transform
+        ivec iscoS(max(user_volume.little_owned_corner(cgrid), min(_iscoS, _iecoS))), iecoS(min(user_volume.big_owned_corner(cgrid), max(_iscoS, _iecoS))); // fix ordering due to to transform
```
mochen4, line 420: A potential bug of #1895 is exposed and fixed by this PR. Previously, the grid_volume ``gvu`` was thought to be always inside the ``big_owned_corner`` of the ``user_volume`` so ``min`` wasn't taken.

stevengj: Would be nice to know the source of this being violated…

PR #2046, opened 2022-04-19
Update documentation showing loading a chunk_layout from file

bencbartlett:
A MEEP user contacted me on twitter and was confused after trying to use the edited demo code. In the case that `chunk_layout=None`, this code would throw an error, as `chunk_layout` did not point to `sim.chunk_layout` after `sim.init_sim()` was called. The modified code should avoid this confusion.


stevengj:
Thanks!

PR #2053, opened 2022-04-30
tutorial and unit test for near-to-far field transformation with symmetries

### LINKED ISSUE #2042: unit test/tutorial for near2far+symmetry for antenna+ground
A nice unit test might be to compute the far-field radiation pattern for an [antenna above a perfectly conducting ground](http://www.vias.org/radioanteng/radio_antenna_engineering_03_06_02.html).

Analytically, this can be computed by the method of images (it is the Green's function minus a mirror image).

Computationally, we'd do something similar, because the near2far transform in Meep only supports the homogeneous-medium Green's function, so it can't explicitly take into account the ground.  So, you'd model it as two antennas of opposite sign (odd mirror symmetric) surrounded by a single near2far box enclosing both sources.  Then impose an odd mirror symmetry in Meep so that it cuts the computational cell in half, effectively equivalent to a PEC boundary condition on one side.

Alternatively, you could have the near2far surface be an infinite plane above the dipole, but in that case you need #1952 or similar.

oskooi:
Closes #2042.

Adds a unit test and tutorial which involves computing the radiation pattern of an antenna positioned a given height above a perfect-electric conductor (PEC) ground plane and validating the result using analytic theory.


oskooi:
@aristeidis-karalis, please take a look at the tutorial "Antenna above a Perfect Electric Conductor Ground Plane" in `Near_to_Far_Field_Spectra.md` of this pull request and see whether it is alright.

aristeidis-karalis:
> @aristeidis-karalis, please take a look at the tutorial "Antenna above a Perfect Electric Conductor Ground Plane" in `Near_to_Far_Field_Spectra.md` of this pull request and see whether it is alright.

The text looks fine to me and it looks like you also get agreement in simulations.

Issue #132, opened 2017-11-21
bug when combining 3 mirror symmetries with periodic boundary conditions in 3d


oskooi:
As has been already reported on the [meep-discuss list](https://www.mail-archive.com/meep-discuss@ab-initio.mit.edu/msg05556.html), there is a bug which leads to field instabilities when combining 3 mirror-symmetry objects with periodic boundary conditions in 3d. 

This is demonstrated by the following simple example of a point source in vacuum:
```scm
(set-param! resolution 10)
(set! geometry-lattice (make lattice (size 1 1 1)))

(set! sources (list (src (make source (src (make gaussian-src (frequency 1) (fwidth 1))) (component Ez) (center 0 0 0)))))

(set! k-point (vector3 0 0 0))

(set! symmetries (list (make mirror-sym (direction X))
                       (make mirror-sym (direction Y))
                       (make mirror-sym (direction Z) (phase -1))))

(define print-field (lambda () (print "ez:, " (meep-time) ", " (magnitude (get-field-point Ez (vector3 0 0 0))) "\n")))

(run-until 500 (at-every 3.4 print-field))
```
For this script, the fields begin to blow up early on:
```
-----------
Initializing structure...
Working in 3D dimensions.
Computational cell is 1 x 1 x 1 with resolution 10
Halving computational cell along direction x
Halving computational cell along direction y
Halving computational cell along direction z
time for set_epsilon = 0.000983953 s
-----------
ez:, 3.4000000000000004, 0.11299371272955527
ez:, 6.800000000000001, 1.3292101237116931
ez:, 10.200000000000001, 2.402271717485686
ez:, 13.600000000000001, 4.484463313924006
ez:, 17.0, 3.8996648343918103
ez:, 20.400000000000002, 0.13634058216335865
ez:, 23.8, 4.629265409516416
ez:, 27.200000000000003, 4.998705102736996
ez:, 30.6, 1.2298673000561804
ez:, 34.0, 2.6981516161770243
ez:, 37.4, 4.377899104931889
ez:, 40.800000000000004, 3.5929413491696733
ez:, 44.2, 15.319503849169674
ez:, 47.6, 1194.3195038491704
ez:, 51.0, 2915498.319503849
ez:, 54.400000000000006, 559645525.6804962
ez:, 57.800000000000004, 4232176173909.6807
```
There are no instabilities if the Z mirror-symmetry object is removed:
```
-----------
Initializing structure...
Working in 3D dimensions.
[...]

PR #2058, opened 2022-05-04
round down to even

mochen4:


mochen4:
I tried the code in https://github.com/NanoComp/meep/issues/132#issuecomment-600212851 and the fields still blew up.

PR #2054, opened 2022-05-02
Add support for `DiffractedPlanewave` to `EigenmodeCoefficient` objective function of adjoint solver

oskooi:
Fixes #2047.

This is a draft and still needs a unit test.

It seems all that is required to enable this feature is to simply reverse (when necessary) the direction of the wavevector of the `DiffractedPlanewave` object (its `g` property) in the `place_adjoint_source` and `__call__` functions of the `EigenmodeCoefficient` class. No other changes to the adjoint solver are necessary.

PR #2069, opened 2022-05-12
`DiffractedPlanewave` objects can only be used with mode decomposition and not mode sources

oskooi:
Related to #2054.

Some of the (unrelated) changes are from rebuilding the documentation using `make python_api_doc`.

### DIFF
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -87,40 +87,40 @@ control various parameters of the Meep computation.
 
 ```python
 def __init__(self,
-             cell_size: Union[meep.geom.Vector3, Tuple[float, ...], NoneType] = None,
-             resolution: float = None,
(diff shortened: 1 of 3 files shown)


stevengj:
Just to be clear, a `DiffractedPlanewave` specification is perfectly mathematically meaningful for an eigenmode source.   However, most of the time you have an incident planewave you will only have a single incident planewave, and it's probably easier to just specify k_point … as long as the plane of incidence is oriented in the xy or xz plane so that you can use the parity feature to select the s or p polarization.

stevengj:
We don't have a tutorial for using it as a source, which makes it harder to use, but I don't think we should remove it from the manual.

stevengj:
To launch a planewave at a particular angle, probably the easiest thing to do is to (a) from the angle and frequency, calculate the `k_point`, and (b) use a DiffractedPlanewave with order 0.  (Basically, you are just using it to specify the s and p amplitudes conveniently.)

PR #2072, opened 2022-05-13

oskooi:
Adds a tutorial and unit test based on the discussion in #2069, particularly this [comment](https://github.com/NanoComp/meep/pull/2069#issuecomment-1124459981).


python/tests/test_diffracted_planewave.py
```diff
@@ -121,17 +138,100 @@ def _pw_amp(x):
...
+                                    eig_band=mp.DiffractedPlanewave((0,0,0),
+                                                                    mp.Vector3(0,1,0),
+                                                                    1,
+                                                                    0))]
+    else:
+      sources = [mp.EigenModeSource(mp.GaussianSource(self.fcen,fwidth=self.df),
+                                    center=self.src_pt,
+                                    size=mp.Vector3(0,self.sy,0),
+                                    direction=mp.NO_DIRECTION,
+                                    eig_kpoint=k,
```
stevengj, line 190: This line seems unnecessary — since the source spans the whole width of the cell, it should automatically set the corresponding k component of the eigenmode source, whereas the orthogonal k component will be found by the mode solver.

oskooi: Actually, it turns out that `eig_kpoint=k` *is* necessary because of the line before it: `direction=mp.NO_DIRECTION`. Removing `eig_kpoint` but keeping `direction=mp.NO_DIRECTION` produces this error:
```
meep: invalid direction in add_eigenmode_source
```
This is because setting `direction=mp.NO_DIRECTION` also requires specifying an `eig_kpoint`.

Removing both `eig_kpoint` and `direction=mp.NO_DIRECTION` causes the test to fail because the default `direction=mp.AUTOMATIC` assumes that the `k_point` is always *perpendicular* to the line monitor which it is not. This leads to a wrong mode to be used for the mode decomposition which causes the test to fail:
```
FAIL: test_mode_source (__main__.TestDiffractedPlanewave)
  File "/meep/python/tests/test_diffracted_planewave.py", line 231, in test_mode_source
    self.assertAlmostEqual(tran_bandnum,
AssertionError: 380.9518159531796 != 325.7552498084876 within 4 places (55.19656614469204 difference)
```
The existing setup is therefore correct.

stevengj: > because the default `direction=mp.AUTOMATIC` assumes that the `k_point` is always perpendicular to the line monitor which it is not.

I don't follow — if the monitor region spans the whole width of the cell, for Bloch-periodic boundaries it sets the parallel components of `k` to those of the boundary conditions, and then solves for the perpendicular component, no?

oskooi: I think the issue is related to these lines in `add_source`:

https://github.com/NanoComp/meep/blob/91d36ba73d1ae0eb135d18c783840cff9fcd71ef/python/simulation.py#L2464-L2468

That is, when creating an `EigenModeSource` with the default `direction` of `mp.AUTOMATIC` (which has a value of -1), the `direction` is assigned to the normal direction of the monitor and thus `eig_kpoint` is ignored.

oskooi: The observed behavior is a feature and not a bug. See [#2107 (comment)](https://github.com/NanoComp/meep/pull/2107#discussion_r963118923).

PR #2073, opened 2022-05-14
fix calculate_fd_gradient to deal with design pixel weights near 0 or 1

mawc2019:
The function `calculate_fd_gradient` does not return the correct finite-difference gradients when the design pixel weight is 0, 1, or too close to them. This PR replaces the central difference with forward or backward differences in these circumstances, and also forbids a too large (≥0.2) step size for finite difference.


mawc2019:
I tested it with uniform design patterns such as `np.zeros(Nx*Ny)`, `np.ones(Nx*Ny)`, `1.5*np.ones(Nx*Ny)`, etc. The adjoint and finite-difference gradients were compared.

PR #2076, opened 2022-05-18
tutorial and unit test for LDOS of planar cavity with lossless metallic walls

oskooi:
Adds a new tutorial ("Planar Cavity with Lossless Metallic Walls") and unit test based on the results in [IEEE J. Quantum Electronics, Vol. 34, pp. 71-76 (1998)](https://ieeexplore.ieee.org/abstract/document/655009).

This replaces the existing unit test for the LDOS feature which just validated the simulated results against hard-coded values.


stevengj:
Two possible follow-up tests:

1. You could try it in cylindrical coordinates.
2. You could try it in a 1d simulation (along x), integrating the result over ky and kz, or just over ky treated as a polar coordinate (i.e. integration weight 2π ky dky)

doc/docs/Python_Tutorials/Local_Density_of_States.md
```diff
@@ -2,21 +2,150 @@
...
+
+An important feature of this cavity system are the lossless metal sidewalls. These mirrors are perfectly reflective at all angles and enable the Purcell enhancement factor to be tuned over a wide range of ~0.5 to 3.0. There are other types of lossless mirrors with reflectivity of +1 (0 phase shift) rather than -1 (π phase shift) as in this example based on e.g., a multilayer stack in which the index of the cavity is *larger* than the index of the first mirror layer. These mirrors tend to have a strong angular dependence of the reflectivity and thus generally a smaller range of variation for the Purcell enhancement. In this demonstration, the cavity thickness is swept over a range of 0.5 to 2.5. Below a thickness of 0.5 there are no guided modes and thus the Purcell enhancement factor is zero.
```
stevengj, line 19:
```suggestion
In this demonstration, the cavity thickness is swept over a range of 0.5 to 2.5. Below a thickness of 0.5 there are no guided modes and thus the Purcell enhancement factor is zero.
```

doc/docs/Python_Tutorials/Local_Density_of_States.md
```diff
@@ -2,21 +2,150 @@
...
+
+Two types of simulations are necessary for computing the Purcell enhancement factor: (1) bulk medium and (2) cavity. The `dft_ldos` featured is used to compute the LDOS in each case at a single wavelength. The Purcell enhancement factor is computed as the ratio of the LDOS measured in (2) to that from (1). Each simulation uses three mirror symmetries to reduce the size of the 3D computation by a factor of eight. The cavity is infinitely extended in the *xy* plane and thus the cell is terminated using PMLs in these two directions. Because Meep uses a default boundary condition of a perfect electric conductor, there is no need to explicitly define the boundaries in the *z* direction. The fields are timestepped until they have decayed away sufficiently due to absorption by the PMLS at the location of the pulsed source.
```
stevengj, line 21:
```suggestion
Two types of simulations are necessary for computing the Purcell enhancement factor: (1) bulk medium and (2) cavity. The `dft_ldos` featured is used to compute the LDOS in each case at a single wavelength. The Purcell enhancement factor is computed as the ratio of the LDOS measured in (2) to that from (1). Each simulation uses three mirror symmetries to reduce the size of the 3D computation by a factor of eight. The cavity is infinitely extended in the *xy* plane and thus the cell is terminated using PMLs in these two directions. Because Meep uses a default boundary condition of a perfect electric conductor, there is no need to explicitly define the boundaries in the *z* direction. The fields are timestepped until they have decayed away sufficiently due to absorption by the PMLs at the location of the pulsed source.
```

PR #2077, opened 2022-05-19
adjoint for ldos 

### LINKED ISSUE #1401: LDOS adjoint
As [previously discussed](#1254), the LDOS is an effective way to indirectly design a cavity.

Given meep's current LDOS functionality, what is the best way to implement its adjoint?

### LINKED ISSUE #2026: differentiable dft_ldos objects
It would be nice to make it easier to optimize the `dft_ldos` by making it a differentiable object for the adjoint solver.

cc @mochen4

mochen4:
Implements the adjoint for ldos. Closes #2026 and closes #1401.

### DIFF
--- a/src/meep.hpp
+++ b/src/meep.hpp
@@ -1416,6 +1416,7 @@ public:
   std::complex<double> *F() const; // returns Fdft
   std::complex<double> *J() const; // returns Jdft
   std::vector<double> freq;
+  mutable double overall_scale;
 
 private:
   std::complex<double> *Fdft;  // Nomega array of field * J*(x) DFT values
--- a/src/meep.hpp
+++ b/src/meep.hpp
@@ -1412,16 +1412,17 @@ public:
   }
 
   void update(fields &f);          // to be called after each timestep
-  double *ldos() const;            // returns array of Nomega values (after last timestep)
+  double *ldos();            // returns array of Nomega values (after last timestep)
   std::complex<double> *F() const; // returns Fdft
   std::complex<double> *J() const; // returns Jdft
   std::vector<double> freq;
-  mutable double overall_scale;
-
+  double overall_scale() const { return saved_overall_scale; }
+  
 private:
   std::complex<double> *Fdft;  // Nomega array of field * J*(x) DFT values
   std::complex<double> *Jdft;  // Nomega array of J(t) DFT values
   double Jsum;                 // sum of |J| over all points
+  double saved_overall_scale;  // saved overall scale for adjoint calculation
 };
 
 // dft.cpp (normally created with fields::add_dft_fields)
--- a/python/adjoint/objective.py
+++ b/python/adjoint/objective.py
@@ -367,6 +367,8 @@ class LDOS(ObjectiveQuantity):
 
     def place_adjoint_source(self, dJ):
         time_src = self._create_time_profile()
+        if dJ.ndim == 2:
+            dJ = np.sum(dJ, axis=1)
         dJ = dJ.flatten()
         sources = []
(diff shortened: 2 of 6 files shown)


stevengj:
I think the derivative of the work done by an eigenmode source (or any Equivalence Principle) source should be zero (with respect to material perturbations that don't intersect the source plane).  So maybe you are just looking at discretization errors (and it's not surprising that they don't agree)?

src/meep.hpp
```diff
@@ -1416,6 +1416,7 @@ class dft_ldos {
   std::complex<double> *F() const; // returns Fdft
   std::complex<double> *J() const; // returns Jdft
   std::vector<double> freq;
+  mutable double overall_scale;
```
stevengj, line 1419: 1. Make this a private member `saved_overall_scale` since it is read-only
2. Add an accessor function `double overall_scale() const { return saved_overall_scale; }`
3. Get rid of `mutable` and remove the `const` qualifier from `dft_ldos::ldos()`.

mochen4:
> 1. In 2D, if the polarization of the EigenModeSource is specified to be TM, everything is fine; however, when the polarization is TE, finite different gradients agree with adjoint gradients at some resolutions (e.g. 15 and 25 in my test case), but differ by a factor at other resolutions (factor of ~0.8 at resolution=20, factor of ~0.9 at resolution =30). It does look like increasing resolution will eventually lead to complete agreement.

Correction: actually, the gradients are consistent using electric components but not magnetic components, regardless of the polarization.

> I think the derivative of the work done by an eigenmode source (or any Equivalence Principle) source should be zero (with respect to material perturbations that don't intersect the source plane). So maybe you are just looking at discretization errors (and it's not surprising that they don't agree)?
> 
> Would be nice to check this, e.g. compute the power P for different perturbations of the material grid and verify that |δP/P| << 1

I found that the gradients with EigenModeSource is around the same magnitude as with GaussianSource, and in both cases, gradients get smaller with larger resolutions. Since the gradients with GaussianSource make sense, I think the gradients with EigenmodeSource also needs to make sense.

> 3. As the case for `FourierFields` and `Near2farFields`, the adjoint simulation requires one more sign flip for adjoint electric sources than adjoint magnetic sources. This also seems to be the case for LDOS with single component, but not for `mp.ALL_COMPONENT`.

I checked that using `mp.ALL_COMPONENTS` was indeed the same as using the corresponding individual components together; however, it was not consistent with the running simulations with individual components separately. Specifically, at ``eig_band = 2``, LDOS with `mp.Hz` alone is 1.705 and with `mp.Ey` alone is 0.2, but for `mp.ALL_COMPONENTS` is 0.236.

smartalecH:
> I found that the gradients with EigenModeSource is around the same magnitude as with GaussianSource, and in both cases, gradients get smaller with larger resolutions. Since the gradients with GaussianSource make sense, I think the gradients with EigenmodeSource also needs to make sense.

By `GaussianSource`, do you actually mean `GaussianBeamSource`? Note that the `EigenModeSource` is a *spatial* definition, while the `GaussianSource` is a *time* profile. As you know, an `EigenModeSource` can also be defined by a `GaussianSource` pulse...

mochen4:
> By `GaussianSource`, do you actually mean `GaussianBeamSource`? Note that the `EigenModeSource` is a _spatial_ definition, while the `GaussianSource` is a _time_ profile. As you know, an `EigenModeSource` can also be defined by a `GaussianSource` pulse...

Sorry for the confusion. By ``GaussianSource`` I mean a plane wave with Gaussian as the time profile. By ``EigenModeSource`` I mean an eigenmode as the space profile and Gaussian as the time profile.

stevengj:
Looks good!  Just needs some documentation?

PR #2079, opened 2022-05-21
fix bug in `plot_boundaries` of `plot2D`

### LINKED ISSUE #2022: 3D simulation 2D visualization PML bug
There appears to be in issue when running 3D simulations and calling plot2D to visualize. It shows PML everywhere, when there shouldn't be. 

```import meep as mp
import meep.adjoint as mpa
import numpy as np

diamond = mp.Medium(index=2.4)  # Lossless diamond
air = mp.Medium(index=1.0)  # Lossless air

# Geometry
radius = 0.88
geometry = [
    mp.Cylinder(radius=radius, height=0.22, material=diamond),  
]

# Source
fcen = 1.67969  # Center frequency
df = 0.05  # pulse frequency width
sources = [
    mp.Source(src=mp.GaussianSource(fcen, fwidth=df), component=mp.Ex, center=mp.Vector3(0.88 - 0.15, 0, 0.00))
]  # Dipole source

# Simulation
sim = mp.Simulation(
    cell_size=mp.Vector3(7*radius, 7*radius, 10),  # Size of the simulation cell
    geometry=geometry,  # Geometry
    boundary_layers=[mp.PML(1.0,direction=mp.X),mp.PML(1.0,direction=mp.Y),mp.PML(1.0,direction=mp.Z)],  # Boundary conditions
    resolution=10,  # Resolution
    sources=sources,  # Sources
)

# Plot the geometry
sim.init_sim()
if mp.am_really_master():
    plt.figure()
    sim.plot2D(output_plane=mp.Volume(center=mp.Vector3(0, 0, 0), size=mp.Vector3(7*radius, 7*radius, 0)))
    plt.savefig("geom")`
```

geom

cc @smartalecH

oskooi:
Fixes #2022.

Currently, the `plot_boundaries` subroutine of `plot2D` produces the incorrect output when the `output_plane` parameter passed to `plot2D` defines an area which does *not* include the PML boundaries. This can be demonstrated using the [`examples/oblique-planewave.py`](https://github.com/NanoComp/meep/blob/master/python/examples/oblique-planewave.py). The output region defined by `output_plane` does not include the PMLs but they incorrectly appear in the wrong location in the output.

Also includes some general formatting tweaks to `visualization.py` to improve readability.

```py
cell_size = mp.Vector3(14,10,0)

pml_layers = [mp.PML(thickness=2,direction=mp.X)]

nonpml_vol = mp.Volume(center=mp.Vector3(), size=mp.Vector3(10,10,0))

sim.plot2D(fields=mp.Ez,
           output_plane=nonpml_vol)
```
pw_bug

With this PR, the output does not include the PML boundaries in the *x* direction.
pw

cc @hammy4815

### DIFF
--- a/python/visualization.py
+++ b/python/visualization.py
@@ -99,7 +99,10 @@ def filter_dict(dict_to_filter, func_with_kwargs):
 
 def place_label(ax, label_text, x, y, centerx, centery, label_parameters=None):
 
-    label_parameters = default_label_parameters if label_parameters is None else dict(default_label_parameters, **label_parameters)
+    if label_parameters is None:
+        label_parameters = default_label_parameters
+    else:
+        label_parameters = dict(default_label_parameters, **label_parameters)
 
     offset = label_parameters['offset']
     alpha = label_parameters['label_alpha']
@@ -215,6 +218,17 @@ def get_2D_dimensions(sim, output_plane):
     sim_center, sim_size = (intersection_vol.center, intersection_vol.size)
     return sim_center, sim_size
 
+
+def box_vertices(box_center, box_size):
+    xmin = box_center.x - 0.5*box_size.x
+    xmax = box_center.x + 0.5*box_size.x
+    ymin = box_center.y - 0.5*box_size.y
+    ymax = box_center.y + 0.5*box_size.y
+    zmin = box_center.z - 0.5*box_size.z
+    zmax = box_center.z + 0.5*box_size.z
+
+    return xmin, xmax, ymin, ymax, zmin, zmax
+
 # ------------------------------------------------------- #
 # actual plotting routines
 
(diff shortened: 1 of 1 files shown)


python/visualization.py
```diff
@@ -336,11 +347,15 @@ def sort_points(xy):
...
+    if eps_parameters is None:
+        eps_parameters = default_eps_parameters
+    else:
+        eps_parameters = dict(default_eps_parameters, **eps_parameters)
 
     # Determine a frequency to plot all epsilon
     if frequency is not None:
-        warnings.warn('The frequency parameter of plot2D has been deprecated. Use the frequency key of the eps_parameters dictionary instead.')
+        warnings.warn("The frequency parameter of plot2D has been deprecated. "
+                      "Use the frequency key of the eps_parameters dictionary instead.")
```
stevengj, line 358: I really wish you wouldn't mix bugfix PRs with a bunch of unrelated changes.  It's a bad habit.

stevengj:
Does it work correctly if the `output_plane` partially intersects the PML?

oskooi:
> Does it work correctly if the output_plane partially intersects the PML?

Yes, it works correctly in this case as well. Here is a demonstration based on the example above with a PML thickness of 2.0. The `output_plane` only intersects the PML over a distance of 1.0 on each side in the *x* direction.
```py
nonpml_vol = mp.Volume(center=mp.Vector3(),
                       size=mp.Vector3(12,10,0))
sim.plot2D(fields=mp.Ez,
           output_plane=nonpml_vol)
```
pw_partial_overlap

For comparison, here is the output using the master branch (5275f4312). The PML thickness incorrectly appears as 2.0.

pw_bug_partial

PR #2095, opened 2022-06-07
Fix FourierFields adjoint for D and B fields in isotropic media

### LINKED ISSUE #2090: Incorrect FourierFields gradients for D components
The FourierFields gradients for D fields is incorrect by a factor of the permittivity ε (at the location of interest).

Here is an example
```
Si = mp.Medium(index=3.4)
resolution = 20
Sx = 5
Sy = 3
cell_size = mp.Vector3(Sx,Sy)
pml_layers = [mp.PML(1.0)]
fcen = 1/1.55
width = 0.3
fwidth = width * fcen
frequencies = [fcen]
source_center  = [-1,0,0]
source_size    = mp.Vector3(0,0.5,0)
source = [mp.Source(src, component=mp.Ez, size = source_size,center=source_center)]
design_region_resolution = 20
Nx = design_region_resolution//2
Ny = design_region_resolution//2

design_variables = mp.MaterialGrid(mp.Vector3(Nx,Ny),mp.air,Si)
design_region = mpa.DesignRegion(design_variables,volume=mp.Volume(center=mp.Vector3(), size=mp.Vector3(0.5, 0.5, 0)))
geometry = [
    mp.Block(center=mp.Vector3(1,0), size = mp.Vector3(0.5, 0.5), material=Si),                 
]
                    boundary_layers=pml_layers,
                    geometry=geometry,
                    sources=source,
                    resolution=resolution)

Dz = mpa.FourierFields(sim,mp.Volume(center=mp.Vector3(1,0),size=mp.Vector3(0, 0.1)),mp.Dz,yee_grid=True)

ob_list = [Dz]                                                                                          
def J(dz):
    return npa.abs(dz[0,0])**2
    simulation=sim,
    objective_functions=J,
    objective_arguments=ob_list,
    design_regions=[design_region],
    frequencies=frequencies)
```
which has the layout:
image

I found the fd_gradients was approximately ε(=12 in the code above) * adjoint_gradients. If the monitor is placed inside the design region, then the factor is the permittivity corresponding to the design parameter weight at that point.

mawc2019:
This PR fixes #2090. `FourierFields` adjoint gradients for D and B fields in isotropic media now become consistent with finite-difference gradients.

### DIFF
--- a/python/tests/test_adjoint_solver.py
+++ b/python/tests/test_adjoint_solver.py
@@ -510,7 +510,8 @@ class TestAdjointSolver(ApproxComparisonTestCase):
 
             ## compare objective results
             print("Dz2 -- adjoint solver: {}, traditional simulation: {}".format(adjsol_obj,Dz2_unperturbed))
-            self.assertClose(adjsol_obj,Dz2_unperturbed,epsilon=1e-4 if mp.is_single_precision() else 1e-6)
+            tol = 1e-5 if mp.is_single_precision() else 1e-6
+            self.assertClose(adjsol_obj,Dz2_unperturbed,epsilon=tol)
 
             ## compute perturbed |Dz|^2
             Dz2_perturbed = forward_simulation_complex_fields(p+dp, frequencies)
@@ -521,7 +522,7 @@ class TestAdjointSolver(ApproxComparisonTestCase):
             adj_scale = (dp[None,:]@adjsol_grad).flatten()
             fd_grad = Dz2_perturbed-Dz2_unperturbed
             print("Directional derivative -- adjoint solver: {}, FD: {}".format(adj_scale,fd_grad))
-            tol = 0.018 if mp.is_single_precision() else 0.002
+            tol = 0.055 if mp.is_single_precision() else 0.002
             self.assertClose(adj_scale,fd_grad,epsilon=tol)
 
     def test_damping(self):
--- a/python/tests/test_adjoint_solver.py
+++ b/python/tests/test_adjoint_solver.py
@@ -510,7 +510,7 @@ class TestAdjointSolver(ApproxComparisonTestCase):
 
             ## compare objective results
             print("Dz2 -- adjoint solver: {}, traditional simulation: {}".format(adjsol_obj,Dz2_unperturbed))
-            self.assertClose(adjsol_obj,Dz2_unperturbed,epsilon=1e-6)
+            self.assertClose(adjsol_obj,Dz2_unperturbed,epsilon=1e-4 if mp.is_single_precision() else 1e-6)
 
             ## compute perturbed |Dz|^2
             Dz2_perturbed = forward_simulation_complex_fields(p+dp, frequencies)
--- a/python/tests/test_adjoint_solver.py
+++ b/python/tests/test_adjoint_solver.py
@@ -522,7 +522,7 @@ class TestAdjointSolver(ApproxComparisonTestCase):
             adj_scale = (dp[None,:]@adjsol_grad).flatten()
             fd_grad = Dz2_perturbed-Dz2_unperturbed
             print("Directional derivative -- adjoint solver: {}, FD: {}".format(adj_scale,fd_grad))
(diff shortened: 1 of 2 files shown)


oskooi:
It would be useful to add a unit test in this PR which should fail using the current master branch since this is a bug fix.

mawc2019:
However, the current version of this PR does not yield correct adjoint gradients if the monitor is placed inside the design region. Should we deal with this case?

smartalecH:
> However, the current version of this PR does not yield correct adjoint gradients if the monitor is placed inside the design region. Should we deal with this case?

That piece of code would need to go here:

https://github.com/NanoComp/meep/blob/91d36ba73d1ae0eb135d18c783840cff9fcd71ef/src/meepgeom.cpp#L3003-L3006

stevengj:
Looks fine, albeit missing a test.

I agree that it would be good to add support for D fields in the design region, but this is nontrivial, and meanwhile this PR is a strict improvement over what we have now.

mawc2019:
> It would be useful to add a unit test in this PR which should fail using the current master branch since this is a bug fix.

I changed an existing test in `test_adjoint_solver.py` to the new test. To be specific, in the functions `forward_simulation_complex_fields`, `adjoint_solver_complex_fields`, and `test_complex_fields`, `Ez` has been changed to `Dz`. I think the new test (for `Dz`) can substitute for the old test (for `Ez`). Besides, in `test_adjoint_solver.py`, another test for `Ez` has already existed.

python/tests/test_adjoint_solver.py
```diff
@@ -505,21 +505,21 @@ def test_complex_fields(self):
...
-            ## compute unperturbed |Ez|^2
-            Ez2_unperturbed = forward_simulation_complex_fields(p, frequencies)
+            ## compute unperturbed |Dz|^2
+            Dz2_unperturbed = forward_simulation_complex_fields(p, frequencies)
 
             ## compare objective results
-            print("Ez2 -- adjoint solver: {}, traditional simulation: {}".format(adjsol_obj,Ez2_unperturbed))
-            self.assertClose(adjsol_obj,Ez2_unperturbed,epsilon=1e-6)
+            print("Dz2 -- adjoint solver: {}, traditional simulation: {}".format(adjsol_obj,Dz2_unperturbed))
+            self.assertClose(adjsol_obj,Dz2_unperturbed,epsilon=1e-6)
```
oskooi, line 513: Looks like this assert statement is failing for the single-precision build. You can try increasing the tolerance for single precision using e.g.:
```
self.assertClose(adjsol_obj,Dz2_unperturbed,epsilon=1e-4 if mp.is_single_precision() else 1e-6)
```

mawc2019: In my place, this line does not cause error. I guess the error message in my place is different from that given by the online check. But I cannot see the detailed error message in the online-check logs. Can I ask where I should find the detailed error message?

oskooi: To find the log of the failing CI run, click on the "Details" button below next to "CI / Test Python 3.10 with MPI (true)" or the "Actions" tab above (next to "Discussions" and "Projects"). From there, for the failing run locate the "Artifacts" section and then download the log file. In this case, this would be [py3.10-tests-mpi-true.log](https://github.com/NanoComp/meep/actions/runs/2513272770) which shows:
```
Ran 7 tests in 307.004s

FAILED (failures=1)
FAIL: test_complex_fields (__main__.TestAdjointSolver)
  File "/home/runner/work/meep/meep/build/meep-1.24.0-beta/_build/sub/python/../../../python/tests/test_adjoint_solver.py", line 525, in test_complex_fields
  File "/home/runner/work/meep/meep/build/meep-1.24.0-beta/python/tests/utils.py", line 28, in assertClose
AssertionError: 0.001776412917294544 not less than or equal to 0.0005871758943843297 : 
```

mawc2019: Thanks! The tolerance for single precision is adjusted further.

PR #2107, opened 2022-06-16

oskooi:
Closes #2072.

For some reason, I was not able to resolve the merge conflicts in `python/tests/test_diffracted_planewave.py` in #2072 during a rebase. This is why I just created this new PR.


oskooi:
There is a test failure due to a segmentation fault related to garbage collection in `python/tests/test_diffracted_planewave.py` which I am not able to reproduce locally.

```
*** Process received signal ***
Signal: Segmentation fault (11)
Signal code:  (128)
Failing at address: (nil)
[ 0] /lib/x86_64-linux-gnu/libc.so.6(+0x43090)[0x7fc847044090]
[ 1] /opt/hostedtoolcache/Python/3.7.13/x64/lib/libpython3.7m.so.1.0(+0x199d88)[0x7fc84739dd88]
[ 2] /opt/hostedtoolcache/Python/3.7.13/x64/lib/libpython3.7m.so.1.0(+0x119982)[0x7fc84731d982]
[ 3] /opt/hostedtoolcache/Python/3.7.13/x64/lib/libpython3.7m.so.1.0(+0x1993e4)[0x7fc84739d3e4]
[ 4] /opt/hostedtoolcache/Python/3.7.13/x64/lib/libpython3.7m.so.1.0(+0x203f0e)[0x7fc847407f0e]
[ 5] /opt/hostedtoolcache/Python/3.7.13/x64/lib/libpython3.7m.so.1.0(PyGC_Collect+0x55)[0x7fc847408005]
[ 6] /opt/hostedtoolcache/Python/3.7.13/x64/lib/libpython3.7m.so.1.0(Py_FinalizeEx+0x67)[0x7fc8473f99e7]
[ 7] /opt/hostedtoolcache/Python/3.7.13/x64/lib/libpython3.7m.so.1.0(Py_Exit+0xc)[0x7fc8473f9cec]
[ 8] /opt/hostedtoolcache/Python/3.7.13/x64/lib/libpython3.7m.so.1.0(+0x1fa543)[0x7fc8473fe543]
[ 9] /opt/hostedtoolcache/Python/3.7.13/x64/lib/libpython3.7m.so.1.0(PyErr_PrintEx+0x3a)[0x7fc8473fd92a]
[10] /opt/hostedtoolcache/Python/3.7.13/x64/lib/libpython3.7m.so.1.0(PyRun_SimpleFileExFlags+0x21d)[0x7fc8473fd4cd]
[11] /opt/hostedtoolcache/Python/3.7.13/x64/lib/libpython3.7m.so.1.0(+0x201811)[0x7fc847405811]
[12] /opt/hostedtoolcache/Python/3.7.13/x64/lib/libpython3.7m.so.1.0(_Py_UnixMain+0x40)[0x7fc8474054a0]
[13] /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf3)[0x7fc847025083]
[14] /opt/hostedtoolcache/Python/3.7.13/x64/bin/python(_start+0x2e)[0x5635a7d9209e]
*** End of error message ***
```

python/tests/test_diffracted_planewave.py
```diff
@@ -137,12 +139,102 @@ def _pw_amp(x):
...
+                                    eig_band=mp.DiffractedPlanewave((0,0,0),
+                                                                    mp.Vector3(0,1,0),
+                                                                    1,
+                                                                    0))]
+    else:
+      sources = [mp.EigenModeSource(mp.GaussianSource(self.fcen,fwidth=0.1*self.fcen),
+                                    center=src_pt,
+                                    size=mp.Vector3(0,sy,0),
+                                    direction=mp.NO_DIRECTION,
+                                    eig_kpoint=k,
```
stevengj, line 194:
```suggestion
```
Should be able to delete these lines as I commented in #2072.  (It's possible you were having problems before that are now fixed by #2114.)

stevengj: If it's not working, make sure the MPB/Meep  verbosity is > 1 and look at the [`NEW KPOINT`](https://github.com/NanoComp/meep/blob/5aac79a41c84cb14a64a844379036a12c14fc51d/src/mpb.cpp#L516) and [`MPB solved for frequency`](https://github.com/NanoComp/meep/blob/5aac79a41c84cb14a64a844379036a12c14fc51d/src/mpb.cpp#L571-L573) verbose outputs — you should see that it is setting the parallel component of `k` by the boundary conditions, and varying the perpendicular component of `k`.

oskooi: Looks like there is a potential bug because printing out `NEW KPOINT` shows that its value is clearly incorrect:
```
NEW KPOINT: 3, 2, 0
```
The *y* component (i.e., the periodic direction of the monitor monitor) should be 1.333... (i.e., 4/3) to match the `k_point` of the `Simulation` object. The reason it is not this value is due to these lines:

https://github.com/NanoComp/meep/blob/405b7e607f87d7dd4a81baf3d5473aa4c28e7a4a/src/mpb.cpp#L426-L433

In this example, `R = (1,1.5,1)` and `kcart = (0,1.33333,0)` which results in `k = (0, 2.0, 0)`.

The fix seems to be to just replace `k[i] = dot_product(R[i], kcart)` with `k[i] = kcart[i]`?

stevengj: No, that code is correct — MPB expects the k point in the reciprocal-lattice basis, so it has to convert from the Cartesian basis of Meep.  (It does the same thing with the `DiffractedPlanewave` object.)

Instead, something weird is going on.  You said the output is:
```
MPB solved for frequency_1(3,1.33333,0) = 2 after 29 iters
```
which means that it *is* taking the nonzero ky into account, but somehow it is getting the wrong kx — somehow, it is acting as though it is passing `ky==0` to MPB, even though we are clearly printing that it is passing a nonzero `k[1]`.

oskooi: I think I have finally gotten to the bottom of this issue. TLDR: the current behavior is a *feature* and not a bug.

Currently, to launch oblique planewaves using an `EigenModeSource` with `eig_band` set to an integer (rather than an `DiffractedPlanewave` object), the user must provide its wavevector `k` (a `meep.Vector3` object) via `direction=mp.NO_DIRECTION` *and* `eig_kpoint=k`. This is because even though internally Meep sets the component of `k` parallel to the line source (in the directions with periodic boundaries) to the components of the `k_point` of the `Simulation` object, the correct perpendicular component $\left|k_\perp\right| = (n\omega)^2 - \left|k_\parallel\right|$ is [*only computed automatically for the case when `eig_band` is a `DiffractedPlanewave` object*](https://github.com/NanoComp/meep/blob/a93a1585b0bd4b3525423c91ed88064d4d54c2da/src/mpb.cpp#L612-L659). 

It would be nice to extend this feature to when `eig_band` is an integer but that can be a separate PR since it is not directly related to this PR.

oskooi:
Replaced by #2226.

python/tests/test_diffracted_planewave.py
stevengj: No.  A `DiffractedPlanewave` object only enables the *analytical* calculation of k.  Otherwise, it uses a Newton solver to find the k where ω(k) = specified frequency.  This of course should be the same thing (but slower, and with more ambiguity about polarization and more difficulty in choosing a specific diffraction order).

oskooi: The wavevector of the oblique planewave in Cartesian coordinates should be: `k = (2.68742, 1.33333, 0.00000)`.  However, the Newton solver shows:
```
MPB solved for frequency_1(3,1.33333,0) = 2 after 29 iters
update_maxwell_data_k:, k=(3.00000, 2.00000, 0.00000)
update_maxwell_data_k:, G[0]=(1.00000, 0.00000, 0.00000)
update_maxwell_data_k:, G[1]=(0.00000, 0.66667, 0.00000)
update_maxwell_data_k:, G[2]=(0.00000, 0.00000, 1.00000)
```

Note that the $y$ component (parallel direction) of the MPB mode (`1.33333`) is correct but the $x$ component (perpendicular direction) (`3`) is not `2.68742`.

oskooi: $|k|=n\omega=(1.5)(2)=3$ but clearly the wavevector from the  Newton solver is wrong since $|3,1.3333,0)| > 3$.

oskooi: Turns out that there was a bug in the *output* of MPB's mode wavevector which is fixed by #2253. (The mode itself was correct.)

With the bug fix, the output from `run_mode_source` of `test_diffracted_planewave.py` in this PR is:
```
MPB solved for frequency_1(2.68742,1.33333,0) = 2 after 23 iters
```
This means `k = (2.68742,1.33333,0)` which is equivalent to the `k_point` of the `Simulation` class object as expected.

PR #2162, opened 2022-07-29
Fix images and equations display issues.

jamesetouma:
- Remove the <center></center> tags
- Remove <p> </p> tags

@oskooi 
I reverted back to markdown images in all the markdown files and removed \<center>\</cener>  for a few equations. I tested using mkdocs. The images render in both md and html files, however, they are not centered. You can't tell that that is the case with large images but with small ones it's pretty obvious.

### DIFF
--- a/doc/docs/Python_Tutorials/Basics.md
+++ b/doc/docs/Python_Tutorials/Basics.md
@@ -51,9 +51,9 @@ geometry = [mp.Block(mp.Vector3(mp.inf,1,mp.inf),
 
 The waveguide is specified by a `Block` (parallelepiped) of size $\infty \times 1 \times \infty$, with $ε=12$, centered at (0,0) which is the center of the cell. By default, any place where there are no objects there is air ($ε=1$), although this can be changed by setting the `default_material` variable. The resulting structure is shown below.
 
-<p align="center">
-  <img src="../images/Python-Tutorial-wvg-straight-eps-000000.00.png">
-</p>
+
+![](../images/Python-Tutorial-wvg-straight-eps-000000.00.png)
+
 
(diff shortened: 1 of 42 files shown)


doc/docs/css/extra.css
```diff
@@ -55,3 +55,36 @@ pre code.python {
...
+img[src*='#left'] {
+    float: left;
+}
+img[src*='#right'] {
+    float: right;
+}
+img[src*='#center'] {
+    display: block;
+    margin: auto;
+}
```
stevengj, line 90: Why do you need both Method 1 and Method 2?

jamesetouma:
I just wanted to see which one you prefer. I like method 2 which is evident
to new editors. I'm fine with either; just let me know which method you'd
like to have in there.

stevengj:
Rebased to fix merge conflicts.

stevengj:
Let's use Method 2.

doc/docs/Python_Tutorials/Basics.md
```diff
@@ -1201,9 +1200,9 @@ from mayavi import mlab
 s = mlab.contour3d(eps_data, colormap="YlGnBu")
 mlab.show()
 ```
-<p align="center">
-  <img src="../images/prism_epsilon.png">
-</p>
+
+![](../images/prism_epsilon.png#center)
```
oskooi, line 1204: With this change, the image reappears in https://meep.readthedocs.io/en/latest/Python_Tutorials/Basics/#visualizing-3d-structures but is *not* centered.

jamesetouma:
yes, I commented on that in a different thread. I'm going to submit another
PR to fix the centering in all the HTML files (readthedocs) but we'll still
have the issue of things not being centered in the MD files. I'll be
looking into this afterwards.

PR #2181, opened 2022-07-30
Fix html md issues

jamesetouma:
Added CSS centering to all the images. Now they are centered in readthedocs . They still are not centered in the MD files. Also, I centered a few equations that were not centered.


oskooi (approved):
LGTM.

Issue #2144, opened 2022-07-21
Prism does weird things when placed with "center=" kwarg


sequoiap:
When defining geometries, Prisms don't seem to build properly when offset with the "center" argument. The z-value changes to some weird and whacky value, even when the z-offset is zero. This has the effect of rendering weird shapes for the dielectric that don't match the defined Prism when using plot2D.

Script to reproduce the problem:

```
import meep as mp
import numpy as np
import matplotlib.pyplot as plt


mp.quiet(quietval=True)
RESOLUTION = 50
OFFSET_Y = 0.5


Si = mp.Medium(index=3.4)
SiO2 = mp.Medium(index=1.44)

WG_W0 = 0.5
WG_W1 = 1.5
WG_RUN_L = 2
WG_TAPER_L = 10

Sx = 2*WG_RUN_L + WG_TAPER_L
Sy = 5
cell_size = (Sx, Sy)

PML_THICKNESS = 1.0
pml_layers = [mp.PML(PML_THICKNESS)]

FCEN = 1/1.55
WIDTH = 0.2
FWIDTH = WIDTH * FCEN
source_center  = mp.Vector3(-Sx / 2 + PML_THICKNESS + 0.5, OFFSET_Y, 0)
source_size    = mp.Vector3(0,1,0)
kpoint = mp.Vector3(1,0,0)
src = mp.GaussianSource(frequency=FCEN, fwidth=FWIDTH)
source = [mp.EigenModeSource(src,
                    eig_band=1,
                    direction=mp.NO_DIRECTION,
                    eig_kpoint=kpoint,
                    size=source_size,
                    center=source_center)]

taper_vtx = [
    mp.Vector3(-WG_TAPER_L/2, WG_W0/2),
    mp.Vector3(WG_TAPER_L/2, WG_W1/2),
    mp.Vector3(WG_TAPER_L/2, -WG_W1/2),
    mp.Vector3(-WG_TAPER_L/2, -WG_W0/2),
]

# Taper (prism) with center specified
geometry_bug = [
    mp.Block(center=mp.Vector3(-(WG_RUN_L/2 + WG_TAPER_L/2), OFFSET_Y), material=Si, size=mp.Vector3(WG_RUN_L, WG_W0)),
    mp.Prism(taper_vtx, mp.inf, center=mp.Vector3(0, OFFSET_Y), material=Si),
[...]

PR #2188, opened 2022-08-05
Prism 2d  correct center calculation

Yaraslaut:
Fixes https://github.com/NanoComp/meep/issues/2144


smartalecH:
I retriggered the tests now that the MPB bug was fixed, but it's still failing (ignore the fact that the first batch of tests is passing... that always passes so long as it builds correctly and we need to fix the test matrix...)

The output of `test_prism.py` might help you debug:

```
Testing Non-Convex Prism #3 using marching squares algorithm...
-----------
Initializing structure...
non-coplanar vertices in init_prism
non-coplanar vertices in init_prism
```

(here's a [related issue](https://github.com/NanoComp/libctl/issues/61) from libctl)

PR #2190, opened 2022-08-08
add/update type hints for JAX wrapper of adjoint solver

oskooi:
This PR adds new and also updates existing type hints for *all* functions which make up the JAX wrapper of the adjoint solver. Several of these changes are necessary because #1855 modified the API for `meep._get_gradients` replacing the forward and adjoint field arguments of type `numpy.ndarray` with `meep.DftFields`.  

The function `gather_design_region_fields` is removed because it is no longer used.

cc @ianwilliamson

### DIFF
--- a/python/adjoint/utils.py
+++ b/python/adjoint/utils.py
@@ -24,22 +24,33 @@ FD_DEFAULT = 1e-3
 
 
 class DesignRegion:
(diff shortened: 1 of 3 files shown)


stevengj:
Since @ianwilliamson is okay with this PR, LGTM.

PR #2191, opened 2022-08-09
Poynting flux Adjoint

### LINKED ISSUE #1307: Adjoint optimization through Poynting Flux

Alex-Kaylor:
Closes #1307,  #1299, #1398 

Initial Implementation of Poynting Flux Adjoint. Uses `add_DFT_fields` objects. Forward run is accurate. Throws an error in `step_db.cpp` on adjoint run: `"Assertion 'changed_materials' failed."` Using this pull request to help debug code changes.


python/adjoint/objective.py
```diff
@@ -483,3 +495,127 @@ def __call__(self):
...
+        works with monitors aligned to a coordinate direction.
+        decimation_factor: Whether to skip points in the time series every 
+        decimation_factor timesteps. See "add_dft_fields" documentation.
+        The biggest warning there is to be careful to avoid aliasing if
+        the fields vary quickly in time.
+        Note on yee_grid: For the Poynting Flux to work, H and E components
+        must lie at the same points. Therefore, the Yee grid will always be false.
+    """
+
+    def __init__(self, sim,volume, decimation_factor=0):
```
smartalecH, line 515: You should have a parameter `scale`, just like we do with the `PoyntingFlux` class.

python/adjoint/objective.py
```diff
@@ -11,6 +11,18 @@
...
+#3 possible components for E x n and H x n 
+#signs are handled in code
+EH_TRANSVERSE    = [ [mp.Hy, mp.Hz, mp.Ey, mp.Ez],
+                     [mp.Hz, mp.Hx, mp.Ez, mp.Ex],
+                     [mp.Hx, mp.Hy, mp.Ex, mp.Ey] ]
+
+#Holds the components for each current source
+#for the cases of x,y, and z normal vectors.
+#This is the same as swapping H and E in the above list
+JK_TRANSVERSE    = [ [mp.Ey, mp.Ez, mp.Hy, mp.Hz],
```
smartalecH, line 23: I think the component directions need to be switched too, right? e.g.

```
[mp.Ey, mp.Ez, mp.Hy, mp.Hz]
```

should be 

```
[mp.Ez, mp.Ey, mp.Hz, mp.Hy]
```

Alex-Kaylor: Great point! I believe this was causing the "Assertion 'changed_materials' failed." bug. New push coming shortly.

python/adjoint/objective.py
```diff
@@ -483,3 +495,123 @@ def __call__(self):
...
+                #TODO: Remove this unnecessary for loop
+                for index,element in enumerate(self.field_component_evaluations[field_t]):
+                    tuple_elements[0,index,0] = element
+                source.append( mp.Source(
+                    src,
+                    component = field,
+                    amplitude=amp,
+                    size=self.volume.size,
+                    center=self.volume.center,
+                    amp_data = tuple_elements
```
smartalecH, line 578: You need to conjugate some of the fields, right?

smartalecH: Some of the current sources have a sign change too. Is that implemented elsewhere?

Alex-Kaylor: These are resolved as of the most recent update

stevengj:
Ideally this would be done in C++ directly from the `dft_flux` object — the problem with forming the Poynting flux from DFT fields in Python is that it requires all the DFT fields to be loaded into a single process, which may be prohibitive for computing the flux through large areas.

In particular, it should be very similar to `dft_fields::fourier_sourcedata` (https://github.com/NanoComp/meep/blob/ea43f83c2f8c3e74aab2f8dae3e2ed96b7347777/src/dft.cpp#L1462), except that the amplitude `dJ` comes not from the user but rather from the Fourier field you are multiplying with (e.g. the Ex adjoint source's "dJ` comes from Hy etcetera).

Note that the `dft_flux` object already has lists of field pairs that are multiplied together (https://github.com/NanoComp/meep/blob/ea43f83c2f8c3e74aab2f8dae3e2ed96b7347777/src/dft.cpp#L537), so for each pair you will add two adjoint sources.

python/adjoint/objective.py
Alex-Kaylor: Added a few commits ago! Thank you for the feedback.

Alex-Kaylor:
Works ok in 2D (see below), fails in 3D (still sticking with a purely Python implementation using FourierFields). I can resolve 3D on my own - I expect the derivative's format isn't what FourierFields expects when we place the adjoint source.


The accuracy doesn't improve monotonically in 2D. The gradient's inaccurate in a few cases. Note sure why, I'll explore it some more once it works in 3D. The same resolutions cause similar errors with different sources. 


Plots from 2D:

Eigenmode source resolution is 10

point_source_res_is_10

Eigenmode source resolution is 22

point_source_res_is_22


Eigenmode source resolution is 26
point_source_res_is_26


Eigenmode source resolution is 30

point_source_res_is_30

Eigenmode source resolution is 42

point_source_res_is_42

point source resolution is 22
point_source_res_is_22

point source resolution is 26
point_source_res_is_26

PR #2203, opened 2022-08-18
add more type hints to simulation.py

oskooi:
Adds more type hints to `simulation.py`. 

Eventually once type hints have been added to the entire Python API, we can enable mypy as a static type checker as part of the GitHub Actions setup for CI.

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -2311,7 +2311,7 @@ class Simulation:
         return not (self.force_complex_fields or cond1 or cond2 or cond5)
 
     def initialize_field(
-        self, cmpnt: int = None, amp_func: Callable[[Vector3Type], complex] = None
+        self, cmpnt: int = None, amp_func: Callable[[Vector3Type], Union[float,complex]] = None
     ):
         """
         Initialize the component `c` fields using the function `func` which has a single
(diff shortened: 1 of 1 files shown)


python/simulation.py
```diff
@@ -2276,7 +2310,9 @@ def using_real_fields(self):
         cond5 = not (cond3 or cond4 or self.k_point == Vector3())
         return not (self.force_complex_fields or cond1 or cond2 or cond5)
 
-    def initialize_field(self, cmpnt, amp_func):
+    def initialize_field(
+        self, cmpnt: int = None, amp_func: Callable[[Vector3Type], complex] = None
```
stevengj, line 2314: Does the `complex` here allow the function to have real outputs?

stevengj:
```suggestion
        self, cmpnt: int = None, amp_func: Callable[[Vector3Type], Union[float,complex]] = None
```

oskooi: Unfortunately, this change exceeds the 80-character line limit which is causing the pre-commit check involving black to fail. In the future, we need to be more careful about committing changes directly from GitHub.

PR #2206, opened 2022-08-25
Add more type hints and regenerate Markdown pages from docstrings

oskooi:
Also fixes the precommit error described in [#2203 (comment)](https://github.com/NanoComp/meep/pull/2203#discussion_r949668835).

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -11,9 +11,9 @@ from collections import OrderedDict, namedtuple
 from typing import Callable, List, Optional, Tuple, Union
 
 try:
-    from collections.abc import Sequence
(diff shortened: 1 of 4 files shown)


python/simulation.py
```diff
@@ -3048,7 +3059,7 @@ def _display_energy(self, name, func, energys):
                     zip(freqs, *[func(f) for f in energys]),
                 )
 
-    def display_electric_energy(self, *energys):
+    def display_electric_energy(self, *energys: List[DftEnergy]):
```
stevengj, line 3062: Why is `energies` misspelled here?

PR #2207, opened 2022-08-25
New Connectivity Constraint

mochen4:
The previous connectivity constraint turned out to be rather stiff in practice: structures may fail to evolve too much if the tolerance is low, or may become disconnected if the tolerance is high. This PR adopts an alternative approach for the constraint and should have better performance. The interface is also streamlined.

The previous constraint solves the heat equation where solid pixels are conductive, and impose $T=T_0$ on the top boundary. The idea was that for connected structure, solid pixels should be heated up, and thus we use the density-weighted p-norm $(\frac{\sum \rho (T-T_0)^p}{\sum \rho})^\frac1{p}$ as the measure of connectivity.

In the new approach, in addition to being conductive, solid pixels are also sources; but $T=0$ on the top boundary. The idea was that for connected structure, heat from solid pixels should be able to diffuse away from the top boundary, and thus we use $\frac{({\sum T^p})^\frac1{p}}{\alpha}-1$  as the measure, where $\alpha$ is some normalization factor.

Here is the comparison of performance:
For the disconnected structure 
image
The previous solver returned 0.6369810999723368 as forward value, and the gradients looked like
image

In comparison, the new solver returns 3.2973988093027025 as forward value, and the gradients look like (notice the change in scale from the colorbar)

For a connected structure
image
The previous forward value was 0.06833918047665405, and the gradients looked like 
image

The new solver returns -0.9379063204901158, and the gradients look like (again, notice the change in scale from the colorbar)

In summary, the forward measure should now be positive for disconnected structure and negative for connected structure (with appropriately chosen hyper-parameters, especially a reasonably large ``cond_s``). The gradients are also much larger for pixels that can "bridge" disconnected components; and smaller overall for connected structures


stevengj:
Great!  Would be good to have a tutorial & documentation; tutorial to start with.

PR #2208, opened 2022-08-25
fix _fit_volume_to_simulation_for_cylindrical for cylindrical coordinates

### LINKED ISSUE #2195: FourierFields in cylindrical coordinates error
Trying to use the FourierFields adjoint objective when the simulation is in 2D cylindrical coordinates returns an error with respect to the volume argument passed to FourierFields. This was found by modifying the example Adjoint FarField code to use cylindrical coordinates (which runs fine), and then try to change the objective to use FourierFields (which then fails).

Trying to call FourierFields with cylindrical coordinates returns the following error for the  Volume (abbreviated):
```
ValueError                                Traceback (most recent call last)
File ~/miniconda3/envs/mp/lib/python3.10/site-packages/meep/adjoint/objective.py:237, in FourierFields.__init__(self, sim, volume, component, yee_grid, decimation_factor)
    235 def __init__(self, sim, volume, component, yee_grid=False, decimation_factor=0):
    236     super().__init__(sim)
--> 237     self.volume = sim._fit_volume_to_simulation(volume)
    238     self.component = component
    239     self.yee_grid = yee_grid

File ~/miniconda3/envs/mp/lib/python3.10/site-packages/meep/simulation.py:1275, in Simulation._fit_volume_to_simulation(self, vol)
   1274 def _fit_volume_to_simulation(self, vol):
-> 1275     return Volume(vol.center, vol.size, dims=self.dimensions, is_cylindrical=self.is_cylindrical)

File ~/miniconda3/envs/mp/lib/python3.10/site-packages/meep/simulation.py:393, in Volume.__init__(self, center, size, dims, is_cylindrical, vertices)
...
    103     return mp.vec(v3.x, v3.y, v3.z)
    104 else:
--> 105     raise ValueError("Invalid dimensions in Volume: {}".format(dims))

ValueError: Invalid dimensions in Volume: -2
```

I could not find the reason for the volume dimension to be set to -2.

A MWE for reproducing:
```
import meep as mp
import meep.adjoint as mpa
pml_layers = [mp.PML(1.0)]
sim = mp.Simulation(
    cell_size=mp.Vector3(3, 0, 3),
    boundary_layers=pml_layers,
    dimensions=mp.CYLINDRICAL,
    m=0,
    default_material=mp.Medium(index=1.0),
    resolution=10
)
Ezvol = mp.Volume(center=mp.Vector3(0, 0, 2.5),
                  size=mp.Vector3(), dims=2, is_cylindrical=True)
Ez = mpa.FourierFields(sim, Ezvol, mp.Ez)
```
The error is the same regardless of if `dims` or `is_cylindrical` are passed to mp.Volume 

Interestingly this error occurs in the current *conda version of meep (1.24.0), but the older Ubuntu package manager version (1.17.1) does not produce any error.

mawc2019:
As discussed in #2195, the function `_fit_volume_to_simulation_for_cylindrical` does not apply to cylindrical coordinates. This small PR fixes the issue.

Fixes #2195.

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -1378,6 +1378,9 @@ class Simulation:
     # on the settings in the Simulation instance. This method must be called on
     # any user-defined Volume before passing it to meep via its `swigobj`.
     def _fit_volume_to_simulation(self, vol: Volume) -> Volume:
+        if self.dimensions == mp.CYLINDRICAL:
+            self.dimensions = 2
+            self.is_cylindrical = True
         return Volume(
             vol.center,
             vol.size,


mawc2019:
The error message is as follows. Shall we simply loosen the tolerance for single precision?
```
  File "/home/runner/work/meep/meep/build/meep-1.25.0-beta/_build/sub/python/../../../python/tests/test_mpb.py", line 570, in test_output_efield_z
    self.compare_h5_files(ref_path, res_path)
  File "/home/runner/work/meep/meep/build/meep-1.25.0-beta/_build/sub/python/../../../python/tests/test_mpb.py", line 236, in compare_h5_files
    compare_arrays(self, ref[k][()], res[k][()], tol=tol)
  File "/home/runner/work/meep/meep/build/meep-1.25.0-beta/python/tests/utils.py", line 17, in compare_arrays
AssertionError: 0.0011886021463879074 not less than 0.001
```

smartalecH:
> Shall we simply loosen the tolerance for single precision?

No probably not... at least not yet. 

Why is the output different than `master`? It should be the same, right?

mawc2019:
> Why is the output different than master? It should be the same, right?

I think the output of `test_mpb.py` in this PR should be the same as that in the master branch, because it seems that `_fit_volume_to_simulation` is not invoked directly or indirectly in `test_mpb.py`.
I do not know why the difference appears, but in my experience, sometimes Meep does not generate the same results even if the same code is run multiple times.

stevengj:
Can we rebase and re-run the tests?

stevengj:
Passing now!

PR #2221, opened 2022-09-01
Allocate ` default_material ` in the MPI shared_memory when copy Numpy arrays to meep

estebvac:
Related to discussion [2158](https://github.com/NanoComp/meep/discussions/2158) This is a possible way in which the memory is not copied multiple times per process, and leverages the use of the MPI shared memmory capabilities to reduce the memory consumption. In the example code showed in the discussion 2158, it was possible to increase the number of process running the simulation from 8 to 24 in a single node, since the memory consumption is reduced.


smartalecH:
Do you want to try fixing the pre-commit errors?

Issue #2130, opened 2022-07-14
Automated script


Yaraslaut:
I've noticed few things in build-meep.sh script that can be updated and enchanced
 * When building without dependencies (-n flag for the script) script is checking for distribution version and throwing error if it is not matches Ubuntu 16.04/18.04 or CentOS 7.x . There is no reason to check distribution if no dependencies will be install through package manager.
 * New version of ubuntu as well as other distribution can be added to install dependencies 
 * It would be great to be able to install meep without guile by simple adding flag
   * One odd thing that I have incountered is that libctl can not be build without guile from source files downloaded directly from github repository and it takes some effort to figured our that.

smartalecH:
From the docs:

> Guile is required in order to use the Scheme interface. If you don't install it, you can only use the C++ and/or Python interfaces.

You can use the flag `--without-scheme` when installing meep.

Yaraslaut:
> From the docs:
> 
> > Guile is required in order to use the Scheme interface. If you don't install it, you can only use the C++ and/or Python interfaces.
> 
> You can use the flag `--without-scheme` when installing meep.

Yes, you are right, it is possible to compile without scheme, the problem that I was facing is that in order to compile meep you need libctl, and if you don’t have guile (scheme) in you system you can not just simple use source files from GitHub repository but you need to unpack tar package of one of the libctl releases and then compile libctl ‘-- without-guile’. That is why I think it might be a good idea to add this option in building script to make compilation easier

smartalecH:
You can also compile libctl without guile.

Yaraslaut:
> You can also compile libctl without guile. But that assumes you are using a prepared package. There's some code generation that's required from the GutHub version.

Yes, i can compile everything manually, i am just saying that it would be great to be able to compile everything without guile using only script from repo

PR #2241, opened 2022-09-20
Update of build-meep script 

Yaraslaut:
Update of build-meep script in order to use it on up-to-date versions of Ubuntu as well as other distibutions.
Short summary:

- fix distribution check if installdeps set to false
- Added support for build without guile
- Added support for build without mpi and openmp
- Added support for build on Ubuntu 22.04 and 20.04


stevengj:
Looks good.

(Personally, I would tend to run this sort of script piece by piece via copy-and-paste, so that if any step fails I can fix it.)

Yaraslaut:
> Looks good.
> 
> (Personally, I would tend to run this sort of script piece by piece via copy-and-paste, so that if any step fails I can fix it.)

I've updated documentation, do you thing i need to change something else to be ready to merge or everything is fine already?
Also, this PR closes https://github.com/NanoComp/meep/issues/2130

Yaraslaut:
@stevengj  do I need to add anuthing to this PR?

PR #2251, opened 2022-09-29
separately delete forward dft chunk for adjoint optimization

### LINKED ISSUE #2097: adjoint optimization with multiple objective functions
Adjoint optimization fails if there are more than one objective functions. https://github.com/NanoComp/meep/blob/91d36ba73d1ae0eb135d18c783840cff9fcd71ef/python/adjoint/utils.py#L54-L58 Somewhere in ``mp._get_gradient`` frees the memory of the fields, causing problems when we reuse the forward fields again for the another objective function.

The issue actually exists before #1855 , where we could make copies of forward fields before passing in ``mp._get_gradient``. Right now, the naive fix could be installing one forward design region monitor for each objective function, so that we have a fresh copy each time we call ``mp._get_gradient``.

mochen4:
Fix #2097. Delete forward dft chunks at the end of all adjoint gradients calculation.

### DIFF
--- a/python/adjoint/optimization_problem.py
+++ b/python/adjoint/optimization_problem.py
@@ -306,9 +306,7 @@ class OptimizationProblem:
 
         for dri in range(self.num_design_regions):
             for i in range(3):
-                mp._delete_dft_chunk(
-                    self.forward_design_region_monitors[dri][i].swigobj
-                )
+                self.forward_design_region_monitors[dri][i].remove()
 
         # Cleanup list of lists
         if len(self.gradient) == 1:
--- a/python/adjoint/optimization_problem.py
+++ b/python/adjoint/optimization_problem.py
@@ -305,11 +305,10 @@ class OptimizationProblem:
         ]
 
         for dri in range(self.num_design_regions):
-            mp._delete_forward_dft_chunk(
-                self.forward_design_region_monitors[dri][0].swigobj,
-                self.forward_design_region_monitors[dri][1].swigobj,
-                self.forward_design_region_monitors[dri][2].swigobj,
-            )
+            for i in range(3):
+                mp._delete_dft_chunk(
+                    self.forward_design_region_monitors[dri][i].swigobj
+                )
 
         # Cleanup list of lists
         if len(self.gradient) == 1:
--- a/python/adjoint/optimization_problem.py
+++ b/python/adjoint/optimization_problem.py
@@ -306,9 +306,9 @@ class OptimizationProblem:
 
         for dri in range(self.num_design_regions):
             mp._delete_forward_dft_chunk(
-                self.forward_design_region_monitors[dri][0],
-                self.forward_design_region_monitors[dri][1],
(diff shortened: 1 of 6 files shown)


src/meepgeom.cpp
```diff
@@ -2999,12 +2999,23 @@ void material_grids_addgradient(double *v, size_t ng, size_t nf,
...
 
+void delete_forward_dft_chunk(std::vector<meep::dft_fields *> fields_f) {
+  for (int i = 0; i < 3; i++) {
+    meep::dft_chunk *current_forward_chunk = fields_f[i]->chunks;
+    meep::dft_chunk *next_forward_chunk;
+    while (current_forward_chunk) {
+      next_forward_chunk = current_forward_chunk->next_in_dft;
+      delete current_forward_chunk;
+      current_forward_chunk = next_forward_chunk;
+    }
```
stevengj, line 3015: Doesn't this need to set `fields_f[i]->chunks = NULL` or something?

stevengj:
Seems like this code should be refactored?  https://github.com/NanoComp/meep/blob/fba8aebb683620016d2f111f329cf8785039da2d/src/fields.cpp#L181-L190

src/meepgeom.cpp
```diff
@@ -2999,12 +2999,23 @@ void material_grids_addgradient(double *v, size_t ng, size_t nf,
   for (int i = 0; i < 3; i++) {
     for (int ii = 0; ii < adjoint_dft_chunks[i].size(); ii++) {
       delete adjoint_dft_chunks[i][ii];
-      delete forward_dft_chunks[i][ii];
     }
   }
 
 } // material_grids_addgradient
 
+void delete_forward_dft_chunk(std::vector<meep::dft_fields *> fields_f) {
```
stevengj, line 3007: It seems like it would make more sense to have a `dft_fields::delete_dft_chunks` method, and then just call it for each element of the array.

python/adjoint/optimization_problem.py
```diff
@@ -304,6 +304,13 @@ def calculate_gradient(self):
             for ar in range(len(self.objective_functions))
         ]
 
+        for dri in range(self.num_design_regions):
+            mp._delete_forward_dft_chunk(
+                self.forward_design_region_monitors[dri][0].swigobj,
```
stevengj, line 309: Just call `self.forward_design_region_monitors[dri][0].delete_dft_chunks()` on each of the objects?

python/adjoint/optimization_problem.py
```diff
@@ -304,6 +304,10 @@ def calculate_gradient(self):
             for ar in range(len(self.objective_functions))
         ]
 
+        for dri in range(self.num_design_regions):
+            for i in range(3):
+                self.forward_design_region_monitors[dri][i].remove()
```
stevengj, line 309:
```suggestion
                self.forward_design_region_monitors[dri][i].remove() # note that dft_fields::remove calls delete on its chunks, and the destructor ~dft_chunk automatically removes it from the fields object
```

PR #2262, opened 2022-10-09
Add more details and a layout schematic to the LED tutorial on dipole emission and reciprocity

oskooi:
Adds details related to the choice of zero `k_point` for a planewave at normal incidence following the discussion in [#2080 (comment)](https://github.com/NanoComp/meep/issues/2080#issuecomment-1272070550) as well as a schematic of the simulation layout to the recently added LED tutorial involving [reciprocity](https://meep.readthedocs.io/en/latest/Python_Tutorials/Custom_Source/#exploiting-reciprocity). Also includes some minor fixes.


smartalecH:
So currently the "backward" run places individual dft monitors for each "forward run" dipole source. Equivalently, we could just place a single line (or plane) dft monitor, right? The line would be discretized in the same fashion and is easier/cleaner to implement.

oskooi:
> Equivalently, we could just place a single line (or plane) dft monitor, right? The line would be discretized in the same fashion and is easier/cleaner to implement.

Recall that the [first tutorial](https://meep.readthedocs.io/en/latest/Python_Tutorials/Custom_Source/#exploiting-linear-time-invariance-of-the-materials-and-geometry) in this series, of which this reciprocity tutorial is an extension, involved $N = 10$ dipole sources. This is why there are 10 DFT monitors in the reciprocity tutorial. The [second tutorial](https://meep.readthedocs.io/en/latest/Python_Tutorials/Custom_Source/#an-efficient-approach-for-large-number-of-dipole-emitters) explicitly addresses the issue of large numbers of dipole emitters.

We could mention as a footnote at the end of the reciprocity tutorial how to account for a large number of dipole emitters and yet keep the demonstration the same.

stevengj:
Since we have discretization error anyway in the reciprocity formulation, it seems cleaner to use a line DFT monitor instead of individual points.

PR #2264, opened 2022-10-10
Fix Animate2D

### LINKED ISSUE #2261: Animate2D update not backwards compatible, breaks tutorials
The latest `Animate2D` update (#2186) added support for TO optimization visualization (thanks @thomasdorch!). In order to do so, we had to allow the `simulation` parameter to be optional, which changed the order of the arguments. Consequently, this breaks existing code, where the position was originally `sim`,`fields`, etc. In fact, the corresponding docstring still lists the wrong order.

This is a simple fix. We can actually make the `fields` argument optional too. Then we can restore the original order for position-dependent legacy code. 

We should update the docstrings too (and add typehints if possible).

@thomasdorch do you have the bandwidth to look into this?

thomasdorch:
Closes #2261 
Fixes and improvements to Animate2D:
- Fix ordering of `fields` and `sim` arguments to `Animate2D`
- Fix issue where fields were not updating in `Animate2D`
- Added more type hints
- Removed `sim` argument in `Animate2D` from all tutorials


stevengj:
I'm not sure what adding `plot_eps` has to do with this?  Should it be called `plot_eps` or `plot2D(fields=mp.Dielectric)`?

thomasdorch:
@stevengj I added it in just for consistency so that the code blocks for plotting epsilon and plotting fields look more similar:

```
if self.update_epsilon:
    ...
    eps = sim.plot_eps(**filtered_plot_eps)
    ...
if self.fields is not None:
    ...
    fields = sim.plot_fields(fields=self.fields, **filtered_plot_fields)
    ...
```
instead of:
```
if self.update_epsilon:
    ...
    eps = plot_eps(sim, **filtered_plot_eps)
    ...
if self.fields is not None:
    ...
    fields = sim.plot_fields(fields=self.fields, **filtered_plot_fields)
    ...
```
Good point about `plot2D`, it might make sense to use it for both actually. The issue with this would be that the `plot_eps` and `plot_fields` aren't actually plotting since no `ax` is passed to them. I think I should drop that commit from this PR and potentially make another PR to refactor a bit of the visualization code and clean this part up. What do you think?

oskooi:
The Python test suite is failing the Python 3.7 build configuration because of the missing `Literal` type in the `typing` module which was only introduced in Python 3.8 with [PEP 586](https://peps.python.org/pep-0586/).


```
FAIL: tests/test_3rd_harm_1d
============================

  File "../../../python/tests/test_3rd_harm_1d.py", line 5, in <module>
    import meep as mp
  File "/home/runner/work/meep/meep/build/meep-1.25.0-beta/_build/sub/python/meep/__init__.py", line 4440, in <mo\
dule>
    from .visualization import (
  File "/home/runner/work/meep/meep/build/meep-1.25.0-beta/_build/sub/python/meep/visualization.py", line 17, in \
<module>
    from typing import Callable, Union, Any, Literal, Tuple, List
ImportError: cannot import name 'Literal' from 'typing' (/opt/hostedtoolcache/Python/3.7.14/x64/lib/python3.7/typ\
ing.py)
```

thomasdorch:
Thanks for pointing that out @oskooi. I dropped the `Literal` type annotation from the commit.

oskooi (approved):
LGTM.

PR #2271, opened 2022-10-15
Add support for `load_minus_flux` to objective functions of adjoint solver

### LINKED ISSUE #2266: adding support for `load_minus_flux` to `EigenmodeCoefficient` objective function of adjoint solver
In calculations involving the $S_{11}$ scattering parameter (i.e., the reflectance into a given port from a source located in the same port), it is often useful prior to the start of the run to first subtract out the incident fields using the [`load_minus_flux`](https://meep.readthedocs.io/en/latest/Python_User_Interface/#flux-spectra) feature, particularly if the reflectance is smaller than e.g. 1e-6. While the computation of $S_{11}$ does [not typically require require a normalization run](https://meep.readthedocs.io/en/latest/Python_Tutorials/GDSII_Import/#when-computing-the-reflection-coefficient-s112-is-it-necessary-to-perform-a-separate-normalization-run-to-obtain-the-incident-fields) because in most telecom applications reflectance values less than 40 dB are considered negligible, adding support for  `load_minus_flux` to the adjoint solver's objective function `EigenmodeCoefficient` (and also `FourierFields` and Poynting flux in #2191)  would be useful as e.g. a means to validate the adjoint gradient against the brute-force finite-difference result for arbitrary values and problems.

oskooi:
Fixes #2266.

Adds a new class member `norm_dft_fields` to the `EigenmodeCoefficient` and `FourierFields` objective functions (and member `norm_near_fields` to class `Near2FarFields`) of the adjoint solver. Also includes a unit test.

### DIFF
--- a/python/adjoint/objective.py
+++ b/python/adjoint/objective.py
@@ -158,10 +158,10 @@ class EigenmodeCoefficient(ObjectiveQuantity):
         kpoint_func_overlap_idx: the index of the mode coefficient to return when
           specifying `kpoint_func`. When specified, this overrides the effect of
           `forward` and should have a value of either 0 or 1.
-        norm_dft_fields: the DFT fields obtained using `get_flux_data` from a
-          previous normalization run. This is subtracted from the DFT fields
+        subtracted_dft_fields: the DFT fields obtained using `get_flux_data` from
+          a previous normalization run. This is subtracted from the DFT fields
           of this mode monitor in order to improve the accuracy of the
-          reflectance calculation (i.e., the $S_{11}$ scattering parameter).
+          reflectance measurement (i.e., the $S_{11}$ scattering parameter).
           Default is None.
     """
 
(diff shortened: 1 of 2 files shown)


python/adjoint/objective.py
```diff
@@ -158,6 +158,11 @@ class EigenmodeCoefficient(ObjectiveQuantity):
         kpoint_func_overlap_idx: the index of the mode coefficient to return when
           specifying `kpoint_func`. When specified, this overrides the effect of
           `forward` and should have a value of either 0 or 1.
+        norm_dft_fields: the DFT fields obtained using `get_flux_data` from a
```
stevengj, line 161: Can re rename this to `subtracted_dft_fields`?

smartalecH (approved):

python/adjoint/objective.py
oskooi: Done.

Issue #1819, opened 2021-11-03
configure script problem


VictorEijkhout:
Configure gives the following error:

```
checking how to activate readline in Guile... ice-9 readline
checking for libctl dir... ./configure: line 29499: $'\357\277\274': command not found
configure: error: could not find libctl files; use --with-libctl=<dir>
```

which I can fix as follows:

```
if test x != x"$LIBCTL_DIR" -a -r "$LIBCTL_DIR/share/libctl/base/ctl.scm"; then
##￼     LIBCTL_DIR="$LIBCTL_DIR/share/libctl"
    LIBCTL_DIR="${LIBCTL_DIR}/share/libctl"
fi
```

But I don't understand why this works, or for that matter where the error comes from to begin with.

VictorEijkhout:
Oh wait, you have some non-ascii white space in that line.

oskooi:
Please note that GitHub issues are for filing bug reports and feature requests. They are not a general help desk. 

It seems the issues you are running into are related to installing software on Unix-like systems and not specific to Meep.

VictorEijkhout:
My problems are specifically with the meep installer. Can you confirm that your own configure script does not have junk on that line?

lockywolf:
meep has a problem with  the `./configure` script, but if you re-generate one with `./autogen.sh`

According to GNU, this is a wrong behaviour, the configure script should be portable. But almost every system now has autotools, so it should not be _much_ of a problem to re-generate one.

oskooi:
Fixed by #2273.

PR #2273, opened 2022-10-17
remove unicode whitespace character

dteal:
Removes the unicode whitespace character (the object replacement character U+FFFC which has byte code 0xEFBFBC (hex) and \357\277\274 (octal) in UTF-8) that is hidden at the beginning of a tab-indented line in configure.ac. This character causes an error when the installation configure script evaluates its line (which only happens occasionally due to the if statement around it).

I think this solves https://github.com/NanoComp/meep/issues/1819.


stevengj:
Thanks, not how that got in there.

PR #2282, opened 2022-10-20
Replace point DFT monitors with a line monitor in LED reciprocity tutorial

oskooi:
Following the suggestion from @smartalecH and @stevengj in #2262, this PR replaces the point DFT monitors with a single line monitor in the LED reciprocity tutorial. There is a noticeable improvement in the agreement with the results from the forward calculation.


smartalecH:
Indeed it looks much better! Nice.

doc/docs/Python_Tutorials/Custom_Source.md
```diff
@@ -542,66 +542,61 @@ def backward(rt: int, is_textured: bool) -> [List, np.ndarray]:
...
 
 A plot of the results from the forward and backward simulations is shown below. There is good agreeement across the entire frequency bandwidth. However, there is a significant difference in computational efficiency: the forward calculation had a runtime which was more than three times that of the backward calculation.
```
stevengj, line 601: The agreement is too good — maybe change the red line to dashed, then you will be able to see the blue line underneath.

oskooi: Done.

Issue #2286, opened 2022-10-27
add epsilon colorbar for plot2D


joamatab:
How could we add an epislon or refractive index colorbar for plot2D? 

See for example the colorbars here
https://gdsfactory.github.io/gdsfactory/notebooks/plugins/tidy3d/00_tidy3d.html

Also we could improve Meeps' 3D visualization using trimesh?

https://gdsfactory.github.io/gdsfactory/notebooks/plugins/meep/001_meep_sparameters.html#Visualizing-the-3D-Geometry

SkandanC:
Meep's current plot3D() function is broken, due to issues with mayavi. 

```
Traceback (most recent call last):
  File "/home/skandanc/plot.py", line 14, in <module>
    sim.plot3D()
  File "/home/skandanc/miniconda3/lib/python3.9/site-packages/meep/simulation.py", line 4213, in plot3D
    return vis.plot3D(self)
  File "/home/skandanc/miniconda3/lib/python3.9/site-packages/meep/visualization.py", line 681, in plot3D
    from mayavi import mlab
  File "/home/skandanc/miniconda3/lib/python3.9/site-packages/mayavi/mlab.py", line 34, in <module>
    from .tools.helper_functions import contour3d, test_contour3d, \
  File "/home/skandanc/miniconda3/lib/python3.9/site-packages/mayavi/tools/helper_functions.py", line 16, in <module>
    from .modules import VectorsFactory, StreamlineFactory, GlyphFactory, \
  File "/home/skandanc/miniconda3/lib/python3.9/site-packages/mayavi/tools/modules.py", line 27, in <module>
    from .filters import new_class
  File "/home/skandanc/miniconda3/lib/python3.9/site-packages/mayavi/tools/filters.py", line 232, in <module>
    _make_functions(locals())
  File "/home/skandanc/miniconda3/lib/python3.9/site-packages/mayavi/tools/filters.py", line 218, in _make_functions
    klass = new_class(
  File "/home/skandanc/miniconda3/lib/python3.9/site-packages/mayavi/tools/filters.py", line 30, in new_class
    return new.classobj(name, bases, dict_)
AttributeError: module 'new' has no attribute 'classobj'
```

I managed to create a rudimentary 3D representation of the Simulation object using vispy. I need to add the source planes and monitor planes, and fine tune the looks. Sim object for an MMI1X2:

[...]

Issue #2267, opened 2022-10-11
Add Poynting components to plot_fields


thomasdorch:
Currently, visualization.plot_fields accepts one of:
```
[mp.Ex, mp.Ey, mp.Ez, mp.Er, mp.Ep, mp.Dx, mp.Dy, mp.Dz, mp.Hx, mp.Hy, mp.Hz]
```
as valid arguments to `fields`. Can we add
```
[mp.Sx, mp.Sy, mp.Sz, mp.Sr, mp.Sp]
```
to this list as well?

smartalecH:
Keep in mind that the poynting flux is an integral over space and is a scalar output. So there's not really anything to visualize. 

You can view things like $E^2$ using function modifiers (as described in the docs).

smartalecH:
Oh I see you are talking about the "derived components" of the Poynting flux, as we call them. 

Sure we can add them to the list. Want to add it to your PR? Might as well add all the other components too. 

(Typically we break things down as granularity as possible, meaning different PRs for different features, but you might as well lump this one in with your other PR)

thomasdorch:
Oh I see how that was unclear, the components were what I was referring to! 

I was thinking of creating another PR for some more type hinting and a fix (if appropriate) for #2268. That way #2264 is more specific to just fixing what's broken. How does that sound?

PR #2289, opened 2022-10-28
Add colorbars for plot2D

thomasdorch:
This PR:
- Adds colorbars for `plot2D` (both in `plot_epsilon` and `plot_fields`) (#2286)
- Adds additional field components (and derived components) to `plot_fields` (#2267)

I think there's more room to improve the look of the colorbars, so let me know if I should make any adjustments!

So far fields look like this:
image
and epsilon looks like this:
image

@smartalecH 
@joamatab 
@SkandanC

### DIFF
--- a/python/visualization.py
+++ b/python/visualization.py
@@ -48,7 +48,6 @@ default_field_parameters = {
     "alpha": 0.8,
     "post_process": np.real,
     "colorbar": False,
-    "colorbar_label": None,
 }
 
 default_eps_parameters = {
@@ -63,11 +62,12 @@ default_eps_parameters = {
 }
 
 default_colorbar_parameters = {
+    "label": None,
     "orientation": "vertical",
+    "extend": None,
     "position": "right",
     "size": "5%",
     "pad": "2%",
-    "clip_values": False,
 }
 
 default_boundary_parameters = {
@@ -463,7 +463,7 @@ def _add_colorbar(
     cmap: str,
     vmin: float,
     vmax: float,
-    label: str,
+    default_label: Optional[str] = None,
     colorbar_parameters: Optional[dict] = None,
 ) -> None:
     """Add a colorbar to the parent Figure of 'ax' by creating an additional Axes."""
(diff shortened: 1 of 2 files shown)


python/visualization.py
```diff
@@ -447,6 +450,31 @@ def sort_points(xy):
...
+    from mpl_toolkits.axes_grid1 import make_axes_locatable
+
+    # Create a map between field/eps values and colors in the colormap.
+    # Note: cm.get_cmap() is deprecated for matplotlib>=3.6, use mpl.colormaps[cmap] instead if necessary.
+    sm = mpl.cm.ScalarMappable(
+        norm=mpl.colors.Normalize(vmin, vmax, clip_values), cmap=mpl.cm.get_cmap(cmap)
+    )
+    plt.colorbar(
+        mappable=sm,
+        cax=make_axes_locatable(ax).append_axes(position="right", size="5%", pad=0.05),
```
oskooi, line 472: Any reason why the `pad` parameter is chosen as an absolute quantity rather than a percentage (e.g., `2%`) similar to `size`?

oskooi (approved):
LGTM.

oskooi:
You should also add this new feature to the `plot2D` docstrings in [`simulation.py`](https://github.com/NanoComp/meep/blob/5fa12b499d4934ea8b352e3e8033762b34f6eb0c/python/simulation.py#L4657-L4673) so that it is documented in the user manual.

smartalecH:
Why not pass the default plotting options used in `plot2D`?

python/visualization.py
thomasdorch: Good catch, I agree that using the relative quantity is better here.

thomasdorch:
Thanks for the feedback! 

I think it's in much better shape now that I've:
- updated the docs in `simulation.py` @oskooi 
- added new argument, `colorbar_parameters`, that let the user specify parameters for the colorbar (including `label`) and pass them into `plot2d` @smartalecH 

Let me know what you think.

smartalecH:
LGTM!

PR #2302, opened 2022-11-03
Tutorial for metalens optimization with FourierFields

mawc2019:
This PR adds a tutorial for metalens optimization with `FourierFields`, which is a counterpart of [Metalens Optimization with Near2Far](https://github.com/NanoComp/meep/blob/master/python/examples/adjoint_optimization/05-Near2Far.ipynb).


stevengj:
Needs some explanation at the beginning to (1) explain what problem you are solving and (2) contrast it with the near2far approach (advantages = greater flexibility, not limited to homogeneous media surrounding near2far, more efficient for fields at nearby points; disadvantages = less efficient for far away points because requires larger computational cell)

Also would be good to show a picture of the CW fields at the the three wavelengths.

Issue #2307, opened 2022-11-10
investigating the error in broadband adjoint gradients


oskooi:
This is an attempt to investigate the errors in the adjoint gradient of a broadband simulation. The test involves comparing the adjoint gradients using two different configurations: 

1. a **multi-frequency** adjoint simulation at wavelengths [ $\lambda_1$, $\lambda_2$, $\lambda_3$ ] 
2. a **single-frequency** adjoint simulation at a wavelength of $\lambda_1$, $\lambda_2$, or $\lambda_3$

The adjoint gradient at wavelength $\lambda_m$ in (1) should match the adjoint gradient $\lambda_m$ in (2) for each $m=1,2,3$. However, we should expect some discrepancy in the results for $\lambda_1$ and $\lambda_3$ (the non-center wavelengths of the pulsed source) because the `EigenmodeSource` uses the mode profile for the center frequency only. While we have already investigated the errors introduced by the broadband mode source at off-center frequencies in [Tutorial/Eigenmode Source](https://meep.readthedocs.io/en/latest/Python_Tutorials/Eigenmode_Source/#what-happens-when-the-source-time-profile-is-a-pulse), we do not seem to have investigated the effect of these errors on the adjoint gradient.

The test involves two objective functions: reflectance and transmittance of a waveguide mode converter. A schematic of the simulation setup is shown in the first figure below. The design region is random.  The results (second figure) demonstrate that the $L_2$ norm of the difference in the adjoint gradient for (1) and (2) is nearly one order of magnitude larger for the non-center wavelengths relative to the center wavelength. The relative error is larger for the reflectance objective function (blue data) than the transmittance objective function (red data). When the absolute magnitude of the gradients in (1) and (2) are compared side by side, they seem to be similar (third and fourth figures).

It would seem that a one order of magnitude reduction in the relative accuracy of the adjoint gradient is large enough that it could degrade the convergence of a topology optimization run.

![waveguide_mode_converter_opt](https://user-images.githubusercontent.com/7152530/201019630-d89f906f-d522-4d75-8efa-863d1d379e8a.png)

![norm_adjoint_gradient](https://user-images.githubusercontent.com/7152530/201021979-fc467265-44a1-410f-804c-b9827b20cf84.png)

**1. Adjoint gradient of reflectance objective function at $\lambda$ = 1.55 μm (center wavelength)**

[...]

PR #2314, opened 2022-11-17
Add two unit tests for adjoint gradients related to source and monitor bandwidths

oskooi:
Related to #2307.


oskooi:
By reducing the source bandwidth by a factor of four, it is possible to then lower the minimum tolerance used to compare the adjoint gradient against the finite-difference approximation by a factor of *two to ten*. A tighter check makes for a stronger test.

Separately, I have added a one-line docstring to all of the tests which summarizes what the test is doing.

Issue #2235, opened 2022-09-14
Computing the gradient of a level-set function


oskooi:
Related to #2011.

This is an initial attempt to use the new subpixel smoothing feature of the adjoint solver to compute the gradient of a structure parameterized by its geometry (a level set) as an alternative to the conventional density-based (`MaterialGrid`) representation. The example involves computing the gradient of the diffraction efficiency of the m=+1 transmitted order of a 1D binary grating with respect to its height and fill factor. The results are to be validated by the usual method of the brute-force gradient computed using a finite difference. Once everything is working correctly, this demonstration will be converted into a tutorial for the user manual and a unit test.

The main feature of this calculation is a *differentiable* wrapper function  (`grating_matgrid` in the script below) that takes the grating height and fill factor as arguments and returns the weights of a `MaterialGrid` for the binary grating. These weights are then passed to the adjoint solver via a `DesignRegion` object in the usual way and the gradient which is computed is then backpropagated through the wrapper function. 

An image of the geometry created using this approach confirms that the structure is set up correctly. However, it seems there is a bug in the current setup because the final gradient is *zero*. Also, there is an important warning from autograd:
```
/.local/lib/python3/site-packages/autograd/tracer.py:14: UserWarning: Output seems independent of input.
  warnings.warn("Output seems independent of input.")
```

This message suggests there is a likely bug in the backpropagation step.

![binary_grating_matgrid](https://user-images.githubusercontent.com/7152530/190065757-a0317b65-70e3-403d-b9ff-3adccda20d04.png)

**output of running the script below**
```
-----------
Initializing structure...
Halving computational cell along direction y
Splitting into 2 chunks by voxels
time for choose_chunkdivision = 0.000121082 s
Working in 2D dimensions.
Computational cell is 8.5 x 0.78 x 0 with resolution 100
time for set_epsilon = 0.024535 s
-----------
field decay(t = 50.005): 0.04304565716016541 / 0.04304565716016541 = 1.0
field decay(t = 100.01): 0.11354452052348654 / 0.11354452052348654 = 1.0
on time step 29645 (time=148.225), 0.000134933 s/step
field decay(t = 150.01500000000001): 1.3975259335832059e-08 / 0.11354452052348654 = 1.2308175922008755e-07
field decay(t = 200.02): 5.861816753051918e-16 / 0.11354452052348654 = 5.162571232875486e-15
run 0 finished at t = 200.02 (40004 timesteps)
Starting forward run...
-----------
[...]

PR #2332, opened 2022-12-07
Make the EigenmodeCoefficient adjont solver accept a DiffractedPlanewave object

mawc2019:
The `mode` argument of the `EigenmodeCoefficient` adjont solver cannot be a `DiffractedPlanewave` object. This PR allows this `mode` argument to be a `DiffractedPlanewave` object.

### DIFF
--- a/python/adjoint/objective.py
+++ b/python/adjoint/objective.py
@@ -150,7 +150,7 @@ class EigenmodeCoefficient(ObjectiveQuantity):
     """A frequency-dependent eigenmode coefficient.
     Attributes:
         volume: the volume over which the eigenmode coefficient is calculated.
-        mode: the eigenmode number.
+        mode: the eigenmode number or the DiffractedPlanewave object.
         forward: whether the forward or backward mode coefficient is returned as
           the result of the evaluation.
         kpoint_func: an optional k-point function to use when evaluating the eigenmode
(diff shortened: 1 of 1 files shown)


oskooi:
Note that #2054 (unfinished) is identical to this PR. Unfortunately, enabling this feature requires more than simply modifying the API for `EigenmodeCoefficient` as explained in this [comment](https://github.com/NanoComp/meep/pull/2054#issuecomment-1126649464).

oskooi:
> Is there a temporary expedient for the DiffractedPlanewave adjoint solver?

The workaround is to specify the wavevector of the diffracted order manually using the `kpoint_func` parameter of the `EigenmodeCoefficient` object. See the example in [#2235(comment)](https://github.com/NanoComp/meep/issues/2235#issue-1372369733). This also requires the hack in `mpb.cpp` in #2285.

oskooi:
> Did you compare this get_eigenmode_coefficients approach mentioned above with the DiffractedPlanewave approach?

We already have a unit test ([`test_diffracted_planewave.py`](https://github.com/NanoComp/meep/blob/master/python/tests/test_diffracted_planewave.py)) which verifies, using a 2d binary grating, that `get_eigenmode_coefficients` produces the same result when called with either a `band_num`/`eig_parity` (but not `kpoint_func`) or `DiffractedPlanewave` object. Once #2285 is resolved, I was planning to add a separate test involving `kpoint_func`.

Anyhow, since the 2d metagrating is a 3d simulation, you need to be careful about specifying the correct symmetry of the mode. I think it should be: `eig_parity = mp.EVEN_Y if P_pol else mp.ODD_Y`.

PR #2333, opened 2022-12-08
Reformulate Simulation.get_sources() and add 2D GaussianBeam

hammy4815:
First, I reformulated get_sources. Before, simulation defined logic for each source type separately with `if isinstance(source, GaussianBeam)`, `elif...` and then provided the specific logic to add the source parameters to the backend.

Now, it simply loops through the list of sources belonging to the simulation and calls add_source on each, passing in self to give the source access to the simulation. Source.py now contains the base logic for adding sources, but `EigenModeSource`, `IndexedSource`, `GaussianBeamSource`, and (new) `GaussianBeam2DSource` overload this logic. 

In addition, a new class `GaussianBeamSource2D` is created. This allows a rigorous solution to Maxwell's equations in 2-dimensions via the complex point source method (similar to 3D). All the logic for finding the solution is provided in the python alone, finding the field at a slice. This is done very similar to `green2d` in near2far.cpp by using Hankel functions of both kinds to generate forward and backward propagating beams when the point source is complex in location. Using this slice, equivalent sources are calculated (also in python). 

Once the equivalent sources are made (python meep objects), `GaussianBeam2dSource.add_source()` simply calls `add_source()` on these equivalent source objects. The fields are normalized to have $|E|=1$ at the location of the point source. 

This is all done in python. Because I wrote a green2d python function and one for equivalent source (for single frequency), we can run our code once in python to generate the sources at the beginning of the run and then pass this information in to the backend before the actual meep runs start. Thus, we do not need to write our code in c++ the way 3D was done. Here is a demonstration of it working, under different conditions. 

Screen Shot 2022-12-07 at 9 32 00 PM

We can change our plane and provide rotations of k, which is done by rotating the imaginary component of the location of the complex point source between x and y.

Screen Shot 2022-12-07 at 9 32 42 PM

We can also change our polarization. In order for this to work, an E source and H source are placed in the same location oriented 90 degrees from each other, similar to what's done for 3D. 

Screen Shot 2022-12-07 at 9 33 19 PM

### DIFF
--- a/python/source.py
+++ b/python/source.py
@@ -761,6 +761,255 @@ class GaussianBeamSource(Source):
         add_vol_src()
 
 
+class GaussianBeam2DSource(GaussianBeamSource):
+    """
+    Identical to `GaussianBeamSource`, except that the beam is defined in 2d. This is useful for 2d simulations, where the 3d beam is not exact.
+
+    The `SourceTime` object (`Source.src`), which specifies the time dependence of the source, should normally be a narrow-band `ContinuousSource` or `GaussianSource`.  (For a `CustomSource`, the beam frequency is determined by the source's `center_frequency` parameter.)
+
+
+            self._beam_x0 = beam_x0
+            self._beam_kdir = beam_kdir
+            self._beam_w0 = beam_w0
+            self._beam_E0 = beam_E0
+
+    """
+
+    def get_fields(self, sim):
+        from scipy.special import hankel1, hankel2, jv
+
+        # Beam parametersd
+        freq = self.src.swigobj.frequency().real
+        eps = sim.fields.get_eps(
+            mp.py_v3_to_vec(sim.dimensions, self.center, sim.is_cylindrical)
+        ).real
+        mu = sim.fields.get_mu(
+            mp.py_v3_to_vec(sim.dimensions, self.center, sim.is_cylindrical)
+        ).real
+        k = 2 * np.pi * freq * np.sqrt(eps * mu)
+
+        # Get this coordinate system
+        center = mp.py_v3_to_vec(sim.dimensions, self.center, sim.is_cylindrical)
+        size = mp.py_v3_to_vec(sim.dimensions, self.size, sim.is_cylindrical)
+        beam_x0 = mp.py_v3_to_vec(sim.dimensions, self.beam_x0, sim.is_cylindrical)
+        beam_kdir = mp.py_v3_to_vec(sim.dimensions, self.beam_kdir, sim.is_cylindrical)
+        beam_E0 = self.beam_E0
+
+        # Check for errors
+        if size.x() and size.y():
+            raise Exception(
+                "GaussianBeam2DSource should be a line source, not a plane. Either set size.x or size.y to zero."
+            )
+
+        # Complex point source
+        x0 = center.x() + beam_x0.x()
+        y0 = center.y() + beam_x0.y()
+        z0 = k * self.beam_w0**2 / 2
+        kdir = beam_kdir.x() + 1j * beam_kdir.y()
+        kdir = kdir / np.abs(kdir)
+        jx0 = 1j * z0 * np.real(kdir)
+        jy0 = 1j * z0 * np.imag(kdir)
+        X0 = np.array([x0 + jx0, y0 + jy0]).astype(complex)
+
+        # Create grid
+        x = (
+            np.linspace(
+                center.x() - size.x() / 2,
+                center.x() + size.x() / 2,
+                int(2 * sim.resolution * size.x()),
+            )
+            if size.x() > 0
+            else np.array([center.x()])
+        )
+        y = (
+            np.linspace(
+                center.y() - size.y() / 2,
+                center.y() + size.y() / 2,
+                int(2 * sim.resolution * size.y()),
+            )
+            if size.y() > 0
+            else np.array([center.y()])
+        )
+        xx, yy = np.meshgrid(x, y)
+        X = np.transpose(np.array([xx, yy])[:, :, :], axes=(0, 2, 1))
+
+        # Find which points are incoming vs outgoing
+        incoming_arg, outgoing_arg, waist_points = self.incoming_mask(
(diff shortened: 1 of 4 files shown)


python/source.py
```diff
@@ -646,6 +729,286 @@ def beam_w0(self):
...
+        ]
+        gaussianbeam = mp.gaussianbeam(*gaussianbeam_args)
+        add_vol_src_args = [self.src.swigobj, where, gaussianbeam]
+        add_vol_src = functools.partial(sim.fields.add_volume_source, *add_vol_src_args)
+        add_vol_src()
+
+
+class GaussianBeam2DSource(GaussianBeamSource):
+    """
+    Identical to `GaussianBeamSource`, except that the beam is defined in 2d. This is useful for 2d simulations, where the 3d beam is not exact.
```
oskooi, line 766: Would be nice to follow the docstring style convention described in https://google.github.io/styleguide/pyguide.html#384-classes. This involves adding a one-line summary as the docstring to all the member functions of this class.

oskooi (commented):
Thanks for putting this together. 

It would be useful to add a unit test for this new feature, perhaps similar to the existing test for `GaussianBeamSource` in https://github.com/NanoComp/meep/blob/master/python/tests/test_gaussianbeam.py.

python/source.py
```diff
@@ -590,12 +621,64 @@ def eig_power(self, freq):
...
 
-    (In 3d, we use a ["complex point-source" method](https://doi.org/10.1364/JOSAA.16.001381) to define a source that generates an exact Gaussian-beam solution.  In 2d, we currently use the simple approximation of taking a cross-section of the 3d beam.  In both cases, the beam is most accurate near the source's center frequency.)
+    (In 3d, we use a ["complex point-source" method](https://doi.org/10.1364/JOSAA.16.001381) to define a source that generates an exact Gaussian-beam solution.  In 2d, we currently use the simple approximation of taking a cross-section of the 3d beam.  In both cases, the beam is most accurate near the source's center frequency.) To use the true solution for a 2d Gaussian Beam, use the `GaussianBeam2DSource` class instead.
```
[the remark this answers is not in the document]
hammy4815: This makes sense to me. My only worry is will people struggle to realize the name change when they next update meep and their old scripts fail to recognize GaussianBeamSource?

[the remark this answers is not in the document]
smartalecH: > they next update meep and their old scripts fail to recognize GaussianBeamSource

I agree. We should probably avoid any breaking changes for something like this... especially since this is preliminary. A future PR will add support for multiple frequencies etc.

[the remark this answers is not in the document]
smartalecH: In other words, we shouldn't need to break the API at all. In the future, we can refactor and call `GaussianBeamSource` in either 2D or 3D and it just "does what it's supposed to do".

Right now that's a challenge due to the complex Hankel functions etc which aren't (readily) compatible with the current 3D codebase in C++.

stevengj:
The most backward-compatible option would be something like:

1. Rename the existing `GaussianBeamSource` to `GaussianBeam3DSource`.  
2. Define a compatibility wrapper called `GaussianBeamSource` which calls `GaussianBeam3DSource`.   *But* if the simulation is in 2d, then this wrapper will print a deprecation warning noting that in the future it will be changed to call `GaussianBeam2DSource` in the 2d case — current users should explicitly call `GaussianBeam3DSource` if they want to keep the old behavior, or call `GaussianBeam2DSource` if they want the new (better) behavior.
3. In a future Meep version, change `GaussianBeamSource` to call `GaussianBeam2DSource` in 2d, with no warning.

hammy4815:
Performed the following changes:

1. Refactored GaussianBeamSource -> GaussianBeam3DSource and created a wrapper GaussianBeamSource that warns the user when running a 2D sim
2. Pulled get_equiv_sources out as a standalone function to be used
3. Added 2D to the gaussianbeam test case. We can still add more complex test cases for both 3D and 2D if we wish such as ensuring the beam waist is correct.

python/tests/test_gaussianbeam.py
```diff
@@ -76,10 +76,11 @@ def gaussian_beam(self, rot_angle):
             f"ratio of the Gaussian beam energy at the focus over the maximum beam energy for the entire cell: {frac}"
         )
 
-        self.assertGreater(frac, 0.99)
+        self.assertGreater(frac, 0.98)
 
     def test_gaussian_beam(self):
-        self.gaussian_beam(-40)
+        self.gaussian_beam(-40, mp.GaussianBeam2DSource)
+        self.gaussian_beam(-40, mp.GaussianBeam3DSource)
```
oskooi, line 83: Since this will just trigger the same 2d test as the previously line, to avoid the redundant calculation it might be better to just test that this function call correctly raises a warning. The procedure for how to do this is described in https://docs.python.org/3/library/warnings.html#testing-warnings.

oskooi: Nevermind. This is correct since this line will execute the 2d simulation using the cross section of the 3d Gaussian beam (the existing test).

hammy4815: We can still test the deprecation warning if we wish. It could look like 

```python
    def test_gaussian_beam(self):

        # 2D Gaussian beam test
        self.gaussian_beam(-40, mp.GaussianBeam2DSource)
        self.gaussian_beam(-40, mp.GaussianBeam3DSource)

        # 3D Gaussian beam deprecation warning test
        with warnings.catch_warnings(record=True) as w:
            warnings.simplefilter("always")
            self.gaussian_beam(-40, mp.GaussianBeamSource)
            self.assertTrue(len(w) == 1)
            self.assertTrue(issubclass(w[-1].category, DeprecationWarning))
            self.assertTrue("deprecated" in str(w[-1].message))
```

Should I include the addition or keep it out? It would add one extra meep run to the test cases. Alternatively we could exclude `self.gaussian_beam(-40, mp.GaussianBeam3DSource)` since calling GaussianBeamSource will test that case.

oskooi (approved):

PR #2350, opened 2022-12-20
pin Numpy version in requirements.txt

oskooi:
Fixes a recent failure in the CI for Python 3.10 related to Numpy 1.24 which [raises a `TypeError` in ufuncs due to an expired deprecation](https://numpy.org/devdocs/release/1.24.0-notes.html#expired-deprecations).

### DIFF
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -3,7 +3,7 @@ h5py
 jax
 jaxlib
 matplotlib
-numpy
+numpy<1.24
 parameterized
 pytest
 scipy


stevengj:
Can we fix it to work with numpy 1.24?   What Meep code is triggering the error?

stevengj:
We can merge this for now, but we should file an issue to fix compatibility with the latest numpy version.

oskooi:
> Can we fix it to work with numpy 1.24? What Meep code is triggering the error?

This PR was indeed a stopgap measure to get the CI passing.

However, this PR is actually not necessary given #2348, #2349, and #2351. We can verify this with a new PR that undoes the change in this PR after those other three PRs are all merged.

Issue #2356, opened 2022-12-27
GaussianBeam2DSource Bug Cases Found


hammy4815:
A few bugs were discovered regarding the new GaussianBeam2DSource implementation.

1. Using materials other than air produces slight issues after creating the equivalent source. 

For example, as can be seen below, after changing the refractive index to 1.5, small issues arise and back propagation below the source region now shows up. I am unsure at this moment if this is an issue in the creation of the beam fields themselves, or in the call to equivalent source, but it appears that a factor of $n$ or $\epsilon$ is either missing or accidentally added. 

<img width="460" alt="Screenshot 2022-12-27 at 2 42 48 PM" src="https://user-images.githubusercontent.com/51729722/209731419-50e148ba-6b2d-4c84-918d-e39957ff9927.png">

Currently, as given [here](https://arxiv.org/pdf/1301.5366.pdf) in equation 4.4, the equivalent current sources do not contain a factor of $\epsilon$ or $\mu$, matching how it is currently implemented in MEEP. 

<img width="702" alt="Screenshot 2022-12-27 at 2 51 39 PM" src="https://user-images.githubusercontent.com/51729722/209732827-c3105d39-ca10-4923-9247-3a65e1a79d2d.png">

2. Under stronger index changes, or a significant wavelength change, an overflow occurs in the normalization of the fields before calling equivalent source, and the time fields remain zero. 

<img width="442" alt="Screenshot 2022-12-27 at 2 46 39 PM" src="https://user-images.githubusercontent.com/51729722/209731641-0f2ea77c-15a0-45bc-8a20-d1520760cda6.png">

This is because calling scipy.special.hankel produces enormous results (on the order of 1e120). When the beam waist is similar in size to the effective wavelength, the order of magnitude is low enough to avoid overflows. We likely need a clever way to fix this issue at this step (calling the hankel function).

oskooi:
The refractive index $n$ (of free space) is implicit in the fields $E_+$ and $H_+$ .There is no need for the index to be accounted for separately in the expression for the equivalent current.

Note that in the original implementation of the `GaussianBeam` object via the complex point source in #711 (and also the [analytic expression for the electric field of a Gaussian beam](https://en.wikipedia.org/wiki/Gaussian_beam#Mathematical_form)), the beam's wavevector $\vec{k}$ in free space includes the index $n$:

[...]

PR #2363, opened 2023-01-05
Bug fixes for GaussianBeam2DSource

hammy4815:
This pull request begins the bug fixes for GaussianBeam2DSource as described under issue #2356 .

Before merging, there is one last fix still required: When the material index is too high, and the beam waist becomes much smaller than the effective wavelength in the material, the overflows occur when passing values into the complex Hankel functions.

After finding these bugs, it might also be useful to add more test cases such as different materials, wavelengths, etc.


hammy4815:
A detailed description of this first commit is found under this comment [here](https://github.com/NanoComp/meep/issues/2356#issuecomment-1371638635).

PR #2364, opened 2023-01-05
connectivity constraint tutorial

mochen4:
Add a tutorial for the connectivity constraint


stevengj:
This is a good first step.  It would eventually be good to show an example using this to perform an simple optimization (e.g. metalens?) with a connectivity constraint.

stevengj:
Also, please add a reference to cite.

stevengj:
Might be good to note that this is typically used in practice along with a lengthscale constraint.

Issue #2344, opened 2022-12-15
Adjoint TO filters not properly accounting for boundary conditions


smartalecH:
As [discussed](https://github.com/NanoComp/photonics-opt-testbed/pull/32), the current adjoint TO filters are not properly accounting for boundary conditions. In theory, we should only need a single pixel around the boundary. The current codebase does an edge-based pad on all dimensions:

https://github.com/NanoComp/meep/blob/ee5324c8da741ad21d2673e01069e6f48746bc1b/python/adjoint/filters.py#L83

Maybe we need an even larger pad? (It's already padded twice what it needs to be to avoid overwriting values due to the circular convolution...)

(cc @oskooi )

PR #2365, opened 2023-01-05
Tutorial for adjoint optimization of broadband waveguide mode converter with minimum feature size

oskooi:
Adds a new tutorial for the adjoint solver demonstrating some of its advanced functionality. This is an initial working version and will likely be modified and expanded in future PRs.

Based mainly on results in NanoComp/photonics-opt-testbed#32.

@smartalecH and @mochen4, would you review the text under the new section "Broadband Waveguide Mode Converter with Minimum Feature Size" in `Adjoint_Solver.md` and make any revisions as necessary?


doc/docs/Python_Tutorials/Adjoint_Solver.md
```diff
@@ -2,13 +2,704 @@
...
+Meep contains an adjoint-solver module for efficiently computing the
+gradient of an arbitrary function of the mode coefficients
+($\mathcal{S}$-parameters), DFT fields, local density of states
+(LDOS), and far fields (Fraunhofer condition) with respect to
```
smartalecH, line 8: The Fraunhofer condition is not required for meep's Near2Far features (regardless if performing adjoint optimization or not). Because we use a Green's function, the result is accurate at any point in space. Of course, as the point approaches the effective currents, this integral becomes more ill-conditioned... but there's no caveat on accuracy here.

If we were using a simple *Fourier transform* to perform the Near-to-far projection, then we would indeed be restricted to the "Fraunhofer regime".

doc/docs/Python_Tutorials/Adjoint_Solver.md
```diff
@@ -2,13 +2,704 @@
...
+distinct timestepping runs are required. The first run is the
+"forward" calculation to compute the objective function and the DFT
+fields of the design region. The second run is the "adjoint"
+calculation to compute the gradient of the objective function with
+respect to the design variables. The adjoint run involves a special
+type of current source distribution used to compute the DFT fields of
+the design region. The gradient is computed in post processing using
+the DFT fields from the forward and adjoint runs. The gradient
+calculation is fully automated. The theoretical and computational
+details of the adjoint-solver module are described in this manuscript:
```
smartalecH, line 22: nit: paper or publication, not manuscript (it's already been peer reviewed and accepted).

doc/docs/Python_Tutorials/Adjoint_Solver.md
```diff
@@ -2,13 +2,704 @@
...
+
+In this example, we use a minimum feature size of 150 nm for the line
+width *and* line spacing. The implementation of these constraints is
+based on [A.M. Hammond et al., Optics Express, Vol. 29, pp. 23916-38
+(2021)](https://doi.org/10.1364/OE.431188).
+
+There are five important items to note in the set up of the
+optimization problem:
+
+- The lengthscale constraint is activated only in the final epoch. The
```
smartalecH, line 83: > The initial design of the final epoch must be binary.

This isn't actually true. For this particular example, it seemed to help. But in the 80+ examples from our paper, *none* of the designs were binary when the lengthscale constraint was activated.

In fact, on feature of the lengthscale constraint is it forces the design to be binary without increasing $\beta$ to a point that breaks differentiability.

doc/docs/Python_Tutorials/Adjoint_Solver.md
```diff
@@ -2,13 +2,704 @@
...
+  initial design of the final epoch must be binary.
+
+- The initial value of the epigraph variable of the final epoch (in
+  which the linewidth constraint is imposed) should take into account
+  the value of the constraint itself to ensure a feasible starting
+  point.
+
+- The edge of the design region is padded by a filter radius (rather
```
smartalecH, line 91: fixing #2344 should resolve this

doc/docs/Python_Tutorials/Adjoint_Solver.md
```diff
@@ -2,13 +2,704 @@
...
+  than e.g., a single pixel) to produce measured lengthscales of the
+  final design that are consistent with the imposed constraint.
+
+- The hyperparameters of the lengthscale constraint function (`glc` in
+  the script below), need to be chosen carefully to produce final
+  designs which do not significantly degrade the performance of the
+  unconstrained designs at the start of the final epoch.
+
+- Damping of the design weights is used for the early epochs in which
```
smartalecH, line 100: I'm not sure I would say the weights are being dampened (the weights are non complex)...

Rather, we introduce a new damping term via a material conductivity which *depends* on the weights meant to encourage binarization.

doc/docs/Python_Tutorials/Adjoint_Solver.md
mochen4: > In fact, on feature of the lengthscale constraint is it forces the design to be binary without increasing β to a point that breaks differentiability.

In my experience, that is exactly the issue: the lengthscale constraint may change the structure too much by binarizing it too much, and the performance is kind of ruined. That's why I would personally wait until the structure is binarized enough before imposing lengthscale constraint.

doc/docs/Python_Tutorials/Adjoint_Solver.md
```diff
@@ -2,13 +2,704 @@
...
+preclude the use of gradient-based optimization algorithms for this
+problem which involves 12 independent functions ($R$ and $1-T$ for
+each of six wavelengths). Fortunately, there is a workaround: the
+problem can be reformulated as a differentiable problem by introducing
+a dummy varaiable $t$ (the
+[epigraph](https://en.wikipedia.org/wiki/Epigraph_(mathematics))) and
+adding each independent function as a new nonlinear constraint. See
+the [NLopt
+documentation](https://nlopt.readthedocs.io/en/latest/NLopt_Introduction/#equivalent-formulations-of-optimization-problems)
+for an overview of this approach.
```
mochen4, line 73: The minimax/epigraph formulation is also mentioned in [tutorial-06](https://nbviewer.org/github/NanoComp/meep/blob/master/python/examples/adjoint_optimization/06-Near2Far-Epigraph.ipynb), and compared with a naive approach in  [tutorial-05](https://nbviewer.jupyter.org/github/NanoComp/meep/blob/master/python/examples/adjoint_optimization/05-Near2Far.ipynb), which simple optimized the average instead of the worst case.

doc/docs/Python_Tutorials/Adjoint_Solver.md
oskooi: True. I have updated this line to mention this even though this point is described [elsewhere in the user manual](https://meep.readthedocs.io/en/latest/Python_User_Interface/#near-to-far-field-spectra):

```
Since the "far" fields are computed using the full Green's functions, they 
should be able to be computed anywhere outside of the near-field surface monitor. 
The only limiting factor should be discretization errors but for any given distannce, 
the "far" fields should converge to the actual DFT fields at that location with resolution 
(assuming the distance separation is >> resolution).
```

doc/docs/Python_Tutorials/Adjoint_Solver.md
oskooi: Fixed.

doc/docs/Python_Tutorials/Adjoint_Solver.md
oskooi: I have updated the text with additional details based on the points raised in this discussion.

doc/docs/Python_Tutorials/Adjoint_Solver.md
oskooi: Acknowledged.

doc/docs/Python_Tutorials/Adjoint_Solver.md
smartalecH: (typo with `distannce` but otherwise LGTM)

doc/docs/Python_Tutorials/Adjoint_Solver.md
```diff
@@ -2,13 +2,715 @@
...
+across all six wavelengths (in the O-band for
+telecommunications). This is known as *minimax* optimization.
+
+The challenge with minimax optimization is that the $\max$ objective
+function is not everywhere differentiable. This property would seem to
+preclude the use of gradient-based optimization algorithms for this
+problem which involves 12 independent functions ($R$ and $1-T$ for
+each of six wavelengths). Fortunately, there is a workaround: the
+problem can be reformulated as a differentiable problem by introducing
+a dummy varaiable $t$ (the
```
stevengj, line 69:
```suggestion
a dummy variable $t$ (the
```

doc/docs/Python_Tutorials/Adjoint_Solver.md
```diff
@@ -2,13 +2,714 @@
...
+In this example, we use a minimum feature size of 150 nm for the line
+width *and* line spacing. The implementation of these constraints is
+based on [A.M. Hammond et al., Optics Express, Vol. 29, pp. 23916-38
+(2021)](https://doi.org/10.1364/OE.431188).
+
+There are five important items to note in the set up of the
+optimization problem:
+
+- The lengthscale constraint is activated only in the final epoch. The
+  initial design of the final epoch should be binary. This is because
```
stevengj, line 87:
```suggestion
- The lengthscale constraint is activated only in the final epoch. It is
   often helpful to mostly binarize the design before this final epoch. This is because
```

PR #2371, opened 2023-01-11
fix hankel overflow from high index/freq

hammy4815:
This should fix the issues with overflows from the hankel functions. I had to reformulate and manipulate a few things so I will explain in detail:

### DIFF
--- a/python/source.py
+++ b/python/source.py
@@ -812,7 +812,8 @@ class GaussianBeam2DSource(GaussianBeam3DSource):
 
     def get_fields(self, sim):
         """Calls green2d under various conditions (incoming vs outgoing) providing the correct Hankel functions and returns the fields at the slice provided by the source."""
-        from scipy.special import hankel1, hankel2, jv
+        from scipy.special import hankel1e, hankel2e, jve
+        from sys import float_info
 
         # Beam parameters
         freq = self.src.swigobj.frequency().real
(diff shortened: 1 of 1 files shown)


hammy4815:
https://github.com/hammy4815/meep/blob/2b7f79f138dc02bde3852fdae0af95ba12d0d8aa/python/source.py#L815

I am now using the exponential hankel functions and pass in the following wrappers that scale by the inverse of the exponential to correctly pass in a hankel function (or bessel at the origin):

https://github.com/hammy4815/meep/blob/2b7f79f138dc02bde3852fdae0af95ba12d0d8aa/python/source.py#L887-L889

In the exponential I am "shifting" by some amount `shift`, which really is just an extra factor being multiplied into the system. This is explained below:

https://github.com/hammy4815/meep/blob/2b7f79f138dc02bde3852fdae0af95ba12d0d8aa/python/source.py#L879-L884

I use `sys.float_info.max` to create a threshold. If the argument into the hankel functions ($kr$) has an imaginary component that is bigger (in magnitude) than this threshold, then I scale my entire system by a factor ($e^{-s}$ for scale s), to bring the magnitude smaller than what will overflow in the exponential. Without this extra `scale` component, reformulating with the exponential bessel functions would be useless because the overflow would instead still happen in my normalizing exponential term $e^{\pm kr}$ rather than in the old hankel functions.

If we want to use another way of deciding the threshold besides `sys.float_info.max` then I can change it.

https://github.com/hammy4815/meep/blob/2b7f79f138dc02bde3852fdae0af95ba12d0d8aa/python/source.py#L967-L968

I created this new function to minimize the amount of repeated code because I now need the term $kr$ in `get_fields` as well as `green2d`. 

https://github.com/hammy4815/meep/blob/2b7f79f138dc02bde3852fdae0af95ba12d0d8aa/python/source.py#L892-L893

I am now passing in my masks for `outgoing_arg`, `incoming_arg`, and `waist_points` before calling green2D to eliminate the overflows that were happening when the wrong hankel was being used for a given argument despite those indices of `X` not  being used. Before I was doing this while constructing the final `fields2D` array.

python/source.py
```diff
@@ -875,34 +876,74 @@ def get_fields(self, sim):
...
 
+        # Find large imaginary argument hankel shift
+        r, _ = self.get_r_rhat(X, X0)
+        kr = (k * r).flatten()
+        max_imag = np.abs(kr.imag[np.argmax(np.abs(kr.imag))])
+        max_float = int(np.log(float_info.max) / 2)
+        shift = 0 if np.abs(max_imag) < max_float else max_imag - max_float
+
+        # Fix hankel functions for large imaginary arguments and remove nans and infs when non-indexed
+        scaled_hankel1 = lambda o, kr: hankel1e(o, kr) * np.exp(1j * kr - shift)
```
hammy4815, line 887: I am now using the exponential hankel functions and pass in the following wrappers that scale by the inverse of the exponential to correctly pass in a hankel function (or bessel at the origin).

In the exponential I am "shifting" by some amount shift, which really is just an extra factor being multiplied into the system. This is explained below:

python/source.py
```diff
@@ -875,34 +876,74 @@ def get_fields(self, sim):
...
+        scaled_hankel2 = lambda o, kr: hankel2e(o, kr) * np.exp(-1j * kr - shift)
+        scaled_jv = lambda o, kr: jve(o, kr) * np.exp(np.abs(kr.imag) - shift)
```
hammy4815, line 889: I am now using the exponential hankel functions and pass in the following wrappers that scale by the inverse of the exponential to correctly pass in a hankel function (or bessel at the origin).

In the exponential I am "shifting" by some amount shift, which really is just an extra factor being multiplied into the system. This is explained below:

python/source.py
```diff
@@ -875,34 +876,74 @@ def get_fields(self, sim):
         )
         waist_xy = np.array([np.real(X0[0]), np.real(X0[1])])[:, np.newaxis, np.newaxis]
 
+        # Find large imaginary argument hankel shift
+        r, _ = self.get_r_rhat(X, X0)
+        kr = (k * r).flatten()
+        max_imag = np.abs(kr.imag[np.argmax(np.abs(kr.imag))])
+        max_float = int(np.log(float_info.max) / 2)
+        shift = 0 if np.abs(max_imag) < max_float else max_imag - max_float
```
hammy4815, line 884: I use `sys.float_info.max` to create a threshold. If the argument into the hankel functions ($kr$) has an imaginary component that is bigger (in magnitude) than this threshold, then I scale my entire system by a factor ($e^{-s}$ for scale s), to bring the magnitude smaller than what will overflow in the exponential. Without this extra `scale` component, reformulating with the exponential bessel functions would be useless because the overflow would instead still happen in my normalizing exponential term $e^{\pm kr}$ rather than in the old hankel functions.

If we want to use another way of deciding the threshold besides `sys.float_info.max` then I can change it.

python/source.py
```diff
@@ -923,23 +964,22 @@ def incoming_mask(self, x0, y0, kx, ky, X):
 
         return incoming_points, outgoing_points, waist_points
 
+    def get_r_rhat(self, X, X0):
+        """Returns r and rhat before normalizing rhat to be used in green2d and get_fields for overflow prediction"""
```
hammy4815, line 968: I created this new function to minimize the amount of repeated code because I now need the term $kr$ in `get_fields` as well as `green2d`.

python/source.py
```diff
@@ -875,34 +876,74 @@ def get_fields(self, sim):
...
+        scaled_hankel1 = lambda o, kr: hankel1e(o, kr) * np.exp(1j * kr - shift)
+        scaled_hankel2 = lambda o, kr: hankel2e(o, kr) * np.exp(-1j * kr - shift)
+        scaled_jv = lambda o, kr: jve(o, kr) * np.exp(np.abs(kr.imag) - shift)
+
         # Get field for outgoing points (hankel2)
-        o_fields2D = self.green2d(X, freq, eps, mu, X0, kdir, hankel2, beam_E0)[
-            :, :, :, np.newaxis
-        ]
+        o_fields2D = self.green2d(
+            X[:, outgoing_arg][..., np.newaxis],
```
hammy4815, line 893: I am now passing in my masks for `outgoing_arg`, `incoming_arg`, and `waist_points` before calling green2D to eliminate the overflows that were happening when the wrong hankel was being used for a given argument despite those indices of `X` not  being used. Before I was doing this while constructing the final `fields2D` array.

hammy4815:
> So this changes the scaling of the resulting fields?
> 
> I would suggest rescaling all the time, not just when `np.abs(max_imag) ≥ max_float`. The reason is that we generally want Meep's outputs to be a continuous function of its inputs, not jump discontinuously when you vary some parameter (e.g. a beam waist) pass some threshold.
> 
> (Unfortunately, this complicates the meaning of the beam amplitude parameter.)

The scaling only changes the order of magnitude of the Hankel amplitudes. Phase is preserved. In addition, after calling green 2D where needed, the final fields are normalized such that the beam waist is amplitude 1. Thus, the scaling only affects the algorithm’s intermediate magnitudes, preserving the magnitude of the final output. Thus I wouldn’t expect a discontinuity in the output.

stevengj (approved):

PR #2373, opened 2023-01-12
Additional results and updates to adjoint-optimization tutorial involving waveguide mode converter

oskooi:
Additional results including two new figures for the tutorial [Broadband Waveguide Mode Converter with Minimum Feature Size](https://meep.readthedocs.io/en/latest/Python_Tutorials/Adjoint_Solver/#broadband-waveguide-mode-converter-with-minimum-feature-size) originally added in #2365.


stevengj:
Would be good to check if the constraint is active (i.e. near zero) for the case with a lengthscale of 225 where the Wenchao's *a posteriori* measurement is returning 275.   If the constraint is active, then this may be limited by the accuracy of the constraint functions?

oskooi:
> Would be good to check if the constraint is active (i.e. near zero) for the case with a lengthscale of 225

Looks like the constraint function is *not* active for the line width and spacing in the final design because the constraint function values are both ~0.001 (i.e., positive). Because the constraint is inactive, this means that the optimizer therefore likely got trapped in a local optima with minimum feature size of 275 nm.

Separately, I have reverted the horizontal axis of the plot of the worst-case reflectance and transmittance back to the imposed constraint from the measured feature size.

doc/docs/Python_Tutorials/Adjoint_Solver.md
```diff
@@ -35,75 +34,86 @@ Python using [autograd](https://github.com/HIPS/autograd) as well as
 The adjoint solver supports inverse design and [topology
 optimization](https://en.wikipedia.org/wiki/Topology_optimization) by
 providing the functionality to wrap an optimization library around the
-gradient computation. This is demonstrated in several tutorials below.
+gradient computation. The adjoint solver also supports enforcing
+minimum feature sizes on the optimal design in 1D (line width and
+spacing) and 2D (circular holes and islands). This is demonstrated in
```
smartalecH, line 39: note that the 2D constraints are for arbitrary "enclosed shapes" (e.g. not restricted to circular).

doc/docs/Python_Tutorials/Adjoint_Solver.md
```diff
@@ -35,75 +34,86 @@ Python using [autograd](https://github.com/HIPS/autograd) as well as
...
+
+3. The edge of the design region is padded by a filter radius (rather
   than e.g., a single pixel) to produce measured lengthscales of the
   final design that are consistent with the imposed constraint.
 
-- The hyperparameters of the lengthscale constraint function (`glc` in
-  the script below), need to be chosen carefully to produce final
-  designs which do not significantly degrade the performance of the
-  unconstrained designs at the start of the final epoch.
+4. The hyperparameters of the lengthscale constraint function (`a1`,
```
smartalecH, line 110: Eventually we should add context here... like what each of these actually does

doc/docs/Python_Tutorials/Adjoint_Solver.md
oskooi: Fixed.

doc/docs/Python_Tutorials/Adjoint_Solver.md
oskooi: Acknowledged.

Issue #2379, opened 2023-01-21
Sources seem to change before and after optimization


Arjun-Khurana:
I am attempting to optimize a vertical grating coupler using a `Source` object defined as:

```python
sources = [
    mp.Source(
        src,
        component=mp.Ez,
        center = src_center,
        size = src_size,
        amp_func = my_func
    )
]
```

where `src` is a standard `GaussianSource` and `my_func` returns an amplitude corresponding to a linearly polarized fiber mode profile. I use this source to inject field into a design region and optimize the field in an output waveguide using an `EigenmodeCoefficient`. 

When I run my optimization, the first iteration (forward run, adjoint run, and gradient calculation) is fine, but when the second forward run starts, I get a segmentation fault. Looking more closely at the adjoint solver, I printed the type of the source objects before and after the optimization problem is called:

```python
if mp.am_really_master(): print(f'Sources (before opt): {[type(src) for src in opt.sim.sources]}')
f0, dJ_du = opt(vb) # returns dJ_du [design_region x design variables x frequency]
if mp.am_really_master(): print(f'Sources (after opt): {[type(src) for src in opt.sim.sources]}')
```

which results in:

```
Sources (before opt): [<class 'meep.source.Source'>]
Sources (after opt): [<class 'meep.source.EigenModeSource'>]
```

on the first iteration, and on the second iteration:

```
Sources (before opt): [<class 'meep.source.EigenModeSource'>]
<<SEGFAULT>>
```

The seg fault then occurs on the `opt` call in the next iteration. It seems like the `OptimizationProblem.adjoint_run()` method does not reset the sources back to the `forward_sources` list after finishing: this is left to the `prepare_forward_run()` method on the next iteration, but it seems like when `prepare_forward_run()` tries to use `change_sources()` to change the source list from the `EigenModeSource` back to my `Source`, it fails. 

Is this a bug or am I doing something incorrectly? Seems related to #2309, so I tried including `force_all_components=True` in the Simulation object and made sure my `eig_parity=mp.EVEN_Z + mp.ODD_Z`, but no luck.

smartalecH:
@Arjun-Khurana I'll address the segfault later. But first, be careful about how you are defining the spatial profile of your source. 

It's tempting to _try_ to excite a source with a particular mode profile by setting its spatial distribution to the expected field profile (as you are doing). But that won't work. You see, the source objects are _currents_ ($J$) not _fields_ ($E$, $H$). So you won't actually launch your mode this way. 

[...]

PR #2381, opened 2023-01-24
removed amp_func argument to add_eig_src method for EigenModeSource

Arjun-Khurana:
Reference issue: https://github.com/NanoComp/meep/issues/2379#issuecomment-1401324620

### DIFF
--- a/python/source.py
+++ b/python/source.py
@@ -671,7 +671,7 @@ class EigenModeSource(Source):
         if isinstance(self.eig_band, mp.DiffractedPlanewave):
             add_eig_src(self.amp_func, diffractedplanewave)
         else:
-            add_eig_src(self.amp_func)
+            add_eig_src()
 
 
 class GaussianBeam3DSource(Source):
--- a/python/source.py
+++ b/python/source.py
@@ -663,9 +663,8 @@ class EigenModeSource(Source):
             self.eig_resolution,
             self.eig_tolerance,
             self.amplitude,
+            self.amp_func
         ]
-        
-        if self.amp_func: add_eig_src_args.append(self.amp_func)
 
         if isinstance(self.eig_band, mp.DiffractedPlanewave):
             add_eig_src_args.append(diffractedplanewave)
--- a/python/source.py
+++ b/python/source.py
@@ -671,7 +671,7 @@ class EigenModeSource(Source):
         if isinstance(self.eig_band, mp.DiffractedPlanewave):
             add_eig_src(self.amp_func, diffractedplanewave)
         else:
-            add_eig_src()
+            add_eig_src(self.amp_func)
 
 
 class GaussianBeam3DSource(Source):
--- a/python/source.py
+++ b/python/source.py
@@ -664,14 +664,13 @@ class EigenModeSource(Source):
             self.eig_tolerance,
             self.amplitude,
         ]
-        add_eig_src = functools.partial(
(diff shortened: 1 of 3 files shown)


smartalecH:
cc @hammy4815

hammy4815:
Would this not prevent `amp_func` from being used at all in the case where an `EigenModeSource` lacks a `DiffractedPlanewave`? I'm not too familiar with how `amp_func` gets used under the hood, but  in `add_source`, `mp.Source` adds a volume source using `amp_func`:

https://github.com/NanoComp/meep/blob/d370f1aa603c5f9f74eae2690952e7a3b3d3cb0d/python/source.py#L153-L154

However, the `EigenModeSource` overloads the `add_source` method. The only other place where I see it fed into the C++ backend is here:

https://github.com/NanoComp/meep/blob/d370f1aa603c5f9f74eae2690952e7a3b3d3cb0d/python/source.py#L671-L674

Which is getting removed in this PR. But if it does get used somewhere else somehow, then I think this is fine as far as the new framework for `add_source` is concerned.

Arjun-Khurana:
Yeah this commit was just my quick and dirty solution for getting rid of the segfault in my example, something like this I think preserves the behavior in the case that an `amp_func` is supplied for an `EigenModeSource`:

```python
if isinstance(self.eig_band, mp.DiffractedPlanewave):
    add_eig_src(self.amp_func, diffractedplanewave) if self.amp_func else add_eig_src(diffractedplanewave)
else:
    add_eig_src(self.amp_func) if self.amp_func else add_eig_src()
```

However I'm not sure an `EigenModeSource` should really ever have a user-specified `amp_func`, since I thought the point of this type of source is that the spatial profile is deduced from the waveguide structure using MPB. If that's the case, I think we can remove the reference to `amp_func` entirely in the overload of `add_source`, if not I believe the above logic should work.

hammy4815:
This will probably work, and maybe we pull it for now. However, I'm wondering if we should fix the bigger issue going on, i.e. why after the adjoint run is the eigenmode source set up incorrectly with regards to `amp_func`? If we can instead patch the memory leak, it might be more optimal of a solution.

stevengj:
> However I'm not sure an EigenModeSource should really ever have a user-specified amp_func, since I thought the point of this type of source is that the spatial profile is deduced from the waveguide structure using MPB.

The user-specified `amp_func` is *multiplied* by the MPB amplitude, e.g. if you want to add some Gaussian envelope to get a localized window of an eigenmode: https://github.com/NanoComp/meep/blob/d370f1aa603c5f9f74eae2690952e7a3b3d3cb0d/src/mpb.cpp#L254

smartalecH:
The issue isn't with the `amp_func`, it's the NULL pointer. So let's just use #2394.

python/source.py
```diff
@@ -663,15 +663,13 @@ def add_source(self, sim):
             self.eig_resolution,
             self.eig_tolerance,
             self.amplitude,
+            self.amp_func
         ]
-        add_eig_src = functools.partial(
-            sim.fields.add_eigenmode_source, *add_eig_src_args
-        )
 
         if isinstance(self.eig_band, mp.DiffractedPlanewave):
```
smartalecH, line 669: We don't need to change this, so let's punt to the simpler PR.

smartalecH:
(we can always re-open if I'm wrong)

PR #2383, opened 2023-01-26
Fix bug for z-PML in cylindrical coordinates for `m` = 0, ±1

### LINKED ISSUE #2182: bug in z-PML in cylindrical coordinates at r=0 for `m`=0, ±1
Based on the discussion in [#2148 (comment)](https://github.com/NanoComp/meep/issues/2148#issuecomment-1200546285). 

In cylindrical coordinates, for a simulation with $m=\pm 1$ in vacuum with PMLs on all sides and an $E_r$ source at $r=0$, the $E_r$ fields at $r=0$ are not absorbed by the PMLs in the $z$ direction. The fields therefore bounce around indefinitely and the total flux emitted by the source measured using a closed three-sided box of DFT monitors never converges with run time.

More details are provided in this [gist](https://gist.github.com/oskooi/c5d2d9f0d2af342ed355abca8043c2eb?permalink_comment_id=4251314#gistcomment-4251314).

oskooi:
Closes #2182.

It turns out there is a bug in the $z$-PML at $r=0$ for $m=\pm 1$. This is due to the current setup of step-db which involves manually zeroing the $D_z$ fields at $r=0$:

https://github.com/NanoComp/meep/blob/d370f1aa603c5f9f74eae2690952e7a3b3d3cb0d/src/step_db.cpp#L342-L347

The $B_r$ fields had been omitted. This meant that the $z$-PML at $r=0$ was incorrect and thus the incident fields were bouncing around indefinitely as reported in #2182.

This PR also includes a new unit test based on the results in #2182.  The test verifies that the flux radiated by an $E_r$ point source at $r=z=0$ in vacuum (schematic below) converges with the simulation runtime. This test is currently failing on master.

cyl_plot2D

### DIFF
--- a/python/Makefile.am
+++ b/python/Makefile.am
@@ -83,6 +83,7 @@ TESTS =                                        \
     $(TEST_DIR)/test_multilevel_atom.py        \
     $(TEST_DIR)/test_n2f_periodic.py           \
     $(TEST_DIR)/test_oblique_source.py         \
+    $(TEST_DIR)/test_pml_cyl.py                \
     $(TEST_DIR)/test_physical.py               \
     $(TEST_DIR)/test_prism.py                  \
     $(TEST_DIR)/test_pw_source.py              \
--- a/python/tests/test_adjoint_cyl.py
+++ b/python/tests/test_adjoint_cyl.py
@@ -201,7 +201,7 @@ class TestAdjointSolver(ApproxComparisonTestCase):
         adj_scale = (dp[None, :] @ adjsol_grad).flatten()
         fd_grad = S12_perturbed - S12_unperturbed
         print(f"Directional derivative -- adjoint solver: {adj_scale}, FD: {fd_grad}")
-        tol = 0.2 if mp.is_single_precision() else 0.1
+        tol = 0.3
         self.assertClose(adj_scale, fd_grad, epsilon=tol)
 
 
--- a/python/tests/test_pml_cyl.py
+++ b/python/tests/test_pml_cyl.py
@@ -4,76 +4,92 @@ import numpy as np
 
 
 class TestPMLCylindrical(unittest.TestCase):
+    @classmethod
+    def setUp(cls):
+        cls.resolution = 25  # pixels/um
+        cls.s = 5.0
+        cls.dpml_r = 1.0
+        cls.dpml_z = 1.0
+        cls.cell_size = mp.Vector3(
+            cls.s + cls.dpml_r,
+            0,
+            cls.s + 2 * cls.dpml_z,
+        )
+        cls.fcen = 1.0
+
     def test_pml_cyl_coords(self):
         """Verifies that the z-PML in cylindrical coordinates properly
         attenuates fields at r=0.
         """
 
-        resolution = 25  # pixels/um
-        s = 5.0
-        dpml_r = 1.0
-        dpml_z = 1.0
-        cell_size = mp.Vector3(s + dpml_r, 0, s + 2 * dpml_z)
-
         pml_layers = [
-            mp.PML(dpml_r, direction=mp.R),
-            mp.PML(dpml_z, direction=mp.Z),
+            mp.PML(self.dpml_r, direction=mp.R),
+            mp.PML(self.dpml_z, direction=mp.Z),
         ]
-
-        fcen = 1.0
 
         sources = [
             mp.Source(
-                src=mp.GaussianSource(fcen, fwidth=0.1 * fcen),
+                src=mp.GaussianSource(self.fcen, fwidth=0.1 * self.fcen),
                 center=mp.Vector3(),
                 component=mp.Er,
             ),
         ]
 
         sim = mp.Simulation(
-            resolution=resolution,
-            cell_size=cell_size,
+            resolution=self.resolution,
+            cell_size=self.cell_size,
             dimensions=mp.CYLINDRICAL,
             m=-1,
-            accurate_fields_near_cylorigin=False,
             sources=sources,
             boundary_layers=pml_layers,
         )
 
         flux_plus_z = sim.add_flux(
-            fcen,
+            self.fcen,
             0,
             1,
             mp.FluxRegion(
-                center=mp.Vector3(0.5 * s, 0, 0.5 * s), size=mp.Vector3(s, 0, 0)
+                center=mp.Vector3(
+                    0.5 * self.s,
+                    0,
+                    0.5 * self.s,
+                ),
+                size=mp.Vector3(self.s, 0, 0),
             ),
         )
 
         flux_plus_r = sim.add_flux(
-            fcen,
-            0,
-            1,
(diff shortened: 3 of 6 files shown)


stevengj:
Doesn't the Yee grid for Bz put the Bz component at r=Δr/2?   Since it is at r≠0, why should it be set to zero?

oskooi:
> Doesn't the Yee grid for Bz put the Bz component at r=Δr/2? Since it is at r≠0, why should it be set to zero?

That's true but we are forcing the $B_r$ fields to be zero at $r=0$, not $B_z$.

stevengj:
No, still doesn't make sense to me — for m = ±1, the radial component *can* be nonzero at r=0.

stevengj:
If the flux in the R direction is not going to zero with increasing radial cell size, I'd like to see a picture of the fields for a large-radius simulation — how are the waves even reaching the R boundary.

oskooi:
> I'd like to see a picture of the fields for a large-radius simulation

Looks like the $E_r$ fields due to an $E_r$ source at $(r,z) = (0,0)$ are decaying to zero in the radial direction for an empty cell in cylindrical coordinates.

cyl_Er_vs_r_z0_s100_loglog

stevengj:
You could also look at Ez as a function of r for z ≠ 0 (since it =0 at z=0 by symmetry), since that is what creates the Poynting flux.   It would also be good to simply plot the Poynting flux (through a plane) vs r.

oskooi:
> You could also look at Ez as a function of r for z ≠ 0 (since it =0 at z=0 by symmetry), since that is what creates the Poynting flux. It would also be good to simply plot the Poynting flux (through a plane) vs r.

For an empty cell and a CW source, the steady-state $E_z$ and $P_r$ (flux in $r$ direction) at $z \neq 0$ both decay with $r$ as expected. 

$P_r$ decays *quadratically* with $r$ as is evident from the second log-log plot below.

cyl_Ez_vs_r_z0 1_t145

cyl_Pr_vs_r_z0 2

For reference, the scripts used to generate these results are provided in two gists: [1 (flux)](https://gist.github.com/oskooi/43391af3ac274c422e5081af374a0174) and [2 (fields)](https://gist.github.com/oskooi/7dcc58035e655d119dca5da771d31be8).

**Summary**

When using a *pulsed* source, the only way to ensure that $P_z$ (flux in the $z$ direction) at $r = 0$ for $m = \pm 1$ is converged with respect to the runtime is, as previously noted, to use a time profile with two properties: (1) narrow bandwidth (`fwidth` $=\Delta f  < ~0.1 f_{cen}$ ) and (2) long cutoff (`cutoff` $> ~10$ ). This requirement is necessary for the empty cell as well as the LED structure.

Unfortunately, this imposes a major computational cost because computing the flux for an $r \neq 0$ point source which requires multiple simulations for $m = 0, \pm 1, \pm 2, ...$ each with a long runtime means that the total runtime is necessarily long. Note that the poor $z$-PML performance is independent of the source location: $r = 0$ and $r \neq 0$ both require a long-duration pulse.

Unless we figure out how to improve the $z$-PML performance at $r = 0$ for $m = \pm 1$, it seems that using cylindrical coordinates may be nearly as costly as doing a single 3D simulation.

oskooi:
A useful reference from Prof. Douglas H. Werner's group at Penn State EE which provides the discretized equations for FDTD in cylindrical coordinates is [Radio Science, Vol. 48, pp. 232-47, (2013)](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1002/rds.20029).

The Yee grid shown in Figure 1 of this paper is identical to our Yee grid.

image

Section 2.2 "Singularity Issues of Electric and Magnetic Fields on the Axis of Symmetry" states:

"Due to the singularity of the field components ( $E_\phi$, $E_z$, $H_r$ ) located on the axis of symmetry, special boundary conditions are required for the update equations in the BOR (body of revolution) FDTD method. For instance, the $E_z$ component is zero for any mode with $m > 0$, while $E_\phi$ and $H_r$ are zero for any mode with $m \neq 1$ ."

This seems to be consistent with our implementation in [`step_db.cpp`](https://github.com/NanoComp/meep/blob/6abb8b3c6692fb17d08a6a5b36b320e1c54f9c55/src/step_db.cpp#L281-L411). This also means that the change proposed in this PR is unnecessary.

Additionally, Appendix B of the paper provides the update equations for the special case of $r=0$:

image

It would be good to compare these equations with our implementation in `step_db.cpp` to make sure we do not have a bug.

oskooi:
Looks like all that is required to fix this bug is to *remove* the special case of $r=0$ in the update $E$ from $D$ step. 

With this one change, the $z$-PML correctly attenuates fields at $r=0$ for simulations with $m=0, \pm 1$. There is no longer any need to use a narrow bandwidth pulse with long cutoff.

stevengj:
Can you check that the field output for Er and Ep looks okay at r=0 after this PR?  The comment on the code says this was originally for visualization, but maybe it was superseded by some later change.

oskooi:
> Can you check that the field output for Er and Ep looks okay at r=0 after this PR? 

The ($E_r$, $E_\phi$, $E_z$) fields at $r=0$ seem to be fine with this PR. 

Here are plots of the steady-state fields at $r=0$ from an $E_r$ CW point source at $(r,z)=(0,0)$ in vacuum for a $m=1$ simulation ([script](https://gist.github.com/oskooi/69581f94472e5547467f6235f26fcbee)).  The only difference in the results between this PR and master is the $E_\phi$ fields at $r=0$ (middle inset of the two figures). The fields from this PR seem to be correct whereas those from the master do not because this PR uses the correct $z$-PML at $r=0$. The $E_r$ and $E_z$ fields for this PR and master are the same.

Based on these results, I think we can go ahead and merge this PR.

cyl_Er_Ep_Ez_real_imag_PR2383

cyl_Er_Ep_Ez_real_imag_master


cyl_plot2D

oskooi:
An addendum in case the results in this PR are referenced in the future. It is important to note that $E_z$ at $r=0$ for a $m=1$ simulation should be zero. This is one of the special cases of the field-update equations shown above. However, the $E_z$ plots (right inset) of the previous comment show that $E_z$ is *nonzero* at $r=0$. This is not a bug. It is an artifact of `get_array` which *always* interpolates the fields to the center of the voxel in 2d. The fields at $r=0$ are therefore interpolated to $r=0.5\Delta r$. 

As a check, I verified that the raw values of the $E_z$ fields array exactly at $r=0$ within [`get_array_slice_chunk_loop`](https://github.com/NanoComp/meep/blob/742cf81861ddec3c198763bf3967b70128e9c8c8/src/array_slice.cpp#L251-L254) of `array_slice.cpp` is indeed zero. I did a similar check for (1) the $E_\phi$ fields at $r=0$ for a $m=0$ simulation and (2) the $E_\phi$ and $E_z$ fields at $r=0$ for a $m=2$ simulation.

PR #2390, opened 2023-02-03
Fix the shapes of 2d filters

mawc2019:
As mentioned [here](https://github.com/NanoComp/meep/pull/2016#issuecomment-1368121521), the 2d filters in `filters.py` should have kernels with circular boundaries, but these boundaries are rectangular in the current master branch. This PR makes them circular. In addition, this PR adds a `round` operation preceding each `int` operation in `filters.py` to avoid unexpected results, as mentioned [here](https://github.com/NanoComp/meep/discussions/2240#discussioncomment-4003077).

### DIFF
--- a/python/adjoint/filters.py
+++ b/python/adjoint/filters.py
@@ -26,7 +26,7 @@ def _proper_pad(arr, pad_to):
 
     Parameters
     ----------
-    arr : 2d input array.
+    arr : 2d input array representing the nonnegative-coordinate ≈1/4 of a filter kernel.
     pad_to : 1d array composed of two integers indicating the total size to be padded to.
     """
 
--- a/python/adjoint/filters.py
+++ b/python/adjoint/filters.py
@@ -20,32 +20,37 @@ def _centered(arr, newshape):
     myslice = [slice(startind[k], endind[k]) for k in range(len(endind))]
     return arr[tuple(myslice)]
 
+
 def _proper_pad(arr, pad_to):
-    '''
+    """
     Parameters
     ----------
     arr : 2d input array.
     pad_to : 1d array composed of two integers indicating the total size to be padded to.
-    '''
+    """
 
-    pad_size = pad_to-2*np.array(arr.shape)+1
+    pad_size = pad_to - 2 * np.array(arr.shape) + 1
 
-    top = np.zeros((pad_size[0],arr.shape[1]))
-    bottom = np.zeros((pad_size[0],arr.shape[1]-1))
-    middle = np.zeros((pad_to[0],pad_size[1]))
+    top = np.zeros((pad_size[0], arr.shape[1]))
+    bottom = np.zeros((pad_size[0], arr.shape[1] - 1))
+    middle = np.zeros((pad_to[0], pad_size[1]))
 
-    top_left = arr[:,:]
-    top_right = npa.flipud(arr[1:,:])
-    bottom_left = npa.fliplr(arr[:,1:])
-    bottom_right = npa.flipud(npa.fliplr(arr[1:,1:])) # equivalent to flip, but flip is incompatible with autograd
+    top_left = arr[:, :]
+    top_right = npa.flipud(arr[1:, :])
+    bottom_left = npa.fliplr(arr[:, 1:])
+    bottom_right = npa.flipud(
+        npa.fliplr(arr[1:, 1:])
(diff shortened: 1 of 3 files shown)


oskooi:
The pre-commit check is failing which likely means you forgot to first install the `pre-commit` package. 

See: https://meep.readthedocs.io/en/latest/Build_From_Source/#meep-for-developers

mawc2019:
I just installed the `pre-commit` and committed again, but the pre-commit check still fails. Does the pre-commit check run automatically when I do `git commit` on a branch, or should I somehow run it manually?

mawc2019:
The error in the failed test is as follows.
```
  File "/home/runner/work/meep/meep/build/meep-1.26.0-beta/_build/sub/python/../../../python/tests/test_adjoint_solver.py", line 685, in test_gradient_backpropagation
    self.assertClose(adj_dd, fnd_dd, epsilon=tol)
  File "/home/runner/work/meep/meep/build/meep-1.26.0-beta/python/tests/utils.py", line 30, in assertClose
AssertionError: 9.003547633008868e-06 not less than or equal to 8.717597578845472e-06 :
```
This error is related to the code below when single precision is used.
```
            if nfrq == 1 and frequencies[0] == self.fcen:
                tol = 2e-4 if mp.is_single_precision() else 5e-6
```
Increasing the tolerance `tol` from `5e-6` to `6e-6` would resolve the issue. Shall we do this?

stevengj:
Sure, a 5% increase in the tolerance is fine.

python/adjoint/filters.py
```diff
@@ -112,21 +131,15 @@ def cylindrical_filter(x, radius, Lx, Ly, resolution):
...
     x = x.reshape(Nx, Ny)  # Ensure the input is 2D
 
     xv = np.arange(0, Lx / 2, 1 / resolution)
     yv = np.arange(0, Ly / 2, 1 / resolution)
 
-    cylindrical = lambda a: np.where(a <= radius, 1, 0)
-    hx = cylindrical(xv)
-    hy = cylindrical(yv)
-
-    h = np.outer(_proper_pad(hx, 3 * Nx), _proper_pad(hy, 3 * Ny))
```
stevengj, line 126: I think we should still be able to call "1d" `_proper_pad` functions along each direction?

That is, when you pad along `x`, you don't need to "know" about the `y` direction — you just need the diameter of the kernel in `x`.  And vice versa for `y`.

python/adjoint/filters.py
```diff
@@ -112,21 +131,15 @@ def cylindrical_filter(x, radius, Lx, Ly, resolution):
...
-
-    # Normalize kernel
-    h = h / np.sum(h.flatten())  # Normalize the filter
+    X, Y = np.meshgrid(xv, yv, sparse=True, indexing="ij")
+    h = np.where(X**2 + Y**2 < radius**2, 1, 0)
```
stevengj, line 142: What happened to the normalization step?

python/adjoint/filters.py
```diff
@@ -36,6 +21,37 @@ def _centered(arr, newshape):
     return arr[tuple(myslice)]
 
 
+def _proper_pad(arr, pad_to):
```
oskooi, line 24: Would be good to add docstrings for all functions following these [guidelines](https://google.github.io/styleguide/pyguide.html#381-docstrings.

python/adjoint/filters.py
stevengj: Ah, it went to line 100

stevengj:
LGTM!

python/adjoint/filters.py
```diff
@@ -36,6 +21,37 @@ def _centered(arr, newshape):
...
+    """
```
stevengj, line 25:
```suggestion
    """Return a zero-padded/expanded version of filter arr (≈1/4 of kernel) to size pad_to, for convolution.

```

python/adjoint/filters.py
```diff
@@ -36,6 +21,37 @@ def _centered(arr, newshape):
...
+    Parameters
+    ----------
+    arr : 2d input array.
```
stevengj, line 28:
```suggestion
    arr : 2d input array representing the nonnegative-coordinate ≈1/4 of a filter kernel.
```

stevengj:
When you get a chance, try re-running the tutorials so that we can update the images (which will presumably change slightly).

PR #2392, opened 2023-02-07
Check for invalid sources at r=0 in cylindrical coordinates

oskooi:
Adds a check within ~~`fields::add_volume_source`~~ `src_vol_chunkloop` for ~~two~~ three different types of invalid sources at $r=0$ in cylindrical coordinates.

### DIFF
--- a/src/sources.cpp
+++ b/src/sources.cpp
@@ -273,7 +273,7 @@ static void src_vol_chunkloop(fields_chunk *fc, int ichunk, component c, ivec is
     amps_array[idx_vol] = IVEC_LOOP_WEIGHT(s0, s1, e0, e1, 1) * amp * data->A(rel_loc);
 
     // check for invalid sources at r=0 in cylindrical coordinates
-    if (fc->gv.dim == Dcyl && loc.r() == 0 && abs(amps_array[idx_vol]) > 0) {
+    if (fc->gv.dim == Dcyl && loc.r() == 0 && amps_array[idx_vol] != 0.0) {
       if (fc->m == 0 && (component_direction(c) == R || component_direction(c) == P))
         meep::abort("Not possible to place a %s source at r=0 in "
                     "cylindrical coordinates for m = 0.",
--- a/src/sources.cpp
+++ b/src/sources.cpp
@@ -284,7 +284,8 @@ static void src_vol_chunkloop(fields_chunk *fc, int ichunk, component c, ivec is
                     component_name(c));
       else
         meep::abort("Not possible to place a source at r=0 in "
-                    "cylindrical coordinates for m = %g.", fc->m);
+                    "cylindrical coordinates for m = %g.",
+                    fc->m);
     }
 
     /* for "D" sources, multiply by epsilon.  FIXME: this is not quite
--- a/src/sources.cpp
+++ b/src/sources.cpp
@@ -278,13 +278,13 @@ static void src_vol_chunkloop(fields_chunk *fc, int ichunk, component c, ivec is
         meep::abort("Not possible to place a %s source at r=0 in "
                     "cylindrical coordinates for m = 0.",
                     component_name(c));
-      else if (fabs(fc->m) == 1.0 && (c == Ez || c == Dz))
+      else if (fabs(fc->m) == 1.0 && component_direction(c) == Z)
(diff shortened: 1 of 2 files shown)


oskooi:
There is a single failing test in `test_adjoint_cyl.py` because of an $E_r$ line source extending in $r$ which *includes* $r=0$ for an $m=1.2$ simulation: 

https://github.com/NanoComp/meep/blob/b86b476c1b85426e18f973a69f24271cce4098be/python/tests/test_adjoint_cyl.py#L39-L42,

The current results therefore do not seem to be accurate because of the $\Delta r$ error that is being introduced by this source.

Perhaps we need to modify this test to exclude $r=0$ in the line source?

cc @mochen4

mochen4:
As Steven pointed out, the sources can be valid if the amplitude is 0 at $r=0$. The test looks good to me, though.

mochen4:
> This check is not reliable, because the user-specified amplitude function `A(x)` could be zero at `r=0`. But I don't know a good way to check that without sampling A at every point along the z axis?
> 

Don't we already have A at every point when we create src_vol to add the sources? Can we do the check here?https://github.com/NanoComp/meep/blob/601609f142e5dfb79be48ac4f69355d373e8b5df/src/sources.cpp#L273

oskooi:
> (Also, note that for m=0 the phi and r components should be zero at r=0.)

Added an additional check for this case.

> Don't we already have A at every point when we create src_vol to add the sources? Can we do the check here?

Good suggestion. I have moved the check from `add_volume_source` to `src_vol_chunkloop`.  With this modification, the case of a user-specified amplitude function is also included.

I tested these changes internally with a couple of tests. It seems to work correctly.

mochen4 (approved):

src/sources.cpp
```diff
@@ -272,6 +272,21 @@ static void src_vol_chunkloop(fields_chunk *fc, int ichunk, component c, ivec is
     vec rel_loc = loc - data->center;
     amps_array[idx_vol] = IVEC_LOOP_WEIGHT(s0, s1, e0, e1, 1) * amp * data->A(rel_loc);
 
+    // check for invalid sources at r=0 in cylindrical coordinates
+    if (fc->gv.dim == Dcyl && loc.r() == 0 && abs(amps_array[idx_vol]) > 0) {
```
stevengj, line 276:
```suggestion
    if (fc->gv.dim == Dcyl && loc.r() == 0 && amps_array[idx_vol] != 0.0) {
```

src/sources.cpp
```diff
@@ -272,6 +272,21 @@ static void src_vol_chunkloop(fields_chunk *fc, int ichunk, component c, ivec is
...
+      if (fc->m == 0 && (component_direction(c) == R || component_direction(c) == P))
+        meep::abort("Not possible to place a %s source at r=0 in "
+                    "cylindrical coordinates for m = 0.",
+                    component_name(c));
+      else if (fabs(fc->m) == 1.0 && (c == Ez || c == Dz))
```
stevengj, line 281:
```suggestion
      else if (fabs(fc->m) == 1.0 && component_direction(c) == Z)
```

stevengj: In practice, Hz is not stored at r=0 on the Yee grid, but this is conceptually clearer.

src/sources.cpp
```diff
@@ -272,6 +272,21 @@ static void src_vol_chunkloop(fields_chunk *fc, int ichunk, component c, ivec is
...
+        meep::abort("Not possible to place a %s source at r=0 in "
+                    "cylindrical coordinates for |m| = 1.0.",
+                    component_name(c));
+      else if (fabs(fc->m) > 1.0)
```
stevengj, line 285:
```suggestion
      else
```

src/sources.cpp
```diff
@@ -272,6 +272,21 @@ static void src_vol_chunkloop(fields_chunk *fc, int ichunk, component c, ivec is
...
+        meep::abort("Not possible to place a source at r=0 in "
+                    "cylindrical coordinates for |m| > 1.0.");
```
stevengj, line 287:
```suggestion
                    "cylindrical coordinates for m = %g.", fc->m);
```

PR #2394, opened 2023-02-09
memory leak fix; Py_XDCREF amp_func set to NULL

### LINKED ISSUE #2379: Sources seem to change before and after optimization

hammy4815:
closes #2379
closes #2381

As discussed, adding NULL assignments in `meep.i` after the calls to `Py_XDECREF` fixes the issue.

### DIFF
--- a/python/meep.i
+++ b/python/meep.i
@@ -714,6 +714,7 @@ meep::volume_list *make_volume_list(const meep::volume &v, int c,
 
 %typemap(freearg) double (*)(const meep::vec &) {
   Py_XDECREF(py_callback);
+  py_callback = NULL;
 }
 
 %typecheck(SWIG_TYPECHECK_POINTER) double (*)(const meep::vec &) {
@@ -738,6 +739,7 @@ meep::volume_list *make_volume_list(const meep::volume &v, int c,
 
 %typemap(freearg) std::complex<double> (*)(const meep::vec &) {
     Py_XDECREF(py_amp_func);
+    py_amp_func = NULL;
 }
 
 // Typemap suite for vector3
@@ -1050,6 +1052,7 @@ void _get_gradient(PyObject *grad, double scalegrad,
 
 %typemap(freearg) std::complex<double> (*)(const meep::vec &) {
     Py_XDECREF(py_amp_func);
+    py_amp_func = NULL;
 }
 
 %typecheck(SWIG_TYPECHECK_POINTER) PyObject *min_max_loc {
--- a/python/tests/test_source.py
+++ b/python/tests/test_source.py
@@ -4,7 +4,7 @@ import unittest
 
 import numpy as np
 from meep.geom import Cylinder, Vector3
-from meep.source import ContinuousSource, EigenModeSource, GaussianSource
+from meep.source import ContinuousSource, EigenModeSource, GaussianSource, Source
 
 import meep as mp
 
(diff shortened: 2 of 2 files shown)


smartalecH:
Nice! Can we add a quick test?

smartalecH:
(In your description, rather than saying "Ref issue" say "closes" so that a GitHub hook will close that issue after this is merged. You can close multiple issues/PRs that way so maybe include #2381)

smartalecH:
(cc @Arjun-Khurana)

hammy4815:
> Nice! Can we add a quick test?

Should the test simply create a Source with an amp_func and quick run, then change sources to an EigenModeSource with a quick run, then change back and ensure a quick run does not break?

stevengj:
First you need something that reliably reproduces the probelm on `master` … then strip it down to as quick a simulation as possible (e.g. set the resolution, cell size, and runtime, to be tiny).

stevengj:
Maybe add something to https://github.com/NanoComp/meep/blob/master/python/tests/test_source.py

Arjun-Khurana:
I wrote a unit test for this on [my fork](https://github.com/Arjun-Khurana/meep/blob/add_eigenmode_source/python/tests/test_amp_func.py).

All it does is spawn a child process, run the MWE from #2379 and assert that the exit code is 0, which is false if the segfault occurs.

#2381 also has these changes reflected and includes that unit test, but since it is poorly named relative to the actual implementation of the fix (and thus closed), you can feel free to include/modify that test in this PR @hammy4815 .

hammy4815:
Thanks @Arjun-Khurana ,

I'll take a look at your unit test and push a test here. I think we're planning to integrate this test into an existing in test_source.py to avoid making a new test file for this case. Once I'm done we can merge and close.

hammy4815:
The bug is fixed as we established, however, I believe something is missing in our understanding. Using the optimization object, we can recreate the issue with ease. To our understanding, this issue should still exist even without the use of an optimization. The following test *should* fail, yet does not. 

```python
 def test_amp_func_change_sources(self):
        src = ContinuousSource(5.0)
        center = Vector3()
        size = Vector3(0, 1, 0)

        ampfunc = lambda X: 1.0
        amp_source = [Source(src, component=mp.Ez, center=center, size=size, amp_func=ampfunc)]
        sim = mp.Simulation(cell_size=Vector3(1, 1, 0), resolution=5, sources=amp_source)
        sim.run(until=1)

        sim.restart_fields()
        sim.clear_dft_monitors()
        default_lattice = [EigenModeSource(src, size=size, center=center)]
        sim.change_sources(default_lattice)
        sim.run(until=1)

        sim.reset_meep()
        sim.change_sources(amp_source)
        sim.run(until=1)

        self.assertTrue((sim.sources[0].amp_func is ampfunc))
        self.assertTrue((sim.sources[0].amp_func is not None))
```

Here, when I call `sim.change_sources(default_lattice)`, the nested calls down to 

https://github.com/NanoComp/meep/blob/395d82104234ba1cf20b6443c5caa74e2962e0d2/python/source.py#L671-L674

gets called and `add_eig_src(self.amp_func)` does not delete the other amp_func like it does when it gets called from the optimization interface. 

From here we can either 1) try to find out what is happening in optimization to allow the trigger of the memory leak down the line, or 2) accept that the leak is happening, but we fixed it anyways with the Null assignments so make the unit test with an optimization object like what's done in Arjun's test.

hammy4815:
For whatever reason, the issue only occurs when the list of sources entering `change_sources` contains at least *two* EigenModeSources:

```python
        default_lattice = [EigenModeSource(src, size=size, center=center),
                           EigenModeSource(src, size=size, center=center)]
        sim.change_sources(default_lattice)
```

I will include this in the test, and then push to call this final, unless we want to explore this.

PR #2411, opened 2023-02-24
Update tutorial for third-order nonlinearity

### LINKED ISSUE #2408: Third Harmonic Generation Tutorial gives different results in current release?
Can somebody please shed some light on this?

[This example](https://github.com/NanoComp/meep/blob/master/python/examples/3rd-harm-1d.ipynb) should produce this spectrum:
3hg_meep1 20

Recent releases of Meep available via Conda produces this instead:
3hg_meep1 24
with a spurious peak around 0.74.

I have gone back to earlier releases and version 1.20 is the newest which gives the intended result.

oskooi:
Fixes #2408.

This PR also revamps the tutorial [Third Harmonic Generation](https://meep.readthedocs.io/en/latest/Python_Tutorials/Third_Harmonic_Generation/) by removing its shell-scripting features. All results and plots are now generated entirely using Python within a single script.


stevengj:
Shouldn't we automatically turn off decimation if the simulation contains a nonlinear material?  The current [default decimation factor](https://github.com/NanoComp/meep/blob/ccb3c372dc34ca75ce01f99c2b50e0e89ad0a203/src/dft.cpp#L208) assumes linearity.

oskooi:
> Shouldn't we automatically turn off decimation if the simulation contains a nonlinear material? 

Good suggestion. Let's address that change in a separate PR since it involves modifying additional files.

stevengj:
I created a draft PR https://github.com/NanoComp/meep/pull/2413 to replace this one.  Please check it out to make sure it fixes the problem.

oskooi:
Now that #2413 has been merged, there is no need to explicitly disable DFT decimation in the tutorial. The tutorial has therefore been updated with this change.

PR #2412, opened 2023-02-24
Connectivity tutorial update

mochen4:
Updated connectivity constraint tutorial for meep adjoint. Also changed the default values of hyper-parameters.


oskooi:
Looks there is are changes to the notebook file in `python/examples/adjoint_optimization/.ipynb_checkpoints` which was added in #2129. Perhaps we can remove this directory in a separate PR?

oskooi:
The updates look nice. Some comments:
1. It would be useful to cite a reference in the notebook tutorial and the docstrings (see comment 2) where the temperature model of the connectivity constraint is described in more detail.
2. Add descriptions for all arguments of the function `constraint_connectivity` including `cond_s`, `thresh` and `p` in `adjoint/connectivity.py` as part of its docstrings formatted according to https://google.github.io/styleguide/pyguide.html#s3.8.3-functions-and-methods.
3. Rename the title of the first section from `Behavior of the Constraint` to `How the Connectivity Constraint Works`.
4. In the first paragraph, consider rephrasing the line "Please also note that the connectivity constraint is independent from the rest of Meep, and may be used independently and  separately. " to something like "The connectivity constraint can be applied independent of the lengthscale constraint in Meep." with a link to the [tutorial involving minimum feature size](https://meep.readthedocs.io/en/latest/Python_Tutorials/Adjoint_Solver/#broadband-waveguide-mode-converter-with-minimum-feature-size).
5. typo in notebook: "Lenth-scale constraints"

mochen4:
Thanks Ardavan! I made changes based on your comments. Regarding the 4th comment:

> 4. In the first paragraph, consider rephrasing the line "Please also note that the connectivity constraint is independent from the rest of Meep, and may be used independently and  separately. " to something like "The connectivity constraint can be applied independent of the lengthscale constraint in Meep." with a link to the [tutorial involving minimum feature size](https://meep.readthedocs.io/en/latest/Python_Tutorials/Adjoint_Solver/#broadband-waveguide-mode-converter-with-minimum-feature-size).

My point was not that the connectivity can be applied independent of the length-scale constraint; rather, I wanted to say the the connectivity constraint is a stand alone feature, and to some extent can be regarded as a separate package and used for density-based topological optimizations with other Maxwell's solvers or even in other subject areas.

oskooi:
Thanks for the improvements. Two additional requests (optional):
1. Type hints for arguments and return values of the two functions `constraint_connectivity` and `cc_fd` in `adjoint/connectivity.py`.
2. Unit test for this feature. One suggestion: pass two different hard-coded 2d arrays, with and without disconnected/isolated features, to the constraint function and verify that it returns positive and negative values, respectively.

PR #2413, opened 2023-02-25
disable default decimation for nonlinear materials

### LINKED ISSUE #2408: Third Harmonic Generation Tutorial gives different results in current release?

stevengj:
See #2411.  Fixes #2408.

### DIFF
--- a/src/dft.cpp
+++ b/src/dft.cpp
@@ -204,10 +204,15 @@ dft_chunk *fields::add_dft(component c, const volume &where, const double *freq,
     double freq_max = 0;
     for (size_t i = 0; i < Nfreq; ++i)
       freq_max = std::max(freq_max, std::abs(freq[i]));
-    if ((freq_max > 0) && (src_freq_max > 0) && !has_nonlinearities())
+    if ((freq_max > 0) && (src_freq_max > 0) && !has_nonlinearities(false))
       decimation_factor = std::max(1, int(std::floor(1 / (dt * (freq_max + src_freq_max)))));
     else
       decimation_factor = 1;
+
+    // with add_srcdata sources, it's possible that not all
+    // sources are present on all chunks, leading us to over-estimate
+    // the allowed decimation_factor -- take minimimum to be sure:
+    decimation_factor = min_to_all(decimation_factor);
   }
   data.decimation_factor = decimation_factor;
 
--- a/src/dft.cpp
+++ b/src/dft.cpp
@@ -192,7 +192,9 @@ dft_chunk *fields::add_dft(component c, const volume &where, const double *freq,
   data.c = c;
   data.vc = vc;
 
-  if (decimation_factor == 0) {
+  if (decimation_factor == 0 && has_nonlinearities())
+    decimation_factor = 1;
+  else if (decimation_factor == 0) {
     double src_freq_max = 0;
     for (src_time *s = sources; s; s = s->next) {
       if (s->get_fwidth() == 0)
@@ -204,7 +206,7 @@ dft_chunk *fields::add_dft(component c, const volume &where, const double *freq,
     double freq_max = 0;
     for (size_t i = 0; i < Nfreq; ++i)
       freq_max = std::max(freq_max, std::abs(freq[i]));
-    if ((freq_max > 0) && (src_freq_max > 0) && !has_nonlinearities())
+    if ((freq_max > 0) && (src_freq_max > 0))
       decimation_factor = std::max(1, int(std::floor(1 / (dt * (freq_max + src_freq_max)))));
     else
       decimation_factor = 1;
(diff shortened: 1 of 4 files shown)


oskooi (approved):
LGTM.

After this PR is merged, I will update the docs, #2411, and the related unit tests where we manually disabled DFT decimation since this is no longer necessary.

oskooi:
For some reason, this PR is causing `python/tests/test_adjoint_solver.py` to stall both on CI and on my local workstation:

```
*** TESTING DFT OBJECTIVE ***
Starting forward run...
-----------
Initializing structure...
time for choose_chunkdivision = 9.9726e-05 s
Working in 2D dimensions.
Computational cell is 5 x 5 x 0 with resolution 30
     block, center = (0,0,0)
          size (1e+20,1,1e+20)
     block, center = (0,0,0)
          size (1.5,1.5,0)
time for set_epsilon = 0.0698151 s
-----------
MPB solved for frequency_1(2.2349,0,0) = 0.657635 after 15 iters
MPB solved for frequency_1(2.19109,0,0) = 0.645164 after 8 iters
MPB solved for frequency_1(2.19108,0,0) = 0.645161 after 1 iters
run 0 finished at t = 430.8333333333333 (25850 timesteps)
Starting adjoint run...
WARNING vol mismatch:, original_vol 2162, looped vol_sum 1081 
WARNING vol mismatch:, original_vol 2162, looped vol_sum 2185 
WARNING vol mismatch:, original_vol 2209, looped vol_sum 1128 
```

src/dft.cpp
```diff
@@ -204,10 +204,15 @@ dft_chunk *fields::add_dft(component c, const volume &where, const double *freq,
     double freq_max = 0;
     for (size_t i = 0; i < Nfreq; ++i)
       freq_max = std::max(freq_max, std::abs(freq[i]));
-    if ((freq_max > 0) && (src_freq_max > 0))
+    if ((freq_max > 0) && (src_freq_max > 0) && !has_nonlinearities(false))
```
oskooi, line 207: Calling `fields::has_nonlinearities(bool parallel)` with a `false` argument means that the `or_to_all` statement in `fields.cpp:670` (below) is never invoked. That would mean `fields::has_nonlinearities(bool parallel)` returns different values for different chunks. This would seem to be a bug.

stevengj: No, because of the subsequent `min_to_all` statement.

oskooi: If I replace `!has_nonlinearities(false)` with `!has_nonlinearities(true)`, `test_adjoint_solver` deadlocks when run with two chunks (as reported previously). This seems to be due to the `or_to_all` call which currently is never invoked. 

It's not obvious why `or_to_all` triggers a deadlock but `min_to_all` does not.

stevengj: The reason why it deadlocks is that `has_nonlinearities` here only gets executed if `(freq_max > 0) && (src_freq_max > 0)` is true, because `&&` is a [short-circuiting operator](https://stackoverflow.com/questions/45848858/what-is-short-circuit-evaluation-in-c).    

In the DFT adjoint simulation, where the adjoint sources get placed with `add_srcdata`, which is only called on chunks where the sources are present, some of the chunks are missing information about the sources and have `src_freq_max == 0`.   In consequence, those chunks are not calling `has_nonlinearities`, so it deadlocks if `parallel == true` and `or_to_all` is called.

In contrast, the `min_to_all` is always called on *every* chunk, because it is outside of these conditionals, so it does not deadlock.

Moreover, this means that the new `min_to_all` potentially fixes bugs even in linear simulations, since it was previously possible that some linear simulations over-estimated the `decimation_factor` on chunks for which no sources were present.

oskooi: I think it would have made the code more readable if, inside the block `if (decimation_factor == 0) { ... }` we had simply inserted:
```cpp
if has_nonlinearities()
   decimation_factor = 1
else {
  // set decimation_factor automatically...
}
```
The `bool` argument to `fields::has_nonlinearities()` is not necessary.

stevengj: The reason I added the `bool` argument is that we need to call `min_to_all` *anyway*, even for linear simulations, to make sure we don't over-estimate the `decimation_factor` in cases where a chunk is missing the source term present only in another chunk.   In that case, doing `or_to_all` is redundant, and is not cheap because it is a second MPI reduction.

PR #2419, opened 2023-03-01
Add function to create nlopt epigraph formulations.

yaugenst:
This PR adds a convenience function to create nlopt epigraph objectives from meep adjoint optimization problems.
I am using this myself currently and thought it would be a neat feature to have, as nlopt is being used in all the adjoint tutorials/examples anyways and it might make it easier for people to get started with epigraph problems.

I'm not sure if this functionality is out of scope for meep, but at least it doesn't require adding a dependency and should still be general enough to not restrict whatever one is trying to do with nlopt.

### DIFF
--- a/python/adjoint/utils.py
+++ b/python/adjoint/utils.py
@@ -4,7 +4,7 @@ import meep as mp
 import numpy as onp
 from autograd import tensor_jacobian_product
 
-from . import ObjectiveQuantity, OptimizationProblem
+from . import ObjectiveQuantity
 
 # Meep field components used to compute adjoint sensitivities
 _ADJOINT_FIELD_COMPONENTS = [mp.Dx, mp.Dy, mp.Dz]
@@ -250,7 +250,7 @@ def create_adjoint_sources(
 
 
 def get_epigraph_nlopt(
-    mpa_opt: OptimizationProblem,
+    mpa_opt,
     mapping: Callable,
     callback: Callable | None = None,
 ) -> tuple[Callable, Callable]:
--- a/python/adjoint/utils.py
+++ b/python/adjoint/utils.py
@@ -1,4 +1,4 @@
-from typing import Callable, Iterable, List, Tuple
+from typing import Callable, Iterable, List, Optional, Tuple
 
 import meep as mp
 import numpy as onp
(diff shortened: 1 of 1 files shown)


mrbaozi:
Sorry about that, the adjoint tests were failing due to circular imports caused by importing `OptimizationProblem` from `utils.py`. I removed the import as I was only using it for a type hint.

oskooi (commented):
Thanks for suggesting this new feature for the adjoint solver. The main limitation that I see with this PR (aside from the essentially new dependency on NLopt) is that it does not provide the flexibility to define a worst-case optimization problem via the epigraph formulation in a general way. It is based on just one specific implementation of the epigraph formulation.

As a specific example, when adding support for a minimum feature size (as demonstrated in e.g. the tutorial [Broadband Waveguide Mode Converter with Minimum Feature Size](https://meep.readthedocs.io/en/latest/Python_Tutorials/Adjoint_Solver/)), one approach is to introduce an additional constraint function for the minimum feature size which does *not* involve any Meep simulations (and only in the final "epoch".) However, the `_constraints` function returned by `get_epigraph_nlopt` invokes an `OptimizationProblem` object (which runs a forward and adjoint simulation) to obtain the objective function and its gradient. This seems like a major limitation to impose on users.

mrbaozi:
Hi, thanks for the review! I agree that NLopt is an "implicit" dependency - nothing in the function depends on NLopt, but of course it will be rather useless without it :smile:

Regarding the limitations - I agree that it does not cover all possible formulations of an epigraph problem. For instance, it also doesn't support multiple design regions at the moment. However, if a user is already thinking along those lines, I would expect them to be familiar enough to implement it on their own. This PR is mostly intended as a basic feature, not a full-blown epigraph toolbox.

Regarding your example, I'm not sure if I agree. The function here simply returns the epigraph dummy objective and the constraints that go along with it. It does not initialize optimization variables or set tolerances. The user still needs to actually set up the NLopt solver, and any additional constraints can still be added, I don't think there is any loss of generality there.

Of course the whole thing is just an idea, if it doesn't fit into Meep then it doesn't :smile:

stevengj:
This is a nice utility function.   Right now, however, our adjoint code is not really tied to any particular optimization package like NLopt.  It might be better to put a function like this into the adjoint-method tutorial that demonstrates the epigraph minimax formulation?

mrbaozi:
Alright, that sounds reasonable. Not sure how useful that would be in the end as I guess it doesn't really make a difference whether the tutorial defines the constraints directly or defines this function. Perhaps it helps a bit with clarity if it's a bit more decoupled from the actual problem. When I have some time I'll have a look at the tutorial and update it in a separate PR. For now, the function will continue to live [here](https://github.com/mrbaozi/photonic_design_examples/blob/main/phodex/optim/nlopt.py) then.

PR #2428, opened 2023-03-08
Add unit test for phase of mode coefficients

### LINKED ISSUE #2409: missing unit test for phase of mode coefficients
The existing unit tests for [`get_eigenmode_coefficients`](https://meep.readthedocs.io/en/latest/Python_User_Interface/#mode-decomposition) including [`test_binary_grating.py`](https://github.com/NanoComp/meep/blob/master/python/tests/test_binary_grating.py), [`test_mode_coeffs.py`](https://github.com/NanoComp/meep/blob/master/python/tests/test_mode_coeffs.py), [`test_mode_decomposition.py`](https://github.com/NanoComp/meep/blob/master/python/tests/test_mode_decomposition.py), [`test_special_kz.py`](https://github.com/NanoComp/meep/blob/master/python/tests/test_special_kz.py), and others all test *purely real* quantities such as the Poynting flux, e.g. $|S_{11}|^2$, etc., which involves the magnitude of the complex mode coefficient. Currently, there are no tests for just the imaginary part of the mode coefficients (or alternatively the real and imaginary parts tested separately).

A possible test could involve comparing the complex reflection and transmission coefficients of a 1d multilayer stack from Meep (2d and 3d simulations) to that computed analytically using a [transfer-matrix approach](https://en.wikipedia.org/wiki/Transfer-matrix_method_(optics)). (We do not have to write our own TMM solver as part of this test. Instead, we can use hard-coded results obtained using a third-party package such as https://github.com/sbyrnes321/tmm.)

oskooi:
Closes #2409.

Adds a missing unit test which validates the phase of the mode-decomposition feature. The test, involving a 2d simulation, is based on computing the phase of a total internal reflected (TIR) mode given an oblique-incident planewave for $S$ and $P$ polarizations. The results are validated using the Fresnel equations. More details are provided in #2409.

A tutorial based on this test will be added to [Tutorials/Mode Decomposition](https://meep.readthedocs.io/en/latest/Python_Tutorials/Mode_Decomposition/) in a separate PR.

### DIFF
--- a/python/tests/test_mode_decomposition.py
+++ b/python/tests/test_mode_decomposition.py
@@ -670,7 +670,7 @@ class TestModeDecomposition(unittest.TestCase):
 
     @parameterized.parameterized.expand(
         [
-            (Polarization.S, 54.3, 0.0),
+            (Polarization.S, 54.3, 0.4),
             (Polarization.P, 48.5, 1.2),
         ]
     )
--- a/python/tests/test_mode_decomposition.py
+++ b/python/tests/test_mode_decomposition.py
@@ -1,10 +1,13 @@
 import cmath
+from enum import Enum
 import math
+import parameterized
 import unittest
 
+import meep as mp
 import numpy as np
 
-import meep as mp
+Polarization = Enum("Polarization", "S P")
 
 
 class TestModeDecomposition(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


python/tests/test_mode_decomposition.py
```diff
@@ -665,6 +668,201 @@ def _pw_amp(x):
         self.assertAlmostEqual(Tsum, Tflux, places=3)
         self.assertAlmostEqual(Rsum + Tsum, 1.00, places=2)
 
+    @parameterized.parameterized.expand(
+        [
+            (Polarization.S, 54.3, 0.0),
```
stevengj, line 673: This should really be a nonzero distance.

We got a bit "lucky" here in getting the correct phase right at the interface, as explained in https://github.com/NanoComp/meep/issues/2409#issuecomment-1462703275, but I don't want to guarantee that for future versions of Meep, and it's not something I would encourage.  (It won't work for the transmitted phase, and it will give you the wrong amplitude in any case.)

oskooi: Fixed.

Issue #2192, opened 2022-08-11
Pade Spectral Extrapolation


Arjun-Khurana:
The following are some initial results for using Pade Approximant-based spectral extrapolation to reduce simulation time for resonant structures (introduced in [this issue](https://github.com/NanoComp/meep/issues/375#issue-330507605) and expanded upon [here](https://github.com/NanoComp/meep/issues/1497#issue-803911576).

The structure used is a simple ring resonator (as in [this issue](https://github.com/NanoComp/meep/issues/2183#issue-1324726574)). Full code is available [here](https://github.com/Arjun-Khurana/series-extrapolation.git).

The time series field data of the ring resonator are sampled (currently with a sampling rate of 1, so every point is considered. Future investigation will decimate the time series data to potentially reduce spatial complexity). 

These data are used as input to `scipy.interpolate.pade` (further work will investigate implementing a more robust or faster Pade algorithm, as this operation is rather slow).  These Pade coefficients are then used with a Fourier basis $e^{i \omega n \Delta t}$ with the same frequencies $\omega$ as the Meep DTFT calculation to construct a spectral extrapolation of the field data which is comparable to the Meep DTFT. 

Comparing the Meep DTFT calculation at 200 simulation time steps (about 5 periods of the ring resonator) with the Pade extrapolation at 200 time steps:

![extrapolation_t=200 0](https://user-images.githubusercontent.com/35874188/184214033-8b930149-b38c-46f6-b813-2110fbaa9633.png)

Clearly there is significantly less spectral leakage in the Pade extrapolation. At very low numbers of time steps (<3 periods of the ring): 

![extrapolation_t=100 0](https://user-images.githubusercontent.com/35874188/184214511-4cff6316-f3a8-473c-a792-073763091a12.png)

There is seemingly not much benefit to using the Pade extrapolation at these time.

Examining the long term convergence behavior of the Meep DTFT, I ran the simulation for 100 periods (4000 time steps) and calculated the DTFT at various points:

![dft-extension](https://user-images.githubusercontent.com/35874188/184214755-019c72e1-19bf-4597-b94b-f48155cfba58.png)

The DTFT has a small amount spectral leakage after approximately 75 periods, and minimal after 100 periods. Comparing the Pade extrapolation after 5 periods with the Meep DTFT after 100 periods:

![pade-winning](https://user-images.githubusercontent.com/35874188/184215363-c1e1aba5-039b-428a-a5cb-61c61af73d2a.png)

[...]

PR #2440, opened 2023-03-14
Pade approximant based spectral extrapolation

Arjun-Khurana:
After too much delay, this is an initial PR for Pade-approximant based spectral extrapolation, hopefully closing #2192. It essentially works in the same way as `Harminv` (thanks @Alex-Kaylor for help with the initial implementation) by aggregating field and computing an analytic frequency response using the rational Pade Approximant.

Like `Harminv`, `Pade` is constructed as a class that holds as a member variable (`data`) the aggregated field data, parameters for sampling the data, and the computed analytic function. 

To initialize, one need only provide a field component `c` and a point `pt` at which to perform the computation. 

Optional arguments include `sample_rate`, `start_time` and `stop_time`, which sample the collected field data according to `samples = self.data[self.start_time : self.stop_time : self.sample_rate]`. 

Additionally, there are `m` and `n`, which respectively set the orders of the numerator and denominator. Alternatively, one can specify `m_frac` or `n_frac`, which specify `m` and `n` as fractions of the length of the collected time series. If none of these are specified, the default behavior is to use half the length of the time series as the order of the numerator `m`, and `N-m-2` as the order of the denominator.

Speaking of the algorithm, the code is currently implemented by calling the builtin scipy `pade` routine. Separately, I have investigated the behavior of the [Trefethen SVD-based algorithm](https://doi.org/10.1137/110853236) and found it to be much harder to get it to converge (though in theory it is more robust to noise in the input data). Thus, the existing implementation only uses the builtin scipy algorithm, but it would be nice to at least have the option of using either algorithm. The way I have implemented the Trefethen algorithm [here](https://github.com/Arjun-Khurana/series-extrapolation/blob/master/trefethen.py) provides straightforward potential interchange of Trefethen and scipy. 

I have also written a test case that compares the center frequency of a peak in the Pade frequency response to the first Harminv mode, included in `test_ring.py`. This case passes when I run it locally with `python3 -m pytest python/tests/test_ring.py` but fails on my fork's CI workflow, I'm sure I'm doing something incorrectly there.

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -857,6 +857,142 @@ class EigenmodeData:
     def amplitude(self, point, component):
         swig_point = mp.vec(point.x, point.y, point.z)
         return mp.eigenmode_amplitude(self.swigobj, swig_point, component)
+    
+class Pade:
+    """
+    Pade is implemented as a class with a [`__call__`](#Pade.__call__) method from scipy,
+    which allows it to be used as a step function that collects field data from a given
+    point and runs [Pade](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.pade.html) 
+    on that data to extract the frequency response
+
+    See [`__init__`](#Pade.__init__) for details about constructing a `Pade`.
+
+    In particular, Pade takes the time series $f(t)$ corresponding to the given field
+    component as a function of time and decomposes it (within the specified bandwidth) as:
+
+    $$f(t) = \\sum_n a_n e^{-i\\omega_n t}$$
+
+    The results are stored in the list `Pade.dtft`, which is a list of tuples holding
+    the frequency, amplitude, and error of each frequency component. Given one of these tuples (e.g.,
+    `first_mode = harminv_instance.modes[0]`), you can extract its various components:
+
+    + **`freq`** — The real part of frequency ω (in the usual Meep 2πc units).
+
+    + **`decay`** — The imaginary part of the frequency ω.
+
+    + **`Q`** — The dimensionless lifetime, or quality factor defined as
+      $-\\mathrm{Re}\\,\\omega / 2 \\mathrm{Im}\\,\\omega$.
+
+    + **`amp`** — The complex amplitude $a$.
+
+    + **`err`** — A crude measure of the error in the frequency (both real and imaginary).
+      If the error is much larger than the imaginary part, for example, then you can't
+      trust the $Q$ to be accurate. Note: this error is only the uncertainty in the signal
+      processing, and tells you nothing about the errors from finite resolution, finite
+      cell size, and so on.
+
+    For example, `[m.freq for m in pade_instance.modes]` gives a list of the real parts
+    of the frequencies. Be sure to save a reference to the `Pade` instance if you wish
+    to use the results after the simulation:
+
+    ```py
+    sim = mp.Simulation(...)
+    h = mp.Pade(...)
+    sim.run(mp.after_sources(h))
+    # do something with h.modes
+    ```
+    """
+
+    def __init__(
+        self,
+        c: int = None,
+        pt: Vector3Type = None,
+        downsampling_rate: int = 1,
+        start_time: Optional[int] =0,
+        stop_time: Optional[int] = -1,
+        max_numerator_order: Optional[float] = None,
+        max_denom_order: Optional[float] = None
+    ):
+        """
+        Construct a Pade object.
+
+        A `Pade` is a step function that collects data from the field component `c`
+        (e.g. $E_x$, etc.) at the given point `pt` (a `Vector3`). Then, at the end
+        of the run, it uses Harminv to look for modes in the given frequency range (center
+        `fcen` and width `df`), printing the results to standard output (prefixed by
+        `harminv:`) as comma-delimited text, and also storing them to the variable
+        `Harminv.modes`. The optional argument `mxbands` is the maximum number of modes to
+        search for. Defaults to 100.
+        
+        m_frac is the order of the top polynomial with respect to thte bottom polynomial
+        """
+        self.c = c
+        self.pt = pt
+        self.downsampling_rate = downsampling_rate
+        self.max_numerator_order = max_numerator_order
+        self.max_denom_order = max_denom_order
+        self.start_time = start_time
+        self.stop_time = stop_time
+        self.data = []
+        self.data_dt = 0
+        self.modes = []
+        self.spectral_density = 1.1
+        self.Q_thresh = 50.0
+        self.rel_err_thresh = mp.inf
+        self.err_thresh = 0.01
+        self.rel_amp_thresh = -1.0
+        self.amp_thresh = -1.0
+        self.step_func = self._pade()
+
+    def __call__(self, sim, todo):
+        """
+        Allows a Pade instance to be used as a step function.
+        """
+        self.step_func(sim, todo)
+
+    def _collect_pade(self):
+        def _collect1(c, pt):
+            self.t0 = 0
+
+            def _collect2(sim):
+                self.data_dt = sim.meep_time() - self.t0
+                self.t0 = sim.meep_time()
+                self.data.append(sim.get_field_point(c, pt))
+
+            return _collect2
+
+        return _collect1
+
+
+    def _analyze_pade(self, sim):
+        # pade_cols = ["frequency", "imag. freq.", "Q", "|amp|", "amplitude", "error"]
+        # display_run_data(sim, "pade", pade_cols)
+
+        dt = sim.fields.dt
+        
+        samples = self.data[self.start_time:self.stop_time:self.downsampling_rate]
+        m = int(len(self.data)/2) if not self.max_numerator_order else self.max_numerator_order
+        
+        n = int(len(self.data)-m-1) if not self.max_denom_order else self.max_denom_order
+        
+        p,q = pade(samples, m, n)
+        
+        #TODO: make compatible with any monitor type (e.g. poynting flux, energy spectra)
+        #TODO: add decimation factor (comes with monitor type)
+        #TODO: print rho, number of zero singular values in SVD
+
+        pn = p.coef
+        qn = q.coef
+
+        P = lambda w: np.sum([pi * np.exp(1j * w * self.downsampling_rate*dt * n) for n,pi in enumerate(np.flip(pn))])
+        Q = lambda w: np.sum([qi * np.exp(1j * w * self.downsampling_rate*dt * n) for n,qi in enumerate(np.flip(qn))])
+
+            
+        return lambda freq: P(2*np.pi*freq)/Q(2*np.pi*freq)
+
+
+    def _pade(self):
+        return _combine_step_funcs(at_end(self._analyze_pade), self._collect_pade(self.c, self.pt))
 
 
 class Harminv:
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -925,10 +925,10 @@ class PadeDFT:
         + **`m` [`Optional[float]`]** — Directly pecifies the order of the numerator $P$. If not specified,
          defaults to the length of aggregated field data times `m_frac`.
         + **`n` [`Optional[float]`]** — Specifies the order of the denominator $Q$. Defaults
-         to length of field data - m - 1. 
-        + **`m_frac` [`float`]** — Method for specifying `m` as a fraction of 
-         field samples to use as order for numerator. Default is 0.5. 
-        + **`n_frac` [`Optional[float]`]** — Fraction of field samples to use as order for 
+         to length of field data - m - 1.
+        + **`m_frac` [`float`]** — Method for specifying `m` as a fraction of
+         field samples to use as order for numerator. Default is 0.5.
+        + **`n_frac` [`Optional[float]`]** — Fraction of field samples to use as order for
          denominator. No default.
         + **`sampling_interval` [`int`]** — Specifies the interval at which to sample the field data.
          Defaults to 1.
@@ -979,9 +979,13 @@ class PadeDFT:
 
         # Infer the desired behavior for m and n from the supplied arguments
         if not self.m:
-            self.m = int(len(samples)*self.m_frac)
+            self.m = int(len(samples) * self.m_frac)
         if not self.n:
-            self.n = int(len(samples)-self.m-1) if not self.n_frac else int(len(samples)*self.n_frac)
+            self.n = (
+                int(len(samples) - self.m - 1)
+                if not self.n_frac
+                else int(len(samples) * self.n_frac)
+            )
 
         # Compute the Padé approximant, store the poly1d instances
         self.p, self.q = pade(samples, self.m, self.n)
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -903,34 +903,31 @@ class Pade:
         self,
         c: int = None,
         pt: Vector3Type = None,
-        m: Optional[float] = None,
-        n: Optional[float] = None,
-        m_frac: float = 0.5,
-        n_frac: Optional[float] = None,
-        sample_rate: int = 1,
+        m: Optional[Union[int, float]] = None,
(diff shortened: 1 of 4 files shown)


smartalecH:
I would punt on the SVD approach. You can always add more PRs with additional functionality later. 

For now, focus on the simplest working routine that's easy to augment later.

smartalecH:
This is why the CI dies (you can download the log as an artifact):

```
ERROR: test_pade (__main__.TestRing)
  File "/home/runner/work/meep/meep/build/meep-1.27.0-beta/_build/sub/python/../../../python/tests/test_ring.py", line 77, in test_pade
    self.init()
  File "/home/runner/work/meep/meep/build/meep-1.27.0-beta/_build/sub/python/../../../python/tests/test_ring.py", line 49, in init
    self.p = mp.Pade(mp.Ez, mp.Vector3(r + 0.1), sample_rate=4)
AttributeError: module 'meep' has no attribute 'Pade'
```

Be sure to include your new module/function in `meep.i` (The swig file).

python/simulation.py
```diff
@@ -859,6 +864,155 @@ def amplitude(self, point, component):
...
+    p = mp.Pade(...)
+    sim.run(p, until=time)
+    # do something with p.freq_response
+    ```
+    """
+
+    def __init__(
+        self,
+        c: int = None,
+        pt: Vector3Type = None,
```
smartalecH, line 904: Any reason we can't do a `Volume` here instead?

python/tests/test_ring.py
```diff
@@ -70,6 +73,28 @@ def test_harminv(self):
         self.assertAlmostEqual(ep, 11.559999999999999, places=places)
         self.assertAlmostEqual(fp, -0.08185972142450348, places=places)
 
+    def test_pade(self):
+        self.init()
+
+        self.sim.run(
+            mp.at_beginning(mp.output_epsilon),
```
oskooi, line 80: Why is it necessary to output epsilon as part of the test?

python/tests/test_ring.py
```diff
@@ -70,6 +73,28 @@ def test_harminv(self):
...
+            self.h.fcen - self.h.df / 2, self.h.fcen + self.h.df / 2, 1000
+        )
+
+        freq_domain = [self.p.freq_response(freq) for freq in freqs]
+
+        idx = find_peaks(
+            np.abs(freq_domain) ** 2 / max(np.abs(freq_domain) ** 2), prominence=1e-4
+        )[0]
+
+        self.assertAlmostEqual(freqs[idx[0]], self.h.modes[0].freq, places=3)
```
oskooi, line 96: Is it possible to test for multiple spectral peaks rather than just one?

python/simulation.py
```diff
@@ -859,6 +864,155 @@ def amplitude(self, point, component):
...
+    See [`__init__`](#Pade.__init__) for details about constructing a `Pade`.
+
+    In particular, Pade stores the discrete time series $\\hat{f}[n]$ corresponding to the given field
+    component as a function of time and expresses it as:
+
+    $$\\hat{f}(\\omega) = \\sum_n \\hat{f}[n] e^{i\\omega n \\Delta t}$$
+
+    The above is a "Taylor-like" polynomial in $$n$$ with a Fourier basis and
+    coefficients which are the sampled field data. We then compute the Pade approximant
+    to the analytic form of this function as:
```
oskooi, line 883: typo: "...to the..." should be "...to be the..."

python/simulation.py
```diff
@@ -859,6 +864,155 @@ def amplitude(self, point, component):
         return mp.eigenmode_amplitude(self.swigobj, swig_point, component)
 
 
+class Pade:
+    """
+    Pade is implemented as a class with a [`__call__`](#Pade.__call__) method,
```
oskooi, line 869: Would be more accurate to write Pade as Padé in the docstrings.

python/simulation.py
```diff
@@ -859,6 +864,155 @@ def amplitude(self, point, component):
...
+    which allows it to be used as a step function that collects field data from a given
+    point and runs [Pade](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.pade.html)
+    on that data to extract an analytic rational function which models the frequency response.
```
oskooi, line 872: change "...which models the frequency response." to "...which approximates the frequency response."

Also, it would be useful to cite a reference on this topic such as the [wikipedia article](https://en.wikipedia.org/wiki/Pad%C3%A9_approximant) or a paper.

python/simulation.py
```diff
@@ -859,6 +864,155 @@ def amplitude(self, point, component):
...
+    $$\\hat{f}(\\omega) = \\sum_n \\hat{f}[n] e^{i\\omega n \\Delta t}$$
+
+    The above is a "Taylor-like" polynomial in $$n$$ with a Fourier basis and
+    coefficients which are the sampled field data. We then compute the Pade approximant
+    to the analytic form of this function as:
+
+    $$R(\\omega) = \\frac{P(\\omega)}{Q(\\omega)}$$
+
+    Where $$P$$ and $$Q$$ are polynomials of degree $$m$$ and $$n$$, and $$m + n + 1$$ is the
+    degree of agreement of the Pade Approximant to the analytic function $$f(\\omega)$$. This
```
oskooi, line 888: typo: "Approximant" should be "approximant"

python/simulation.py
```diff
@@ -859,6 +864,155 @@ def amplitude(self, point, component):
...
+        + **`c` [`Component`]** — Specifies the field component to use for extrapolation.
+         No default.
+        + **`pt` [`Vector3`]** — Specifies the location to accumulate fields. No default.
+        + **`m` [`Optional[float]`]** — Specifies the order of the numerator $$P$$. Defaults
+         to half the length of the sampled field data.
+        + **`n` [`Optional[float]`]** — Specifies the order of the denominator $$Q$$. Defaults
+         to length of sample data - m - 2.
+        + **`m_frac` [`float`]** — Alternative method for specifying `m` as a fraction of
+         field samples to use as order for numerator. Default 0.5.
+        + **`n_frac` [`Optional[float]`]** — Fraction of field samples to use as order for
```
oskooi, line 930: Suggestion: rather than have separate keyword arguments (`m`, `n`) and (`m_frac`, `n_frac`) which describe the same thing, it might improve the API to have just (`m`, `n`) and then within the constructor use the *type* of `m` and `n`  (either `int` `float` type) to infer how to define these two values.

python/simulation.py
```diff
@@ -859,6 +864,155 @@ def amplitude(self, point, component):
...
+         denominator. No default.
+        + **`sample_rate` [`int`]** — Specifies the rate at which to sample the field data. Default 1.
```
oskooi, line 932: Might be better to rename `sample_rate` to `sampling_interval`. A `sample_interval` of 1 (the default) therefore means collecting the field data at every timestep.

oskooi (changes requested):
Good work. We should add a tutorial for this feature after this is merged.

python/tests/test_ring.py
Arjun-Khurana: Probably not necessary. Out of curiosity, why does the `test_harminv` (in the same test file) do this?

python/tests/test_ring.py
Arjun-Khurana: It is possible, but the `scipy.signal.find_peaks` algorithm I used requires some coaxing to only find the same modes that `Harminv` finds, so it feels a bit contrived. Is there a better way to identify peaks and compare them to the `Harminv` modes?

python/simulation.py
Arjun-Khurana: resolved in https://github.com/NanoComp/meep/pull/2440/commits/6dd0939cac5f12cd11ece3b659ddd524f121f41e

python/simulation.py
Arjun-Khurana: https://github.com/NanoComp/meep/pull/2440/commits/6dd0939cac5f12cd11ece3b659ddd524f121f41e

python/simulation.py
Arjun-Khurana: https://github.com/NanoComp/meep/pull/2440/commits/6dd0939cac5f12cd11ece3b659ddd524f121f41e

python/simulation.py
Arjun-Khurana: https://github.com/NanoComp/meep/pull/2440/commits/6dd0939cac5f12cd11ece3b659ddd524f121f41e

python/simulation.py
Arjun-Khurana: Great suggestion, very elegant solution. This is updated in https://github.com/NanoComp/meep/pull/2440/commits/711c4500d1329368e9075d2b2aa16e08d7000320

python/simulation.py
Arjun-Khurana: Resolved in https://github.com/NanoComp/meep/pull/2440/commits/711c4500d1329368e9075d2b2aa16e08d7000320

python/simulation.py
smartalecH: Yep there is certainly a tradeoff with storage. But it's useful to be able to do this. One could simply pass a point volume if that's all they care about. 

For adjoint sensitivities, we'll have to do this across the entire design region.

python/tests/test_ring.py
oskooi: Perhaps we just need to use a different ring resonator? For example, the ring resonator in this [tutorial](https://meep.readthedocs.io/en/latest/Python_Tutorials/Basics/#modes-of-a-ring-resonator) contains three closely spaced resonant modes.

python/simulation.py
```diff
@@ -859,6 +864,168 @@ def amplitude(self, point, component):
...
+    """
+
+    def __init__(
+        self,
+        c: int = None,
+        pt: Vector3Type = None,
+        m: Optional[Union[int, float]] = None,
+        n: Optional[Union[int, float]] = None,
+        sampling_interval: int = 1,
+        start_time: Optional[int] = 0,
```
oskooi, line 909: The [`Optional`](https://docs.python.org/3/library/typing.html#typing.Optional) type is used when `None` is an option which is not the case here. Hence, the type should be just `int`.

oskooi (commented):
Getting there. Some more suggested changes.

oskooi:
You will also need to create a new entry for the `Pade` class in `doc/docs/Python_User_Interface.md.in` (similar to `Harminv`): 

https://github.com/NanoComp/meep/blob/c376694f8b094d9aa50499ee4667d4388f98f6dd/doc/docs/Python_User_Interface.md.in#L651-L655

This is necessary for converting its docstrings into Markdown for the readthedocs manual.

python/simulation.py
```diff
@@ -859,6 +864,168 @@ def amplitude(self, point, component):
...
+
+    def _analyze_pade(self, sim):
+        # Sample the collected field data and possibly truncate the beginning or
+        # end to avoid transients
+        samples = self.data[self.start_time : self.stop_time : self.sampling_interval]
+
+        # Infer the desired behavior for m and n from the types of the supplied arguments
+        if not self.m:
+            self.m = int(len(samples) / 2)
+        elif type(self.m) is float and (0 <= self.m <= 1):
```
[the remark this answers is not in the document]
Arjun-Khurana: I considered this and thought this type of error message might be confusing if a user supplies an argument which is `float` but intended to specify the `int` implementation of `m`, e.g. `128.0`. The combined error message lacks specificity in the type of error but provides detail about the full range of acceptable values.

stevengj:
In a separate PR, it would be good to show how to use this to calculate transmission and reflection spectra in problems with single-mode ports and sharp resonances.  For example, to compute the transmission/reflection spectrum through N layers of a 2d photonic crystal at normal incidence.

python/simulation.py
stevengj: This seems like a good feature to have. 

Your return value would then be a lambda that returns a numpy array of the spatial data (exactly in the shape of the `get_array` data.)

python/simulation.py
```diff
@@ -859,6 +864,168 @@ def amplitude(self, point, component):
         return mp.eigenmode_amplitude(self.swigobj, swig_point, component)
 
 
+class Pade:
```
stevengj, line 867: Maybe `PadeDFT`?

python/simulation.py
```diff
@@ -859,6 +864,168 @@ def amplitude(self, point, component):
...
+        self.c = c
+        self.pt = pt
+        self.m = m
+        self.n = n
+        self.sampling_interval = sampling_interval
+        self.start_time = start_time
+        self.stop_time = stop_time
+        self.data = []
+        self.data_dt = 0
+        self.freq_response: Callable = None
```
stevengj, line 945: Maybe `self.dft`

python/simulation.py
```diff
@@ -859,6 +864,168 @@ def amplitude(self, point, component):
...
+            pass
+        else:
+            raise TypeError(
+                "Order of denominator n must be positive integer or float between 0 and 1."
+            )
+
+        print(self.m, self.n)
+
+        # Compute the Padé approximant
+        p, q = pade(samples, self.m, self.n)
```
stevengj, line 998: Would be good to store these in instance variables.

python/simulation.py
```diff
@@ -859,6 +864,168 @@ def amplitude(self, point, component):
...
+            [
+                pi * np.exp(1j * w * self.sampling_interval * sim.fields.dt * i)
+                for i, pi in enumerate(p.coef)
+            ]
+        )
+        Q = lambda w: np.sum(
+            [
+                qi * np.exp(1j * w * self.sampling_interval * sim.fields.dt * i)
+                for i, qi in enumerate(q.coef)
+            ]
```
stevengj, line 1015: As long we are using numpy here, might as well vectorize this sum?

python/simulation.py
Arjun-Khurana: I have a quick and dirty
```python
p_arr = []
q_arr = []
for k in range(np.shape(samples)[-1]):
     p,q = pade(samples[:,k], self.m, self.n)
     p_arr.append(p)
     q_arr.append(q)
```
which works fine if the sample data is 2-dimensional (time x 1 spatial dimension), but not if the sample is higher dimensional. I tried using `np.apply_along_axis()` but got ragged array issues when the length of the numerator `poly1d` was different from the length of the denominator `poly1d`. Any insight on how to do this efficiently (hopefully not just a triple for loop over the volume)?

python/simulation.py
Arjun-Khurana: [resolved](https://github.com/NanoComp/meep/pull/2440/commits/19f434f0c3833401e348968b6ebe21a26855c0a0)

python/simulation.py
```diff
@@ -859,6 +864,184 @@ def amplitude(self, point, component):
...
+        # If the supplied volume is actually a point, np.atleast_2d will have
+        # shape (1, T), we want (T, 1)
+        if np.shape(self.data)[0] == 1:
+            self.data = self.data.T
+
+        # Sample the collected field data and possibly truncate the beginning or
+        # end to avoid transients
+        samples = np.array(
+            self.data[self.start_time : self.stop_time : self.sampling_interval]
+        )
```
oskooi, line 993: Let's add a TODO for enhancing performance by rewriting this section using a custom run function to collect only the necessary field points (we have all the machinery) rather than collecting the fields at every timestep.

python/simulation.py
```diff
@@ -859,6 +864,184 @@ def amplitude(self, point, component):
...
+
+        # Infer the desired behavior for m and n from the supplied arguments
+        if not self.m:
+            self.m = int(len(samples) * self.m_frac)
```
oskooi, line 997: Would be helpful to raise an exception (`ValueError`) if `self.m_frac` is not defined.

oskooi (changes requested):
Just two final comments/suggestions.

python/simulation.py
oskooi: For formatting, see: https://google.github.io/styleguide/pyguide.html#s3.12-todo-comments

PR #2452, opened 2023-03-23
Tutorial for nonaxisymmetric point-dipole sources in cylindrical coordinates

### LINKED ISSUE #2108: tutorial for non-axisymmetric sources in cylindrical coordinates
It would be nice to have a tutorial showing how to use non-axisymmetric sources in cylindrical coordinates by expanding the source in a exp(imφ) Fourier series.  For example:

* An off-axis planewave can be handled with the Jacobi–Anger expansion, as mentioned in #902.  Another example can be found in [this paper eqs (5–10)](https://math.mit.edu/~stevenj/papers/LinCa21.pdf) and many other papers.

* LDOS for off-axis points.  This is a point source, so when expanded in a Fourier series it is just a constant as a function of m.   Power orthogonality means that you can then just sum the LDOS's for different m's.

oskooi:
Closes #2108.

This is an initial and incomplete draft intended for early review and feedback.


doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
```diff
@@ -585,3 +585,26 @@ Note that the volume specified in `get_farfields` via `center` and `size` is in
...
+
+2. Each $m$-simulation is independent of the others. Each $m$-simulation can therefore be executed simultaneously using an embarassingly parallel approach.
+
+As a demonstration, we will repeat the calculation of the extraction efficiency of the LED using a point-dipole source at $r > 0$. The radiated flux of the point-dipole source (the numerator in the expression for the extraction efficiency) is computed using the Fourier--series-based method described above. A schematic of the simulation layout is shown below.
+
+![](../images/cyl_nonaxisymmetric_source_layout.png#center)
+
+![](../images/cyl_nonaxisymmetric_source_flux_vs_m.png#center)
+
+To convert the radiated flux at $r=0$ in cylindrical coordinates to 3d, the result must be multiplied by $2 / (\pi \Delta r)^2$ where $\Delta r$ is the grid size ($1/resolution$).
```
stevengj, line 606: There is a factor of 1/(2πr) missing in the current source in cylindrical coordinates, because that is Jacobian factor that should go into a cylindrical delta function — Meep uses δ(r), when it should really be δ(r)/(2πr).   This is then squared in the power.   Because of the Yee grid, an Er source at r=0 is actually placed at at Yee grid point at r=Δr/2.

The factor of 2 might come from a couple of different.  To get a linearly polarized source, you need (right+left circular)/2.  So there is a factor of 2 there.      There is also probably an interpolation weight: when you ask for a source at r=0, Meep will interpolate this onto the nearest grid points, so it might put 1/2 at Δr/2 and 1/2 at –Δr/2; however, because of the mirror symmetry in r, it's possible that the latter gets combined with Δr/2 and there ends up being a weight of 1; we'd have to check.

stevengj: Maybe add a print statement in `add_source` to see if the currents are added with weight 1 or 1/2.  (Or added twice with weight 1/2 each so that it totals to 1.)

doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
```diff
@@ -585,3 +585,26 @@ Note that the volume specified in `get_farfields` via `center` and `size` is in
...
+
+The relationship between the radiated flux from a point dipole at $r = 0$ ($P(0)$) and $r > 0$ ($P(r)$) is $(\pi / (\Delta r)^2) P(0) = P(r) / (\pi r)^2$.
```
stevengj, line 608: There are several sources of constant factors that we would have to track down in order to get this right.  
* If there is a 1/2 factor from interpolation at Δr/2 (above) from r=0 but not at r > 0, that would give a factor of 4.
* When you sum the Fourier series m, you are including both ±m already so that changes a factor of 2 compared to above.
* In order to convert from a point source at an r=0 to a Fourier series of "circle" sources, there are some normalization factors of 1/π and 1/2π that may potentially show up.  You have to go through the algebra carefully to get these right.

However, the (1/2πr)^2 factor from the δ(r) mentioned above should be the same for both r=Δr/2 and other values of r, so this should only give you factors of (r/Δr)^2 with no pi factors.

smartalecH:
This looks great, @oskooi. Nice job!

The idea here, of course, is that we can efficiently simulate the effects of a single dipole source *in 3D* using a few (or more than a few) relatively cheap cylindrical simulations. As we've discussed, it may not *actually* be that much cheaper (but hopefully it is, especially since it parallelizes much better!)

It might be worth it to have a bit of dialogue discussing this nuance. And maybe link to the other [tutorial](https://meep.readthedocs.io/en/latest/Python_Tutorials/Custom_Source/#exploiting-linear-time-invariance-of-the-materials-and-geometry) which exploits linear time invariance to compute the LEE of an entire structure using several dipoles. In other words, you can do various simulations at multiple r positions (each with multiple m numbers) and compute the LEE for the whole 3D device. 

That naturally begs the question: how easily we can adapt the *other* LEE tutorial ([using custom basis functions](https://meep.readthedocs.io/en/latest/Python_Tutorials/Custom_Source/#an-efficient-approach-for-large-numbers-of-dipole-emitters)) within cylindrical coordinates?

Finally, (as another todo) we could add a tutorial that does relatively simple adjoint optimization on the LEE. These results, together with @mochen4's LDOS adjoint should enable that (in fact, the LDOS saves us from having to do an adjoint solve, right, Mo?) There would be a lot going on with this kind of thing... but it would be rather interesting!

stevengj:
I don't like putting an empirical scale factor in the tutorial.  If we want to claim this is correct, someone should go through the algebra and derive it — that pi^5 factor looks weird.  The scale factor for r=0 is much simpler to explain (though even then more work is required to get the factor of 2).

Better to keep the tutorial focused on dimensionless quantities like extraction efficiency or Purcell enhancement so that these scale factors are irrelevant.

doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
```diff
@@ -585,3 +585,202 @@ Note that the volume specified in `get_farfields` via `center` and `size` is in
...
+Nonaxisymmetric Dipole Sources
+-------------------------------
+
+In [Tutorial/Local Density of States/Extraction Efficiency of a Light-Emitting Diode (LED)](Local_Density_of_States.md#extraction-efficiency-of-a-light-emitting-diode-led), the extraction efficiency of an LED was computed using an axisymmetric point-dipole source at $r = 0$. This involved a single simulation with $m = \pm 1$. Simulating a point-dipole source at $r > 0$ (as shown in the schematic below) is more challenging because it is nonaxisymmetric: any point source at $r > 0$ is equivalent to a ring source in cylindrical coordinates. In order to simulate a point-dipole source at $r > 0$, it is necessary to represent the $\delta(r)$ (Dirac delta function) current source as a Fourier-series expansion of $\phi$ (the field angular dependence $\exp(im\phi)$). This computational approach involves two parts: (1) performing a series of simulations for $m = 0, 1, 2, ..., M$ for some cutoff $M$ of the Fourier series (described below), and (2) because of power orthogonality, summing the results from each $m$-simulation in post processing where the $m \neq 0$ terms are multiplied by two to account for the $-m$ solutions.
```
stevengj, line 592: What you want is a source that is a 3d delta function:
```
δ(x)δ(y)δ(z) = δ(r)δ(φ)δ(z) / 2πr
```
and then to frame this in terms of axisymmetric simulations with exp(imφ) dependence you need to expand δ(φ) in a Fourier series
```
\delta(\phi) = \frac{1}{2\pi} \sum_m e^{im\phi}
```
up to a normalization.

smartalecH:
We mentioned that one possible reason for the discrepancy w.r.t source position (along `r`) could be related to how we are placing the (delta) source.

A few thoughts to add to this: we are currently placing the source in cylindrical coordinates the same as we do in cartesian. That is, we use a *restriction*, which is the transpose of a linear interpolation. However, this *same* restriction (i.e. using the same code) in cylindrical coordinates will place an unequal weight for different points along `r` (the `z` direction restriction should be fine... it's just the `r` direction).

We may need to scale the restriction using something like the bilinear transform (or maybe just transform the restriction using the Jacobian of the coordinate transform?).

doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
stevengj: Regarding power orthogonality:

* Physically, the total field E(x,y,z) is a sum of E_m(r,z)e^{im\phi} terms, one for the solution at each m (similarly for H).
* When you compute things like the total Poynting flux, however, you are integrating \Re [ E \times H^*] over a surface that includes an integral over φ.  In doing so, you'll see that all of the cross terms of different m's in E \times H^* integrate to zero (thanks to Fourier orthogonality), so you end up just getting a sum over the Poynting fluxes calculated separately for each m.   In short, the total power is the sum of the powers for each m.

stevengj:
@smartalecH, yes — see https://github.com/NanoComp/meep/blob/master/src/loop_in_chunks.cpp#L112-L113

doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
stevengj: Consider also the case of an x-polarized source, i.e. J proportional to `x̂`.  Then you have:
```
δ(x)δ(y)δ(z) x̂ = δ(r)δ(φ)δ(z) r̂ / 2πr
```
where the key point is that even though `r̂` is φ-dependent, because of the `δ(φ)` you are only evaluating `r̂` at φ=0 where it is the *same* as `x̂`.

This is different from point an `x̂` source at r=0, because then you write it with no `δ(φ)` as `x̂ = r̂ cos(ϕ) - ϕ̂ sin(ϕ)` and the cos and sin terms are why you do the calculation at m=1.

oskooi: I have expanded the discussion to address all of these points.

oskooi:
> Better to keep the tutorial focused on dimensionless quantities like extraction efficiency or Purcell enhancement so that these scale factors are irrelevant.

I have replaced the calculation of the dimensionful radiated flux with the dimensionless extraction efficiency.

Another useful demonstration that we may want to consider adding to this tutorial is computing the extraction efficiency for some narrow cone of angles (rather than all angles). See related discussion in #2368.

smartalecH:
> Another useful demonstration that we may want to consider adding to this tutorial is computing the extraction efficiency for some narrow cone of angles (rather than all angles).

Yes! As mentioned in that issue, a classic near2far makes this rather trivial. However, Our current near2far is overly expensive for this kind of calculation. The current implementation uses the Green's function to project to any arbitrary point in space. But what we care about (in this particular case) is the result as function *of angle*.

Now you *could* use some trig to convert your near2far response so long as it's sufficiently far... but this is needlessly complicated. When evaluating *infinitely far away*, the Green's function simplifies significantly resulting in a simple Fourier transform integral (with some nuances in cylindrical coordinates...). This integral is not only cheaper to evaluate, but directly provides the far field as a function on angle (no additional post processing). You could trivially implement this in python even (as done [here](https://github.com/flexcompute/tidy3d/blob/cc91d56e50be14c33e625a067894485e71356920/tidy3d/components/field_projection.py#L402)).

smartalecH:
> We mentioned that one possible reason for the discrepancy w.r.t source position (along r) could be related to how we are placing the (delta) source.

Here's an idea: rather than place a single *dipole* source, why not place a *line* source along the waveguide (really a circular plane source)? This way, only  *one* `r` weight is needed, and it's far from the center anyway (so it's error will be minimal).

Since we are already computing extraction efficiency, we might as well demonstrate it with the *basis function approach* (method 3) in the tutorials. For this case though, just simulate a *single* basis function (the DC term). The computational cost is the same as the current tutorial (for both the cylindrical and 3D case) but now the error is *effectively eliminated*.

If we have a discrepancy between the cylindrical and 3D case, then we know the primary error is *not* with the interpolation weights, but something else (although those weights should still be fixed eventually)

doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
```diff
@@ -589,60 +589,54 @@ Shown below is the far-field energy-density profile around the focal length for
...
 
-- To convert the radiated flux from a point dipole at $r=0$ in cylindrical coordinates to 3d, the result must be multiplied by $2 / (\pi \Delta r)^2$ where $\Delta r$ is the grid size ($1/resolution$).
+| `rpos` | **extraction efficiency** | **relative error** |  `M` (number of `m`-simulations) |
+|:------:|:-------------------------:|:------------------:|:--------------------------------:|
+|    0   |          0.310405         |        0.042       |                 1                |
+|   3.5  |          0.299407         |        0.041       |                56                |
+|   6.7  |          0.300334         |        0.036       |                101               |
+|   9.5  |          0.302234         |        0.028       |                141               |
```
smartalecH, line 619: Oof that's a lot of `m` simulations...

doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
```diff
@@ -693,102 +686,84 @@ def radiated_flux_cyl(dmat: float, h: float, rpos: float, m: int) -> float:
...
+        m,
+    )
+    ext_eff = flux_air / flux_src
+    print(f"exteff:, {rpos}, {ext_eff:.6f}")
+
+    # r > 0 source requires Fourier-series expansion of φ
+    flux_thresh = 1e-5  # threshold value for flux for truncating expansion
+    rpos = [3.5, 6.7, 9.5]
+    for rp in rpos:
+        cutoff_M = int(2 * rp * 2 * np.pi * fcen * n)
```
smartalecH, line 748: We don't really mention where this comes from in the tutorial description... at the very least a comment in the code would be useful.

oskooi: Added some notes in the main text to describe how this expression was derived.

doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
oskooi: In this example, I intentionally used a large value (`9.5`) for the source position in $r$ (which requires a good number of individual simulations as you have noted) simply to demonstrate convincingly that the approach involving a Fourier-series expansion works even when many `m`-simulations are involved. For most practical applications, it is probably unlikely that such a large value would be used.

Also, note that it is likely that I could have probably used fewer simulations without sacrificing accuracy in the converged results. The threshold flux value used to determine when to truncate the expansion is chosen somewhat arbitrarily as `flux_tol = 1e-5` in this tutorial. We can try increasing this value.

doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
```diff
@@ -585,3 +585,186 @@ Note that the volume specified in `get_farfields` via `center` and `size` is in
...
+
+![](../images/cyl_nonaxisymmetric_source_layout.png#center)
+
+A point-dipole source at $r > 0$ can be represented as a Dirac delta function in space: $\delta(r)\delta(\phi)\delta(z) / 2\pi r$. (The $2\pi r$ factor in the denominator is necessary to ensure correct normalization.) In order to set up such a source using only axisymmetric simulations, it is necessary to expand the $\delta(\phi)$ term as a Fourier series of $\phi$: $\delta(\phi) = \frac{1}{2\pi} \sum_m e^{im\phi}$. (The Fourier transform of a Dirac delta function is a [constant](https://en.wikipedia.org/wiki/Dirac_delta_function#Fourier_transform). Each spectral component has equal weighting in its Fourier-series expansion.)
```
stevengj, line 596:
```suggestion
A point-dipole source at $r_0 > 0$ can be represented as a Dirac delta function in space: $\delta(r - r_0)\delta(\phi)\delta(z) / r_0$. (The $r_0$ factor in the denominator is [necessary to ensure correct normalization](https://math.stackexchange.com/questions/398777/dirac-delta-in-polar-coordinates).) In order to set up such a source using only axisymmetric simulations, it is necessary to expand the $\delta(\phi)$ term as a Fourier series of $\phi$: $\delta(\phi) = \frac{1}{2\pi} \sum_m e^{im\phi}$. (The Fourier transform of a Dirac delta function is a [constant](https://en.wikipedia.org/wiki/Dirac_delta_function#Fourier_transform). Each spectral component has equal weighting in its Fourier-series expansion.)
```

doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
```diff
@@ -585,3 +585,186 @@ Note that the volume specified in `get_farfields` via `center` and `size` is in
...
+
+Simulating a point-dipole source involves two parts: (1) perform a series of simulations for $m = 0, 1, 2, ..., M$ for some cutoff $M$ of the Fourier-series expansion (the solutions for $\pm$ are simply complex conjugates, so we can omit $m <0$ and simply double the powers for $m > 0$ in the sum), and (2) because of power orthogonality, sum the results from each $m$-simulation in post processing where the $m \neq 0$ terms are multiplied by two to account for the $-m$ solutions. This procedure is described in more detail below.
```
stevengj, line 598:
```suggestion
Simulating a point-dipole source involves two parts: (1) perform a series of simulations for $m = 0, 1, 2, ..., M$ for some cutoff $M$ of the Fourier-series expansion (the solutions for $\pm m$ are simply complex conjugates, so we can omit $m <0$ and simply double the powers for $m > 0$ in the sum), and (2) because of power orthogonality, sum the results from each $m$-simulation in post processing where the $m \neq 0$ terms are multiplied by two to account for the $-m$ solutions. This procedure is described in more detail below.
```

doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
```diff
@@ -585,3 +585,186 @@ Note that the volume specified in `get_farfields` via `center` and `size` is in
...
+
+Simulating a point-dipole source involves two parts: (1) perform a series of simulations for $m = 0, 1, 2, ..., M$ for some cutoff $M$ of the Fourier-series expansion (the solutions for $\pm m$ are simply complex conjugates, so we can omit $m <0$ and simply double the powers for $m > 0$ in the sum), and (2) because of power orthogonality, sum the results from each $m$-simulation in post processing where the $m \neq 0$ terms are multiplied by two to account for the $-m$ solutions. This procedure is described in more detail below.
```
stevengj, line 598:
```suggestion
Simulating a point-dipole source involves two parts: (1) perform a series of simulations for $m = 0, 1, 2, ..., M$ for some cutoff $M$ of the Fourier-series expansion (the solutions for $\pm m$ are simply complex conjugates), and (2) because of power orthogonality, sum the results from each $m$-simulation in post processing, where the $m > 0$ terms are multiplied by two to account for the $-m$ solutions. This procedure is described in more detail below.
```

doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
```diff
@@ -585,3 +585,186 @@ Note that the volume specified in `get_farfields` via `center` and `size` is in
...
+
+A note regarding the source polarization at $r > 0$. The $\hat{x}$ polarization in 3d (the "in-plane" polarization) corresponds to the $\hat{r}$ polarization in cylindrical coordinates. An $\hat{r}$-polarized point-dipole source involves $\hat{r}$-polarized point sources in the $m$-simulations. Even though $\hat{r}$ is $\phi$-dependent, $\hat{r}$ is only evaluated at $\phi = 0$ because of $\delta(\phi)$. $\hat{r}$ is therefore equivalent to $\hat{x}$. This property does not hold for an $\hat{x}$-polarized point source at $r = 0$. A source at $r = 0$ does not involve $\delta(\phi)$. $\hat{x}$ is therefore defined in a more general way as $\hat{x} = \hat{r}\cos(\phi) - \hat{\phi}\sin(\phi)$. The $\sin$ and $\cos$ terms of this definition are why simulations for $m = \pm 1$ are necessary. See also [Tutorial/Scattering Cross Section of a Finite Dielectric Cylinder](#scattering-cross-section-of-a-finite-dielectric-cylinder).
```
stevengj, line 602:
```suggestion
A note regarding the source polarization at $r > 0$. The $\hat{x}$ polarization in 3d (the "in-plane" polarization) corresponds to the $\hat{r}$ polarization in cylindrical coordinates. An $\hat{r}$-polarized point-dipole source involves $\hat{r}$-polarized point sources in the $m$-simulations. Even though $\hat{r}$ is $\phi$-dependent, $\hat{r}$ is only evaluated at $\phi = 0$ because of $\delta(\phi)$. $\hat{r}$ is therefore equivalent to $\hat{x}$. This property does not hold for an $\hat{x}$-polarized point source at $r = 0$ (where $
\delta(\phi)$ is replaced by $1/2\pi$). A source at $r = 0$ does not involve $\delta(\phi)$. $\hat{x}$ is therefore defined in a more general way as $\hat{x} = \hat{r}\cos(\phi) - \hat{\phi}\sin(\phi)$. The $\sin$ and $\cos$ terms of this definition are why simulations for $m = \pm 1$ are necessary. See also [Tutorial/Scattering Cross Section of a Finite Dielectric Cylinder](#scattering-cross-section-of-a-finite-dielectric-cylinder).
```

stevengj:
Note that for an r=0 source, I don't think it should matter (modulo factors of 2) whether you do an rhat source or an rhat+iφhat source.

doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
```diff
@@ -585,3 +585,186 @@ Note that the volume specified in `get_farfields` via `center` and `size` is in
...
... r \omega$ where $\omega$ is the angular frequency of the source within the source medium. This relationship comes from expressing the maximum phase of the fields of a *guided* mode in terms of the wavevector of a planewave in free space: $M = kr$ where $k = n\omega_0/c$ is the light line in a medium with index $n$ and $\omega_0$ is the angular frequency in vacuum ($c = 1$ in Meep units). For $m > M$, the field oscillations tend to be too rapid and the current source radiates power into non-guided (i.e., free space) modes. As an example, a point-dipole source with wavelength of $1.0 \mu m$ at a radial position of $r = 1.0 \mu m$ within a medium of $n = 2.4$ would require roughly $M = 16$ simulations. (In practice, however, we can usually truncate the Fourier-series expansion earlier without significantly degrading accuracy whenever the radiated flux at some $m$ has dropped to some small fraction of its maximum value in the summation.) The plot below shows the radiated flux vs. $m$ for three different source positions used in this tutorial example. Generally, the farther the point source is from $r = 0$, the more simulations are required for the Fourier-series summation to converge.
```
stevengj, line 606:
```suggestion
1. Convergence of the Fourier series may require only a small number ($M + 1$) of simulations. For a given source position $r$, $M$ can be estimated analytically as $M \approx k r$ where $k = n\omega/c$ is the wavenumber of the source within the source medium. This comes from the fact that a source $\sim e^{im\phi}$ at $r$ oscillates in the angular direction with a spatial frequency $m/r$, but $m/r > k$ waves are evanescent, so for $m \gtrsim kr$ the radiated power tends to drop exponentially. As an example, a point-dipole source with wavelength of $1.0 \mu m$ at a radial position of $r = 1.0 \mu m$ within a medium of $n = 2.4$ would require roughly $M = 16$ simulations. (In practice, however, we can usually truncate the Fourier-series expansion earlier without significantly degrading accuracy whenever the radiated flux at some $m$ has dropped to some small fraction of its maximum value in the summation.) The plot below shows the radiated flux vs. $m$ for three different source positions used in this tutorial example. Generally, the farther the point source is from $r = 0$, the more simulations are required for the Fourier-series summation to converge.
```

doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
```diff
@@ -585,3 +585,186 @@ Note that the volume specified in `get_farfields` via `center` and `size` is in
...
+            1e-8,
+        ),
+    )
+
+    flux_air = mp.get_fluxes(flux_air_mon)[0]
+
+    if rpos == 0:
+        dV = np.pi / (resolution**3)
+    else:
+        dV = 2 * np.pi * rpos / (resolution**2)
```
smartalecH, line 717: Might be good to have a comment (or a note in the tutorial) that describes that this is the volume of the pixel (as briefly mentioned in the EE tutorial). However, there is an extra nuance with this approach (the pixel at r=0 is different than the rest of the grid).

smartalecH: (Maybe also explain where the $2\pi r$ comes from, and how to extend this to arbitrary sources (like a line).

oskooi:
> Note that for an r=0 source, I don't think it should matter (modulo factors of 2) whether you do an rhat source or an rhat+iφhat source.

Using the necessary bug fix in #2459, I verified for the case of a point source at $r=0$ with $m=1$ that the two polarizations $\hat{r}$ and $\hat{r}+i\hat{\phi}$ produce the *same* result for the extraction efficiency.

doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
oskooi: I am planning to create a dedicated page at `Features/Cylindrical Coordinates` after this PR is merged which describes this feature in more detail. This will include the Maxwell equations, Yee grid, special boundary conditions at $r = 0$, and scaling factors. This information is currently scattered in various tutorials, issues, and PRs. It would be good to collect everything into a single document.

smartalecH: Perfect, I'm happy to help.

PR #2462, opened 2023-04-06
Reverse engineer filtered and projected design variables

mochen4:
Sometimes, it is desirable to have some specific initial guess for the topological optimization. With the presence of filtering and projection function, we can do a mini-optimization to reverse engineer the design variables whose filtered projection matches the desired structure.


mochen4:
For a completely binary target structure:
image

The difference (target-mapping(target)) after mapping would be:
image

With the mini-optimization, the difference (target-mapping(optimized x)) looks like:
image

where mapping includes filter with radius=0.1 and projection with beta=8.

python/adjoint/reverse_design.py
```diff
@@ -0,0 +1,27 @@
+import nlopt
+from autograd import numpy as npa
+from autograd import grad
+
+
+def reverse_design(target, processing, maxiter=50):
```
oskooi, line 6: Would be nice to have docstrings and typehints.

python/adjoint/reverse_design.py
```diff
@@ -0,0 +1,52 @@
+import nlopt
```
stevengj, line 1: Maybe change this to a lazy import inside the function, so that NLopt is only required if you want to run this particular function.

stevengj:
Probably add a test.

python/tests/test_adjoint_solver.py
```diff
@@ -943,6 +943,38 @@ def test_mode_source_bandwidth(self):
                 self.assertLessEqual(rel_err, tol)
                 print(f"PASSED: fwidth={fwidth:.5f}, m={m}, err={rel_err:.10f}")
 
+    def test_reverse_design(self):
+        """Verifies that the reverse_design finds better initialization for specific
+        design than that design itself as initialization."""
+        print("*** TESTING reverse_design ***")
```
stevengj, line 949: This is too loose of a test.

I would do a [gold master test](https://codeclimate.com/blog/gold-master-testing) where you just verify that you reproduce the results of a previous run that you checked manually.

stevengj:
Should pass the new `periodic_axes` keyword from #2465 through to the filter functions.

python/tests/test_adjoint_solver.py
```diff
@@ -1060,6 +1060,25 @@ def test_periodic_design(self):
...
+        expected = np.load(os.path.join(data_dir, "mpa_reverse_design.npy"))
+        target = np.load(os.path.join(data_dir, "mpa_reverse_design_target.npy"))
+
+        def processing(x):
+            filtered_field = mpa.conic_filter(x, 0.1, 2, 2, 200)
+            projected_field = mpa.tanh_projection(filtered_field, 8, 0.5)
+            return projected_field.flatten()
+
+        reverse_x = mpa.reverse_design(target, processing)
+        tol = 1e-5
```
oskooi, line 1078: Is this value supposed to match the `ftol = 1e-5` in `reverse_design.py`?

python/adjoint/reverse_design.py
```diff
@@ -0,0 +1,53 @@
+from autograd import numpy as npa
+from autograd import grad
+from typing import Callable, List
+
+
+def reverse_design(target: List[float], processing: Callable, maxiter: int = 100):
```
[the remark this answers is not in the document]
mochen4: Good point. Thanks!

python/tests/test_adjoint_solver.py
mochen4: Not really. This value is just a small number to ensure future versions will have equally satisfying performance. I change it to a smaller value.

oskooi (approved):
LGTM.

PR #2470, opened 2023-04-12
Update integration weights to account for cylindrical coordinates

### LINKED ISSUE #2467: Implement proper integration weights in cylindrical coordinates
As discussed in #2452, the integration weights in cylindrical coordinates are implemented identically as those in cartesian coordinates (which is [incorrect](https://github.com/NanoComp/meep/pull/2452#issuecomment-1492366501)). The procedure to fix this is outlined in the code itself:

https://github.com/NanoComp/meep/blob/1fe38999997f1825054fc978e473327c77169671/src/loop_in_chunks.cpp#L112-L113

but it's not trivial to implement, particularly with cases at or crossing `r=0` (which happens a lot...)

Anecdotally, this doesn't seem to be an issue if R is rather large. However, for smaller cells, the error compounds substantially (and doesn't "easily" disappear with resolution).

smartalecH:
closes #2467.

This PR attempts to compute the proper integration weights for $r$ in cylindrical coordinates. There are a few things to be aware of:

1. @oskooi I don't think this fixes the issue with dipoles near $r=0$. Primarily because the interpolation formula for a dipole in cylindrical coordinates appears to be the *same* as Cartesian (at least according to the in-repo notes). So while this PR should help with some of the edge elements of monitors, it's not the fix we're looking for.
2. I added an `abort` if the user attempts to place a source or monitor that spans $r<0$. Physically, I don't even know what this means (a double source/monitor?) and it's a pain to deal with in the code (as described in the notes). So we punt.
3. Dealing with the weights right at $r=0$ is tricky. We don't want to divide by zero, so I include the `dV` term in the weight itself here. This required some additional modification to the `loop_in_chunks()` function, but doesn't seem to affect the `IVEC_LOOP_WEIGHT` macro (contrary to what the notes suggest) so I left that alone
4. Getting the indexing and resolution just right is a bookkeeping nightmare... so we should double check this.
5. The current test isn't a suitable test, since this doesn't fix that issue (plus it takes far too long to run with all of the `m` simulations sequentially, even for just two points relatively close to $r=0$.

I think this PR is more in the noise floor when it comes to comparing accuracy of a cylindrical simulation and a true 3D simulation with dipole sources.

@stevengj, since the issue isn't with source grid weights themselves, can you think of any other factors that would impact a dipole source in cylindrical coordinates?

### DIFF
--- a/python/tests/test_cyl_weights.py
+++ b/python/tests/test_cyl_weights.py
@@ -1,0 +1,145 @@
+"""Tests that the integration weights are computed properly in cylindrical coordinates.
+
+"""
+
+import unittest
+from typing import Tuple
+import meep as mp
+import numpy as np
+
+
+resolution = 80  # pixels/μm
+n = 2.4  # refractive index of dielectric layer
+wvl = 1.0  # wavelength (in vacuum)
+fcen = 1 / wvl  # center frequency of source/monitor
+
+
+def led_flux(dmat: float, h: float, rpos: float, m: int) -> Tuple[float, float]:
+    """Computes the radiated and total flux of a point source embedded
+       within a dielectric layer above a lossless ground plane.
+
+    Args:
+       dmat: thickness of dielectric layer.
+       h: height of dipole above ground plane as a fraction of dmat.
+       rpos: position of source in radial direction.
+       m: angular φ dependence of the fields exp(imφ).
+
+    Returns:
+       The radiated and total flux as a 2-Tuple.
+    """
+    L = 20  # length of non-PML region in radial direction
+    dair = 1.0  # thickness of air padding
+    dpml = 1.0  # PML thickness
+    sr = L + dpml
+    sz = dmat + dair + dpml
+    cell_size = mp.Vector3(sr, 0, sz)
+
+    boundary_layers = [
+        mp.PML(dpml, direction=mp.R),
+        mp.PML(dpml, direction=mp.Z, side=mp.High),
+    ]
+
+    src_pt = mp.Vector3(rpos, 0, -0.5 * sz + h * dmat)
+    sources = [
+        mp.Source(
+            src=mp.GaussianSource(fcen, fwidth=0.1 * fcen),
+            component=mp.Er,
+            center=src_pt,
+        ),
+    ]
+
+    geometry = [
+        mp.Block(
+            material=mp.Medium(index=n),
+            center=mp.Vector3(0, 0, -0.5 * sz + 0.5 * dmat),
+            size=mp.Vector3(mp.inf, mp.inf, dmat),
+        )
+    ]
+
+    sim = mp.Simulation(
+        resolution=resolution,
+        cell_size=cell_size,
+        dimensions=mp.CYLINDRICAL,
+        m=m,
+        boundary_layers=boundary_layers,
+        sources=sources,
+        geometry=geometry,
+    )
+
+    flux_air_mon = sim.add_flux(
+        fcen,
+        0,
+        1,
+        mp.FluxRegion(
+            center=mp.Vector3(0.5 * L, 0, 0.5 * sz - dpml),
+            size=mp.Vector3(L, 0, 0),
+        ),
+        mp.FluxRegion(
+            center=mp.Vector3(L, 0, 0.5 * sz - dpml - 0.5 * dair),
+            size=mp.Vector3(0, 0, dair),
+        ),
+    )
+
+    sim.run(
+        mp.dft_ldos(fcen, 0, 1),
+        until_after_sources=mp.stop_when_fields_decayed(
+            50.0,
+            mp.Er,
+            src_pt,
+            1e-8,
+        ),
+    )
+
+    flux_air = mp.get_fluxes(flux_air_mon)[0]
+
+    if rpos == 0:
+        dV = np.pi / (resolution**3)
+    else:
+        dV = 2 * np.pi * rpos / (resolution**2)
+
+    # total flux from point source via LDOS
+    flux_src = -np.real(sim.ldos_Fdata[0] * np.conj(sim.ldos_Jdata[0])) * dV
+
+    print(f"flux-cyl:, {rpos:.2f}, {m:3d}, {flux_src:.6f}, {flux_air:.6f}")
+
+    return flux_air, flux_src
+
+class TestCylWeights(unittest.TestCase):
+    def test_weights(self):
+        layer_thickness = 0.7 * wvl / n
+        dipole_height = 0.5
+
+        rpos = [0.5, 1.5]
+        flux_tol = 1e-5  # threshold flux to determine when to truncate expansion
+        normalized_flux = []
+        for rp in rpos:
+            # analytic upper bound on m based on coupling to free-space modes
+            # in light cone of source medium
+            cutoff_M = int(rp * 2 * np.pi * fcen * n)
+            ms = range(cutoff_M + 1)
+            flux_src_tot = 0
+            flux_air_tot = 0
+            flux_air_max = 0
+            for m in ms:
+                flux_air, flux_src = led_flux(
+                    layer_thickness,
+                    dipole_height,
+                    rp,
+                    m,
+                )
+                flux_air_tot += flux_air if m == 0 else 2 * flux_air
+                flux_src_tot += flux_src if m == 0 else 2 * flux_src
+                if flux_air > flux_air_max:
+                    flux_air_max = flux_air
+                if m > 0 and (flux_air / flux_air_max) < flux_tol:
+                    break
+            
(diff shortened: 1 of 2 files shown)


stevengj:
> Primarily because the interpolation formula for a dipole in cylindrical coordinates appears to be the same as Cartesian (at least according to the in-repo notes). 

I think that might be wrong — for a dipole source, what you want to do is not quite interpolation, I think.  Really you should think of it as a certain integral you need to match (e.g. a finite-width delta approximation you are trying to sample, such that the integral remains the same).

stevengj:
Would be good to check that it computes ∫ₐᵇf(r)rdr exactly for any affine f(r).  It should suffice to check that it integrates 1 and r exactly for various intervals [a,b]

smartalecH:
Hmm I'm still not convinced that there's a problem with the current implementation of a dipole in cylindrical coordinates. 

First, consider the problem in **cartesian coordinates**:

```
| -------o-----|
a        x     b 
```

where $a$ and $b$ are grid points, and $x$ is the location of our dipole in continuous space. The weights we use at grid points $a$ and $b$ we'll call $w_a$ and $w_b$ respectively.

Meep currently defines the weights using simple linear interpolation, such that:
$$\Delta x = b-a$$
$$w_a + w_b = 1$$
$$w_a=\frac{1}{\Delta x}\left(b-x\right)$$
$$w_b=\frac{1}{\Delta x}\left(x-a\right)$$

(The $\Delta x$ term isn't applied until the `IVEC_LOOP_WEIGHT` multiplies by `dV`).

Now, consider the problem in **cylindrical coordinates**:

```
| -------o-----|
a        r     b 
```

where the only difference here is $r$. But now, meep does a slightly different operation to determine the weights because of the `dV0` and `dV1` terms used to compute the $r$ in the integral:
$$w_a=\frac{1}{\Delta r}a\left(b-r\right)$$
$$w_a=\frac{1}{\Delta r}b\left(r-a\right)$$

Let's see what happens when we "integrate" over the weights by summing them together:
$$w_a + w_b$$
$$\frac{1}{\Delta r}a\left(b-r\right)+\frac{1}{\Delta r}b\left(r-a\right)$$
$$=\frac{b-a}{\Delta r} r$$
$$=r$$

...which is exactly what we want! Whereas in cartesian coordinates, we expect the integral of our delta function to sum to 1, we expect the integral in cylindrical coordinates for our delta function to sum to $r$, because we don't incorporate that factor by default in meep (and it's part of the definition of a delta function in cylindrical coordinates).

So again, I'm not sure the issue is with the weights of the dipole source...

smartalecH:
> So again, I'm not sure the issue is with the weights of the dipole source...

... and yet, experimentally, the code doesn't seem to be behaving as intended. I ran a simple test where I have a 1D cylindrical simulation, and I pull the weights of a dipole that I "slide" along $R$. Ideally, the sum of the weights should be $r$ itself... but there's a lot of noise...

```
import meep as mp
import numpy as np

sim = mp.Simulation(
    resolution=10,
    cell_size=mp.Vector3(100,0,0),
    dimensions=mp.CYLINDRICAL,
    m=1,
)
sim.init_sim()
r_bank = np.linspace(0.5,100,50)
w_bank = []
for r in r_bank:
    (x,y,z,w) = sim.get_array_metadata(mp.Volume(center=mp.Vector3(r,0,0),size=mp.Vector3(0,0,0)),)
    w_bank.append(w)


plt.figure()
plt.plot(r_bank,r_bank,color='k',label='ideal weights')
plt.plot(r_bank,w_bank, color='r',label='measured weights')
plt.xlabel("$r$ postion")
plt.ylabel("sum of weights")
plt.legend()
plt.savefig("weights.png")
plt.show()
```
weights

(Now I'm using `get_array_metadata()` to pull the weights, which automatically "collapses" the weights of a dipole by summing it together. There could be a bug here rather than the actual weight computation routine?)

oskooi:
As a useful reference, here is a schematic of ΔV in cylindrical coordinates from slide 2 of these [lecture notes](https://www.math.arizona.edu/~mgilbert/Math_223/Lecture_Notes/Section_16.5_Lecture_Notes_223.pdf):

image

oskooi:
As I previously commented in [#2108](https://github.com/NanoComp/meep/issues/2108#issuecomment-1445769513), I think there could be a bug in `fields::add_volume_source` related to the normalization of the $\delta$-function source in cylindrical coordinates:

https://github.com/NanoComp/meep/blob/aedc7949c0e4f816c5ee7d548db95b12990d078c/src/sources.cpp#L481-L484

The normalization is $\Delta r * \Delta z$ which is the same as 2D: $\Delta x * \Delta y$. A $\delta(r)$ source in cylindrical coordinates would seem to require an additional $r$ term in the normalization which is missing.

As a test, I modified the point source in [python/examples/point_dipole_cyl.py](https://github.com/NanoComp/meep/blob/aedc7949c0e4f816c5ee7d548db95b12990d078c/python/examples/point_dipole_cyl.py#L49-L55) to include `amplitude=1/rpos` for `rpos > 0`. However, this did not affect the plot of `flux_air_tot / rpos**2`  flux vs. `rpos`. The normalized radiated flux is still not a constant independent of `rp`.

smartalecH:
> The normalization is $\Delta r * \Delta z$ which is the same as 2D: $\Delta x * \Delta y$. A $\delta(r)$ source in cylindrical coordinates would seem to require an additional $r$ term in the normalization which is missing.

Wait, I thought this was already known (per our previous discussion). In fact, I thought this was *why* the test divides by $r^2$ in the first place (to account for the missing $r$ in the sources)?

This isn't a "bug", just a matter of convention. My above [comment](https://github.com/NanoComp/meep/pull/2470#issuecomment-1507726847) discusses this.

smartalecH:
@oskooi you can try this patch to see if this resolves the behavior you're seeing: `https://github.com/smartalecH/meep/tree/cyl_dipoles`

smartalecH:
> There are still rpos-dependent values for flux_air_tot / rpos**2 in the modified test based on point_dipole_cyl.py using your cyl_dipoles

Bummer.

> The extraction-efficiency values are slightly different compared to the master branch which is unexpected

Why is this unexpected? I would expect them to be different if the dipole is changing.

> Finally, three of the unit tests in python/tests are failing using the cyl_dipoles branch

Yes this is expected. I didn't write any logic to check for dipoles in the chunk code (which is much harder). I wanted to test if this fixed the issue first.

oskooi:
> Why is this unexpected? I would expect them to be different if the dipole is changing.

Since the extraction efficiency is a dimensionless quantity, scaling the amplitude of a point source by a constant factor should not affect its result. The scale factor would only affect dimensionful quantities such as the radiated flux (numerator) and source power (denominator) which indeed it does.

smartalecH:
> scaling the amplitude of a point source by a constant factor should not affect its result.

But the point of this is that the point source is *changing*. LEE is highly dependent on the type of source (position, size, polarization, etc). So the LEE should in fact change (otherwise, we wouldn't bother fixing the bug).

stevengj:
To be clear, there are two separate issues here:

* Do we define a cylindrical point source as δ(r) or δ(r)/r (or δ(r)/2πr)?  This is mostly a matter of convention.   As long as we document what we are doing, the user can easily multiply or divide by r as needed (in the source amplitude or in post-processing).   Currently, we define it as δ(r); we may want to change this in the future, but I suggest that we stick with the current convention for now.
* How do we discretize δ(r) onto the grid?  Currently, we are using just linear interpolation, but I think that introduces a first-order error in cylindrical coordinates.  That is what we would like to fix, and is independent of whether there is a 1/r in the amplitude convention.

smartalecH:
> How do we discretize δ(r) onto the grid? Currently, we are using just linear interpolation, but I think that introduces a first-order error in cylindrical coordinates. That is what we would like to fix, and is independent of whether there is a 1/r in the amplitude convention.

But from my comment [above](https://github.com/NanoComp/meep/pull/2470#issuecomment-1507726847), I think we already interpolating in cylindrical coordinates *correctly*.

oskooi:
> A symptom of the fact that the latter is just about discretization is that you should find that it goes away as you crank up the resolution. i.e. try your test from https://github.com/NanoComp/meep/issues/2108 at a higher resolution, and you should find that it approaches a constant at a smaller radius.

Looks like the discretization error for a point source $\delta(r-r_0)$ does *not* go away with increasing resolution. 

As a demonstration, here are results for dipole emission in vacuum (same test in #2108) plotting radiated flux / $r_0^2$ vs. $r_0$ (dipole position) for three different resolutions. This is using the master branch (not #2483). The results in the plot seem to be converged for the lowest resolution of 50 since there is no significant change at the higher resolutions.

freespace_cyl_flux_vs_rpos_tol5_master

Simulation layout with dipole at $r_0 = 0$.

cyl_sim_dpmlr1 0_dpmlz1 0

PR #2472, opened 2023-04-13
Document computation of extraction efficiency within angular cone in cylindrical coordinates

oskooi:
This PR expands [Tutorial/Nonaxisymmetric Dipole Sources](https://meep.readthedocs.io/en/latest/Python_Tutorials/Cylindrical_Coordinates/#nonaxisymmetric-dipole-sources) to include a description of the computation of the extraction efficiency within an angular cone using a near-to-far field transformation.


smartalecH:
So I know what is trying to be said. But I think there's a misunderstanding. 

We should still be able to sum far field *powers* from different $m$ simulations if the far field is measured and integrated around $\phi$. _Even_ if we only care about a subset of $\theta$ angles (eg a cone).

doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
```diff
@@ -618,6 +618,18 @@ As a demonstration, we compute the [extraction efficiency of an LED](https://mee
...
+
+The procedure for computing the extraction efficiency within an angular cone for a dipole source at $r > 0$ involves five steps:
+
+1. For each simulation in the Fourier-series expansion ($m = 0, 1, ..., M$), compute the far fields $\vec{E}_m$, $\vec{H}_m$ for $N$ points along the circumference of a quarter circle spanning the angular range [0°, 90°] with "infinite" radius (i.e., $R \gg \lambda$) using a near-to-far field transformation.
+2. Obtain the *total* far fields by summing the far fields from (1): $\vec{E}_{tot} = \vec{E}_{m=0} + 2\sum_{m=1}^M \Re\{\vec{E}_m\}$ and $\vec{H}_{tot} = \vec{H}_{m=0} + 2\sum_{m=1}^M \Re\{\vec{H}_m\}$.
```
stevengj, line 626: If you have a non-axisymmetric source, then this needs to be a function of φ as well — you need to include exp(imφ) terms in the sum.

PR #2476, opened 2023-04-20
Correct the weights for dipole sources in cylindrical coordinates

smartalecH:
As discussed, it's challenging to get the weights just right for a dipole source in cylindrical coordinates. This PR attempts to address that.

First, I modified `sources.cpp` to include the `r` factor from the `dV0` and `dV1` weight factors, just like what's implemented when processing DFT fields. Previously, this factor was completely omitted.

Second, for dipole sources, I include the `1/r` term that is necessary for a proper delta function in cylindrical coordinates.

Finally, I added some convenience functions that sum up all the source amplitudes (i.e. perform a sort of "integration") to check if what we're doing is right.

Using these three changes, we should be able to confirm that for a dipole source, it's integral should always be 1 (modulo the resolution scale factor) even for dipoles that lie between grid points. Previously, that wasn't the case. Now, it is. I ran a [simple experiment](https://gist.github.com/smartalecH/6bbd9ced25a18fb29c0a9798e159ee24) that sweeps a dipole between two grid points and sum up the corresponding weights. Indeed, it sums to 1 for all values of $r$ (there's some ambiguity here at $r=0$).

weights.

Currently, some of the tests are failing. This could be due to a few different factors. Before we tackle that, I want to ensure we get the weight formulation down first.

### DIFF
--- a/src/sources.cpp
+++ b/src/sources.cpp
@@ -270,7 +270,12 @@ static void src_vol_chunkloop(fields_chunk *fc, int ichunk, component c, ivec is
     loc += shift * (0.5 * inva);
 
     vec rel_loc = loc - data->center;
-    amps_array[idx_vol] = IVEC_LOOP_WEIGHT(s0, s1, e0, e1, 1) * amp * data->A(rel_loc);
+
+    if (fc->gv.dim == Dcyl)
+      amps_array[idx_vol] =
+          IVEC_LOOP_WEIGHT(s0, s1, e0, e1, dV0 + dV1 * loop_i2) * amp * data->A(rel_loc);
+    else
+      amps_array[idx_vol] = IVEC_LOOP_WEIGHT(s0, s1, e0, e1, 1) * amp * data->A(rel_loc);
 
     // check for invalid sources at r=0 in cylindrical coordinates
     if (fc->gv.dim == Dcyl && loc.r() == 0 && amps_array[idx_vol] != 0.0) {
@@ -479,7 +484,7 @@ void fields::add_volume_source(component c, const src_time &src, const volume &w
   data.A = A ? A : one;
   data.amp = amp;
   LOOP_OVER_DIRECTIONS(gv.dim, d) {
-    if (where.in_direction(d) == 0.0 && !nosize_direction(d)) // delta-fun
+    if (gv.dim != Dcyl && where.in_direction(d) == 0.0 && !nosize_direction(d)) // delta-fun
       data.amp *= gv.a; // correct units for J delta-function amplitude
   }
   sources = src.add_to(sources, &data.src);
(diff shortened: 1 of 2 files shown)


src/sources.cpp
```diff
@@ -270,7 +270,13 @@ static void src_vol_chunkloop(fields_chunk *fc, int ichunk, component c, ivec is
...
 
     vec rel_loc = loc - data->center;
-    amps_array[idx_vol] = IVEC_LOOP_WEIGHT(s0, s1, e0, e1, 1) * amp * data->A(rel_loc);
+
+    // in cylindrical coordinates, we need to account for the r term in the integral.
+    if (fc->gv.dim == Dcyl)
+      amps_array[idx_vol] =
+          IVEC_LOOP_WEIGHT(s0, s1, e0, e1, dV0 + dV1 * loop_i2) * amp * data->A(rel_loc);
+    else
+      amps_array[idx_vol] = IVEC_LOOP_WEIGHT(s0, s1, e0, e1, 1) * amp * data->A(rel_loc);
```
oskooi, line 279: This can block can be more cleanly expressed as:
```cpp
double interp_weight = fc->gv.dim == Dcyl ? dV0 + dV1 * loop_i2 : 1.0;
amps_array[idx_vol] = IVEC_LOOP_WEIGHT(s0, s1, e0, e1, interp_weight) * amp * data->A(rel_loc);
```

src/sources.cpp
```diff
@@ -755,4 +768,29 @@ diffractedplanewave::diffractedplanewave(int g_[3], double axis_[3], std::comple
...
+std::complex<double> fields::sum_sources(field_type ft) {
+  std::complex<double> total_sources = 0.0;
+
+  for (int i = 0; i < num_chunks; i++)
+    if (chunks[i]->is_mine()) total_sources += chunks[i]->sum_sources(ft);
+
+  return sum_to_all(total_sources);
+}
+
+std::complex<double> fields_chunk::sum_sources(field_type ft) {
```
oskooi, line 785: Is this function just used for debugging and was forgotten to be removed from the commit?

src/sources.cpp
```diff
@@ -478,9 +484,16 @@ void fields::add_volume_source(component c, const src_time &src, const volume &w
...
   data.A = A ? A : one;
   data.amp = amp;
+  // correct units for J delta-function amplitude
   LOOP_OVER_DIRECTIONS(gv.dim, d) {
-    if (where.in_direction(d) == 0.0 && !nosize_direction(d)) // delta-fun
-      data.amp *= gv.a; // correct units for J delta-function amplitude
+    if (where.in_direction(d) == 0.0 && !nosize_direction(d)) { // delta-fun
+      if (d == R && where.center().in_direction(d) > 0)
+        // in cylindrical coordinates, we need to scale by $r$ for a dipole.
+        data.amp *= 1.0 / where.center().in_direction(d);
```
oskooi, line 492: Why was the scaling by $\Delta r \times \Delta z$ removed?

src/sources.cpp
smartalecH: No I think we should keep it, as mentioned in the code comment above and the blurb above:

> Finally, I added some convenience functions that sum up all the source amplitudes (i.e. perform a sort of "integration") to check if what we're doing is right.

src/sources.cpp
smartalecH: See the code comment below:

```
        // we handle the resolution scaling at the chunk level for cylindrical coordinates.
```

src/sources.cpp
smartalecH: Sure, but let's make sure things work before we golf the code.

stevengj:
It might be worthwhile to think this through more carefully from a finite-element perspective (thinking of finite differences as equivalent to finite elements with first-order elements and a uniform grid, which is typically the case).  Then the normalization is an integral of the currents "tent functions" times r.

This also gives you a better prescription for what to do with an r=0 gridpoint, which in FEM would be "half a tent" function.

oskooi:
It seems this PR does not produce the expected result for the test in #2108. The radiated flux normalized by the dipole position $r^2$ is not a constant value independent or $r$. The results do not change with increasing resolution similar to the master branch as demonstrated in [#2470 (comment)](https://github.com/NanoComp/meep/pull/2470#issuecomment-1521157748).

freespace_cyl_flux_vs_rpos_PR2476

PR #2483, opened 2023-04-24
Improvements to tutorial on dipole sources in cylindrical coordinates

oskooi:
Two main improvements to the recent [Tutorial/Nonaxisymmetric Dipole Sources](https://meep.readthedocs.io/en/latest/Python_Tutorials/Cylindrical_Coordinates/#nonaxisymmetric-dipole-sources):

1. Mention that for simulation of a dipole at $r = 0$, only a *single* $\hat{r}$-polarized point source is necessary rather than a combination of $\hat{r}$ and $\hat{\phi}$ sources as would be expected theoretically. See also [#2452 (comment)](https://github.com/NanoComp/meep/pull/2452#issuecomment-1496880309).
2. In the simulation script, rather than perform $M$ simulations where $M$ is determined analytically as $kr$, perform as many simulations as necessary until the radiated flux has decreased by some threshold relative to its maximum value. This modification improves the convergence of the results since internal testing has revealed that using a predetermined $M$ may truncate the expansion too early and therefore produce inaccurate results.

Also includes some minor fixes and improvements to the rest of the text and script.


doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
```diff
@@ -597,9 +597,9 @@ A point-dipole source at $r_0 > 0$ can be represented as a Dirac delta function
...
+A note regarding the source polarization at $r > 0$. The $\hat{x}$ polarization in 3d (the "in-plane" polarization) corresponds to the $\hat{r}$ polarization in cylindrical coordinates. An $\hat{r}$-polarized point-dipole source involves $\hat{r}$-polarized point sources in the $m$-simulations. Even though $\hat{r}$ is in fact $\phi$-dependent, $\hat{r}$ is only evaluated at $\phi = 0$ because of $\delta(\phi)$. $\hat{r}$ is therefore equivalent to $\hat{x}$. This property does not hold for an $\hat{x}$-polarized point source at $r = 0$ (where $\delta(\phi)$ is replaced by $1/2\pi$): in that case, we write $\hat{x} = \hat{r}\cos(\phi) - \hat{\phi}\sin(\phi)$, and the $\sin$ and $\cos$ terms yield simulations for $m = \pm 1$. See also [Tutorial/Scattering Cross Section of a Finite Dielectric Cylinder](#scattering-cross-section-of-a-finite-dielectric-cylinder) which demonstrates setting up a linearly polarized planewave using a similar approach. However, in practice, a single $\hat{r}$-polarized point source at $r = 0$ is necessary for $m = \pm 1$.
```
stevengj, line 602:
```suggestion
A note regarding the source polarization at $r > 0$. The $\hat{x}$ polarization in 3d (the "in-plane" polarization) corresponds to the $\hat{r}$ polarization in cylindrical coordinates. An $\hat{r}$-polarized point-dipole source involves $\hat{r}$-polarized point sources in the $m$-simulations. Even though $\hat{r}$ is in fact $\phi$-dependent, $\hat{r}$ is only evaluated at $\phi = 0$ because of $\delta(\phi)$. $\hat{r}$ is therefore equivalent to $\hat{x}$. This property does not hold for an $\hat{x}$-polarized point source at $r = 0$ (where $\delta(\phi)$ is replaced by $1/2\pi$): in that case, we write $\hat{x} = \hat{r}\cos(\phi) - \hat{\phi}\sin(\phi)$, and the $\sin$ and $\cos$ terms yield simulations for $m = \pm 1$. See also [Tutorial/Scattering Cross Section of a Finite Dielectric Cylinder](#scattering-cross-section-of-a-finite-dielectric-cylinder) which demonstrates setting up a linearly polarized planewave using a similar approach. However, in practice, a single $\hat{r}$-polarized point source at $r = 0$ is necessary for $m = \pm 1$, because that gives a circularly polarized source that emits the same power as a linearly polarized source.
```

Issue #1734, opened 2021-08-17
Build PDF documentation.


lockywolf:
Would it be possible to add an option to build a PDF version of the documentation? My e-book reader is not good at showing webpages, and generating pdf files indirectly from html is also no lossless.

stevengj:
Hmm, I added [PDF output](https://docs.readthedocs.io/en/stable/downloadable-documentation.html) 
 to the readthedocs config (https://github.com/NanoComp/meep/commit/cd2a18b000e84cbb4e32422abcca7d8e4450aac0), but I'm not seeing a PDF link in the documentation and none of the suggestions in the stackoverflow discussion work (https://stackoverflow.com/questions/50425740/how-to-find-the-pdf-version-of-a-read-the-docs-project).   Any ideas?

oskooi:
@stevengj perhaps you also need to update the "Repository" setting in https://readthedocs.org/projects/meep/?

Note that if you navigate to https://meep.readthedocs.io/en/latest/, at the bottom right corner of this page you will see a button next to a book icon labeled "v: latest" which expands to reveal several options (none of which are PDF output). When the PDF output feature is finally working, this is where it will be found.

lockywolf:
FWIW, I have managed to create an improvised version of the documentation by using this plugin to mkdocs: [[https://github.com/orzih/mkdocs-with-pdf]] and tweaking `requirements.txt`

I am feeling that the font generated is too small, but at least I do have a pdf file now.

oskooi:
Unfortunately, `mkdocs` (which we are using to create the documentation for readthedocs) does not support building additional formats such as `pdf` or `epub` as described in https://docs.readthedocs.io/en/stable/config-file/v2.html#formats. (These formats are only supported by `sphinx`.)

PR #2505, opened 2023-05-06
Add configuration file for readthedocs

oskooi:
Adds a [configuration file](https://docs.readthedocs.io/en/stable/config-file/index.html) for readthedocs to the source tree. This should [improve reproducibility of the build environments](https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html) which hopefully fixes the random [build failures](https://docs.readthedocs.io/en/stable/faq.html#why-does-my-project-have-status-failing) we experience occasionally (such as the current failure). Note that the YAML file must have a `.yaml` rather than `.yml` extension (even though the other YAML files in the source tree use `.yml`).

Also, as described in [#1734 (comment)](https://github.com/NanoComp/meep/issues/1734#issuecomment-1537042784), `mkdocs` does *not* support building `pdf` or `epub` versions of the docs (only `html` is supported) and so the `formats` option is removed from `mkdocs.yml` because it does not do anything.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,13 @@
+version: 2
+
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.10"
+
+mkdocs:
+  configuration: mkdocs.yml
+
+python:
+   install:
+   - requirements: doc/requirements.txt
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -26,10 +26,6 @@ markdown_extensions:
 extra_javascript:
  - 'https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js?config=TeX-AMS_HTML'
 
-formats:
-  - epub
-  - pdf
-
 nav:
 - 'Manual': 'index.md'
 - 'Introduction': 'Introduction.md'


stevengj:
Still getting an error:
```
  File "/home/docs/.pyenv/versions/3.7.9/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/docs/.pyenv/versions/3.7.9/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/docs/checkouts/readthedocs.org/user_builds/meep/envs/latest/lib/python3.7/site-packages/mkdocs/__main__.py", line 16, in <module>
    from mkdocs import config                                 # noqa: E402
  File "/home/docs/checkouts/readthedocs.org/user_builds/meep/envs/latest/lib/python3.7/site-packages/mkdocs/config/__init__.py", line 2, in <module>
    from mkdocs.config.defaults import DEFAULT_SCHEMA
  File "/home/docs/checkouts/readthedocs.org/user_builds/meep/envs/latest/lib/python3.7/site-packages/mkdocs/config/defaults.py", line 1, in <module>
    from mkdocs.config import config_options
  File "/home/docs/checkouts/readthedocs.org/user_builds/meep/envs/latest/lib/python3.7/site-packages/mkdocs/config/config_options.py", line 7, in <module>
    from mkdocs import utils, theme, plugins
  File "/home/docs/checkouts/readthedocs.org/user_builds/meep/envs/latest/lib/python3.7/site-packages/mkdocs/theme.py", line 6, in <module>
    from mkdocs.utils import filters
  File "/home/docs/checkouts/readthedocs.org/user_builds/meep/envs/latest/lib/python3.7/site-packages/mkdocs/utils/filters.py", line 11, in <module>
    @jinja2.contextfilter
AttributeError: module 'jinja2' has no attribute 'contextfilter'
```

stevengj:
Same as https://github.com/readthedocs/readthedocs.org/issues/9064

stevengj:
In particular, we might try: https://github.com/wishopt/ReFramed/commit/a9338b6205e1050b8f12b3590f604baa8b2f1992

oskooi:
Looks like this broken dependency was fixed in [mkdocs 1.2.4](https://github.com/mkdocs/mkdocs/releases/tag/1.2.4) in Mar. 2022. However, we should be using version [1.4.3](https://github.com/NanoComp/meep/blob/master/doc/requirements.txt#L2) which does not require pinning `jinja2` as we are doing now. The build error could mean that `requirements.txt` is not actually being used for the build. To confirm this, we would need to know from the log what version of mkdocs is being used. My guess is it is likely < 1.2.4. 

(Note that this error is consistent with `.readthedocs.yaml` which pins the Python version to [3.10](https://github.com/NanoComp/meep/blob/28ee9f6bb35ad58c46ff4286339a5116356f085d/.readthedocs.yaml#L6) but yet the build log shows Python 3.7. The configuration files `.readthedocs.yaml` and `docs/requirements` checked into this repo both seem to be ignored.)

PR #2538, opened 2023-05-26
Special field updates for r = 0 and m = 0, ±1 in cylindrical coordinates

### LINKED ISSUE #2489: Radiation pattern of a dipole emitter does not converge with resolution in cylindrical coordinates
As part of a convergence analysis of the radiation pattern of a dipole emitter in cylindrical coordinates (based on the calculation procedure described in #2472), it seems the results do not converge with resolution. Not sure whether this is an indication of a bug somewhere or the way the calculation has been set up.

The test structure involves a point source at $r = 0$ in a finite-size "disc" (a modification of the structure in the [LED tutorial](https://meep.readthedocs.io/en/latest/Python_Tutorials/Local_Density_of_States/#extraction-efficiency-of-a-light-emitting-diode-led) which is infinitely extended). There are two different configurations of the closed box of near-field monitors (see schematics below). By Poynting's theorem, the two monitor configurations should produce *identical* results in the limit of infinite resolution. 

While the extraction efficiency over *all* angles computed using the Poynting flux is in fact identical (= 1.0), there are small but noticeable differences in the radiation pattern computed using a near-to-far field transformation. These discrepancies do not go away with increasing resolution.  The script used to generate the results is [here](https://gist.github.com/oskooi/07bf0800ef2c8146c7602c10b710f476).

Here is a side-by-side comparison of the extraction efficiency (really the fractional flux $P_\theta$ in the far field) within four angular cones at a resolution of 200. The final column is the absolute difference in the results (i.e., a measure of the error).

| **half angle of collection cone** | **Config. 1** ext. eff. | **Config. 2** ext. eff. | \|ee<sub>1</sub>-ee<sub>2</sub>\| |
|:---------------------------------:|:-----------------------:|:-----------------------:|:---------------------------------:|
|                10°                |          0.4539         |          0.4421         |               0.0118              |
|                20°                |          0.7451         |          0.7480         |               0.0029              |
|                30°                |          0.8683         |          0.8642         |               0.0041              |
|                40°                |          0.9143         |          0.9154         |               0.0011              |

Same results at a resolution of 400. The difference in the results (final column) is not decreasing as would be expected.

| **half angle of collection cone** | **Config. 1** ext. eff. | **Config. 2** ext. eff. | \|ee<sub>1</sub>-ee<sub>2</sub>\| |
|:---------------------------------:|:-----------------------:|:-----------------------:|:---------------------------------:|
|                10°                |          0.4524         |          0.4399         |               0.0125              |
|                20°                |          0.7420         |          0.7446         |               0.0026              |
|                30°                |          0.8644         |          0.8601         |               0.0043              |
|                40°                |          0.9110         |          0.9118         |               0.0008              |

**Configuration 1**

led_disc_cyl_plot2D_res200_1

*resolution = 200*

led_disc_cyl_radpattern_h0 5_res200_polar_1

*resolution = 400*

led_disc_cyl_radpattern_h0 5_res400_polar_1

**Configuration 2**

led_disc_cyl_plot2D_res200_3

*resolution = 200*

led_disc_cyl_radpattern_h0 5_res200_polar_3

*resolution = 400*

led_disc_cyl_radpattern_h0 5_res400_polar_3

oskooi:
Fixes #2489.

While #2383 fixed a bug for the $z$-PML at $r = 0$ for $m = \pm 1$, it seems to have introduced a separate bug for the $m = 0$ case. This is because #2383 actually [never tested](https://github.com/NanoComp/meep/pull/2383#issuecomment-1432614706) the $m = 0$ case which is why it went undetected until #2489. 

This PR reverts the change to `update_eh.cpp` in #2383 but *only* for $m = 0$. It is not yet clear to me why this fixes the issue reported in #2489 but the test results suggests it does.

A new unit test for $m = 0$ (currently missing) based on the results in #2489 will also be added to this PR in a subsequent commit.

Note: two test cases in `test_adjoint_cyl.py` involving $m = 0$ required adjusting the tolerance of their error check.


oskooi:
Added a new function `step_update_EDHB0` in `step_generic.cpp` based on the [suggestion](https://github.com/NanoComp/meep/issues/2489#issuecomment-1568734230) from @stevengj. This seems to fix the problem with one caveat.

As shown below, this fix produces the correct results whereby the DFT fields for $E_r$ and $E_z$ match the near-to-far fields for three test cases of $m = 0, 1, 2$. Results shown for $E_r$ only.

src_er_mon_er_dft_vs_n2f_m0_dair1 0_res100

src_er_mon_er_dft_vs_n2f_m1_dair1 0_res100

src_er_mon_er_dft_vs_n2f_m2_dair1 0_res100

The caveat is there is still one case which does show a discrepancy: $H_r$ at $r = 0$ for $|m| = 1$. This case is not necessarily a bug because the $H_r$ from $B_r$ update is *not* affected by `step_update_EDHB0` since $H_r = B_r$ (i.e., no PML for $H_r$ at $r = 0$). Fixing this would require storing $H_r$ as a separate array but the additional storage over the entire cell is probably not worth it just to fix the output at $r = 0$.

src_er_mon_hr_dft_vs_n2f_m1_dair1 0_res100

src/meep_internals.hpp
```diff
@@ -153,6 +158,11 @@ void step_beta_stride1(realnum *f, component c, const realnum *g, const grid_vol
                        sigw, kapw);                                                                \
   } while (0)
 
+#define STEP_UPDATE_EDHB0(f, fc, gv, is, ie, g, u, s, chi2, chi3, fw, dsigw, sigw, kapw)           \
+  do {                                                                                             \
+    step_update_EDHB0(f, fc, gv, is, ie, g, u, s, chi2, chi3, fw, dsigw, sigw, kapw);              \
+  } while (0)
```
stevengj, line 164: Why do we need a macro here rather than just calling `step_update_EDHB0` directly?

stevengj: Actually, this line https://github.com/NanoComp/meep/blob/e4504ed58eec75eea57aebc342d5b63683c89edf/src/Makefile.am#L37
is generating a `stride1` version called (confusingly) `STEP_UPDATE_EDHB_stride10`, which we should call from the macro similar to `STEP_UPDATE_EDHB`

stevengj:
I assume it still fixes #2182?

oskooi:
> I assume it still fixes https://github.com/NanoComp/meep/issues/2182?

`test_pml_cyl.py` added in #2383 to verify that the $z$-PML is working correctly is failing for $|m| = 1$. The fields at $r = 0$ are oscillating in time long after the source as turned off as they were prior to #2383. The change introduced in #2383 to make the $z$-PML work correctly was to *remove* the step-update at $r = 0$ in `update_eh.cpp`.

The $z$-PML at $r = 0$ is working correctly for $m = 0$ but that is not part of the existing tests in `test_pml_cyl.py`. This test is to be added in this PR.

oskooi:
Perhaps there is separate bug in the step-curl update for $B_r$ (= $H_r$) at $r = 0$ for $|m| = 1$ in `fields_chunk::step_db` of `step_db.cpp`:

https://github.com/NanoComp/meep/blob/e4504ed58eec75eea57aebc342d5b63683c89edf/src/step_db.cpp#L313-L347

src/update_eh.cpp
```diff
@@ -187,12 +187,19 @@ bool fields_chunk::update_eh(field_type ft, bool skip_w_components) {
...
                            gvs_eh[ft][i].big_corner(), dmp[dc][cmp], dmp[dc_1][cmp], dmp[dc_2][cmp],
                            s->chi1inv[ec][d_ec], dmp[dc_1][cmp] ? s->chi1inv[ec][d_1] : NULL,
                            dmp[dc_2][cmp] ? s->chi1inv[ec][d_2] : NULL, s_ec, s_1, s_2, s->chi2[ec],
                            s->chi3[ec], f_w[ec][cmp], dsigw, s->sig[dsigw], s->kap[dsigw]);
+
+          // special update for r=0 in cylindrical coordinates
+          STEP_UPDATE_EDHB0(f[ec][cmp], ec, gv, gvs_eh[ft][i].little_owned_corner(ec),
+                            gvs_eh[ft][i].big_corner(), dmp[dc][cmp], s->chi1inv[ec][d_ec], s_ec,
+                            s->chi2[ec], s->chi3[ec], f_w[ec][cmp], dsigw, s->sig[dsigw],
+                            s->kap[dsigw]);
```
stevengj, line 201: It occurs to me that we don't need a special `STEP_UPDATE_EDHB0` routine at all.  We just need to call `STEP_UPDATE_EDHB` and pass `NULL` for the off-diagonal terms.
```suggestion
          if (gv.dim == Dcyl) {
            vec is(gvs_eh[ft][i].little_owned_corner(ec));
            if (is.r() == 0) {
                vec ie(gvs_eh[ft][i].big_corner());
                ie.set_direction(R, 0);
                /* pass NULL for off-diagonal terms since they must be
                   zero at r=0 for an axisymmetric structure: */
                STEP_UPDATE_EDHB(f[ec][cmp], ec, gv, gvs_eh[ft][i].little_owned_corner0(ec),
                           gvs_eh[ft][i].big_corner(), dmp[dc][cmp], NULL, NULL,,
                           s->chi1inv[ec][d_ec], NULL, NULL, s_ec, s_1, s_2, s->chi2[ec],
                           s->chi3[ec], f_w[ec][cmp], dsigw, s->sig[dsigw], s->kap[dsigw]);
            }
          }
        }
```

stevengj:
Would be good to see Ep (before and after this PR) for m=1.   Without this PR, it seems like Ep at r=0 would never be updated, which seems like it would screw up the Hr update at r=0, so I'm confused about why removing the Ep update would help.  (Unless Ep is getting updated somewhere else?)

stevengj:
@mochen4 should also check to make sure that he didn't change what portion of the array is "owned" for cylindrical coordinates when he was updating loop_in_chunks (#1895), since that would affect the `step_db` code.

oskooi:
**with 2538**

src_er_mon_ep_dft_vs_n2f_m1_dair1 0_res50

**before 2538**

src_er_mon_ep_dft_vs_n2f_m1_dair1 0_res50

src_er_mon_hr_dft_vs_n2f_m1_dair1 0_res50

stevengj:
This is confusing to me because it doesn't seem like Ep should be updated *at all* at r=0 without this patch.  (The previous update_EDHB line skips r=0 for Ep because of the little_owned_corner0.)

Maybe put in some printf calls to see where the Ep is being changed for r=0?

stevengj:
Another quick check is to try setting μ = 1 + 1e-14 (i.e. force it to store and update H separately), in case the missing H update at r=0 (which affects the r=0 PML) is actually the real source of #2182.

oskooi:
> This is confusing to me because it doesn't seem like Ep should be updated at all at r=0 without this patch. (The previous update_EDHB line skips r=0 for Ep because of the little_owned_corner0.)

> Maybe put in some printf calls to see where the Ep is being changed for r=0?

Looks like the reason the DFT and N2F fields for $E_\phi$ and $H_r$ do not exactly match at $r = 0$ for $|m| = 1$ (as shown in the figures above) is not because of the changes introduced in this PR but rather because `get_dft_array` is interpolating the fields to the *center* of the voxel. This means that `ep` and `hr` which are defined on the Yee grid at $r = 0$ are being interpolated to $r = \Delta / 2$. The key point is that **we should not be using `get_dft_array` at $r =  0$ as part of the validation check**.

I inspected the time-domain (and not the DFT) fields at $r = 0$ by printing out the raw values (see code snippet below). The results are consistent with the expected values at $r = 0$ for $|m| = 1$ for which $E_\phi$ and $H_r$ can be nonzero.

**Summary of Results**
- $\varepsilon = 1$ everywhere
  - pre 2538: `ep` is zero at $r = 0$  whereas `hr` is nonzero at $r = 0$
  - 2538: `ep` and `hr` are nonzero at $r = 0$

- $\varepsilon = 1$, $\mu = 1.01$ everywhere via `default_material=mp.Medium(mu=1.01)`
  - pre 2538: `ep` and `hr` are zero at $r = 0$
  - 2538: `ep` and `hr` are nonzero at $r = 0$ 

**Notes**
- for $\varepsilon = 1$ and pre 2538:
   - `dp` at $r = 0$ is nonzero due to step-curl at $r = 0$ while `ep` is zero (never changed from its initial value)
   - `hr` is nonzero at $r = 0$ is expected because `hr = br` and `br` is nonzero due to step-curl at $r = 0$

```cxx
diff --git a/src/update_eh.cpp b/src/update_eh.cpp
index 31e1c9e6..9b6d0c88 100644
--- a/src/update_eh.cpp
+++ b/src/update_eh.cpp
@@ -207,6 +207,20 @@ bool fields_chunk::update_eh(field_type ft, bool skip_w_components) {
             }
           }
         }
+
+        if (ec == Ep || ec == Hr) {
+          realnum *E = f[ec][cmp];
+          const int yee_idx = gv.yee_index(ec);
+          const int sR = gv.stride(R), nZ = gv.num_direction(Z);
+          bool zero_everywhere = true;
+          for (int iZ = 0; iZ < nZ; iZ++) {
+            const int i = yee_idx + iZ - sR;
+            if (E[i] != 0)
+              zero_everywhere = false;
+          }
+          master_printf("zero-at-r0?, %s, %d, %d\n", component_name(ec), cmp, zero_everywhere);
+        }
+
       }
     }
   }
```

It seems therefore that this PR is producing the expected results (in the non-PML region anyway). However, we still need to explain why the $z$-PML at $r = 0$ is not working correctly due to the failing test in `test_pml_cyl.py`.

mochen4:
> @mochen4 should also check to make sure that he didn't change what portion of the array is "owned" for cylindrical coordinates when he was updating loop_in_chunks (#1895), since that would affect the `step_db` code. (Also check whether #2182 was present before #1895.)

I checked and found the ownership isn't changed by #1895, and #2182 was present before #1895.

oskooi:
A couple of changes:
1. Added a new test case for `m = 0` which fails without this PR.
2. Modified the `m = -1` test case to use a narrowband source similar to the fix used in [#2182 (comment)](https://github.com/NanoComp/meep/issues/2182#issuecomment-1224792986) in order to get the unit test to pass.

This PR does fix a bug for the `m = 0` but uncovers the bug for `m = -1` originally reported in #2182. 

There are two options for how to proceed: (1) we merge this PR as-is and then work on a bug fix for `|m| = 1` or (2) include the additional bug fix in this PR.

stevengj:
As discussed today, we went over the r=0 updates at 

https://github.com/NanoComp/meep/blob/955dcaa61faa50cec2909a0c090855312b14ecb6/src/step_db.cpp#L313-L346

and so far they look correct, but we didn't check the PML terms yet.  Would be good to compare those to the ones (for r≠0) in step_generic.cpp, since the PML equations should be the same (they don't involve spatial derivatives).

oskooi:
I also verified that we are not adding the `i*m/r` term *twice* at $r = 0$ during the update for `d(Br)/dt = d(Ep)/dz - i*m*Ez/r`.  There are two separate blocks where this happens:

https://github.com/NanoComp/meep/blob/eb8ff3aff135f28e5171e2417817b6bbd9c98e7f/src/step_db.cpp#L156-L157

https://github.com/NanoComp/meep/blob/eb8ff3aff135f28e5171e2417817b6bbd9c98e7f/src/step_db.cpp#L281-L282

In the first instance, the loop over the $r$ axis is set up such that it *never* starts at $r = 0$ for the case in which the field chunk includes $r = 0$: 

https://github.com/NanoComp/meep/blob/eb8ff3aff135f28e5171e2417817b6bbd9c98e7f/src/step_db.cpp#L174

https://github.com/NanoComp/meep/blob/eb8ff3aff135f28e5171e2417817b6bbd9c98e7f/src/step_db.cpp#L182

oskooi:
> we didn't check the PML terms yet. Would be good to compare those to the ones (for r≠0) in step_generic.cpp, since the PML equations should be the same (they don't involve spatial derivatives).

The special step-curl update equation with $z$-PML for `Dp` and `|m| = 1` at $r = 0$ in `step_db.cpp` does *not* seem to match the step-curl update equations for $r \neq 0$ in `step_generic.cpp`.

Specifically, in `step_db.cpp` we have `cc = Dp`, `d_c = P`, `f_p = f[Hr][cmp]`, `f_m = f[Hz][cmp]`, `dsig = Z`, `siginv = s->siginv[Z]`, `dsigu = R`, `siginvu = 0`, `fu = 0`, `the_f = f[Dp][cmp]`,  `s_d = +1`, `f_m_mult = 2`, and `fcnd = 0`. Using these parameters, the update equations are:

https://github.com/NanoComp/meep/blob/eb8ff3aff135f28e5171e2417817b6bbd9c98e7f/src/step_db.cpp#L334-L341

The key line number is 339 involving updating `the_f` at some $z$ position along $r = 0$.

For comparison, in `step_generic.cpp` we have:

https://github.com/NanoComp/meep/blob/eb8ff3aff135f28e5171e2417817b6bbd9c98e7f/src/step_generic.cpp#L197-L204

Only the second term matches `step_db.cpp` (although the sign is flipped). The first term in this equation does *not* appear in `step_db.cpp`.

(The step-curl equation for `Br` contains `dsig = P` which means it does not involve $z$-PML. `Br` is therefore omitted from this analysis.)

oskooi:
I have tried to add the correct step-db update equations for $z$-PML and the absorber at $r = 0$ for $|m| = 1$ to `step_db.cpp` by simply copying and over the update equations from `step_generic.cpp` (with some modifications). 

However, there seems to be a bug somewhere in this implementation since the fields blow up for the unit test for `m = -1` in `test_pml_cyl.py`.

oskooi:
```
step_db.cpp: In member function 'bool meep::fields_chunk::step_db(meep::field_type)':
step_db.cpp:349:29: error: operands to '?:' have different types 'const meep::realnum*' {aka 'const float*'} and 'meep::realnum' {aka 'float'}
  349 |           the_f[iz] = ((kap ? : kap[k] - sig[k] : 1) * the_f[iz] + dfcnd) * (siginv ? siginv[k] : 1);
      |                         ~~~~^~~~~~~~~~~~~~~~~~~
step_db.cpp:349:48: error: expected ')' before ':' token
  349 |           the_f[iz] = ((kap ? : kap[k] - sig[k] : 1) * the_f[iz] + dfcnd) * (siginv ? siginv[k] : 1);
      |                        ~                       ^~
      |                                                )
step_db.cpp:349:101: error: expected ')' before ';' token
  349 |           the_f[iz] = ((kap ? : kap[k] - sig[k] : 1) * the_f[iz] + dfcnd) * (siginv ? siginv[k] : 1);
      |                       ~                                                                             ^
      |                                                                                                     )
step_db.cpp:350:81: error: 'i' was not declared in this scope; did you mean 'pi'?
  350 |           if (fu) fu[iz] = siginvu[ku] * ((kapu ? kapu[ku] - sigu[ku] : 1) * fu[i] + the_f[iz] - fprev);
      |                                                                                 ^
      |                                                                                 pi
step_db.cpp:325:24: warning: unused variable 'siginv' [-Wunused-variable]
  325 |         const realnum *siginv = s->sigsize[dsig] > 1 ? s->siginv[dsig] : 0;
```

oskooi:
There is a bug in the special step-db update equations at $r = 0$ for $|m| = 1$ introduced by the latest commit because the unit test (`TestLDOS.test_ldos_ext_eff` of `test_ldos.py`) which verifies that the extraction efficiency of an LED is identical in 3d and cylindrical coordinates is failing:

```
extraction efficiency (cyl):, 3.440227
extraction efficiency (3D):, 0.327121
```

Clearly, the results for the cylindrical coordinates is wrong (the extraction efficiency cannot be larger than one). 

I went over the update equations again and compared them to the "most general case" of `step_generic.cpp` from which they were adapted. The update equations seem correct, however, and it is not obvious where the bug could be.

https://github.com/NanoComp/meep/blob/95ddacdda9754b0c6613b2a25a1154e87ab427e9/src/step_generic.cpp#L218-L230

oskooi:
There was a bug in the sign of the curl terms and strides which involved a simple fix:

```
@@ -346,7 +346,7 @@ bool fields_chunk::step_db(field_type ft) {
         const int dku = gv.iyee_shift(cc).in_direction(dsigu);
         realnum *fu = siginvu && f_u[cc][cmp] ? f[cc][cmp] : 0;
         realnum *the_f = fu ? f_u[cc][cmp] : f[cc][cmp];
-        int sd = ft == D_stuff ? +1 : -1;
+        int sd = ft == D_stuff ? -1 : +1;
         realnum f_m_mult = ft == D_stuff ? 2 : (1 - 2 * cmp) * m;
         realnum dt2 = dt * 0.5;
```

stevengj:
I think the old `sd` was correct, as we just verified by looking at the update equations and Yee grid.

If a test is failing now and was passing with the old code in `step_db`, what I would suggest is simply changing the r=0 update equations back into the old ones term-by-term.   At some point the test will pass, and that will let us zoom in on the new term that is causing the problem.

oskooi:
It seems there was a bug in the `Dp` update whereby the loop over the $z$ coordinates involving the index `iz` was one larger than it needed to be:
```
--- a/src/step_db.cpp
+++ b/src/step_db.cpp
@@ -351,7 +350,7 @@ bool fields_chunk::step_db(field_type ft) {
         realnum f_m_mult = ft == D_stuff ? 2 : (1 - 2 * cmp) * m;
         realnum dt2 = dt * 0.5;
 
-        for (int iz = (ft == D_stuff); iz < nz + (ft == D_stuff); ++iz) {
+        for (int iz = (ft == D_stuff); iz < nz; ++iz) {
           realnum fprev = the_f[iz];
           realnum dfcnd = (sd * Courant) * (f_p[iz] - f_p[iz - sd] - f_m_mult * f_m[iz]);
           if (fcnd) {
```

oskooi:
With the latest commit, this PR finally fixes #2489. 

The [test](https://gist.github.com/oskooi/8946d61da945f5ab68ce26cf028c6cb0) for $m = 1$ involves an $E_r$ point source at $r = 0$ within a dielectric disc and two different sets of near-field monitors. The computed far fields are the same regardless of the near-field monitor configuration.

led_disc_cyl_discR1_L10 0_dpmlr1 0_dpmlz1 0_plot2D

led_disc_cyl_radpattern_rp0_res50 0_L10 0_dpmlr1 0_dpmlz1 0

src/step_db.cpp
```diff
@@ -351,7 +350,7 @@ bool fields_chunk::step_db(field_type ft) {
         realnum f_m_mult = ft == D_stuff ? 2 : (1 - 2 * cmp) * m;
         realnum dt2 = dt * 0.5;
 
-        for (int iz = (ft == D_stuff); iz < nz + (ft == D_stuff); ++iz) {
+        for (int iz = (ft == D_stuff); iz < nz; ++iz) {
```
stevengj, line 353: This is a little surprising to me.  I thought that `iz == nz` was owned for the `Dp` component?   This means that `Dp` has only `nz-1` owned points, whereas `Hr` has `nz` owned points?

Would be worth printing out `gv.little_owned_corner(Dp)` and `gv.big_owned_corner(Dp)` and checking how those values relate to these (after you divide by 2) in the `z` component.

oskooi: Adding some `printf` statements to `step_db.cpp` for outputting the properties of the chunk's `grid_volume` `gv` for the PML and non-PML regions used during the `Dp` and `Hr` field updates:
```
--- a/src/step_db.cpp
+++ b/src/step_db.cpp
@@ -350,6 +350,22 @@ bool fields_chunk::step_db(field_type ft) {
         realnum f_m_mult = ft == D_stuff ? 2 : (1 - 2 * cmp) * m;
         realnum dt2 = dt * 0.5;
 
+        if (sig || sigu)
+          printf("PML region\n");
+        else
+          printf("non PML region\n");
+        printf("vol:, (%0.2f, %0.2f), (%0.2f, %0.2f)\n",
+               gv.surroundings().in_direction_min(R), gv.surroundings().in_direction_min(Z),
+               gv.surroundings().in_direction_max(R), gv.surroundings().in_direction_max(Z));
+        printf("grid_vol:, Dp, (%d, %d), (%d, %d), %d, %d\n",
+               gv.little_owned_corner(Dp).r(), gv.little_owned_corner(Dp).z(),
+               gv.big_owned_corner(Dp).r(), gv.big_owned_corner(Dp).z(),
+               gv.big_owned_corner(Dp).z() - gv.little_owned_corner(Dp).z() + 1, nz);
+        printf("grid_vol:, Hr, (%d, %d), (%d, %d), %d, %d\n",
+               gv.little_owned_corner(Hr).r(), gv.little_owned_corner(Hr).z(),
+               gv.big_owned_corner(Hr).r(), gv.big_owned_corner(Hr).z(),
+               gv.big_owned_corner(Hr).z() - gv.little_owned_corner(Hr).z() + 1, nz);
+
```
Then running the unit test `python3 test_ldos.py TestLDOS.test_ldos_ext_eff` (`Er` source at $r = 0$ for $m = -1$) shows the following output:

```
Working in Cylindrical dimensions.
Computational cell is 6.52 x 0 x 1.72 with resolution 25
     block, center = (0,0,-0.75)
          size (1e+20,1e+20,0.208333)
          dielectric constant epsilon diagonal = (5.76,5.76,5.76)
time for set_epsilon = 0.0203028 s
-----------
Meep: using complex fields.
PML region
vol:, (0.00, 0.32), (5.96, 0.88)
gv:, Dp, (0, 18), (298, 44), 27, 14
gv:, Hr, (0, 17), (298, 43), 27, 14
non PML region
vol:, (0.00, -0.84), (5.96, 0.32)
gv:, Dp, (0, -40), (298, 16), 57, 29
gv:, Hr, (0, -41), (298, 15), 57, 29
```

led_plot2D

The `Dp` and `Hr` field arrays each contains 14 and 29 grid points in the $z$ direction within the PML and non-PML regions, respectively. The difference in the $z$ coordinate of the `big_owned_corner` and `little_owned_corner` is 27 and 57, respectively.

oskooi:
As a demonstration that this PR produces the correct result for $\textbf{m = 0}$, here are the results from a [test](https://gist.github.com/oskooi/8946d61da945f5ab68ce26cf028c6cb0) which compares the radiated fields in air computed using two different methods of (1) DFT monitor and (2) near-to-far field transformation, from an $E_r$ point source at $r = 0.1$ within a dielectric disc above a perfect-metal ground plane. (1) and (2) should produce *identical* results.

Results are shown below for the master branch and this PR.

disc_n2f_dair1 0_plot2D

**1. master branch (5a1dd2c)**

led_disc_er_dft_vs_n2f_rpos0 1_m0_dair1 0_res100_master

**PR 2538**

led_disc_er_dft_vs_n2f_rpos0 1_m0_dair1 0_res100_PR2538

stevengj:
What concerns me here is that it seems like the loop has to be  `for (iz = 1; iz < nz+1; ++iz)` to get all of the owned points for the `Dp` component.   So, we are intentionally introducing a bug by not updating the `iz = nz` owned point (albeit maybe a second-order error), based on the empirical observation that it somehow works around some other problem (we hope … we aren't sure).   This is an unreliable approach …

You say that changing the loop to `nz+1` messes up the LED extraction simulation.   Could we see some snapshots of the time domain fields with and without the `+1` to see why the update rule at a single grid point is causing such a big difference?  e.g. is something blowing up in the PML?

Also might be good to check the step_boundaries code — add a printf to see what grid points and components are being set by boundary conditions, to make sure that this "owned" point is not accidentally being overwritten by a boundary condition.

oskooi:
> Could we see some snapshots of the time domain fields with and without the +1 to see why the update rule at a single grid point is causing such a big difference? e.g. is something blowing up in the PML?

I plotted the DFT fields for `Dp` vs. $z$ at $r = 0$ without (blue plot) and with (red plot) the `+1` term applied to the ending index of the `Dp` field array. There is a large difference in the DFT fields. The fields without the `+1` term seem correct. The fields with the `+1` seem clearly wrong.

In this setup, the $E_r$ point source is at $z = -1.0$ and the $z$-PML starts at $z = 0.15$.

One obvious difference in these results is that the $z$-PML in the `+1` case does not seem to be working properly: the fields only start to decay well into the PML at $z > 0.15$. 

Dp_dft_vs_z_r0_PR2538

Dp_dft_vs_z_r0_PR2538_plus

mochen4:
> I plotted the DFT fields for `Dp` vs. z at r=0 without (blue plot) and with (red plot) the `+1` term applied to the ending index of the `Dp` field array. There is a large difference in the DFT fields. The fields without the `+1` term seem correct. The fields with the `+1` seem clearly wrong.
> 

Both plots seem incorrect to me. I agree that fields should start decaying at 0.15, but they should also decay slowly as in the second plot, not stop abruptly as in the first plot. Moreover, the bottom boundary is metallic right? Shouldn't the fields go to zero there?

I have been looking at the indices that are looped over in ``step_db``, and I noticed some inconsistencies (if I understand correctly):
There are ``nz+1`` (from ``iz=0`` to ``iz=nz``) field points at ``ir=0``. And indeed, earlier in the ``step_db`` code, the script handling PML (and other stuff) loops from ``iz=0`` to ``iz=nz`` to update ​``the_f`` for all cases, e.g.

https://github.com/NanoComp/meep/blob/133570ccbd5ba622a407e31d74b512a7c8e459f3/src/step_db.cpp#L184

On the other hand, when handling the ``Dz``, ``Dp``, ``Hr`` fields at ``r=0`` for ``m=-1,0,1``, (if I keep the ``+1`` for ``Dp``,) the loop only goes through ``nz`` points (either ``iz=0`` to ``nz-1`` or ``iz=1`` to ``nz``). And I am confused how these start and end points were chosen.
As Ardavan [noted](https://github.com/NanoComp/meep/pull/2538#discussion_r1251305041), only ``nz`` points are owned at ``ir=0``, so I guess we should indeed only update ``nz`` points? And I think it should be the first ``nz`` points ``iz=0`` to ``nz-1``?

stevengj:
Regarding @mochen4's comments:

1. Remember that we aren't outputting the Yee points, but by default the DFT is interpolated to the center of the voxels.  So at z=0 we don't see the Yee point where Ep is exactly zero, but an interpolation to a nearby point, which is nonzero.  It looks like it is going to zero, so it looks okay.
2. In the case where it looks like it is suddenly going to zero in the PML, this is where I think @oskooi's loop is going to `nz-1` instead of all the way to `nz`.  So effectively it is not updating one of the "owned" points, which effectively puts a pixel of PEC at the boundary between the PML and non-PML region.  This would make it suddenly zero.  But I think it would be informative to have a full 2d image (either of a DFT field or a time-domain movie), to see why what should be just a single pixel seems to matter so much.
3. The loop that goes unconditionally from `iz=0` to `iz=nz` is in the code adding the `i*m/r` terms.   Indeed, this is updating both owned and not-owned points.  But updating the not-owned points here should be harmless because they will just get overwritten by `step_boundaries`.   Doing it this way makes the logic simpler, but it might be a nice exercise to compute the correct owned bounds using `LOOP_OVER_VOL_OWNED`, or by calculating them from the `little_owned_corner` and `big_owned_corner`.

oskooi:
Regarding @stevengj's first comment, this is indeed what I observed in [#2383 (comment)](https://github.com/NanoComp/meep/pull/2383#issuecomment-1435488769).

mochen4:
I added print statement after these lines to check the non-owned points. 
https://github.com/NanoComp/meep/blob/133570ccbd5ba622a407e31d74b512a7c8e459f3/src/boundaries.cpp#L461-L462

In my setup, at ``r=0``: the two ``grid_volume`` owned points from -112 to 14 and 16 to 116 for Dp, and -113 to 13 and 15 to 115 for Br; the non-owned points are (0, -114) and (0, 14) for Dp, and (0,15) and (0,117) for Br. It looks correct that the first of Dp and last of Br are not owned.

oskooi:
Following @stevengj's third suggestion, I tried to rewrite the $+im/r$ update loop in `step_db.cpp` using `LOOP_OVER_VOL_OWNED` which replaces the manual loops over the $r$ and $z$ grid points. The setup required using the function [`ptrdiff_t index(component, const ivec &)`](https://github.com/NanoComp/meep/blob/5fdec2d3b6d3e6eb151ee4b2193532b8080014e0/src/meep/vec.hpp#L1056) to determine the array indices for the $r$ and $z$ grid points given an `ivec` grid-point location.

This is what it looks like for the most general case:

```
diff --git a/src/step_db.cpp b/src/step_db.cpp
index 794d7ab6..99a91ae6 100644
--- a/src/step_db.cpp
+++ b/src/step_db.cpp
@@ -179,97 +179,95 @@ bool fields_chunk::step_db(field_type ft) {
           if (siginvu) { // PML + fu
             if (cndinv)  // PML + fu + conductivity
               //////////////////// MOST GENERAL CASE //////////////////////
-              for (int ir = ir0 == 0; ir <= gv.nr(); ++ir) {
+              LOOP_OVER_VOL_OWNED(gv, cc, i) {
+                IVEC_LOOP_ILOC(gv, here);
+                const ptrdiff_t ir = gv.index(direction_component(cc, R), here);
+                const ptrdiff_t iz = gv.index(direction_component(cc, Z), here);
                 realnum rinv = the_m / (ir + ir0);
-                for (int iz = 0; iz <= gv.nz(); ++iz) {
-                  ptrdiff_t idx = ir * sr + iz;
-                  int k = dk + 2 * (dsig == Z ? iz : ir);
-                  int ku = dku + 2 * (dsigu == Z ? iz : ir);
-                  realnum df, dfcnd = rinv * g[idx] * cndinv[idx];
-                  fcnd[idx] += dfcnd;
-                  fu[idx] += (df = dfcnd * siginv[k]);
-                  the_f[idx] += siginvu[ku] * df;
-                }
+                int k = dk + 2 * (dsig == Z ? iz : ir);
+                int ku = dku + 2 * (dsigu == Z ? iz : ir);
+                realnum df, dfcnd = rinv * g[i] * cndinv[i];
+                fcnd[i] += dfcnd;
+                fu[i] += (df = dfcnd * siginv[k]);
+                the_f[i] += siginvu[ku] * df;
               }
```

Unfortunately, these changes are causing the fields to blow up for a cylindrical-coordinate simulation involving PMLs.

oskooi:
Regarding @stevengj 's second comment, I plotted the DFT fields (linear and log scales) for $|D_\phi|$ over the *entire* cell. There does not seem to a noticeable difference in the fields for the two `Dp` loop cases of (1) `iz < nz` and (2) `iz < nz + (ft == D_stuff)`.

Even though the fields are similar in the two cases, the extraction efficiency is very different. Case 1 is correct, case 2 is not.

image

image

oskooi:
After rebasing to master following #2585 and replacing the custom loops for $r = 0$ with `LOOP_OVER_IVECS` for $m = 0, \pm 1$, there is now just a *single* failing test: `TestLDOS.test_ldos_ext_eff` in `python/tests/test_ldos.py`.

The thing that is causing this test to fail is the flux computed in the $z$ direction which is different than the value computed by the master branch. The LDOS (used to compute the total flux) and the flux in the $r$ direction are *unchanged*. This is an important finding which suggests one thing: **the field updates are correct but there is a bug in the $z$-flux calculation at $r = 0$** probably because of the way the fields $(E_r, E_\phi, H_r, H_\phi)$ are all interpolated to the center of the voxel at $r = 0.5\Delta r$ which violates the [special field values for |m| = 1](https://meep.readthedocs.io/en/latest/Exploiting_Symmetry/#cylindrical-symmetry). Note that $E_\phi$ and $H_r$ are defined at $r = 0$ in the Yee grid and can have nonzero values.

@mochen4, any thoughts?

image

mochen4:
> After rebasing to master following #2585 and replacing the custom loops for r=0 with `LOOP_OVER_IVECS` for m=0,±1, there is now just a _single_ failing test: `TestLDOS.test_ldos_ext_eff` in `python/tests/test_ldos.py`.

If I understand correctly, replacing with `LOOP_OVER_IVECS`  in #2585 actually made a difference? So the original loop incorrectly updates some points that actually were not overridden? If that is the case, maybe something similar might happen somewhere else?

oskooi:
Replacing the custom loops with `LOOP_OVER_IVECS` did make a difference. Everything seems to be working fine now except for the $z$-flux calculation at $r = 0$. I confirmed this by repeating the calculation in the extraction efficiency unit test but modifying the $z$-flux line monitor to *exclude* $r = 0$. The results are *identical* for this branch and master. Since the $z$-flux value is different for this branch and master only when the line monitor includes $r = 0$, this must mean the problem is with the $z$-flux calculation at $r = 0$.

I suspect there is a bug in the field-interpolation step of `dft_chunk::update_dft` because of the way the $r = 0$ case is being handled:

https://github.com/NanoComp/meep/blob/ee3bd80a419ceead4f1cd4c1fd5ae7ed82c6de46/src/dft.cpp#L274-L287

The problem is that we need $P_z$ exactly at $r = 0$. However, in these lines the fields are *always* interpolated to the center of the voxel (i.e., $r = \Delta r$). It seems we need to add a special case here just for $r = 0$?

mochen4:
> Replacing the custom loops with `LOOP_OVER_IVECS` did make a difference. Everything seems to be working fine now except for the z-flux calculation at r=0. I confirmed this by repeating the calculation in the extraction efficiency unit test but modifying the z-flux line monitor to _exclude_ r=0. The results are _identical_ for this branch and master. Since the z-flux value is different for this branch and master only when the line monitor includes r=0, this must mean the problem is with the z-flux calculation at r=0.

I agree. My point was that, since we actually didn't expect replacing `LOOP_OVER_IVECS` would make a difference based on @stevengj 's comment, there might be something that we missed. Thankfully it is fixed now, but we might want to keep it in mind when we encounter another bug involving custom loops in the future.

> The problem is that we need Pz exactly at r=0. However, in these lines the fields are _always_ interpolated to the center of the voxel (i.e., r=Δr). It seems we need to add a special case here just for r=0?

That sounds plausible. If the error comes from interpolation, can we confirm it by checking the error vs. resolution?

mochen4:
I checked and found that, if r=0 was included, the flux increased (roughly) quadratically with respect to the resolution, which means a factor of resolution for each field component at r=0?

stevengj:
It would still be nice to see what the time-domain fields look like (either a movie or just a few typical snapshots)

mochen4:
DFT fields of ``Ep`` and ``Hr`` at ``r=0`` position of the flux monitor indeed grow linearly with respect to the resolution. 

Using ``sim.run(until=10)`` and ``sim.run(until=100)``, and ``sim.plot2D``, I plotted ``Ep`` field at **time=10**:
at_time10

**and at time=100**
at_time100

The fields at ``r=0`` at the start of simulation don't make sense to me, but it looks fine later at time=100, so I guess the time evolution probably is correct, but the initialization is wrong somehow?

mochen4:
In comparison, on master branch,

**At time = 10**:
master_at_time10

**And at time = 100**:
master_at_time100

stevengj:
The t=10 plot shows something weird at r=0.  Would be good to check:
* Does this increase with resolution?
* Is it only the fields at r=0, not any other pixel?
* Does it show up in both Ep and Dp, or just Ep?  If it's just Ep, maybe chi1inv is screwed up at r=0?
* Does it show up if you add a printf to output the raw r=0 Yee-grid values, or does it only show up after interpolation?

The only way I can think of for the fields to be screwed up at r=0 and not be screwed up everywhere is for the problem to be in Ez and maybe Ep, since there was a comment (removed in #2383) that these E field components at r=0 don't actually affect the H = curl E timestepping, they only affect field output and DFTs where E is specifically requested.   Note that the present PR puts back the E-from-D updates at r=0, albeit in a different way.

oskooi:
In the failing unit test for the extraction efficiency which involves computing the radiated flux using `add_dft` (which is based on interpolating the DFT fields to the center of the voxel that we have identified as containing a potential bug at $r = 0$), computing the radiated flux instead from the **radiation pattern of the far fields** via `add_near2far` (which [does *not* involve interpolating the DFT voxels to the center of the voxels](https://github.com/NanoComp/meep/blob/cd9a7ebe8fb4ccd47cbf9eeb1bc8d8f71b1cea1a/src/near2far.cpp#L634-L635)) the test passes: the extraction efficiency computed in cylindrical coordinates matches the 3d Cartesian result.

I think the change from `add_dft` to `add_near2far` provides further evidence of the interpolation bug for the DFT fields at $r = 0$. 

We can probably go ahead and merge this PR now and then file a bug report that we can address with a separate PR.

mochen4:
It looks like that the raw Dp fields at r=0 actually do increase with resolution. At t=5, the fields are
Screen Shot 2023-08-09 at 13 32 03

In addition, the fields oscillates twice as fast. For the same period of time=5, I have 500 snapshots of fields at resolution=25 and 1000 snapshots of fields at resolution=50. Therefore, frames 498-500 at resolution 25 should correspond to frames 996-1000 at resolution 50. However, plotting those frames, it looks like that the fields are oscillating at the rate of time stepping:

Screen Shot 2023-08-09 at 13 50 20

Screen Shot 2023-08-09 at 13 50 32

mochen4:
@oskooi pointed out that, when resolution is doubled, pixel size is halved, so time-domain fields getting doubled would make sense. The evolution of fields in time is reasonable and consistent with what we expect from a correct time stepping. The next step is to check that there is no jump in the raw field as r approaches 0.

oskooi:
> when resolution is doubled, pixel size is halved, so time-domain fields getting doubled 

For reference, due to the finite discretization a $\delta$-function current source has an amplitude of $1 / \Delta x$ where $\Delta x$ is the voxel dimension. This is implemented in:

https://github.com/NanoComp/meep/blob/dad3e5273e92f45f66991397e05e65c821cee55a/src/sources.cpp#L480-L484

Doubling the grid resolution should therefore result in a doubling of the field amplitudes in the time domain. This is what we are observing here.

mochen4:
> In addition, the fields oscillates twice as fast. For the same period of time=5, I have 500 snapshots of fields at resolution=25 and 1000 snapshots of fields at resolution=50. Therefore, frames 498-500 at resolution 25 should correspond to frames 996-1000 at resolution 50. However, plotting those frames, it looks like that the fields are oscillating at the rate of time stepping:

I realized that I was saving both real and imaginary parts of the fields together. If I focus on the real part (i.e. frames with odd number index), the fields evolution would make sense. Here is an animation for the real part of Dp at r=0 for t=0 to 5:


https://github.com/NanoComp/meep/assets/25192039/e235a709-f0e5-479f-a7a5-daf5524d268d

mochen4:
On the other hand, using ``get_array`` with ``snap=True`` to get the fields on Yee grid, I did find a discontinuity along r at r=0, consistent with the previous plot in https://github.com/NanoComp/meep/pull/2538#issuecomment-1663226841
Dp_along_r

oskooi:
Instead of a bug in the $z$-flux calculation at $r = 0$, I think the bug is in the "restriction" procedure used to define an $E_r$ point source at $r = 0$ since $E_r$ is *not* defined at $r = 0$ on the Yee grid. An $E_r$ point source defined at $r = 0$ should be "restricted" to its nearest Yee-grid locations at $r = 0.5 Δr$ but this does not seem to be happening correctly.

In the failing unit test of the extraction efficiency, if the position of the source is shifted by a voxel away from $r = 0$ the test passes using the flux calculation via `add_flux`. I modified this test in the latest commit and added this note in the comments:

```py
	# Because (1) Er is not defined at r=0 on the Yee grid, and (2) there                  
	# seems to be a bug in the restriction of an Er point source at r = 0,                 
	# the source is placed at r = ~Δr (just outside the first voxel).                      
	# This incurs a small error which decreases linearly with resolution.                  
	src_pt = mp.Vector3(1.5 / self.resolution, 0, -0.5 * sz + h * dmat)
```

Note that this hack only seems to work if the source position in $r$ is placed **outside of the first voxel**. In this example, it is placed in the center of the *second* voxel (i.e., `1.5 * Δr`).

stevengj:
Might be worth trying an integrated source, since that adds the current in a completely different place.   Might help to determine whether this is problem with the Er field update.

stevengj:
One thing to make sure of is that we are doing the transformation across r=0 correctly when interpolating the source in src_vol_chunkloop.  Two things to try:

1. Around [this line](https://github.com/NanoComp/meep/blob/8d2fd8db23a4f55b613ba2c7780773d583dfcebf/src/sources.cpp#L273) add a printf statement to see (1) the amplitude `amps_array[idx_vol]`, (2) the coordinates `iloc` and `loc`, and (3) the `shift_phase`
2. I have a suspicion that we may be subtracting currents from `r < 0` instead of adding.  You could check that forcing it to always add, by changing [this line](https://github.com/NanoComp/meep/blob/8d2fd8db23a4f55b613ba2c7780773d583dfcebf/src/sources.cpp#L259) to remove the `shift_phase` factor (just to `complex<double> amp = data->amp;`) whether that helps.

stevengj:
If the problem is with sources that touch the r=0 pixel, would be good to check whether the problem is simply an overall normalization or something deeper.

One way to check this is to look at the spectrum, e.g. radiated power, for some nontrivial structure (e.g. a waveguide or sphere), and compare it for a source at r=0 and a source at r=1.5Δx.  They should be the same, up to a first-order O(Δx) error, but if there is an overall normalization problem at r=0 they should still be nearly proportional to one another.

oskooi:
Plot of DFT flux for two source configurations.

flux_compare_r0_r1 5

stevengj:
This looks just like a scale factor (i.e. the ratio is nearly constant over most of the bandwidth).   Presumably as you increase the resolution it should get more and more flat.

That means that it's not a problem with the timestepping or anything like that — it's just an issue with the normalization when the source is placed.

mochen4:
Here is a plot of field after 100 wavelength

Figure_1

stevengj:
Some related problem might occur for the `dft_ldos` calculation if you have a point source at r=0, since again it involves calculating sources at the origin.

So, for example, if you did the extraction-efficiency tutorial but normalized by a `dft_flux` rather than `dft_ldos`, then it might be fine since a normalization error in the source would cancel.

mochen4:
> @mochen, would you update your [field plots](https://github.com/NanoComp/meep/pull/2538#issuecomment-1760271451) above for an Er source at r=0 and r=1.5Δr to show scale bars.

Here are the plots of fields with colorbars. I fixed their range to be the same:

Er source at r=0:
Figure_0

Er source at r=1.5Δr:
Figure_1

oskooi:
> One thing to make sure of is that we are doing the transformation across r=0 correctly when interpolating the source in src_vol_chunkloop.  Two things to try:
>
> 1. Around [this line](https://github.com/NanoComp/meep/blob/8d2fd8db23a4f55b613ba2c7780773d583dfcebf/src/sources.cpp#L273) add a printf statement to see (1) the amplitude `amps_array[idx_vol]`, (2) the coordinates `iloc` and `loc`, and (3) the `shift_phase` .   Do it for a source at r=0 and at r=1.5Δr.   In both cases the point source should be mapped to **4** grid points, at ± 1 pixel in Δz and Δr, but in the r=0 case the sources should be at ±Δr/2, which should both get mapped via `r_to_minus_r` symmetry to +Δr/2 (hopefully with the same sign).

I added the `printf` statement within `src_vol_chunkloop` of `sources.cpp` to output the relevant quantities described above. 

The output for the $E_r$ source at $r = 1.5 \Delta r$ and $r = 0$ (of the parameter `src_pt` of the function `ext_eff_cyl` of `python/tests/test_ldos.py`) is shown below. I ran the test using:

```
$ mpirun -np 2 python3 test_ldos.py TestLDOS.test_ldos_ext_eff
```

For both cases, there are only **2** grid points at ± 1 pixel in Δz. There is **no restriction** of the $E_r$ point source occurring in $r$, regardless of its value. This seems like an obvious bug.

```
diff --git a/src/sources.cpp b/src/sources.cpp
index e4ef6d1b..6608c3d9 100644
--- a/src/sources.cpp
+++ b/src/sources.cpp
@@ -271,6 +271,9 @@ static void src_vol_chunkloop(fields_chunk *fc, int ichunk, component c, ivec is
 
+    master_printf("src_vol_chunkloop:, %0.5f%+0.5fj, (%d, %d), (%0.5f, %0.5f),  %0.5f%+0.5fj\n",
+                  real(amps_array[idx_vol]), imag(amps_array[idx_vol]), iloc.r(), iloc.z(), loc.r(), loc.z(),
+                  real(shift_phase), imag(shift_phase));
 
     // check for invalid sources at r=0 in cylindrical coordinates
     if (fc->gv.dim == Dcyl && loc.r() == 0 && amps_array[idx_vol] != 0.0) 
```

1. $r = 1.5 \Delta r$
```
src_vol_chunkloop:, 468.75000+0.00000j, (3, -38), (0.06000, -0.76000),  1.00000+0.00000j
src_vol_chunkloop:, 156.25000+0.00000j, (3, -36), (0.06000, -0.72000),  1.00000+0.00000j
```

2. $r = 0$
```
src_vol_chunkloop:, 234.37500+0.00000j, (1, -38), (0.02000, -0.76000),  1.00000+0.00000j
src_vol_chunkloop:, 78.12500+0.00000j, (1, -36), (0.02000, -0.72000),  1.00000+0.00000j
```

However, if I change the source component to $E_\phi$ (`src_cmpt = mp.Ep`), the output shows **4** grid points for the point source at $r = 1.5 \Delta r$ and **2** grid points $r = 0$. Something is clearly wrong.

1. $r = 1.5 \Delta r$

```
src_vol_chunkloop:, 234.37500+0.00000j, (2, -38), (0.04000, -0.76000),  1.00000+0.00000j
src_vol_chunkloop:, 78.12500+0.00000j, (2, -36), (0.04000, -0.72000),  1.00000+0.00000j
src_vol_chunkloop:, 234.37500+0.00000j, (4, -38), (0.08000, -0.76000),  1.00000+0.00000j
src_vol_chunkloop:, 78.12500+0.00000j, (4, -36), (0.08000, -0.72000),  1.00000+0.00000j
```

2. $r = 0$

```
src_vol_chunkloop:, 468.75000+0.00000j, (0, -38), (0.00000, -0.76000),  1.00000+0.00000j
src_vol_chunkloop:, 156.25000+0.00000j, (0, -36), (0.00000, -0.72000),  1.00000+0.00000j
```

stevengj:
1.5Δr will be exactly on the Yee grid point for Er, so you expect only 2 current points.  It would be good to try 2.0Δr, in which case we expect 4 points.

It would also be useful to add some print statements to `loop_in_chunks` (or step through with a debugger).

1. After https://github.com/NanoComp/meep/blob/20de9c7754c8a7b4b22b16c6f450875c9323574c/src/loop_in_chunks.cpp#L358-L359 to print out `is, ie` and the weights `s0, e0, s1, e1` — this is the bounding box for the `where` volume (a single point in this case), which may extend outside the computational cell.  In the r=0 case, this should include ±Δr/2 points.
2. After https://github.com/NanoComp/meep/blob/20de9c7754c8a7b4b22b16c6f450875c9323574c/src/loop_in_chunks.cpp#L379, to see `gvS`, the grid volume transformed by symmetry — here, the only symmetry should be `r_to_minus_r` so we should see a loop iteration where the grid is transformed to negative radii
3. After https://github.com/NanoComp/meep/blob/20de9c7754c8a7b4b22b16c6f450875c9323574c/src/loop_in_chunks.cpp#L423-L424 to see the transformed volume `iscoS, iecoS` for each chunk

oskooi:
> 1.5Δr will be exactly on the Yee grid point for Er, so you expect only 2 current points. It would be good to try 2.0Δr, in which case we expect 4 points.

Findings:

1. $E_r$ source placed at $r = \Delta r$ and $r = 2.0\Delta r$ (edge of the Yee-grid voxel) yields 4 restricted grid points. 

2. $E_r$ source at $r = 0.5\Delta r$ and $r = 2.5\Delta r$ (center of the Yee-grid voxel) yields 2 restricted grid points. 
 
In these four cases (output shown below), the sum of the amplitudes of the grid points is *always* 625.00. This indicates "charge conservation." These results suggest that the restriction for $E_r$ is working correctly for sources placed at $r > 0$.

However, the sum of the amplitudes for an $E_r$ source at $r = 0$ with 2 restricted grid points is 312.5 (output shown above). This is exactly **half** the expected value of 625.00 due to the two missing restricted grid points at $r < 0$.

1. $r = \Delta r$ (edge of voxel)
```
src_vol_chunkloop:, 234.37500+0.00000j, (1, -38), (0.02000, -0.76000), 1.00000+0.00000j
src_vol_chunkloop:, 78.12500+0.00000j, (1, -36), (0.02000, -0.72000), 1.00000+0.00000j
src_vol_chunkloop:, 234.37500+0.00000j, (3, -38), (0.06000, -0.76000), 1.00000+0.00000j
src_vol_chunkloop:, 78.12500+0.00000j, (3, -36), (0.06000, -0.72000), 1.00000+0.00000j
``` 

2. $r = 2.0\Delta r$ (edge of voxel)
```
src_vol_chunkloop:, 234.37500+0.00000j, (3, -38), (0.06000, -0.76000), 1.00000+0.00000j
src_vol_chunkloop:, 78.12500+0.00000j, (3, -36), (0.06000, -0.72000), 1.00000+0.00000j
src_vol_chunkloop:, 234.37500+0.00000j, (5, -38), (0.10000, -0.76000), 1.00000+0.00000j
src_vol_chunkloop:, 78.12500+0.00000j, (5, -36), (0.10000, -0.72000), 1.00000+0.00000j
```

6. $r = 0.5\Delta r$ (center of voxel)
```
src_vol_chunkloop:, 468.75000+0.00000j, (1, -38), (0.02000, -0.76000), 1.00000+0.00000j
src_vol_chunkloop:, 156.25000+0.00000j, (1, -36), (0.02000, -0.72000), 1.00000+0.00000j
```

7. $r = 2.5\Delta r$ (center of voxel)
```
src_vol_chunkloop:, 468.75000+0.00000j, (5, -38), (0.10000, -0.76000), 1.00000+0.00000j
src_vol_chunkloop:, 156.25000+0.00000j, (5, -36), (0.10000, -0.72000), 1.00000+0.00000j
```

oskooi:
Additional output from items 1-3 listed above. The main finding is that `iscoS` and `iecoS` do *not* contain negative values for the $r$ coordinate for an $E_r$ source at $r = 0$ as expected.

```
--- a/src/loop_in_chunks.cpp
+++ b/src/loop_in_chunks.cpp
@@ -357,6 +357,12 @@ void fields::loop_in_chunks(field_chunkloop chunkloop, void *chunkloop_data, con
 
   vec s0(gv.dim), e0(gv.dim), s1(gv.dim), e1(gv.dim);
   compute_boundary_weights(gv, where, is, ie, snap_empty_dimensions, s0, e0, s1, e1);
+  printf("is=(%d, %d), ie=(%d, %d)\n",
+         is.r(), is.z(), ie.r(), ie.z());
+  printf("s0=(%0.4f, %0.4f), e0=(%0.4f, %0.4f)\n",
+         s0.r(), s0.z(), e0.r(), e0.z());
+  printf("s1=(%0.4f, %0.4f), e1=(%0.4f, %0.4f)\n",
+         s1.r(), s1.z(), e1.r(), e1.z());

   int original_vol = 1;
@@ -377,6 +383,11 @@ void fields::loop_in_chunks(field_chunkloop chunkloop, void *chunkloop_data, con
   for (int sn = 0; sn < (use_symmetry ? S.multiplicity() : 1); ++sn) {
     component cS = S.transform(cgrid, -sn);
     volume gvS = S.transform(gv.surroundings(), sn);
+
+    printf("gvS:, min_corner=(%0.5f, %0.5f), max_corner=(%0.5f, %0.5f)\n",
+           gvS.get_min_corner().r(), gvS.get_min_corner().z(),
+           gvS.get_max_corner().r(), gvS.get_max_corner().z());
     vec L(gv.dim);
     ivec iL(gv.dim);
 
@@ -422,6 +433,8 @@ void fields::loop_in_chunks(field_chunkloop chunkloop, void *chunkloop_data, con
         ivec _iecoS(S.transform(gvu.big_owned_corner(cS), sn));
         ivec iscoS(max(user_volume.little_owned_corner(cgrid), min(_iscoS, _iecoS))),
             iecoS(max(_iscoS, _iecoS)); // fix ordering due to to transform
+        printf("iscoS=(%d, %d), iecoS=(%d, %d)\n",
+               iscoS.r(), iscoS.z(), iecoS.r(), iecoS.z());
```

1. $r = 0$
```
is=(-1, -38), ie=(1, -36)
s0=(0.5000, 0.7500), e0=(0.5000, 0.2500)
s1=(0.5000, 0.2500), e1=(0.5000, 0.7500)
```
2. $r = 0.5\Delta r$
```
is=(1, -38), ie=(1, -36)
s0=(1.0000, 0.7500), e0=(1.0000, 0.2500)
s1=(1.0000, 0.2500), e1=(1.0000, 0.7500)
```
3. $r = 1.0\Delta r$
```
is=(1, -38), ie=(3, -36)
s0=(0.5000, 0.7500), e0=(0.5000, 0.2500)
s1=(0.5000, 0.2500), e1=(0.5000, 0.7500)
```
4. $r = 1.5\Delta r$
```
is=(3, -38), ie=(3, -36)
s0=(1.0000, 0.7500), e0=(1.0000, 0.2500)
s1=(1.0000, 0.2500), e1=(1.0000, 0.7500)
```
5. $r = 2.0\Delta r$
```
is=(3, -38), ie=(5, -36)
s0=(0.5000, 0.7500), e0=(0.5000, 0.2500)
s1=(0.5000, 0.2500), e1=(0.5000, 0.7500)
```

For all five test cases, the output is *always*:
```
gvS:, min_corner=(0.00000, -0.84000), max_corner=(6.52000, 0.88000)

iscoS=(299, 18), iecoS=(325, 44)
iscoS=(299, -40), iecoS=(325, 16)
iscoS=(1, 18), iecoS=(297, 44)
iscoS=(1, -40), iecoS=(297, 16)
```

oskooi:
Results from a separate experiment involving an $E_r$ point source at various positions along $r$ and investigating the convergence with resolution of the radiated flux via `add_flux` (column labeled `out_flux`), total flux via `dft_ldos` (column labeled `total_flux`), and extraction efficiency (column labeled `ext_eff`). The findings demonstrate that the extraction efficiency is incorrect whenever $E_r$ is restricted to any grid point within the **first voxel**.

The output is organized into four columns:
```
resolution, out_flux, total_flux, ext_eff
```
Resolutions tested were 25, 50, 100, and 200 pixels/μm.


1. $r = 0$, 2 restricted points (*should be 4 restricted points*)  [ incorrect result ]

```
25,  3.369926, 0.979566, 3.440224
50,  3.040429, 0.239534, 12.693084
100, 2.996429, 0.062598, 47.867953
200, 2.987436, 0.016272, 183.593285
```

2. $r = 0.5\Delta r$, 2 restricted points [ incorrect result ]

```
25,  13.479703, 3.918263, 3.440224
50,  12.161716, 0.958137, 12.693084
100, 11.985716, 0.250391, 47.867953
```

3. $r = 1.0\Delta r$, 4 restricted points [ incorrect result ]

```
25,  7.945538, 14.760318, 0.538304
50,  4.183601,  3.775316, 1.108146
100, 3.271576,  0.997740, 3.278987
200, 3.056390,  0.260119, 11.749961
```

4. $r = 1.3\Delta r$, restricted points [ incorrect result ]

```
25,  8.698912, 24.393588, 0.356606
50,  2.534997,  6.344131, 0.399581
100, 0.973209,  1.683729, 0.578008
200, 0.601769,  0.439425, 1.369446
```

5. $r = 1.4\Delta r$, 4 restricted points [ incorrect result ]

```
25,  9.629016, 28.103474, 0.342627
50,  2.508571,  7.345236, 0.341524
100, 0.695439,  1.951878, 0.356292
200, 0.263794,  0.509565, 0.517685
```

6. $r = 1.5\Delta r$, 2 restricted points  [ correct result ]

```
25,  10.898609, 32.056495, 0.339981
50,   2.743680,  8.418313, 0.325918
100,  0.661845,  2.239738, 0.295501
```

7. $r = 1.8\Delta r$, 4 restricted points  [ correct result ]

```
25, 14.878816, 42.740855, 0.348117
50,  3.897262, 11.874070, 0.328216
100, 0.949759,  3.207856, 0.296073
```

8. $r = 2.0\Delta r$, 4 restricted points  [ correct result ]

```
25, 17.881969, 50.531040, 0.353881
50,  4.778999, 14.487701, 0.329866
100, 1.170510,  3.948137, 0.296471
```

Cases 6-8 produce the correct result for the extraction efficiency because they do *not* involve restricting to $E_r$ grid points within the first voxel unlike Cases 1-5.

oskooi:
Results for the experiment above but computing the total flux using `add_flux` rather than `dft_ldos`. The columns in the output for the various runs shown below are:
```
resolution, radiated flux, total flux, (radiated flux) / (total flux)
```

The ratio of the radiated to the total flux is converging with resolution to the same constant value for all dipole positions. 

1. $r = 0$
```
25,  3.369926, 4.007910, 0.840819
50,  3.040429, 3.201351, 0.949733
100, 2.996429, 3.040487, 0.985510
```

2. $r = 0.5\Delta r$
```
25,  13.479703, 16.031640, 0.840819
50,  12.161716, 12.805405, 0.949733
100, 11.985716, 12.161949, 0.985510
```
3. $r = 1.0\Delta r$
```
25,  7.945538, 17.276357, 0.459908
50,  4.183601,  6.702135, 0.624219
100, 3.271576,  3.972616, 0.823532
200, 3.056390,  3.242241, 0.942678
```

stevengj:
Since the remaining problems are simply an overall normalization factor for Er delta function sources at (or near) r=0, I think we can go ahead an merge.

PR #2544, opened 2023-06-07
Add type hints and update docstrings for adjoint filters

oskooi:
Various readability improvements for the adjoint filters.

### DIFF
--- a/python/adjoint/filters.py
+++ b/python/adjoint/filters.py
@@ -1,35 +1,51 @@
 """
-General filter functions to be used in other projection and morphological transform routines.
+A collection of routines for use in topology optimization comprising
+filters, projection operators, and morphological transforms.
 """
+from typing import List, Union
+
+from autograd import numpy as npa
 import meep as mp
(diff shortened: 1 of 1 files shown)


python/adjoint/filters.py
```diff
@@ -1,35 +1,51 @@
...
-from autograd import numpy as npa
 from scipy import signal, special
 
+ArrayLikeType = Union[List, Tuple, np.ndarray]
+
 
-def _centered(arr, newshape):
-    """Helper function that reformats the padded array of the fft filter operation.
-    Borrowed from scipy:
+def _centered(arr: np.ndarray = None, newshape: ArrayLikeType = None) -> np.ndarray:
```
smartalecH, line 15: Why all the default `None` arguments? Shouldn't these just be required args (no default)?

oskooi: Fixed.

smartalecH (approved):
LGTM so long as tests pass. Thanks for doing this.

PR #2556, opened 2023-06-21
add docstrings to adjoint solver

mochen4:
Add docstrings to ``objective.py`` and ``optimization_problem.py``

### DIFF
--- a/python/adjoint/objective.py
+++ b/python/adjoint/objective.py
@@ -533,11 +533,17 @@ class Near2FarFields(ObjectiveQuantity):
 class LDOS(ObjectiveQuantity):
     """A differentiable LDOS"""
 
-    def __init__(self, sim: mp.Simulation, **kwargs):
+    def __init__(
+        self, sim: mp.Simulation, decimation_factor: Optional[int] = 0, **kwargs
+    ):
         """Initialize a differentiable LDOS instance
 
         Args:
           sim: the Meep simulation object of the problem.
+          decimation_factor: An integer used to specify the number of timesteps between updates of
+            the DFT fields. The default is 0, at which the value is automatically determined from the
+            Nyquist rate of the bandwidth-limited sources and the DFT monitor. It can be turned off
+            by setting it to 1.
         """
         super().__init__(sim)
         self.srckwarg = kwargs
--- a/python/adjoint/objective.py
+++ b/python/adjoint/objective.py
@@ -147,7 +147,7 @@ class ObjectiveQuantity(abc.ABC):
 
 
 class EigenmodeCoefficient(ObjectiveQuantity):
-    """A frequency-dependent eigenmode coefficient.
+    """A differentiable frequency-dependent eigenmode coefficient.
     Attributes:
         volume: the volume over which the eigenmode coefficient is calculated.
         mode: the eigenmode number.
@@ -158,6 +158,10 @@ class EigenmodeCoefficient(ObjectiveQuantity):
         kpoint_func_overlap_idx: the index of the mode coefficient to return when
           specifying `kpoint_func`. When specified, this overrides the effect of
           `forward` and should have a value of either 0 or 1.
+        decimation_factor: an integer so that the DFT fields are updated at every
+          decimation_factor timesteps. The default is 0, at which the value is
+          automatically determined from the Nyquist rate of the bandwidth-limited
+          sources and the DFT monitor. It can be turned off by setting it to 1.
         subtracted_dft_fields: the DFT fields obtained using `get_flux_data` from
           a previous normalization run. This is subtracted from the DFT fields
           of this mode monitor in order to improve the accuracy of the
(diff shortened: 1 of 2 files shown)


python/adjoint/objective.py
```diff
@@ -284,11 +293,30 @@ def __call__(self):
 
 
 class FourierFields(ObjectiveQuantity):
+    """A differentiable frequency-dependent Fourier Fields (dft_fields)
+    Attributes:
```
oskooi, line 297: Replace "Attributes:" with "Args:" and separate it with the one-line summary above it with a line break following the style guide: https://google.github.io/styleguide/pyguide.html#s3.8.3-functions-and-methods.

python/adjoint/objective.py
```diff
@@ -284,11 +293,30 @@ def __call__(self):
...
+        volume: the volume over which the Fourier Fields are calculated. Due to an unsolved bug,
```
oskooi, line 298: "Fields" should be lowercase. Change "unsolved" to "unresolved."

python/adjoint/objective.py
```diff
@@ -284,11 +293,30 @@ def __call__(self):
...
+          the size must not be zero in at least one direction.
+        component: field component (e.g. mp.Ex, mp.Hz, etc.)
+        yee_grid: if True, the Fourier fields evaluated at corresponding Yee grid point
```
oskooi, line 301: ...at *the* corresponding Yee grid point...

python/adjoint/objective.py
```diff
@@ -284,11 +293,30 @@ def __call__(self):
...
+          are returned; otherwise, interpolated fields at the center of each voxel
+          are returned
+        decimation_factor: an integer so that the DFT fields are updated at every
```
oskooi, line 304: An integer used to specify the number of timesteps between updates of the DFT fields.

python/adjoint/objective.py
```diff
@@ -385,6 +420,24 @@ def __call__(self):
 
 
 class Near2FarFields(ObjectiveQuantity):
+    """A differentiable near2far field transformation
+    Attributes:
```
oskooi, line 424: Args:

python/adjoint/optimization_problem.py
```diff
@@ -21,6 +21,27 @@ class OptimizationProblem:
     calculation) and optionally its gradient (adjoint calculation).
     This is done by the __call__ method.
 
+    Attributes:
```
oskooi, line 24: Args:

oskooi (changes requested):

python/adjoint/objective.py
oskooi: `Attributes` is used for classes and `Args` for functions.

For the `Attributes` section of the docstrings for the `EigenmodeCoefficient` class, we would need to list its *public* members, almost anything with a `self.` prefix defined in the constructor excluding `self._` which are generally used for private members.

We should also add a separate `Args` section to the constructor. The argument descriptions in the constructor docstrings should hopefully be different than the attributes description in the class docstring. The attributes description should be a general summary whereas the argument description is less so.

mochen4: I added separate ``Args`` sections to the constructors. I tried to make the attribute description more concise and general, and argument description more detailed and elaborated. But they still seem a little redundant to me.

python/adjoint/optimization_problem.py
```diff
@@ -123,7 +166,29 @@ def __call__(
...
+              and gradients will then be evaluated at this configuration of design variables.
+            need_value: whether forward evaluations are needed. Default is True.
+            need_gradient: whether adjoint and gradients evaluatiosn are needed. Default is True.
+            beta: the strength of projection of rho_vector. Default to None.
+
+        Returns:
+            A tuple (f0, gradient) where:
+            f0 is the list of objective functions values when design variables
+              are set to rho_vector
+            gradient is the lists of gradients of objective functions with respect to
```
stevengj, line 183: Clarify that `gradient` is a list (over objective functions) of lists (over design regions) of 2d arrays (design vars x frequencies) of derivatives.   If there is only a single objective function, the outer 1-element list is replaced by just that element, and similarly if there is only one design region, while if there is only 1 frequency then the innermost array is squeezed to a 1d array.  (So, for example, if you have only a single objective function, a single design region, and a single frequency, then `gradient` is simply a 1d array of the derivatives.)

python/adjoint/optimization_problem.py
```diff
@@ -21,6 +21,26 @@ class OptimizationProblem:
     calculation) and optionally its gradient (adjoint calculation).
     This is done by the __call__ method.
 
+    Attributes:
+        sim: the corresponding Meep `Simulation` object of the problem
+        objective_functions: list of objective functions on objective_arguments
```
stevengj, line 26:
```suggestion
        objective_functions: list of differentiable functions (callable objects) whose arguments are given by objective_arguments
```

PR #2569, opened 2023-06-29
Describe an additional potential fix for divergent fields due to negative-epsilon materials

oskooi:
Adds an addendum to one of the five proposed solutions of the FAQ: [Why are the fields blowing up in my simulation?](https://meep.readthedocs.io/en/latest/FAQ/#why-are-the-fields-blowing-up-in-my-simulation)


doc/docs/FAQ.md
```diff
@@ -203,6 +203,8 @@ Usage: Fields
...
 
+Regarding (3), if the fields are still blowing up when overlapping the susceptibility material with a PML or Absorber, consider *removing* the absorbing boundary layer on that side to preclude its overlap with the material. If the material is strongly reflective or absorptive, there are likely no transmitted fields. In this case, a PML or Absorber is unnecessary.
```
stevengj, line 206:
```suggestion
As another possibility, if an *entire* boundary consists of a material where $\varepsilon$ has a negative real part (and $\mu$ is positive), or is sufficiently lossy, then realize that you don't need a PML at all — all waves are exponentially decaying in such a material, so you just need to make the material sufficiently thick at the boundary.
```

PR #2583, opened 2023-07-20
refactor KDEF macros into meep_internals.hpp

stevengj:
We want to use them in `step_db.cpp` for the cylindrical updates.


stevengj:
Regarding https://github.com/NanoComp/meep/pull/2538#issuecomment-1639485464, after this PR you should just need `DEF_k` and `DEF_ku` to replace `int k = ...` and `int ku = ...`

PR #2585, opened 2023-07-21
Replace custom loops for im/r update of `step_db.cpp` with `LOOP_OVER_VOL_OWNED0`

oskooi:
Following a suggestion by @stevengj in [#2538 (comment)](https://github.com/NanoComp/meep/pull/2538#issuecomment-1634805342) which had #2583 as a prerequisite, this PR replaces the eight separate custom loops for the $+im/r$ update of `step_db.cpp` with `LOOP_OVER_VOL_OWNED0`.

### DIFF
--- a/src/step_db.cpp
+++ b/src/step_db.cpp
@@ -171,105 +171,96 @@ bool fields_chunk::step_db(field_type ft) {
         const int dku = gv.iyee_shift(cc).in_direction(dsigu);
         const realnum the_m =
             m * (1 - 2 * cmp) * (1 - 2 * (ft == B_stuff)) * (1 - 2 * (d_c == R)) * Courant;
-        const realnum ir0 = gv.origin_r() * gv.a + 0.5 * gv.iyee_shift(cc).in_direction(R);
-        int sr = gv.nz() + 1;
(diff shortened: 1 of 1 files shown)


oskooi:
I think the implementation is now correct (mainly because it is rather straightforward). However, some of the tests are failing because the results have changed slightly and not, importantly, because the fields are blowing up. 

The slight change in results may be an indication that these test failures are not due to the changes introduced by this PR but rather due to a bug somewhere else, i.e. in the special updates equations for $|m| =1$. 

What is unexpected is that looping over just the **owned** grid points (as in this PR) rather than also the not-owned grid points was expected *not* to change the results per @stevengj in [#2538 (comment)](https://github.com/NanoComp/meep/pull/2538#issuecomment-1634805342):

```
But updating the not-owned points here should be harmless because they 
will just get overwritten by `step_boundaries`. 
```

oskooi:
The pre-commit failure is not related to this PR since the logs indicate that the problem is with the notebooks in the `example` folder. Also, I was not able to reproduce the pre-commit failures on the my local machine by running `pre-commit run --all-files` per instructions in the logs. I think we can therefore ignore these errors.

PR #2599, opened 2023-07-31
Allow design-grid resolution to accept a tuple

mawc2019:
This PR allows the `resolution` argument in [filters.py](https://github.com/NanoComp/meep/blob/master/python/adjoint/filters.py) to accept a tuple, so that different resolutions along different directions can be supported. This PR also refactors the code in filters.py.

### DIFF
--- a/python/adjoint/filters.py
+++ b/python/adjoint/filters.py
@@ -235,7 +235,8 @@ def mesh_grid(
 
     if Nx <= 1 and Ny <= 1:
         raise AssertionError(
-            "The grid size is improper. Check the size and resolution of the design region.")
+            "The grid size is improper. Check the size and resolution of the design region."
+        )
 
     xv = np.arange(0, Lx / 2, 1 / resolution[0]) if resolution[0] > 0 else [0]
     yv = np.arange(0, Ly / 2, 1 / resolution[1]) if resolution[1] > 0 else [0]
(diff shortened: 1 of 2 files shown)


python/adjoint/filters.py
```diff
@@ -182,12 +181,90 @@ def convolve_design_weights_and_kernel(
...
+    """Converts input design-grid resolution to the acceptable format.
+
+    Args:
+        resolution: number of list of numbers representing design-grid
+                    resolution, allowing anisotropic resolution.
+
+    Returns:
+        A two-element tuple composed of the resolution in x and y directions.
+    """
+    if isinstance(resolution, (tuple, list, np.ndarray)):
```
oskooi, line 194: `tuple` and `list` should be `Tuple` and `List`, respectively.

oskooi (changes requested):
Would be good to include a unit test for this feature. One suggestion would be to simply verify that defining a filter with the resolution tuple `(res, res)` produces the same result as passing in just `res`.

mawc2019:
> One suggestion would be to simply verify that defining a filter with the resolution tuple `(res, res)` produces the same result as passing in just `res`.

I have added a different but more stringent test, in which one of the elements of the `resolution` tuple is zero, which defines a 1d design pattern. Suppose a 1d design pattern is defined in the x direction. Whether this 1d design pattern is periodic in the y direction or not should make no difference in the filtered 1d patterns. Likewise, suppose a 1d design pattern is defined in the y direction. There should be no difference whether this 1d design pattern is periodic in the x direction. The new test function checks this identity based on lengthscale constraint functions.

python/tests/test_adjoint_solver.py
```diff
@@ -1152,6 +1152,99 @@ def test_periodic_design(self):
 
             print(f"PASSED: filter function = {selected_filter.__name__}")
 
+    def test_anisotropic_resolution(self):
```
oskooi, line 1155: Suggest renaming to `test_nonuniform_grid_resolution` to be more precise and avoid confusion with more common usage of anisotropic to refer to ε.

oskooi (commented):
LGTM.

python/tests/test_adjoint_solver.py
mawc2019: Your suggested name is clearer, but it still seems misleading because nonuniform resolution may also refer to something else, such as the case below. Shall we use a longer name such as `test_unequal_horizontal_vertical_resolution`?
image

oskooi (approved):

PR #2609, opened 2023-08-15
bfast algorithm for fixed angle broadband

### LINKED ISSUE #1991: Fixed-angle broadband simulations
**Lumerical FDTD Solutions** supports BFAST source, which enable broad band simulation with oblique incident wave.

[BFAST](https://support.lumerical.com/hc/en-us/articles/360034902273-Broadband-Fixed-Angle-Source-Technique-BFAST-)

Is there any similar implementation in meep?

Dan2357:
Fixes #1991 

This is a work in progress which implements the bfast algorithm adapted for MEEP's UPML formulation. Will post mathematical notes soon.

To do:
- [x] get it working on a minimal test case
- [ ] add tests 
- [ ] documentation
- [x] remove changes to example files
- [x] convert mathematical documentation to markdown


oskooi:
Thanks for working on this feature. One suggestion for a quick test in 1d would be computing the reflectance from a flat interface and verifying the result using the Fresnel equations as demonstrated in this [tutorial](https://meep.readthedocs.io/en/latest/Python_Tutorials/Basics/#angular-reflectance-spectrum-of-a-planar-interface).

oskooi:
In case you missed it, there is a type-conversion error in `step_db.cpp` in the latest commit when running `make` for this branch during the CI as reported in the [test logs](https://github.com/NanoComp/meep/actions/runs/5992048927/job/16251330469?pr=2609):

```
../../../src/step_db.cpp: In member function ‘bool meep::fields_chunk::step_db(meep::field_type)’:
../../../src/step_db.cpp:129:36: error: conversion from ‘vector<double>’ to non-scalar type ‘vector<float>’ requested
  129 |           std::vector<realnum> k = bfast_k_bar;
      |                                    ^~~~~~~~~~~
``` 
You can view these test logs by clicking on the `Details` button for the failing test.

stevengj:
@Dan2357, great to see that you are making progress on this.  Let me know if you want to meet sometime.

stevengj:
@oskooi said he will write the user documentation / tutorial and a test (based on your example — just comparing the reflection to the analytical Fresnel coefficients).

@Dan2357 is going to tweak the technical writeup a bit — expand the intro, and add a section about how this was implemented in Meep as a separate `step_bfast` function that just adds the new terms, so that we don't need to modify the main timestepping code.     He'll also delete the obsolete TeX and PDF files from this PR since everything is now in the `.md` file.

We should also finalize the API — probably the `needs_bfast` option can be removed.  `kbar=None` will the be default, and if you pass any other `kbar` that will enable the bfast option.  Not sure if we should call it `kbar` or something else?  (It's just k/ω)

(The technical writeup will then be linked into the main manual somewhere.)

Dan2357:
I have edited the technical writeup and deleted the unnecessary files.

oskooi:
As reported in #2722, I think there could be a bug in the current implementation related to not taking into account the *index of the source medium* when specifying the Bloch-periodic wavevector internally. Note that in the notebook demonstration in this branch, the source medium is air which produces the correct reflectance matching the Fresnel equations whereas for the unit test in #2722 the source medium is $n = 1.4$ and the results do not agree.

As additional information, in the unit test in #2722 I am using:
```
bfast_k_bar = (np.sin(theta_rad), 0, 0)
```
which produces the wrong results. If I modify this to include the source medium:
```
bfast_k_bar = (self.n1 * np.sin(theta_rad), 0, 0)
```
the fields blow up. 

The bug fix therefore likely needs to be made internally rather than by the user when specifying `bfast_k_bar`.

doc/bfast/fixed_angle_broadband_simulations_in_Meep.md
```diff
@@ -0,0 +1,154 @@
...
+```math
+\vec{K} = -i\omega(1+\frac{i\sigma_{D}}{\omega})\vec{C} - i\omega\vec{F}. \tag{9}
+```
+This new field satisfies the equation:
+```math
+\vec{F} = \vec{\bar{k}}\times\vec{H}, \tag{10}
+```
+where
+```math
+\vec{\bar{k}} =\frac{1}{\omega}\begin{pmatrix} k_{x} \\ k_{y}\\0 \end{pmatrix} \ =\begin{pmatrix} \sin{\theta}\cos{\phi} \\ \sin{\theta}\sin{\phi}\\0 \end{pmatrix} \ \tag{11}
```
oskooi, line 82: I think this equation is missing $n$ (the refractive index of the source medium) in the numerator.

Dan2357: Thanks, I've added that now.

stevengj:
Note that, in the longer run, it would be good to have a option that lets you put the exp(ikx) factor back into the Fourier-transformed fields, e.g. for mode extraction or visualization.  (This can only be done in the DFT fields, not the time-domain fields, because `k = k̄ω` is ω-dependent.)

Not something for this PR, just something to keep in mind for the future.

stevengj:
Could we correct the Courant factor automatically when kbar is specified?

doc/bfast/fixed_angle_broadband_simulations_in_Meep.md
```diff
@@ -0,0 +1,154 @@
...
+```
+Similarly, all other fields which depend on the value of $\vec{C}$ have new terms added to them, multiplied by the relevant conductivity. Therefore the original functions in MEEP remain unchanged.
+
+Stability
+=========
+
+As the incident angle increases, the maximum possible $\Delta t$ value
+decreases, following the formula:
+```math
+\frac{c\Delta t}{\Delta x} \leq \frac{(1-sin(\theta))}{\sqrt{D}} \tag{17}
```
stevengj, line 137:
```suggestion
\frac{c\Delta t}{\Delta x} \leq \frac{(1-\sin(\theta))}{\sqrt{D}} \tag{17}
```

doc/bfast/fixed_angle_broadband_simulations_in_Meep.md
```diff
@@ -0,0 +1,154 @@
...
+Similarly, all other fields which depend on the value of $\vec{C}$ have new terms added to them, multiplied by the relevant conductivity. Therefore the original functions in MEEP remain unchanged.
+
+Stability
+=========
+
+As the incident angle increases, the maximum possible $\Delta t$ value
+decreases, following the formula:
+```math
+\frac{c\Delta t}{\Delta x} \leq \frac{(1-\sin(\theta))}{\sqrt{D}} \tag{17}
+```
```
stevengj, line 138: It's a little surprising to me that this depends on sin(θ) and not sin(θ)^2 — I would have thought that the stability condition would be the same for θ and –θ.

Let's double-check this formula against the bfast literature?

stevengj: Nope, that is exactly the formula in the paper:

image

stevengj:
Let's merge this for now, and maybe update the API in future PR to (1) eliminate `needs_bfast`, (2) change `bfast_kbar` to `bfast_scaled_k`, and (3) automatically update the Courant factor.

PR #2612, opened 2023-08-17
Remove `reinterpret_cast` to `std::complex<double> *` in fields::process_incoming_chunk_data`

### LINKED ISSUE #2611: Occasional errors in `fields::process_incoming_chunk_data` from multiplying two `std::complex<realnum>` numbers
Occasionally, I detect errors in the unit test [`python/tests/test_refl_angular.py`](https://github.com/NanoComp/meep/blob/master/python/tests/test_refl_angular.py) which is traced to line 190 of `fields::process_incoming_chunk_data` involving the multiplication of two complex numbers:

https://github.com/NanoComp/meep/blob/dad3e5273e92f45f66991397e05e65c821cee55a/src/step.cpp#L172-L196

The problem, I suspect, is actually in the call to `reinterpret_cast` which is supposed to convert `const realnum` to `const std::complex<realnum>` and then use the result in the multiplication on line 190:

https://github.com/NanoComp/meep/blob/dad3e5273e92f45f66991397e05e65c821cee55a/src/step.cpp#L182-L183

A snippet of the stack trace shows where the error is triggered:
```
...
on time step 103902 (time=129.877), 0.00696845 s/step
on time step 104413 (time=130.516), 0.00783793 s/step
on time step 105022 (time=131.278), 0.00657031 s/step
*** SIGTERM received by PID 1500 (TID 1500) on cpu 68 from PID 1; stack trace: ***
PC: @     0x556870995664  (unknown)  std::__u::operator*<>()
    @     0x55688065707a       1152  FailureSignalHandler()
    @     0x7f9de4deb1c0  (unknown)  (unknown)
    @     0x556870995664          8  std::__u::operator*<>()
    @     0x556870a15111        128  meep::fields::process_incoming_chunk_data()
    @     0x556870a14edd        208  meep::fields::step_boundaries()
    @     0x556870a141f1        112  meep::fields::step()
    @     0x7f9dde21fbaf         80  _wrap_fields_step()
    @     0x55688052b589         32  cfunction_call
    @     0x5568804e860d         96  _PyObject_MakeTpCall
    @     0x5568805ecbe1        112  call_function
    @     0x5568805e8866        400  _PyEval_EvalFrameDefault
    @     0x5568805e3aa4         48  _PyEval_Vector
...
```

oskooi:
Fixes #2611.

According to this [Stackoverflow thread](https://stackoverflow.com/questions/69591371/reinterpreting-array-of-doubles-as-array-of-stdcomplexdouble), the C++ standard does not define the conversion of a `double *` to `std::complex<double> *` using `reinterpret_cast`. (The conversion in the *reverse* direction is defined, however.) This is probably why certain compilers have encountered the error reported in #2611.

### DIFF
--- a/src/step.cpp
+++ b/src/step.cpp
@@ -179,15 +179,15 @@ void fields::process_incoming_chunk_data(field_type ft, const chunk_pair &comm_p
     const comms_key key = {ft, CONNECT_PHASE, comm_pair};
     size_t num_transfers = get_comm_size(key) / 2; // Two realnums per complex
     if (num_transfers) {
-      const std::complex<realnum> *pair_comm_block_complex =
-          reinterpret_cast<const std::complex<realnum> *>(pair_comm_block);
       const std::vector<realnum *> &incoming_connection =
           chunks[this_chunk_idx]->connections_in.at(key);
       const std::vector<std::complex<realnum> > &connection_phase_for_ft =
           chunks[this_chunk_idx]->connection_phases[key];
 
       for (size_t n = 0; n < num_transfers; ++n) {
-        std::complex<realnum> temp = connection_phase_for_ft[n] * pair_comm_block_complex[n];
+        std::complex<realnum> temp =
+            connection_phase_for_ft[n] *
+            std::complex<realnum>(pair_comm_block[2 * n], pair_comm_block[2 * n + 1]);
         *(incoming_connection[2 * n]) = temp.real();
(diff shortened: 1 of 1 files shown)


stevengj:
I'm confused as to why this cast would not be allowed?

stevengj:
In any case, the new version is fine and has the virtue of eliminating pointer casts.

PR #2615, opened 2023-08-25
Provide additional details for calculating phase of reflection coefficient in tutorial

oskooi:
Adds two important details to  [Tutorial/Mode Decomposition/Phase of a Total Internal Reflected Mode](https://meep.readthedocs.io/en/latest/Python_Tutorials/Mode_Decomposition/#phase-of-a-total-internal-reflected-mode) involving: (1) the definition of the phase as the argument of the complex mode coefficient and (2) the convention used to define the phase based on the sign of the mode coefficients.


doc/docs/Python_Tutorials/Mode_Decomposition.md
```diff
@@ -148,13 +148,13 @@ Phase of a Total Internal Reflected Mode
...
+The calculation of the reflection coefficient requires two separate runs: (1) a normalization run involving just the source medium ($n_1$) to obtain the incident fields, and (2) the main run including the interface whereby the incident fields from (1) are first subtracted from the monitor to obtain only the reflected fields. The mode coefficient in (2) divided by (1) is, by definition, the reflection coefficient. The phase is the [argument](https://en.wikipedia.org/wiki/Argument_(complex_analysis)) of the complex reflection coefficient. Note: there is a [convention](https://en.wikipedia.org/wiki/Total_internal_reflection#Phase_shifts), followed by Meep, used to define the phase which is based on the *sign* of the mode coefficients, in particular for the $P$ polarization.
```
stevengj, line 157:
```suggestion
The calculation of the reflection coefficient requires two separate runs: (1) a normalization run involving just the source medium ($n_1$) to obtain the incident fields, and (2) the main run including the interface whereby the incident fields from (1) are first subtracted from the monitor to obtain only the reflected fields. The mode coefficient in (2) divided by (1) is, by definition, the reflection coefficient. The phase is the [argument](https://en.wikipedia.org/wiki/Argument_(complex_analysis)) of the complex reflection coefficient.  Whenever one defines a phase, it is relative to some choice of reference — in particular, the phase of the mode profile, which is chosen so that the electric field is purely real and its sign is chosen following a [typical convention](https://en.wikipedia.org/wiki/Total_internal_reflection#Phase_shifts) used for Fresnel coefficients.
```

PR #2616, opened 2023-08-25
clarify forward/backward mode coefficients

stevengj:
edit as needed

### DIFF
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -1419,7 +1419,7 @@ following fields:
 + `alpha`: the complex eigenmode coefficients as a 3d NumPy array of size
   (`len(bands)`, `flux.nfreqs`, `2`). The last/third dimension refers to modes
   propagating in the forward (+) or backward (-) directions defined relative to
-  the sign of the propagation constant β of the mode (the wavevector component in the direction perpendicular to the mode-monitor plane)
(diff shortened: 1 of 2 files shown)


doc/docs/Python_User_Interface.md
```diff
@@ -1419,7 +1419,7 @@ following fields:
 + `alpha`: the complex eigenmode coefficients as a 3d NumPy array of size
   (`len(bands)`, `flux.nfreqs`, `2`). The last/third dimension refers to modes
   propagating in the forward (+) or backward (-) directions defined relative to
-  the mode's dominant wavevector.
+  the sign of the propagation constant β of the mode (the wavevector component in the direction perpendicular to the mode-monitor plane)
```
oskooi, line 1422: This looks good but the place to make this change is not here but actually in the docstrings:

https://github.com/NanoComp/meep/blob/f12c7af91e1b095ba3282c2a17d76daa65ad2e2c/python/simulation.py#L4146-L4148

`Python_User_Interface.md` is a generated file (using `Python_User_Interface.md.in`) and so any changes will be overwritten.

oskooi (changes requested):

doc/docs/Python_User_Interface.md
stevengj, line 1422:
```suggestion
  the sign of the propagation constant β of the mode (the wavevector component in the direction perpendicular to the mode-monitor plane).
```

PR #2629, opened 2023-09-07
Tutorial for computing the radiation pattern and total flux of a disc in cylindrical coordinates

oskooi:
Adds a new tutorial demonstrating the computation of the total flux from the the radiation pattern of a disc in cylindrical coordinates.

Meep Adjoint Meetings (4)

Meep Adjoint Meetings (5)


stevengj:
If you want to do this, you'll want to have a *finite* structure (or no structure at all, just vacuum), so that all of the power is radiated.  (If you have a structure with a slab waveguide extending to infinity, like the LED examples, you will have a hard time disentangling the radiated power and the power in the waveguide mode, especially since the near2far approach cannot give you the latter).

oskooi:
> If you want to do this, you'll want to have a finite structure (or no structure at all, just vacuum), so that all of the power is radiated.

This seems to be working after I replaced the infinitely extended LED with a finite-sized disc. I also modified the polar-coordinate plot of the radiation pattern to specify 0° as the $+z$ direction (previously it was the $+r$ direction). This conforms with the standard convention used in spherical coordinates.

stevengj:
Maybe try it without a ground plane to avoid any trickiness in the near-to-far calculation arising from that?

oskooi:
> Maybe try it without a ground plane to avoid any trickiness in the near-to-far calculation arising from that? Basically, you don't have a closed "near-field surface". 

Simulating the same structure using a *closed* near-field surface by removing the ground plane and duplicating the disc and the source below the $z = 0$ plane (the method of images) does indeed produce the expected result: the difference in the flux computed from the near and far fields goes to zero with resolution.

I have added a note which describes this effect as well as  the possible workarounds mentonoed above.

led_method_of_images_simulation_layout

Issue #2614, opened 2023-08-22
Plot labels for geometric objects


Yaraslaut:
At the moment plot can add additional labels to the figure, it will point out to sources and monitors
It would be great if user can provide label for geometry entity so plot will shows this labels of the objects as well.

stevengj:
I guess an interface for this would involve:

1. add an optional string `label` field to `GeometricObject`
2. add a `labelgeometry=True` flag to `sim.plot2D`, which would simply plot the given object labels (if any) at the (2d-projected) centers of the corresponding objects.  (Or in 3d, maybe only plot them if they intersect the 2d slice.)

Yaraslaut:
> 

I can try to make a PR for this

oskooi:
> 2. add a labelgeometry=True flag to sim.plot2D, which would simply plot the given object labels (if any) at the (2d-projected) centers of the corresponding objects. (Or in 3d, maybe only plot them if they intersect the 2d slice.)

Following the [naming conventions](https://google.github.io/styleguide/pyguide.html#3162-naming-conventions) for method parameters, it would be good to use `label_geometry` rather than `labelgeometry` as the new parameter to `sim.plot2D`.

PR #2631, opened 2023-09-08
Add label for geometrical object

Yaraslaut:
Closes https://github.com/NanoComp/meep/issues/2614

PR adds labels for geometrical objects to show them on plots

Example 
```
import meep as mp
import argparse

resolution = 50   # pixels/um
lk = mp.Block(size=mp.Vector3(mp.inf,5.0,mp.inf), material=mp.Medium(epsilon=3.0), label="main block")
geometry = [blk]

for i in range(1,2):
    geometry.append(mp.Cylinder(0.2, center=mp.Vector3(2 * i), label = "right hole"))
    geometry.append(mp.Cylinder(0.2, center=mp.Vector3(-2 * i), label = "left hole"))

src = [mp.Source(mp.GaussianSource(1.0, fwidth=0.5),
                     component=mp.Ex,
                     center=mp.Vector3(0.0, -3.0,0.0),
                     size=mp.Vector3(5,0))]


sim = mp.Simulation(cell_size=mp.Vector3(10,10,0),
                    geometry=geometry,
                    boundary_layers= [mp.PML(1.0)],
                    sources= src,
                    resolution=resolution)

sim.run(until=0)
sim.plot2D(label_geometry=True,labels=True)
plt.show()

```

image

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -4874,7 +4874,7 @@ class Simulation:
             output_plane=output_plane,
             fields=fields,
             labels=labels,
-            label_geometry = label_geometry,
+            label_geometry=label_geometry,
             eps_parameters=eps_parameters,
             boundary_parameters=boundary_parameters,
             source_parameters=source_parameters,
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -4745,6 +4745,7 @@ class Simulation:
         output_plane: Optional[Volume] = None,
         fields: Optional = None,
         labels: bool = False,
+        label_geometry: bool = False,
         eps_parameters: Optional[dict] = None,
         boundary_parameters: Optional[dict] = None,
         source_parameters: Optional[dict] = None,
@@ -4796,6 +4797,8 @@ class Simulation:
           `mp.Hz`) to superimpose over the simulation geometry. Default is `None`, where
           no fields are superimposed.
         * `labels`: if `True`, then labels will appear over each of the simulation
+          elements. Defaults to `False`.
+        * `label_geometry`: if `True`, then labels will appear over each of the geometry
           elements. Defaults to `False`.
         * `eps_parameters`: a `dict` of optional plotting parameters that override the
           default parameters for the geometry.
@@ -4871,6 +4874,7 @@ class Simulation:
             output_plane=output_plane,
             fields=fields,
             labels=labels,
+            label_geometry = label_geometry,
             eps_parameters=eps_parameters,
             boundary_parameters=boundary_parameters,
             source_parameters=source_parameters,
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -4745,7 +4745,7 @@ class Simulation:
         output_plane: Optional[Volume] = None,
         fields: Optional = None,
         labels: bool = False,
-        label_geometry: bool = False,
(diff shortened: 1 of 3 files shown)


python/visualization.py
```diff
@@ -615,6 +629,33 @@ def plot_eps(
...
+                elif sim_size.y == 0:
+                    center_first = el.center.y
+                    center_second = el.center.z
+                    sim_first = sim_center.y
+                    sim_second = sim_center.z
+                elif sim_size.z == 0:
+                    center_first = el.center.x
+                    center_second = el.center.y
+                    sim_first = sim_center.x
+                    sim_second = sim_center.y
```
stevengj, line 648: Seems like we should have a function for this kind of coordinate projection?

python/visualization.py
```diff
@@ -615,6 +629,33 @@ def plot_eps(
...
+                ax = place_label(
```
stevengj, line 649: Shouldn't this only be called if `el.label` is not `None`?

Yaraslaut: Updated

python/simulation.py
```diff
@@ -4797,6 +4798,8 @@ def plot2D(
           no fields are superimposed.
         * `labels`: if `True`, then labels will appear over each of the simulation
           elements. Defaults to `False`.
+        * `label_geometry`: if `True`, then labels will appear over each of the geometry
+          elements. Defaults to `False`.
```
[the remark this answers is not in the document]
Yaraslaut: I changed default behavior

python/visualization.py
Yaraslaut: I didn't find any helper function to get this done, in `plot_volume` similar structure exist

python/visualization.py
```diff
@@ -615,6 +629,34 @@ def plot_eps(
...
+                if sim_size.x == 0:
+                    center_first = el.center.y
+                    center_second = el.center.z
+                    sim_first = sim_center.y
+                    sim_second = sim_center.z
+                elif sim_size.y == 0:
+                    center_first = el.center.y
+                    center_second = el.center.z
+                    sim_first = sim_center.y
+                    sim_second = sim_center.z
```
oskooi, line 643: This is a bug. The set of lines within `elif sim_size.y == 0` should be:
```
                    center_first = el.center.x
                    center_second = el.center.z
                    sim_first = sim_center.x
                    sim_second = sim_center.z
```
However, note that the first two `if` / `elif` statements in this series only apply when taking a 2D cross section of a 3D cell volume corresponding to `sim.dimensions` of 3.

For the case of `sim.dimensions == 2`, `sim_size.x` and `sim_size.y` must be nonzero.

oskooi (changes requested):
Would you provide some examples which demonstrate this feature is working as intended?

Yaraslaut:
> Would you provide some examples which demonstrate this feature is working as intended?

See example in the PR description

python/visualization.py
Yaraslaut: Sorry it take a while for me to realize that I missed this comment :( 
I fixed this mistake.
I am not sure what I should do about validation of the size in the case of 2D system

oskooi (approved):
LGTM.

Issue #2644, opened 2023-09-19
numerical stability in cylindrical coordinate


mochen4:
``RuntimeError: meep: simulation fields are NaN or Inf`` occured at the beginning of field updates when the source includes $r=0$ point at $|m| > 1$. I used a custom amplitude function such that the amplitude was 0 at $r=0$, and this [check](https://github.com/NanoComp/meep/blob/b065eae0491a7987983779c16115200acf0085d2/src/sources.cpp#L276) did pass.

This issue doesn't occur at $m = -1, 0, 1$. And since the amplitude should be 0 at $r=0$ for $|m| > 1$, we could just exclude the point when we define sources for now.

stevengj:
One thing to try would be to add a line after [this line](https://github.com/NanoComp/meep/blob/b065eae0491a7987983779c16115200acf0085d2/src/sources.cpp#L273C5-L273C88):
```cxx
if (amps_array[idx_vol] == 0) continue;
```

mochen4:
Here is a simple script that shows the issue:

```
import numpy as np
import meep as mp

resolution = 25
frq, dfrq = 1, 0.2
dpml, dair = 1, 1
sr, sz = dair + dpml, dpml + dair + dpml
pml_layers = [mp.PML(thickness=dpml)]
cell_size = mp.Vector3(sr, 0, sz)

Jr = lambda v3: np.sin(v3.x)
offset = 0

sources = [mp.Source(mp.GaussianSource(frq, fwidth=dfrq), component=mp.Er,
        center=mp.Vector3(0.5 * dair, 0, -0.5 * sz + dpml + 0.25*dair), size=mp.Vector3(dair-offset),
        amp_func = Jr)]
sim = mp.Simulation(cell_size=cell_size, boundary_layers=pml_layers,
    resolution=resolution, sources=sources,dimensions=mp.CYLINDRICAL, m=2,)
sim.run(until_after_sources=10)
```

The issue goes aways if there is some offset between the source and ``r=0``, e.g. ``offset = 0.15``

stevengj:
You could also just try a smaller Courant number following this comment: https://github.com/NanoComp/meep/blob/7e747d2c4de03329506dd2d31bfcc53e12460029/src/step_db.cpp#L436C30-L441

stevengj:
See also the `accurate_fields_near_cylorigin` option in the Python API, which documents this behavior.

mochen4:
[...]

Issue #2604, opened 2023-08-09
Bad interpolation of DFT fields at `r = 0` in cylindrical coordinates


oskooi:
As described in [#2538 (comment)](https://github.com/NanoComp/meep/pull/2538#issuecomment-1649118476), there seems to be a bug in the interpolation of the DFT fields to the center of the voxel at $r = 0$ in cylindrical coordinates. This bug appears in cases involving $E_\phi$ and $H_r$ at $r = 0$ for $|m| = 1$ simulation (which have [special boundary conditions](https://meep.readthedocs.io/en/latest/Exploiting_Symmetry/#cylindrical-symmetry)), e.g [`get_dft_array`](https://github.com/NanoComp/meep/pull/2538#issuecomment-1572674318) and Poynting-flux calculations (via `add_flux`) in the $z$ direction. It also affects the time-domain fields obtained via `get_array` since this routine also [interpolates fields to the center of the voxel](https://github.com/NanoComp/meep/blob/cd9a7ebe8fb4ccd47cbf9eeb1bc8d8f71b1cea1a/src/array_slice.cpp#L679).

For flux calculations involving $r = 0$, a workaround is to compute the flux manually in post processing using a near-to-far field transformation (via `add_near2far`) since this routine does not involve any interpolation. See [#2538 (comment)](https://github.com/NanoComp/meep/pull/2538#issuecomment-1669994747) and the modified function `ext_eff_cyl` in `test_ldos.py` of #2538 for an example.

cc @mochen4

oskooi:
Replaced by #2704.

PR #2663, opened 2023-10-05
tutorial for oblique planewave in cylindrical coordinate 

### LINKED ISSUE #2630: tutorial for off-axis planewaves in cylindrical coordinates
Following up on #2108, it would be good to have a cylindrical test case that shows how to make an oblique planewave in cylindrical coordinates using a [Jacobi–Anger expansion](https://en.wikipedia.org/wiki/Jacobi%E2%80%93Anger_expansion), as found in equations 5–7 of [this paper](https://math.mit.edu/~stevenj/papers/LinCa21.pdf), for example.

A simple problem would be Mie scattering from a sphere, which we already have a tutorial on.  This is nice because if you look at the scattering cross section at an oblique angle it should match that from normal incidence, so it's easy to validate.

cc @mochen4

mochen4:
Documentation and tutorial for how to have oblique incident planewave in cylindrical coordinate. Compared the scattering flux of a sphere under different angles of incidence. Closes #2630. 

The setup of the tutorial looks like
oblique_cyl

Part of the source around $r=0$ is currently excluded for $|m|>1$ due to #2644. In addition, as noted in https://github.com/NanoComp/meep/pull/2538#issuecomment-1686430983, source at $r=0$ for $|m|=1$ may also not be accurate. On the other hand, both issues should only introduce 1st order errors, which should go away with resolution. 

Here is a table of the results at various incidence angles at various cutoff of ``|m|`` at ``resolution=100``. 

|  Angle of incidence |  $$\|m\|=0$$ | $$\|m\| \le 1$$ |  $$\|m\| \le 2$$ |  $$\|m\| \le 3$$ |  $$\|m\| \le 4$$ | 
| --- | --- | --- | --- | --- | --- |
|  $$0^\circ$$ |  0.      |   17.83312867  | 17.83312867  | 17.83312867  | 17.83312867 | 
|  $$5^\circ$$ | 0.26219472  | 16.28351528  | 16.8657832   | 16.99266891 |  17.02334217 | 
|  $$7.5^\circ$$ | 0.32401024 | 15.14771059  | 16.08409591  | 16.12975925  | 16.17413882 | 
|  $$15^\circ$$ | 1.16801433 | 12.15571072  | 14.8310367   | 15.03711602  | 15.13986574 | 

A higher cutoff should have better convergence, but as mentioned in https://github.com/NanoComp/meep/issues/2644#issuecomment-1743208682,  a higher cutoff currently required excluding more pixels around $r=0$.

### DIFF
--- a/doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
+++ b/doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
@@ -455,10 +455,10 @@ As shown below, the results for the scattering cross section computed using cyli
 ![](../images/cylinder_cross_section.png#center)
 
 
-Scattering of Sphere with Oblique Planewave 
+Scattering of Sphere with Oblique Planewave
 -------------------------------------------
 
-It is also possible to launch an oblique incident planewave in cylindrical coordinate by decomposing the planewave $A_xe^{ik_xx+ik_yy}\hat{x} + A_ye^{ik_xx+ik_yy}\hat{y}$ into $\sum_m (J_r(r, m)\hat{r} + J_\phi(r, m)\hat{\phi})e^{im\phi}$. The exact expressions of $J_r(r,m)$ and $J_\phi(r,m)$ are given [here](http://github.com/zlin-opt/axisym_meta3d_inverse_design/blob/master/Implementation_of_FDFD_with_Cylindrical_Coordinates.pdf) by Zin Lin. In the simplest case of normal incidence, $J_r(r,m)$ and $J_\phi(r,m)$ are nonzero only when $m = \pm 1$, as shown in the [previous tutorial](https://meep.readthedocs.io/en/latest/Python_Tutorials/Cylindrical_Coordinates/#scattering-cross-section-of-a-finite-dielectric-cylinder). 
+It is also possible to launch an oblique incident planewave in cylindrical coordinate by decomposing the planewave $A_xe^{ik_xx+ik_yy}\hat{x} + A_ye^{ik_xx+ik_yy}\hat{y}$ into $\sum_m (J_r(r, m)\hat{r} + J_\phi(r, m)\hat{\phi})e^{im\phi}$. The exact expressions of $J_r(r,m)$ and $J_\phi(r,m)$ are given [here](http://github.com/zlin-opt/axisym_meta3d_inverse_design/blob/master/Implementation_of_FDFD_with_Cylindrical_Coordinates.pdf) by Zin Lin. In the simplest case of normal incidence, $J_r(r,m)$ and $J_\phi(r,m)$ are nonzero only when $m = \pm 1$, as shown in the [previous tutorial](https://meep.readthedocs.io/en/latest/Python_Tutorials/Cylindrical_Coordinates/#scattering-cross-section-of-a-finite-dielectric-cylinder).
 
 Given the decomposition of planewave into the sum of different current sources at each $m$, we can run individual simulations at each $m$ with their corresponding source amplitudes and record the relevant physical quantities. For quantities such fields, linearity implies that we can simply sum the results from each simulations; for quantities such as flux, orthogonality implies cross terms will be zero, and we can again simply sum the results. Moreover, simulations
 at each $m$ values are embarrassingly parallel so they can be run simultaneously.
@@ -480,7 +480,7 @@ frq_cen = 1 / wvl
 dfrq = 0.2
 nfrq = 1
 resolution, dair_fac, mrange = 50, 10, 5
-src_offset = 3/resolution # a small offset in source size 
+src_offset = 3/resolution # a small offset in source size
 dpml = 0.5 * wvl
 dair = 1.0 * wvl
 pml_layers = [mp.PML(thickness=dpml)]
@@ -506,7 +506,7 @@ for alpha_i in range(alpha_range):
     if abs(cur_m) > 1:
         src_cen = 0.5 * (r + dair_fac*dair) + 0.5*src_offset
     else:
-        src_cen = 0.5 * (r + dair_fac*dair) #+ 0.5*src_offset
+        src_cen = 0.5 * (r + dair_fac*dair)
     Jpm = lambda v3: coeff_p1 * special.jv(cur_m+1, kxy * (v3.x+src_cen)) + coeff_m1 * special.jv(cur_m-1, kxy * (v3.x+src_cen))
     Jrm = lambda v3: 1j * coeff_p1 * special.jv(cur_m+1, kxy * (v3.x+src_cen)) - 1j * coeff_m1 * special.jv(cur_m-1, kxy * (v3.x+src_cen))
 
@@ -538,7 +538,7 @@ for alpha_i in range(alpha_range):
                   center=mp.Vector3(src_cen, 0, -0.5 * sz + dpml),
                   size=mp.Vector3(r + dair_fac*dair),
                   amp_func = Jpm),]
-    
+
     sim = mp.Simulation(
         cell_size=cell_size,
         boundary_layers=pml_layers,
--- a/doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
+++ b/doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
@@ -458,14 +458,16 @@ As shown below, the results for the scattering cross section computed using cyli
 Scattering of Sphere with Oblique Planewave
 -------------------------------------------
 
-It is also possible to launch an oblique incident planewave in cylindrical coordinate by decomposing the planewave $A_xe^{ik_xx+ik_yy}\hat{x} + A_ye^{ik_xx+ik_yy}\hat{y}$ into $\sum_m (J_r(r, m)\hat{r} + J_\phi(r, m)\hat{\phi})e^{im\phi}$. The exact expressions of $J_r(r,m)$ and $J_\phi(r,m)$ are given [here](http://github.com/zlin-opt/axisym_meta3d_inverse_design/blob/master/Implementation_of_FDFD_with_Cylindrical_Coordinates.pdf) by Zin Lin. In the simplest case of normal incidence, $J_r(r,m)$ and $J_\phi(r,m)$ are nonzero only when $m = \pm 1$, as shown in the [previous tutorial](https://meep.readthedocs.io/en/latest/Python_Tutorials/Cylindrical_Coordinates/#scattering-cross-section-of-a-finite-dielectric-cylinder).
+It is also possible to launch an oblique incident planewave in cylindrical coordinate by decomposing the planewave $A_xe^{ik_xx+ik_yy}\hat{x} + A_ye^{ik_xx+ik_yy}\hat{y}$ into $\sum_m (J_r(r, m)\hat{r} + J_\phi(r, m)\hat{\phi})e^{im\phi}$ through [Jacobi-Anger expansion](https://en.wikipedia.org/wiki/Jacobi%E2%80%93Anger_expansion). The exact expressions of $J_r(r,m)$ and $J_\phi(r,m)$ are given [here](http://github.com/zlin-opt/axisym_meta3d_inverse_design/blob/master/Implementation_of_FDFD_with_Cylindrical_Coordinates.pdf) by Zin Lin. In the simplest case of normal incidence, $J_r(r,m)$ and $J_\phi(r,m)$ are nonzero only when $m = \pm 1$, as shown in the [previous tutorial](https://meep.readthedocs.io/en/latest/Python_Tutorials/Cylindrical_Coordinates/#scattering-cross-section-of-a-finite-dielectric-cylinder).
 
-Given the decomposition of planewave into the sum of different current sources at each $m$, we can run individual simulations at each $m$ with their corresponding source amplitudes and record the relevant physical quantities. For quantities such fields, linearity implies that we can simply sum the results from each simulations; for quantities such as flux, orthogonality implies cross terms will be zero, and we can again simply sum the results. Moreover, simulations
+Given the decomposition of planewave into the sum of different current sources at each $m$, we can run individual simulations at each $m$ with their corresponding source amplitudes and record the relevant physical quantities. For some quantities such as fields, linearity implies that we can simply sum the results from each simulations; for some other quantities such as flux, orthogonality implies cross terms will be zero, and we can again simply sum the results. Moreover, simulations
 at each $m$ values are embarrassingly parallel so they can be run simultaneously.
 
-On the other hand, because the source amplitudes $J_r(r,m)$ and $J_\phi(r,m)$ are generally not constant and extend to infinity, we used the principle of equivalence (for reference, see [Electromagnetic wave source condition](https://arxiv.org/pdf/1301.5366.pdf)) to create equivalent sources that are of finite sizes.
-
 We present an example below that calculates the scattered flux of a sphere. Because of the spherical symmetry, incidence at different angle should have identical results. We can thus use this feature to check our approach. Note that because of the axial symmetry in the cylindrical coordinates, we cannot distinguish different azimuthal angles but we can distinguish different polar angles. We thus simply choose our incidence to be of form $E_ye^{ik_xx}$, and we can vary the angle of incidence by varying $k_x$.
+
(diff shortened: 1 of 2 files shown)


doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
```diff
@@ -455,6 +455,152 @@ As shown below, the results for the scattering cross section computed using cyli
...
+
+Given the decomposition of planewave into the sum of different current sources at each $m$, we can run individual simulations at each $m$ with their corresponding source amplitudes and record the relevant physical quantities. For quantities such fields, linearity implies that we can simply sum the results from each simulations; for quantities such as flux, orthogonality implies cross terms will be zero, and we can again simply sum the results. Moreover, simulations
```
oskooi, line 463: Change "such fields" to "such as the DFT fields".

Change "implies cross terms will" to "implies cross terms ($\vec{E_m}^* \times \vec{H_n}$ where $m \neq n$) will""

doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
```diff
@@ -455,6 +455,152 @@ As shown below, the results for the scattering cross section computed using cyli
...
+at each $m$ values are embarrassingly parallel so they can be run simultaneously.
+
+On the other hand, because the source amplitudes $J_r(r,m)$ and $J_\phi(r,m)$ are generally not constant and extend to infinity, we have to make sure the sources are wide enough to accurately approximate the actual incident wave.
```
oskooi, line 466: Change "are wide enough" to "sufficiently extended in space"

doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
```diff
@@ -455,6 +455,152 @@ As shown below, the results for the scattering cross section computed using cyli
...
+
+We present an example below that calculates the scattered flux of a sphere. Because of the spherical symmetry, incidence at different angle should have identical results. We can thus use this feature to check our approach. Note that because of the axial symmetry in the cylindrical coordinates, we cannot distinguish different azimuthal angles but we can distinguish different polar angles. We thus simply choose our incidence to be of form $E_ye^{ik_xx}$, and we can vary the angle of incidence by varying $k_x$.
```
oskooi, line 468: Change "to be of form" to "to be of the form"

doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
```diff
@@ -455,6 +455,152 @@ As shown below, the results for the scattering cross section computed using cyli
...
+mp.verbosity(0)
+r = 0.7  # radius of sphere
+h = 2 * r  # height/diameter of sphere
+
+wvl = 2 * np.pi * r / 4
+frq_cen = 1 / wvl
+dfrq = 0.2
+nfrq = 1
+resolution, dair_fac, mrange = 50, 10, 5
+src_offset = 3/resolution # a small offset in source size
```
oskooi, line 483: The comment should reference the bug reports #2538 and #2644 for the $E_r$ source at $r = 0$ for |m| > 0.

oskooi (commented):
Thanks for putting this together.

It would be good to provide some comments on the rate of convergence of the result with $m$ in the series expansion for increasing angles of the incident planewave. We could cite the analysis from a [different tutorial](https://meep.readthedocs.io/en/latest/Python_Tutorials/Cylindrical_Coordinates/#nonaxisymmetric-dipole-sources).

doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
mochen4: I wonder whether we should try to fix #2538 or #2644 first, since It seems that they really affect the accuracy. I was hoping maybe we could fix some bugs and I can update the tutorial later; that's why I didn't reference the issues.

oskooi: Based on the workaround described in #2604, would you try sizing the line source along $r$ to *exclude* $r < 1.5 \Delta r$? In other words, the line source should extend from $r = 1.5 \Delta r$ to the edge of the cell. This just requires changing `src_offset = 3 / resolution` to `src_offset = 1.5 / resolution`. This should hopefully improve the accuracy of your results.

mochen4: I tried that before but there wasn't much improvement. Due to #2644, additional pixels have to be excluded for |m|>1, and the larger |m| is, the more pixels needed to be excluded, which also implies that having more terms in the expansion may not help.

mochen4:
Ensuring stability(#2644) and forcing complex fields (inspired by https://github.com/NanoComp/meep/pull/2726#issuecomment-1899803540) produce much better agreement:

|  Angle of incidence |  $$\|m\|=0$$ | $$\|m\| \le 1$$ |  $$\|m\| \le 2$$ |  $$\|m\| \le 3$$ |  $$\|m\| \le 4$$ | 
| --- | --- | --- | --- | --- | --- |
|  $$0^\circ$$ |  0.       |   8.89879305 | 8.89879305 | 8.89879305 | 8.89879305 |
|  $$5^\circ$$ | 0.38449451  | 8.43871285  | 8.71133751  | 8.71249049  | 8.7718544 |
|  $$10^\circ$$ | 0.66191608  | 8.28861876  | 8.71832748   | 8.72696144  | 8.75597351 |
|  $$15^\circ$$ | 1.84922004  | 7.4380022  |  8.64164363  | 8.75756958  | 8.79085675  |


Also notice that the size of source is increased to accurately represent a planewave, since there is no real planewave in cylindrical coordinate: 

Screen Shot 2024-01-25 at 09 48 09

oskooi:
Good to see this is finally working as expected. A few comments and questions:

- don't the angles from `alpha_list = [0, np.pi/36, np.pi/24, np.pi/12]` in the script correspond to 0°, 5°, 7.5°, and 15° rather than 0°, 5°, 10°, 15° in your table?
- why choose a maximum value of `m` of `M = 4` for each angle? should `M` be a constant independent of the angle of the planewave?
- in the Jacobi-Anger expansion, is it necessary to perform a sum of `m` from `-M` to `M`? why not sum from `0` to `M` where the results for the flux for the `m > 1` runs are multiplied by two? (Running the script on my machine with 2 Intel Xeon cores at 4.2 GHz takes ~30 minutes.)
- it would be useful to explain why it is necessary to set `Courant = 0.2` for all the simulations in the expansion. Does the `Courant` factor need to be a constant? What about making its value a function of `m`?

mochen4:
> Good to see this is finally working as expected. A few comments and questions:
> 
> * don't the angles from `alpha_list = [0, np.pi/36, np.pi/24, np.pi/12]` in the script correspond to 0°, 5°, **7.5°**, and 15° rather than 0°, 5°, **10°**, 15° in your table?

Thanks! I fixed the value.

> * why choose a maximum value of `m` of `M = 4` for each angle? should `M` be a constant independent of the angle of the planewave?

I chose `M=4` for simplicity. I don't think larger angles would require more terms to converge, but I am not sure.

> * in the Jacobi-Anger expansion, is it necessary to perform a sum of `m` from `-M` to `M`? why not sum from `0` to `M` where the results for the flux for the `m > 0` runs are multiplied by two? (Running the script on my machine with 2 Intel Xeon cores at 4.2 GHz takes ~40 minutes.)

There might be a correspondence between m and -m terms; I will go through the math and make sure I have the correct signs and factor.

> * it would be useful to explain why it is necessary to set `Courant = 0.2` for all the simulations in the expansion. Does the `Courant` factor need to be a constant? What about making its value a function of `m`?

Thanks! I can make it as a function of m.

> * should the line source extend into the PML to the edge of the cell?

For Cartesian cells, sources extending into the PML with the periodic BC will generate planewave; but in cylindrical coordinate that is not the case. I am not sure which way would be better.

stevengj:
In the radial direction, it would be to make the computation converge faster with the cell radius R so that you can use a smaller cell.  A basic issue is that you are just truncating the current source at R, so you are introducing an error that converges as 1/R.

One way to get faster convergence would be to make the source decay slowly in R, similar to the windowed-Green's function approach, e.g. by multiplying it by a [bump function](https://en.wikipedia.org/wiki/Bump_function).

However, an even more efficient approach would be to use the principle of equivalence: have a current source surface that *completely surrounds the object*.  Then you should be able to make the computational cell quite small (the only limitation being the inexactness of our PML in the r direction).

stevengj:
(Eventually, we would like a plot of the error vs. m, and/or the magnitude of the contributions vs m.  I would naively expect that it will converge with a power law, so plot on a log–log scale, but it will get worse for increasingly oblique angles.)

mochen4:
I managed to create the equivalent planewave sources in cylindrical coordinate. Here is the plot of Ep field at m=1, for example:

Screen Shot 2024-02-08 at 09 17 55

In comparison, with one long line source of $-\frac1{2}E_p + \frac{i}{2}E_r$ (corresponding to m=1 term of a $E_y$ source), the plot of Ep looks like:

Screen Shot 2024-02-08 at 09 28 29

Notice that the colorbar suggests that the amplitude is 0.5 for the equivalent source, but is around 0.3 for the line source. There are also more artifacts for the E fields compared to the H fields; here is a plot of the Hp field, for example:

Screen Shot 2024-02-08 at 09 29 38

On the other hand, if I use a line source of H instead, the E field will then look cleaner than the H field.

stevengj:
It should be exactly a factor of 2 smaller if you had an infinite line source, with just an electric current equal to the planewave amplitude — i.e. if you start with an infinite line equivalent-current source and then set the magnetic currents to zero.

For an oblique planewave, you would:
1. Write the E and H fields of the planewave exp(ikx)
2. Expand each exp(ikx) term in a Jacobi–Anger expansion.  Now you have the E and H fields in terms of Bessel functions
3. For each m, use the Bessel E and H fields to construct equivalent sources

mochen4:
I created the equivalent sources and the current agreement looks good:

|  Angle of incidence |  $$\|m\|=0$$ | $$\|m\| \le 1$$ |  $$\|m\| \le 2$$ |  $$\|m\| \le 3$$ |  $$\|m\| \le 4$$ | $$\|m\| \le 5$$ | 
| --- | --- | --- | --- | --- | --- | --- |
|  $$0^\circ$$ |  0.       |   34.07353489 | 34.07353489 | 34.07353489 | 34.07353489 | 34.07353489 |
|  $$5^\circ$$ | 1.17112887 | 33.20208782 | 33.95312684 | 33.9596888 | 33.95971151 | 33.95971153 |
|  $$7.5^\circ$$ | 2.498413 | 32.20815769 | 33.80061922 | 33.8324966 | 33.83274769 | 33.83274833 |
|  $$10^\circ$$ | 4.12241563 | 30.97946575 | 33.58485588 | 33.67994189 | 33.68129492 | 33.68130108 |
|  $$15^\circ$$ | 7.49597576 | 28.34131234 | 32.96926081 | 33.37707925 | 33.39075734 | 33.39090213 |

But the errors still increase slowly with angles, so there still might be a small bug somewhere. Out of an abundance of caution, is there any other test I can check?

stevengj:
Might be worth doubling the resolution to see if that reduces the angle dependence of the results.

mochen4:
Here are the results at resolution=100 and the agreement looks pretty good with relative error less than 1%

|  Angle of incidence |  $$\|m\|=0$$ | $$\|m\| \le 1$$ |  $$\|m\| \le 2$$ |  $$\|m\| \le 3$$ |  $$\|m\| \le 4$$ | $$\|m\| \le 5$$ | 
| --- | --- | --- | --- | --- | --- | --- |
|  $$0^\circ$$ |  0.        | 34.29346804 | 34.29346804  |  34.29346804 |  34.29346804  | 34.29346804 |
|  $$5^\circ$$ | 1.26307288  | 33.45160194  | 34.24236427  | 34.24941477  | 34.24943828  | 34.2494383 | 
|  $$7.5^\circ$$ | 2.69179486  | 32.49031421 |  34.16626883  | 34.20051478  | 34.20077483  | 34.20077546 | 
|  $$10^\circ$$ | 4.43472468  | 31.29992238  | 34.03998618  | 34.1421106   | 34.1435132   | 34.14351927 | 
|  $$15^\circ$$ |8.02426432  | 28.73136299  | 33.58593652  | 34.02343398  | 34.03764361 |  34.03778666 | 
|  $$30^\circ$$ | 10.4188277   | 24.67896853  | 30.80532103  | 33.60631157  | 34.085187   |  34.10838805 | 
|  $$45^\circ$$ |  8.48250479  | 21.42266376 |  28.44485281  | 32.1072768   | 33.9610158   | 34.23488559 |

doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
```diff
@@ -455,6 +455,156 @@ As shown below, the results for the scattering cross section computed using cyli
...
+            box_z1_flux = mp.get_fluxes(box_z1)
+            box_z2_flux = mp.get_fluxes(box_z2)
+            box_r_flux = mp.get_fluxes(box_r)
+
+            scatt_flux_m[alpha_i, cur_m] = box_z1_flux[0] - box_z2_flux[0] - box_r_flux[0]
+            sim.reset_meep()
+
+scatt_power_m = np.zeros((alpha_range, mrange+1))
+for i in range(mrange+1):
+  scatt_power_m[:,i] = - 2*np.sum(scatt_flux_m[:,0:(i+1)], axis=1) + scatt_flux_m[:,0]
```
oskooi, line 603: The two-space indentation is inconsistent with the four-space indentation in the rest of the script. This will produce a runtime error. Please fix.

doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
```diff
@@ -455,6 +455,156 @@ As shown below, the results for the scattering cross section computed using cyli
...
+        if alpha!=0 or cur_m == 1:
+            coeff_p1 = 0.5 * (1j)**(cur_m+1)
+            coeff_m1 = 0.5 * (1j)**(cur_m-1)
+
+            src_cen = src_size_tb/2
+            Jpm = lambda v3: coeff_p1 * special.jv(cur_m+1, kxy * (v3.x+src_cen)) + coeff_m1 * special.jv(cur_m-1, kxy * (v3.x+src_cen))
+            Jrm = lambda v3: 1j * coeff_p1 * special.jv(cur_m+1, kxy * (v3.x+src_cen)) - 1j * coeff_m1 * special.jv(cur_m-1, kxy * (v3.x+src_cen))
+            Jside = (1j)**cur_m * special.jv(cur_m, kxy*src_size_tb) * kxy/k_cen
+
+
```
oskooi, line 521: Would be helpful to add a comment to describe the setup of the equivalent sources based on the equations derived in the reference from Zin Lin.

doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
```diff
@@ -455,6 +455,156 @@ As shown below, the results for the scattering cross section computed using cyli
...
+                cell_size=cell_size,
+                geometry=geometry,
+                boundary_layers=pml_layers,
+                resolution=resolution,
+                sources=sourcesp,
+                dimensions=mp.CYLINDRICAL,
+                m=cur_m,
+                force_complex_fields = True,
+                accurate_fields_near_cylorigin=True,
+                Courant=min(0.5, 1/(abs(cur_m)+0.5)))
```
oskooi, line 577: Would be useful to add a note in the text of the tutorial text explaining the scaling of the Courant factor with `m`.

doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
```diff
@@ -455,6 +455,156 @@ As shown below, the results for the scattering cross section computed using cyli
...
+            sim = mp.Simulation(
+                cell_size=cell_size,
+                geometry=geometry,
+                boundary_layers=pml_layers,
+                resolution=resolution,
+                sources=sourcesp,
+                dimensions=mp.CYLINDRICAL,
+                m=cur_m,
+                force_complex_fields = True,
+                accurate_fields_near_cylorigin=True,
```
oskooi, line 576: Would be helpful to add a comment to explain why this is necessary.

oskooi (changes requested):

oskooi (approved):
LGTM.

PR #2695, opened 2023-10-31
Fixed use_output_directory clearing filename_prefix in all instances

bhacha:
The `simulation.use_output_directory()` function currently clears the `filename_prefix` variable whenever executed. The documentation says this should only happen when `dname` is not specified. I moved the clearing step into the if condition, so the `filename_prefix` is only cleared when `dname` is not specified.


oskooi (approved):
LGTM.

PR #2708, opened 2023-11-09
Prepare for 1.28 release

oskooi:
The next tarball release is tentatively scheduled for Thursday November 9, 2023.


stevengj:
Should be 1.27.1 since it is just bug fixes?

PR #2725, opened 2023-11-21
Add labels and use equal aspect ratio for axes in plots of radiation pattern in tutorial

oskooi:
Applies two changes to the axes in the plots of the radiation pattern in polar and 3D coordinates in [Tutorial/Radiation Pattern of a Disc in Cylindrical Coordinates](https://meep.readthedocs.io/en/latest/Python_Tutorials/Near_to_Far_Field_Spectra/#radiation-pattern-of-a-disc-in-cylindrical-coordinates): 
1. Add missing labels.
2. Use an *equal* aspect ratio for the *xyz* axes in 3D rather than the [default 4:4:3]( https://matplotlib.org/stable/api/_as_gen/mpl_toolkits.mplot3d.axes3d.Axes3D.set_box_aspect.html#mpl_toolkits.mplot3d.axes3d.Axes3D.set_box_aspect).


stevengj:
`black` wants double quotes in Python strings.

PR #2726, opened 2023-11-24
Tutorial for extraction efficiency of a collection of dipoles in a disc in cylindrical coordinates

### LINKED ISSUE #2690: LED tutorial in cylindrical coordinates
It might be nice to extend [this tutorial](https://meep.readthedocs.io/en/latest/Python_Tutorials/Custom_Source/#exploiting-linear-time-invariance-of-the-materials-and-geometry) to show how a similar calculation could be performed in cylindrical coordinates, e.g. for random currents distributed uniformly over a disk.

Even if you were doing the calculation in Cartesian coordinates, you would still only need one dipole source per radius r.  Given its emitted power P(r), the total emitted power would be ∫P(r) 2πr dr — note the r factor!   To do the same thing in cylindrical coordinates, you additionally may need to be careful about the normalization of the cylindrical "dipole" (actually ring) sources.

oskooi:
Closes #2690.

This script seems to be working and generates the radiation pattern below. The extraction efficiency computed from this radiation pattern is `0.9569`.

disc_dipoles_radiation_pattern


stevengj:
The fact that your vacuum radiation patterns are not independent of the location of the source looks wrong.

Not sure what the bug is — maybe you should have summed E and H over m *before* computing the E\*xH power?

stevengj:
Would be good to overlay the analytical formula (for a circularly polarized source).

stevengj:
A good check is that summing the fields over m and then computing the Poynting flux `ExH*` should give the same answer as computing the Poynting flux and then summing over m.

oskooi:
> A good check is that summing the fields over m and then computing the Poynting flux ExH* should give the same answer as computing the Poynting flux and then summing over m.

The procedure for summing the fields first and then computing the Poynting flux is described in [Tutorial/Nonaxisymmetric Dipoles Sources](https://meep.readthedocs.io/en/latest/Python_Tutorials/Cylindrical_Coordinates/#nonaxisymmetric-dipole-sources): 

image

Note that in Step 2, the $|m| > 0$ terms in the expansion involve taking the *real* part of the fields. Also, the fields $E_{tot}$ and $H_{tot}$ and flux $P(\theta, \phi)$ are always computed at $\phi = 0$.

Using this approach, the radiation pattern of the dipoles at $r > 0$ is significantly different than the alternative approach (shown previously) of summing the Poynting flux of the individual $m$ calculations:

image

The same simulation parameters (resolution, field decay threshold for termination, and flux threshold for truncating the Fourier-series expansion) as before were used. For the $r = 0.5 \mu m$ case, there are $M + 1 = 14$ terms in the expansion, and for $r = 1.0 \mu m$ there are $M + 1 = 23$ terms.

oskooi:
There is a subtlety in the calculation of the radiation pattern $P(\theta)$ based on summing the results for the Poynting flux $P_{r,m}$ and $P_{z,m}$ using the Fourier-series expansion over $m$. There are two ways to obtain $P(\theta)$ as shown below. The plots of the radiation pattern shown above were obtained using the method on the left. For comparison, I also tried the method on the right. The results were *identical*. This was unexpected.

Screenshot 2023-12-14 at 07 12 49

doc/docs/Python_Tutorials/Near_to_Far_Field_Spectra.md
```diff
@@ -635,6 +635,350 @@ if __name__ == "__main__":
...
+                0,
+                FARFIELD_RADIUS_UM * math.cos(theta_rad[n]),
+            ),
+        )
+        e_field[n, :] = [np.conj(far_field[j]) for j in range(3)]
+        h_field[n, :] = [far_field[j + 3] for j in range(3)]
+
+    flux_r = np.real(e_field[:, 1] * h_field[:, 2] - e_field[:, 2] * h_field[:, 1])
+    flux_z = np.real(e_field[:, 0] * h_field[:, 1] - e_field[:, 1] * h_field[:, 0])
+    flux_rz = np.sqrt(np.square(flux_r) + np.square(flux_z))
```
stevengj, line 768:
```suggestion
    flux_rz = flux_r * np.sin(theta_rad) + flux_z * np.cos(theta_rad)
```
you should be taking the dot product with `n̂` here.  (Though I think it should give the same results, since at a large radius the flux should be almost all radially outwards, I think?)

oskooi:
> Would be good to overlay the analytical formula (for a circularly polarized source).

I have added the radiation pattern for a circularly polarized electric dipole obtained using antenna theory to the plot in the [previous comment](https://github.com/NanoComp/meep/pull/2726#issuecomment-1846498776). There is good agreement between the theoretical result and the simulated radiation pattern for the dipole at $r_0 \approx 0$. This confirms that the simulation is set up correctly.

doc/docs/Python_Tutorials/Near_to_Far_Field_Spectra.md
oskooi: I verified that this suggested change does not in fact change the results.

oskooi:
It seems that to obtain agreement in the radiation pattern in vacuum for dipoles at $r > 0$ it is necessary to set `accurate_fields_near_cyl_origin` to `True` when performing the Fourier-series expansion for terms with $|m| > 1$. Also, as described in #2644, the Courant factor needs to be made smaller with $m$.  

These two modifications are:
```
    if abs(m) > 1:
        accurate_fields_near_cylorigin = True
        Courant = 1 / (abs(m) + 1)
    else:
        accurate_fields_near_cylorigin = False
        Courant = 0.5
```

With these changes, the agreement in the radiation pattern is noticeably improved compared to the [original result](https://github.com/NanoComp/meep/pull/2726#issuecomment-1846498776).

image

oskooi:
For debugging purposes, here is the script I used to generate the radiation patterns (shown above) for a dipole in vacuum at $r > 0$ using a Fourier-series expansion of the fields via $e^{im\phi}$.
```py
import math

import matplotlib
matplotlib.use('agg')
import meep as mp
import numpy as np


RESOLUTION = 200  # pixels/μm

PML_UM = 1.0
PADDING_UM = 8.0

SIZE_R = PADDING_UM + PML_UM
SIZE_Z = PML_UM + PADDING_UM + PML_UM

WAVELENGTH_UM = 1.0
FREQUENCY_CENTER = 1 / WAVELENGTH_UM

NUM_FARFIELD_PTS = 500
FARFIELD_ANGLES = np.linspace(0, 0.5 * math.pi, NUM_FARFIELD_PTS)
FARFIELD_RADIUS = 1e6 * WAVELENGTH_UM

DEBUG_OUTPUT = False


def plot_radiation_pattern_polar(radiation_pattern: np.ndarray):
    """Plots the radiation pattern in polar coordinates.                                                          
                                                                                                                  
    The angles increase clockwise with zero at the top (+z direction).                                            
                                                                                                                  
    Args:                                                                                                         
        radiation_pattern: radial flux of the far fields in polar coordinates.                                    
    """
    radiation_pattern_theory = ((np.cos(FARFIELD_ANGLES)**2 + 1) *
                                radiation_pattern[0] * 0.5)

    fig, ax = plt.subplots(subplot_kw={"projection": "polar"}, figsize=(6,6))
    ax.plot(
        FARFIELD_ANGLES,
        radiation_pattern,
        "b-",
        label="Meep"
    )
    ax.plot(
        FARFIELD_ANGLES,
        radiation_pattern_theory,
        "r-",
        label="theory"
    )
    ax.set_theta_direction(-1)
    ax.set_theta_offset(0.5 * math.pi)
    ax.set_thetalim(0, 0.5 * math.pi)
    ax.grid(True)
    ax.set_rlabel_position(22)
    ax.set_ylabel("radial flux (a.u.)")
    ax.set_title("radiation pattern in polar coordinates")
    ax.legend()

    if mp.am_master():
        fig.savefig(
            "vacuum_cyl_radpattern.png",
            dpi=150,
            bbox_inches="tight",
        )

def radiation_pattern_flux(radiation_pattern: np.ndarray) -> float:
    """Computes the Poynting flux from the radiation pattern."""

    dtheta = 0.5 * math.pi / (NUM_FARFIELD_PTS - 1)
    dphi = 2 * math.pi
    flux_far = (np.sum(radiation_pattern * np.sin(FARFIELD_ANGLES)) *
                FARFIELD_RADIUS * FARFIELD_RADIUS * dtheta * dphi)

    # We need to multiply the result by two because this integral is only
    # for θ in [0, π/2] whereas we need it to be [0, π].
    flux_far *= 2

    return flux_far


def radiation_pattern_from_farfields(
        sim: mp.Simulation, n2f_mon: mp.DftNear2Far
) -> np.ndarray:
    """Computes the radiation pattern from the far fields.                                                        
                                                                                                                  
    Args:                                                                                                         
        sim: a `Simulation` object.                                                                               
        n2f_mon: a `DftNear2Far` object returned by `Simulation.add_near2far`.                                    
                                                                                                                  
    Returns:                                                                                                      
        Array of radial Poynting flux, one for each point on the circumference of                                 
        a quarter circle with angular range of [0, π/2]. 0 is the +z direction (the "pole")
        and π/2 is +r (the "equator").
    """
    e_field = np.zeros((NUM_FARFIELD_PTS, 3), dtype=np.complex128)
    h_field = np.zeros((NUM_FARFIELD_PTS, 3), dtype=np.complex128)
    for n in range(NUM_FARFIELD_PTS):
        ff = sim.get_farfield(
            n2f_mon,
            mp.Vector3(
                FARFIELD_RADIUS * math.sin(FARFIELD_ANGLES[n]),
                0,
                FARFIELD_RADIUS * math.cos(FARFIELD_ANGLES[n])
            )
        )
        e_field[n, :] = [ff[j] for j in range(3)]
        h_field[n, :] = [ff[j + 3] for j in range(3)]

    x_flux = (np.real(np.conj(e_field[:, 1]) * h_field[:, 2] -
                      np.conj(e_field[:, 2]) * h_field[:, 1]))
    z_flux = (np.real(np.conj(e_field[:, 0]) * h_field[:, 1] -
                      np.conj(e_field[:, 1]) * h_field[:, 0]))
    r_flux = np.sqrt(np.square(x_flux) + np.square(z_flux))

    return r_flux


def point_dipole_in_vacuum(rpos: float, m: int) -> np.ndarray:
    cell_size = mp.Vector3(SIZE_R, 0, SIZE_Z)

    pml_layers = [mp.PML(thickness=PML_UM)]

    src_pt = mp.Vector3(rpos, 0, 0)
    sources = [
        mp.Source(
            src=mp.GaussianSource(
                FREQUENCY_CENTER, fwidth=0.2 * FREQUENCY_CENTER
            ),
            component=mp.Er,
            center=src_pt,
        )
    ]

    sim = mp.Simulation(
        resolution=RESOLUTION,
        cell_size=cell_size,
        dimensions=mp.CYLINDRICAL,
        sources=sources,
        boundary_layers=pml_layers,
        m=m,
    )

    # flux monitor                                                                                                
    flux_mon = sim.add_flux(
        FREQUENCY_CENTER,
        0,
        1,
        mp.FluxRegion(
            center=mp.Vector3(
                0.5 * (SIZE_R - PML_UM), 0, 0.5 * SIZE_Z - PML_UM
            ),
            size=mp.Vector3(SIZE_R - PML_UM, 0, 0)
        ),
        mp.FluxRegion(
            center=mp.Vector3(SIZE_R - PML_UM, 0, 0),
            size=mp.Vector3(0, 0, SIZE_Z - 2 * PML_UM),
        ),
        mp.FluxRegion(
            center=mp.Vector3(
                0.5 * (SIZE_R - PML_UM), 0, -0.5 * SIZE_Z + PML_UM
            ),
            size=mp.Vector3(SIZE_R - PML_UM, 0, 0),
            weight=-1.0,
        ),
    )

    # near-field monitor                                                                                          
    n2f_mon = sim.add_near2far(
        FREQUENCY_CENTER,
        0,
        1,
        mp.Near2FarRegion(
            center=mp.Vector3(
                0.5 * (SIZE_R - PML_UM), 0, 0.5 * SIZE_Z - PML_UM
            ),
            size=mp.Vector3(SIZE_R - PML_UM, 0, 0)
        ),
        mp.Near2FarRegion(
            center=mp.Vector3(SIZE_R - PML_UM, 0, 0),
            size=mp.Vector3(0, 0, SIZE_Z - 2 * PML_UM),
        ),
        mp.Near2FarRegion(
            center=mp.Vector3(
                0.5 * (SIZE_R - PML_UM), 0, -0.5 * SIZE_Z + PML_UM
            ),
            size=mp.Vector3(SIZE_R - PML_UM, 0, 0),
            weight=-1.0,
        ),
    )

    if DEBUG_OUTPUT:
        fig, ax = plt.subplots()
        sim.plot2D(ax=ax)
        if mp.am_master():
            fig.savefig(
                "cyl_simulation_layout.png", dpi=150, bbox_inches="tight"
            )

    sim.run(
            50, mp.Er, src_pt, 1e-9
        )
    )

    flux_near = mp.get_fluxes(flux_mon)[0]

    radiation_pattern = radiation_pattern_from_farfields(sim, n2f_mon)
    flux_far = radiation_pattern_flux(radiation_pattern)

    err = abs(flux_far - flux_near) / flux_near

    print(
        f"flux-m:, {rpos}, {m}, {flux_near:.6f} (near), {flux_far:.6f} (far), "
        f"{err:.6f} (error)"
    )

    return radiation_pattern


if __name__ == "__main__":
    # An Er source at r=0 must be slightly offset due to #2704.
    # Requires a single simulation with m = ±1.
    rpos = 1.5 / RESOLUTION
    m = 0
    radiation_pattern = point_dipole_in_vacuum(rpos, m)

    plot_radiation_pattern_polar(FARFIELD_RADIUS * FARFIELD_RADIUS * radiation_pattern)

    # An Er source at r > 0 requires Fourier-series expansion of the fields via exp(imϕ).
    rpos = 0.35                                                                                                   
    m = 1                                                                                                      
    flux_max = 0                                                                                                  
    flux_thresh = 1e-2                                                                                            
    radiation_pattern_total = np.zeros(NUM_FARFIELD_PTS)                                                          
    while True:                                                                                                   
        radiation_pattern = point_dipole_in_vacuum(rpos, m)                                                       
        radiation_pattern_total += radiation_pattern * (1 if m == 0 else 2)                                       
                                                                                                                  
        flux_far = radiation_pattern_flux(radiation_pattern)                                                      
        if flux_far > flux_max:                                                                                   
            flux_max = flux_far                                                                                   
                                                                                                                  
        if m > 0 and (flux_far / flux_max) < flux_thresh:                                                         
            break                                                                                                 
        else:                                                                                                     
            m += 1                                                                                                
                                                                                                                  
    flux_far = radiation_pattern_flux(radiation_pattern_total)                                                    
    print(f"flux-tot:, {rpos}, {m}, {flux_far}")                                                                  
                                                                                                                  
    plot_radiation_pattern_polar(                                                                                 
        FARFIELD_RADIUS * FARFIELD_RADIUS * radiation_pattern_total                                               
    )
```

```

**Comment by oskooi** [comment]
I figured out why the radiation pattern for an $E_r$ point-dipole current source in vacuum computed at various positions for $r > 0$ is not equivalent to the analytic result. It has to do with the way we are combining the results from multiple independent simulations using the Fourier-series expansion of the fields with $e^{im\phi}$.

Specifically, in [Tutorial/Nonaxisymmetric Dipole Sources](https://meep.readthedocs.io/en/latest/Python_Tutorials/Cylindrical_Coordinates/#nonaxisymmetric-dipole-sources),  the Fourier-series expansion is described as:

image

This approach produces incorrect results as shown above in the plots of the radiation pattern at various source positions. In order to produce the correct result which matches the radiation pattern from antenna theory, the Fourier-series expansion must be changed to:

image

Note: the $m=0$ term has twice the weight of the $m \neq 0$ terms. This is the exact *opposite* of what we have been using.

This requires changing a single line in the script above:
```py
radiation_pattern_total += radiation_pattern * (1 if m == 0 else 2)
```
to 
```py
radiation_pattern_total += radiation_pattern * (2 if m == 0 else 1)
```

With this change, the radiation pattern at e.g. $r = 0.35$ μm matches the analytic result as expected.

vacuum_cyl_radpattern_weights_rpos0 35_res100

oskooi:
With the correction described in the previous comment to the way the Fourier-series expansion for $e^{im\phi}$ is computed, it is now possible to demonstrate that the radiation pattern of a dipole in vacuum at different positions with $r > 0$ is equivalent. This involves using the scaling factor $s(r) = (r_0 / r)^2$ for a dipole at position $r$ where $r_0$ is the dipole location at the origin (= `1.5 / resolution`).

The simulation script as well as the description in the text of the tutorial have been updated with these changes.

vacuum_cyl_radpattern_scale_factor

stevengj:
I think the issue is that, as an optimization, Meep only simulates the real parts of the field for m=0 (where the real and imaginary parts are decoupled), which effectively halves the current source for the m=0.  To compensate, you could *either* use `force_complex_fields=True` for the `m=0` simulation *or* multiply the m=0 power by a factor of 4.  (This corresponds to 2x the empirical formula you wrote above.)

stevengj:
For the emission at 90°, you might want to make sure that `greencyl` is doing the angular integral accurately.  e.g. force it to use more angular points by setting `N0 = 128` at https://github.com/NanoComp/meep/blob/4f67cb69592e7899877cabd3dc15dce27ee33f95/src/near2far.cpp#L291C5-L291C20

stevengj:
If you want to validate this, an ideal tool here would be SCUFF, which should be very efficient for this type of problem.

stevengj:
Note that a dielectric disk will have very high Q resonance, especially at higher frequencies.  This might cause the fields to decay very slowly if you have a broadband source.  If these aren't the frequencies of interest, then you might want to either (a) not excite the resonances by using a narrow-band source or (b) use `stop_when_dft_decayed` (or whatever it's called) to only look at the frequency components of interest when stopping the simulation.  Or both.

oskooi:
> For the emission at 90°, you might want to make sure that greencyl is doing the angular integral accurately. e.g. force it to use more angular points by setting N0 = 128

Increasing `N0` from 4 (currently) to 128 considerably reduces the discretization artifacts in the radiation pattern. The figure below shows these artifacts near $\theta \approx 90^\circ$ for `N0 = 4` (blue line) but not for `N0 = 128` (red) which is much smoother and matches the theoretical result at all angles.

Based on these results, it might therefore be useful to make `N0` a user parameter in the routines `get_farfield` and `get_farfields` rather than a hard-coded internal value.

vacuum_cyl_radpattern_greencyl_num_points

oskooi:
> I think the issue is that, as an optimization, Meep only simulates the real parts of the field for m=0 (where the real and imaginary parts are decoupled), which effectively halves the current source for the m=0. To compensate, you could either use force_complex_fields=True for the m=0 simulation or multiply the m=0 power by a factor of 4. (This corresponds to 2x the empirical formula you wrote above.)

As suggested, setting `force_complex_fields=True` in the `Simulation` constructor for the `m = 0` run fixes the issue. With this change, using the existing expression for the Fourier-series expansion involving a weight of 1 for the results of the `m = 0` run and 2 for the `m ≠ 0` runs produces identical radiation patterns for dipoles at $r > 0$ in vacuum. Obtaining this agreement requires using the scaling factor $s(r) = 0.5(r_0/r)^2$.

vacuum_cyl_radpattern_scale_factor

doc/docs/Python_Tutorials/Near_to_Far_Field_Spectra.md
```diff
@@ -656,6 +656,340 @@ if __name__ == "__main__":
...
+
+$$P(\theta) = \frac{1}{R} \int_0^R P(r,\theta) s(r) 2\pi rdr = \frac{1}{N} \sum_{n=0}^{N-1} P(r_n,\theta) s(r_n) 2\pi r_n \Delta r$$,
+
+where $s(r)$ is a weighting function necessary for ensuring equal contribution from all dipoles relative to the dipole at $r = 0$. Note: an $E_r$ dipole at $r = 0$ must be placed at $r_0 = 1.5\Delta r$ due to an [interpolation bug](https://github.com/NanoComp/meep/issues/2704). $s(r)$ can be determined empirically by computing the emission in vacuum for a set of dipoles at different radial positions. The emission profile of a dipole in vacuum is a constant independent of its position. This criteria is used to obtain $s(r) = 0.5(r_0/r)^2$. This weighting function is also used to average the flux emitted by each dipole (obtained using using the LDOS feature). This quantity is the denominator in the expression for the extraction efficiency.
```
stevengj, line 667:
```suggestion
where $s(r)$ is a weighting function necessary for ensuring equal contribution from all dipoles relative to the dipole at $r = 0$. Note: an $E_r$ dipole at $r = 0$ must be placed at $r_0 = 1.5\Delta r$ due to an [interpolation bug](https://github.com/NanoComp/meep/issues/2704). $s(r)$ can be determined empirically by computing the emission in vacuum for a set of dipoles at different radial positions. The emission profile of a dipole in vacuum is a constant independent of its position. This criteria is used to obtain $s(r) = 0.5(r_0/r)^2$.  (A $1/r^2$ dependence is expected because a cylindrical delta function should really have a $1/r$ Jacobian factor, but Meep's currently does not, and power goes like the square of the current amplitude.) This weighting function is also used to average the flux emitted by each dipole (obtained using using the LDOS feature). This quantity is the denominator in the expression for the extraction efficiency.
```

doc/docs/Python_Tutorials/Near_to_Far_Field_Spectra.md
```diff
@@ -656,6 +656,340 @@ if __name__ == "__main__":
...
+
+The example uses the same setup as the [previous tutorial](#radiation-pattern-of-a-disc-in-cylindrical-coordinates) involving a dielectric disc above a lossless-reflector ground plane. The dipoles are arranged on a line extending from $r = 0$ to $r = R$ where $R$ is the disc radius. The height of the dipoles ($z$ coordinate) within the disc is fixed. The radiation pattern $P(r,\theta)$ for a dipole at $r$ is computed using a Fourier-series expansion in $\phi$. The *total* radiation pattern $P(\theta)$ for $N$ dipoles with equal spacing $\Delta r$ is the average of the per-dipole results:
+
+$$P(\theta) = \frac{1}{R} \int_0^R P(r,\theta) s(r) 2\pi rdr = \frac{1}{N} \sum_{n=0}^{N-1} P(r_n,\theta) s(r_n) 2\pi r_n \Delta r$$,
```
stevengj, line 665:
```suggestion
The example uses the same setup as the [previous tutorial](#radiation-pattern-of-a-disc-in-cylindrical-coordinates) involving a dielectric disc above a lossless-reflector ground plane. The dipoles are arranged on a line extending from $r = 0$ to $r = R$ where $R$ is the disc radius. The height of the dipoles ($z$ coordinate) within the disc is fixed. The radiation pattern $P(r,\theta)$ for a dipole at $r$ is computed using a Fourier-series expansion in $\phi$. The *total* radiation pattern $P(\theta)$ for an ensemble of incoherent dipoles is just the integral of the individual dipole powers, which we can approximate by a sum::

$$P(\theta) = \int_0^R P(r,\theta) s(r) 2\pi rdr = \sum_{n=0}^{N-1} P(r_n,\theta) s(r_n) 2\pi r_n \Delta r$$,
```

Issue #1854, opened 2021-12-12
evaluating the error convergence of the gradient from the adjoint solver


oskooi:
While we have already demonstrated second-order accuracy for subpixel smoothing of the `MaterialGrid` in a "forward" simulation (see [results](https://github.com/NanoComp/meep/pull/1503#issuecomment-781110073) in #1503), following #1780 which enabled support for backpropagating the gradients from the Yee grid to the `MaterialGrid`, we have yet to validate the second-order accuracy of the gradients computed using an adjoint simulation. For completeness, we also need to verify that no smoothing produces linear convergence.

The following 2d test computes the directional derivative using the adjoint gradient over a design region consisting of a ring structure which connects an input and output port/waveguide. (This is a similar setup to the existing unit tests for the adjoint solver in which the design region contains random pixel values and the directional derivative is computed in a random direction.) The objective function is a single DFT field point in the output waveguide at a single frequency (the red dot in the schematic below). (We would also eventually like to adapt this test to the other two objective functions: mode coefficients and near-to-far fields.) The ring structure is a `MaterialGrid` at a fixed resolution of 640 pixels/μm while the Meep grid resolution is repeatedly doubled: 20, 40, 80, and 160 pixels/μm. The directional derivative should be converging to a fixed value with increasing resolution but this does not appear to be the case. The output from running the script below shows the resolution in the second column and directional derivative in the third column.

```
grad:, 20.0, -8.3007345310497
grad:, 40.0, -9.140322608657096
grad:, 80.0, -8.32203124442412
grad:, 160.0, -7.0622182341423985
```

Before we can validate the accuracy of the gradients, we need to ensure that the results are converging to a fixed value.

![ring_matgrid_opt](https://user-images.githubusercontent.com/7152530/145698404-e80b2d10-dbdf-4aae-8da3-7eca7719b311.png)

```py
import meep as mp
import meep.adjoint as mpa
import numpy as np
from autograd import numpy as npa

silicon = mp.Medium(epsilon=12)

sxy = 5.0
cell_size = mp.Vector3(sxy,sxy,0)

dpml = 1.0
pml_xy = [mp.PML(thickness=dpml)]

eig_parity = mp.EVEN_Y + mp.ODD_Z

design_region_size = mp.Vector3(1.5,1.5)
design_region_resolution = 640
Nx = int(design_region_resolution*design_region_size.x) + 1
Ny = int(design_region_resolution*design_region_size.y) + 1

w = 1.0
[...]

PR #2741, opened 2023-12-15
First-order subpixel smoothing for density-based TO

smartalecH:
Here we implement first-order subpixel smoothing for density-based TO. This approach allows us to treat the density formulation as a level set, such that the user can now continuously increase β→∞.

This approach is 100% in python and leverages `autograd` to do all of the backpropagation. It's very straightforward, but currently only works for 2D degrees of freedom. Adding capability for 1D and 3D is trivial -- we just need to ensure the filters work in those dimension, and we need to add in the right fill factor kernel (analytically derived by assuming the smoothing kernel is a sphere). This is a "simple version" of what's implemented in #1951

Here's an example:

image

While the approach works well, it is still sensitive to numerical roundoff errors. Autograd doesn't really have any tooling in place to track where things are breaking down (particularly in the backward pass). So we'll have to get creative. Here's a plot that shows the norm of the gradient which breaks down due to numerical error with increasing β:

image

Also, still needs a test.

### DIFF
--- a/python/adjoint/filters.py
+++ b/python/adjoint/filters.py
@@ -1100,6 +1100,7 @@ def gray_indicator(x):
     """
     return npa.mean(4 * x.flatten() * (1 - x.flatten())) * 100
 
+
 def hybrid_levelset(
     x: ArrayLikeType,
     beta: float,
@@ -1147,7 +1148,7 @@ def hybrid_levelset(
             periodic. Default is None (all axes are non-periodic).
     Returns:
         The projected and smoothed output.
-    
+
     Example:
         >>> Lx = 2
         >>> Ly = 2
@@ -1196,7 +1197,8 @@ def hybrid_levelset(
     # with array-based AD tracers, apparently. See here:
     # https://github.com/google/jax/issues/1052#issuecomment-5140833520
     fill_factor = npa.where(
-        (d <= pixel_radius) & (npa.abs(d / pixel_radius) <= 1), # domain of arccos() and sqrt()
+        (d <= pixel_radius)
+        & (npa.abs(d / pixel_radius) <= 1),  # domain of arccos() and sqrt()
         (1 / (npa.pi * pixel_radius**2))
         * (
             pixel_radius**2
--- a/python/adjoint/filters.py
+++ b/python/adjoint/filters.py
@@ -700,6 +700,153 @@ def tanh_projection(x: np.ndarray, beta: float, eta: float) -> np.ndarray:
     )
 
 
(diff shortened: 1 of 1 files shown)


oskooi:
> Also, still needs a test.

I suppose a test for this feature would be based on @stevengj's proposal in [#1854 (comment)](https://github.com/NanoComp/meep/issues/1854#issuecomment-1499544800) but this time demonstrating *first-order* convergence.

smartalecH:
> but this time demonstrating first-order convergence.

I don't think we really need to test the first-order convergence, since that's not really the feature here. Rather we should probably check the accuracy of the gradient for a few finite betas, and when `beta = np.inf`. Maybe a few tests that check for correctness, type sanity too, etc.

The real novelty with this PR is the ability to differentiate with any beta. So that's probably the test we should implement.

stevengj:
> So we'll have to get creative. Here's a plot that shows the norm of the gradient which breaks down due to numerical error with increasing β:

What's the source of this error?

smartalecH:
> Isn't the point of smoothing to get a finite nonzero gradient?

Right, exactly. I would have expected it to be monotonic too, but maybe there's no reason for it to be monotonic. 

> What's the source of this error?

So during the backward pass, autograd's numpy throws some overflow warnings for a few different functions. But again, the tooling and logging capabilities for autograd are really nonexistent, so debugging will be pretty manual. 

But if I had to guess, here are some areas where things could blow up:
* Computing `d`, the distance to an interface, which requires _dividing_ by the norm of the spatial gradient. 
* The meep `tanh_projection` function has, as an argument to the tanh, the product of beta (a big number) with other small numbers. There could be some stuff going on here downstream (particularly with the backward pass). 
* The fill factor formula has some ugly `arccos` and `sqrt` business. We use a numpy `where` function to filter out domain violations, but still...

smartalecH:
Alrighty, I think I resolved all the above issues. Just had to implement a few "double where" tricks to sanitize the backprop. (also had a bug in the effective material part of the algorithm). The smoothed materials themselves look much better:

image

And, if we do a convergence check as  β→∞, we see that the norm of the gradient converges smoothly to a finite, non-zero value: 

image

I've also checked the pathological case where we have a uniform design field (such that the spatial gradient is zero). Normally, this would create a divide by zero error when computing the distance to the interface, and further complicate the backward pass. But I've sanitized things enough that everything looks good.

This should be a drop-in replacement for the `tanh_projection()` function, with the requirement of an additional `resolution` argument (to compute the spatial gradient).

Any ideas for a test?

PR #2742, opened 2023-12-16
Fixing compile issue if hdf5 not installed

theogdoctorg:
When trying to compile from source, if HDF5 is not enabled, the "HAVE_HDF5" identifier is false and certain sections within h5file.cpp are ignored. Some types (H5T_NATIVE_DOUBLE, "H5T_NATIVE_FLOAT)  are used within a few h5file::***_chunk 
functions, and without HDF5, are undefined. This prevents the file from compiling. Added #ifdef blocks to get file to compile without HDF5.


oskooi:
Note that both `_write_chunk(...)` and `_read_chunk(...)` have `#if def HAVE_HDF5` macros as their first line:

https://github.com/NanoComp/meep/blob/295fa633cefb016a8ffd52d98e393ae7781e2bc7/src/h5file.cpp#L623-L626

https://github.com/NanoComp/meep/blob/295fa633cefb016a8ffd52d98e393ae7781e2bc7/src/h5file.cpp#L773-L775

This means these functions should compile without HDF5 but will fail at *runtime* due to:

https://github.com/NanoComp/meep/blob/295fa633cefb016a8ffd52d98e393ae7781e2bc7/src/h5file.cpp#L689-L691

This means that `make` will likely succeed but `make check` will fail when running [`python/tests/test_dump_load.py`](https://github.com/NanoComp/meep/blob/master/python/tests/test_dump_load.py) (which invokes `dump_structure` and `load_structure` which in turn invoke `structure::dump` and `structure::load` of `structure_dump.cpp` which in turn invoke `_write_chunk` and `_read_chunk`, respectively):

https://github.com/NanoComp/meep/blob/295fa633cefb016a8ffd52d98e393ae7781e2bc7/python/Makefile.am#L61

To get `make check` to pass even when HDF5 is not installed, we would need to exclude this test.

stevengj:
I would rather fix this by editing this block https://github.com/NanoComp/meep/blob/425aa398faf1bf28b21f92c63c94fcb587819aa0/src/h5file.cpp#L48-L50

to add dummy definitions
```cxx
#  define H5T_NATIVE_FLOAT  0  // ignored
#  define H5T_NATIVE_DOUBLE 0  // ignored
```
That way, it will still fall through to the error exception if HDF5 is missing.

Issue #2757, opened 2024-01-12
Accuracy of adjoint gradients in 3D


oskooi:
This is an initial attempt to verify the accuracy of the adjoint gradients in 3D. Once we are able to get this example working, we can consider converting it into a unit test for #2661 as well as a tutorial example.

The example involves a **power splitter** (shown in the schematic below) at $\lambda$ = 1.55 μm for a silicon waveguide on a silicon dioxide substrate. This particular SOI waveguide is identical to an [existing tutorial](http://www.simpetus.com/projects.html#mpb_waveguide) with the schematic and band diagram also shown below. The nice thing about the power splitter is that it is an extension of a similar [example in 2D](https://nbviewer.jupyter.org/github/NanoComp/meep/blob/master/python/examples/adjoint_optimization/04-Splitter.ipynb) and can be modified to use various types of objective arguments (`EigenmodeCoefficient`, `FourierFields`, `LDOS`) and a broad bandwidth objective function. The design region is a 2D grid with the pixels extruded in the $z$ direction. The example uses either a constant or random design region. The entire simulation runs in about 25 minutes using 3 Xeon 4.2 GHz cores.

The test involves the usual comparison of the directional derivative computed using (1) a brute-force finite difference and (2) the adjoint gradient. (This calculation is similar to the 2D simulation of a 1D grating in [#2054 (comment)](https://github.com/NanoComp/meep/pull/2054#issuecomment-1873657678).) The MPB output of the simulation shows that the correct waveguide mode is being computed (i.e., matching the values for ($k$, $\omega$) at the red dot in the band diagram below).

At a resolution of 20 pixels/μm, there is a large discrepancy in the finite-difference and adjoint-gradient results:

```
directional-derivative:, [-0.14618432] (finite difference), [-0.00242355] (adjoint)
```

This discrepancy does not seem to decrease with increasing resolution. @smartalecH, any thoughts?

![power_splitter_snapshot_3](https://github.com/NanoComp/meep/assets/7152530/6b1e74a5-c72c-407f-a387-37404734c4b9)

![image](https://github.com/NanoComp/meep/assets/7152530/1282ac9a-db81-4b6a-8607-7930dd4e40ed)

```py
"""Validates the adjoint gradient of a power splitter in 3d."""

from autograd import numpy as npa
import meep as mp
import meep.adjoint as mpa
import numpy as np


RESOLUTION = 20  # pixels/μm                                                                                      

WAVELENGTH_UM = 1.55
[...]

PR #2758, opened 2024-01-12
Disable subpixel smoothing by default in `MaterialGrid` class objects

oskooi:
As described in #2757, there seems to be a bug in the subpixel smoothing feature for the `MaterialGrid` for 3D structures. Until this is resolved, it is better to disable subpixel smoothing by default.


smartalecH (approved):
LGTM!

PR #2766, opened 2024-01-20
Increase initial number of grid points used in phi quadrature for near-to-far field transformation in cylindrical coordinates

### LINKED ISSUE #2764: Make the initial number of azimuthal grid points in near-to-far field transformation in cylindrical coordinates a user parameter
Currently, the computation of the far fields in cylindrical coordinates involves an integration in $\phi$ (azimuthal angle) which is based on repeatedly doubling a grid of $\phi$ points using a *hard-coded* initial value:

https://github.com/NanoComp/meep/blob/04fa3053650d936114b70331e971031c6ae3eb87/src/near2far.cpp#L288-L291

The value of 4 turns out to be too small in some cases which produces discretization artifacts in the output as demonstrated in [#2726 (comment)](https://github.com/NanoComp/meep/pull/2726#issuecomment-1899753136). As such, it would be useful to make the initial number of $\phi$ grid points a user parameter in the routines `get_farfield` and `get_farfields`.

oskooi:
Closes #2764.

Rather than make the number of grid points in the $\phi$ quadrature a user parameter in `get_farfields`, it is simpler just to increase the hard-coded value for `N0` set internally in `greencyl`. The impact of integrating over a larger number of grid points should be insignificant when compared to the accuracy improvements enabled by this modification as this integral converges exponentially fast.

### DIFF
--- a/src/near2far.cpp
+++ b/src/near2far.cpp
@@ -288,7 +288,7 @@ void greencyl(std::complex<double> *EH, const vec &x, double freq, double eps, d
   /* Perform phi integral.  Since phi integrand is smooth, quadrature with equally spaced points
      should converge exponentially fast with the number N of quadrature points.  We
      repeatedly double N until convergence to tol is achieved, re-using previous points. */
-  const int N0 = 4;
+  const int N0 = 128;
   double dphi = 2.0 / N0; // factor of 2*pi*r is already included in add_dft weight
   for (int N = N0; N <= 65536; N *= 2) {
     std::complex<double> EH_sum[6];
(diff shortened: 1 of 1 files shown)


stevengj:
128 might be a little on the large side?

Probably it should increase proportional to `m`.

Would e.g. `16 + int(4*abs(m))` be sufficient here?

oskooi:
> Would e.g. 16 + int(4*abs(m)) be sufficient here?

Verified that this produces nearly the same results as `N0 = 128`.

PR #2767, opened 2024-01-23
Add `eig_vol` argument to `EigenModeSource` class constructor

### LINKED ISSUE #2755: Inconsistency in interface for `get_eigenmode_coefficients` and `EigenModeSource` constructor affecting adjoint solver
There is an inconsistency in the interface for the related functions `get_eigenmode_coefficients` and the constructor for the `EigenModeSource` class object involving the argument `eig_vol`. 

`get_eigenmode_coefficients` contains an argument `eig_vol` of type `meep.Volume`:

https://github.com/NanoComp/meep/blob/0dc44daf9fd4bfea17c209bcd25f110c08341080/python/simulation.py#L4133-L4143

The constructor for the `EigenModeSource` class object does *not* have an `eig_vol` argument and instead uses `eig_lattice_size` and `eig_lattice_center`:

https://github.com/NanoComp/meep/blob/0dc44daf9fd4bfea17c209bcd25f110c08341080/python/source.py#L459-L475

These two arguments are used to create an `eig_vol` object in a separate function `add_source`:

https://github.com/NanoComp/meep/blob/0dc44daf9fd4bfea17c209bcd25f110c08341080/python/source.py#L646-L651

`eig_vol` and (`eig_lattice_center`, `eig_lattice_size`) are doing the same thing. It would be good to remove this inconsistency in the interface which currently prevents using `eig_vol` to define an `meep.adjoint.EigenmodeCoefficient` objective function because it is not possible to create an adjoint source. The problem is that the keyword arguments passed to `EigenmodeCoefficient` are passed directly to `EigenModeSource` in the function `place_adjoint_source`:

https://github.com/NanoComp/meep/blob/0dc44daf9fd4bfea17c209bcd25f110c08341080/python/adjoint/objective.py#L262-L272

oskooi:
Fixes #2755.

### DIFF
--- a/python/source.py
+++ b/python/source.py
@@ -567,20 +567,10 @@ class EigenModeSource(Source):
         self.eig_resolution = eig_resolution
         self.eig_tolerance = eig_tolerance
 
-        if eig_vol:
+        if eig_vol is not None:
             self.eig_vol = eig_vol.swigobj
-        elif self.eig_lattice_center and self.eig_lattice_size:
-            self.eig_vol = mp.Volume(
-                self.eig_lattice_center,
-                self.eig_lattice_size,
-                sim.dimensions,
-                is_cylindrical=sim.is_cylindrical,
-            ).swigobj
         else:
-            raise ValueError(
-                "only one of eig_vol or (eig_lattice_center, eig_lattice_size) "
-                "can be specified."
-            )
+            self.eig_vol = None
 
     @property
     def eig_lattice_size(self):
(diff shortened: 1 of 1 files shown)


oskooi:
Note: I used this PR along with the one-line change in `mpb.cpp` in #2285 to verify the accuracy of the adjoint gradients for the `EigenmodeCoefficient` objective function using the 2d test involving a 1d binary grating described in [#2054 (comment)](https://github.com/NanoComp/meep/pull/2054#issuecomment-1873657678). After this PR and #2285 are merged, we may therefore add the 1d diffraction grating as a new unit test for the adjoint solver.

stevengj:
LGTM.

PR #2771, opened 2024-01-26
Update tutorial for launching planewaves in arbitrary directions using `EigenModeSource`

oskooi:
Updates [Tutorial/Planewaves in Homogeneous Media](https://meep.readthedocs.io/en/latest/Python_Tutorials/Eigenmode_Source/#planewaves-in-homogeneous-media) to demonstrate launching planewaves in *arbitrary* directions. This is made possible by #2767. Previously, it was not possible to launch planewaves with incident angles larger than 90° (i.e., propagating in the backward direction).


doc/docs/Python_Tutorials/Eigenmode_Source.md
```diff
@@ -229,62 +229,97 @@ Increasing the size of the cell improves results at the expense of a larger simu
...
-           output_plane=nonpml_vol)
+There are several items to note in this script.
 
-if mp.am_master():
-    plt.axis('off')
-    plt.savefig('pw.png',bbox_inches='tight')
-```
+1. The line source spans the *entire* length of the cell. Planewave sources extending into the PML region must include `is_integrated=True` in the source definition.
+2. The wavevector of the planewave $\vec{k}$ is defined using the `k_point` which specifies Bloch-periodic boundaries. PML boundaries are used only along the $x$ direction.
+3. The `eig_vol` parameter of the `EigenModeSource` defines a volume that is one-pixel thick in the direction of the line monitor. This is necessary to ensure that the planewave's wavevector `eig_kpoint` which is defined in the continuum rather than for a system with discrete translational symmetry is correctly interpreted by MPB.
```
stevengj, line 318:
```suggestion
3. The `eig_vol` parameter of the `EigenModeSource` defines a volume that is one-pixel thick in the direction of the line monitor. This is necessary to ensure that the`eig_kpoint` is interpreted by MPB as the wavevector of a planewave — that is, so it is the wavevector as defined for a homogeneous material (continuous translational symmetry) rather than as a Bloch wavevector for a system with discrete translational symmetry.
```

PR #2792, opened 2024-02-27
Tutorial example for computing the adjoint gradient of a level set

### LINKED ISSUE #2235: Computing the gradient of a level-set function

oskooi:
Closes #2235.

In order to obtain early feedback, this is the first commit which currently just includes two explanatory figures and an image showing simulation output along with the simulation script in what will eventually become a new tutorial for the adjoint solver. Text and additional results are forthcoming.


stevengj:
If you are willing to do finite difference, in principle you should be able to skip the subpixel smoothing entirely.   You can instead do:

1. Construct your level set function φ at a very high resolution for a given width h.   Call this φ(h)
2. Tell numpy to downsample (with bilinear interpolation) to a lower resolution, comparable to the Meep resolution.  Call this ρ(φ(h)).  This will "anti-alias" the boundaries.  This is the material grid you input to Meep.
3. Meep gives you the derivative with respect to ρ of your objective function.  To get the derivative with respect to h, you just need a custom vJp that computes dρ/dh, e.g. by finite differences.

oskooi:
> 1. Construct your level set function φ at a very high resolution for a given width h. Call this φ(h)

Unfortunately, even using this new approach (downsampling from a grid for the level set $\phi(h)$ that that has **10X** the resolution of the Yee grid) produces similar results to what I had previously: good agreement in the directional derivative for the adjoint gradient and finite difference for the $\mathcal{S}$ but *not* the $\mathcal{P}$ polarization.

The test uses a resolution of **400** pixels/um which is fairly high. It is not obvious why the results for the $\mathcal{P}$ polarization are inaccurate.

**1. $\mathcal{S}$ polarization**
```
directional-deriv:, -0.00096108 (finite difference), -0.00090948 (adjoint), 0.053690 (error)
```

**2. $\mathcal{P}$ polarization**
```
directional-deriv:, -0.00015576 (finite difference), -0.00112504 (adjoint), 6.222962 (error)
```

Script used to generate these results is in this [gist](https://gist.github.com/oskooi/339ac363036c89007e71fa5efd41f705).

stevengj:
When you carry out the chain rule, you multiply the Meep d/dρ with the level-set dρ/dh derivative.  The latter is only nonzero on the boundary pixels.  So, if the Meep derivative is inaccurate right on the boundary, that would be exacerbated by this d/dh derivative.    

My hypothesis at the moment is that the p-polarization Meep derivatives are less accurate at the boundary (due to the the differentiate–then–discretize errors), since the p polarization has a discontinuity at the boundary, compared to the s polarization (which is continuous).   

To check this, try checking the Meep d/dρ derivatives against finite differences, but instead of using a random perturbation to ρ, do a perturbation right at the the boundary pixels.

oskooi:
> My hypothesis at the moment is that the p-polarization Meep derivatives are less accurate at the boundary (due to the the differentiate–then–discretize errors), since the p polarization has a discontinuity at the boundary, compared to the s polarization (which is continuous).

> To check this, try checking the Meep d/dρ derivatives against finite differences, but instead of using a random perturbation to ρ, do a perturbation right at the the boundary pixels.

The P polarization Meep derivatives are indeed less accurate at the boundaries compared to the S polarization. To verify this, I used the example of the usual 1D binary grating and added a **2-pixel perturbation** to the boundary (see [gist](https://gist.github.com/oskooi/f8d19c8db45a25ff33ca87b65eda1412)). 

As shown in the results below, the error in the directional derivative is small and decreases with resolution for the S polarization but *not* the P polarization.

We can explain this feature in this tutorial. For now, the tutorial is restricted to just the S polarization.

**S polarization**
```
RESOLUTION_UM = 50
grad:, -5.3573335767609365e-09 (finite difference), -5.358572385464965e-09 (adjoint), 0.000231 (error)

RESOLUTION_UM = 100
grad:, -2.244694430331151e-08 (finite difference), -2.2446890114290335e-08 (adjoint), 0.000002 (error)
```

**P polarization**
```
RESOLUTION_UM = 50
grad:, 2.913486456534997e-07 (finite difference), 2.6559604840176923e-07 (adjoint), 0.088391 (error)

RESOLUTION_UM = 100
grad:, 4.6565183708935365e-08 (finite difference), 4.157217179759396e-08 (adjoint), 0.107226 (error)
```

oskooi:
The test failures are due to Autograd's dependence on deprecated Numpy features as reported (and fixed but not yet merged) in HIPS/autograd#618.

stevengj:
Would be good to check the in-plane polarization for the $\beta = \infty$ subpixel smoothing as well.

Maybe there is a bug in what @smartalecH calls the "recombination" step (the overlap between adjoint and forward fields) that is lowering the accuracy.

oskooi:
@stevengj, this is ready to be merged pending final review.

doc/docs/Python_Tutorials/Adjoint_Solver.md
```diff
@@ -804,6 +804,392 @@ if __name__ == "__main__":
         )
 ```
 
+Adjoint Gradient of a Level Set
+-------------------------------
+
+It is also possible to compute the gradient of a [level set](https://en.wikipedia.org/wiki/Level_set) (i.e., a discontinuous function) using the density-based adjoint solver. This is useful for applications involving [shape optimization](https://en.wikipedia.org/wiki/Shape_optimization) or geometry parameterization.
```
stevengj, line 810:
```suggestion
It is also possible to compute the derivative of the Meep outputs with respect to a geometric parameter via a [level-set](https://en.wikipedia.org/wiki/Level_set) formulation (an implicit representation of the interface between two materials) using the density-based adjoint solver. This is useful for applications involving [shape optimization](https://en.wikipedia.org/wiki/Shape_optimization) of explicitly parameterized geometries.
```

doc/docs/Python_Tutorials/Adjoint_Solver.md
```diff
@@ -804,6 +804,392 @@ if __name__ == "__main__":
         )
 ```
 
+Adjoint Gradient of a Level Set
+-------------------------------
+
+It is also possible to compute the derivative of the Meep outputs with respect to a geometric parameter via a [level-set](https://en.wikipedia.org/wiki/Level_set) formulation (an implicit-function representation of a material discontinuity) using the density-based adjoint solver. This is useful for applications involving [shape optimization](https://en.wikipedia.org/wiki/Shape_optimization) of explicitly parameterized geometries.
+
+As a demonstration, we will compute the gradient of the diffraction efficiency of the first transmitted order with $\mathcal{S}$ polarization (electric field out of the plane, i.e. $E_z$) of a 1D binary grating with respect to the grating height. The accuracy of the adjoint gradient is validated using a brute-force finite difference via the directional derivative.
```
stevengj, line 812:
```suggestion
As a demonstration, we will compute the derivative of the diffraction efficiency of the first transmitted order with $\mathcal{S}$ polarization (electric field out of the plane, i.e. $E_z$) of a 1D binary grating with respect to the grating height. The accuracy of the adjoint derivative is validated using a brute-force finite difference approximation.
```

doc/docs/Python_Tutorials/Adjoint_Solver.md
```diff
@@ -804,6 +804,392 @@ if __name__ == "__main__":
...
+
+The calculation of the gradient of the diffraction efficiency ($F$) with respect to the grating height ($h$) involves using the chain rule to obtain a product of two derivatives: (1) $\partial F / \partial \rho$ and (2) $\partial \rho / \partial h$, where $\rho$ are the density weights (2D grid of $N_x \times N_y = N$ points) used to represent the binary grating as a levelset. (1) is the adjoint gradient computed by Meep. (2) requires the levelset to be *smoothed* to ensure it is differentiable. A non-smoothed levelset is generally not differentiable. (2) is implemented in Autograd using a custom vector-Jacobian product (vJP) which is used as part of reverse-mode differentiation (backpropagation) to compute $\partial F / \partial h$. An overview of this calculation is shown below.
```
stevengj, line 816:
```suggestion
The calculation of the derivative of the diffraction efficiency ($F$) with respect to the grating height ($h$) involves using the chain rule to obtain a product of two derivatives (Jacobians): (1) $\partial F / \partial \rho$ and (2) $\partial \rho / \partial h$, where $\rho$ are the density weights (2D grid of $N_x \times N_y = N$ points) used to represent the binary grating as a level set. (1) is the adjoint gradient computed by Meep. (2) requires the level set to be *smoothed* to ensure it that the derivative $\partial \rho / \partial h$ exists. (2) is implemented in Autograd using a custom vector–Jacobian product (vJP) which is used as part of reverse-mode differentiation (backpropagation) to compute $\partial F / \partial h$. An overview of this calculation is shown below.
```

doc/docs/Python_Tutorials/Adjoint_Solver.md
```diff
@@ -804,6 +804,392 @@ if __name__ == "__main__":
...
+
+![](../images/levelset_gradient_backpropagation.png#center)
+
+The derivative $\partial \rho / \partial h$ can be approximated by a finite difference using a one-pixel perturbation applied to the grating height. The calculation is summarized in the schematic below. There are two function evaluations of $\rho(h)$ for which the cost is negligible. Smoothing of the levelset can be performed using a number of different methods including a [signed-distance function](https://en.wikipedia.org/wiki/Signed_distance_function) or convolution filter. In this example, the smoothing is based on downsampling the levelset from a high-resolution grid (10X the resolution of the simulation grid) to the lower-resolution simulation grid using bilinear interpolation. Note that only the boundary pixels are nonzero in the Jacobian matrix in this case.
```
stevengj, line 820:
```suggestion
The derivative $\partial \rho / \partial h$ can be approximated by a finite difference using a one-pixel perturbation applied to the grating height; this is computationally convenient because it greatly simplifies the construction of $\rho(h)$ as explained below. (A finite difference involves two function evaluations of $\rho(h)$, but the cost for this is negligible since it involves no Meep simulations.) The construction of $\rho(h)$ involves two steps: first, constructing a simple binary image $b(h)$ at a high resolution; and second, smoothing $b(h)$ into a continuous level-set function $\rho(h)$.  This smoothing of the image can be performed using a number of different methods including a [signed-distance function](https://en.wikipedia.org/wiki/Signed_distance_function) or convolution filter. In this example, the smoothing is based simply on downsampling the image from a high-resolution grid (10X the resolution of the simulation grid) to the lower-resolution simulation grid using bilinear interpolation, which leads to "gray" pixels at the boundaries between materials in a way that changes continuously with $h$. Only these boundary pixels have nonzero derivatives in the Jacobian $\partial\rho/\partial h$ in this case. This calculation is summarized in the schematic below. 
```

doc/docs/Python_Tutorials/Adjoint_Solver.md
```diff
@@ -804,6 +804,392 @@ if __name__ == "__main__":
...
+RESOLUTION_UM = 400
+dir-deriv:, -0.00140221 (finite difference), -0.00132065 (adjoint), 0.058165 (error)
+
+RESOLUTION_UM = 800
+dir-deriv:, -0.00069606 (finite difference), -0.00067547 (adjoint), 0.029583 (error)
+```
+
+A logarithmic plot of the relative error vs. grid resolution based on these results demonstrates linear convergence. This is expected for fields right on the boundary of a discontinuous interface.
+
+Note: because of a likely bug in the adjoint gradients for the $\mathcal{P}$ polarization (electric field in the $xy$ plane), there is a large error in the adjoint gradient which [does not decrease with resolution](https://github.com/NanoComp/meep/pull/2792#issuecomment-2171942477). This is why this demonstration only involved the $\mathcal{S}$ polarization.
```
stevengj, line 849:
```suggestion
Currently, we recommend using this procedure only for the $\mathcal{S}$ ($E_z$) polarization, since for the for the $\mathcal{P}$ polarization (electric field in the $xy$ plane), there appear to be large discretization errors in the adjoint gradient which [we are currently investigating](https://github.com/NanoComp/meep/pull/2792#issuecomment-2171942477). This is why this demonstration only involved the $\mathcal{S}$ polarization.
```

doc/docs/Python_Tutorials/Adjoint_Solver.md
```diff
@@ -804,6 +804,392 @@ if __name__ == "__main__":
         )
 ```
 
+Adjoint Gradient of a Level Set
+-------------------------------
```
stevengj, line 808:
```suggestion
Derivatives with Respect to Shape Parameters
--------------------------------------------
```

PR #2816, opened 2024-04-07
Fix typo in expression for total fields from Fourier-series summation

oskooi:
Fixes a typo in [Tutorial/Cylindrical Coordinates/Nonaxisymmetric Dipole Sources](https://meep.readthedocs.io/en/latest/Python_Tutorials/Cylindrical_Coordinates/#nonaxisymmetric-dipole-sources) involving the expression for the total fields based on the Fourier-series summation which incorrectly included taking the real part of the $m > 0$ terms.

**original**

image

**modified**

image


stevengj:
The original looks correct to me — you take twice the real part because you are only summing positive m.   Summing a positive m and a negative m term is equivalent to taking twice the real part of the positive-m term, since they are complex conjugates: `z + z̄ = 2ℜ[z]`

oskooi:
If I take the Fourier-series expansion for $\vec{E}_{tot}(\theta, \phi)$ and simplify it based on $\vec{E}_m(\theta)e^{im\phi}$ being a complex conjugate for $m$ and $-m$, the resulting expression is:

image

oskooi:
I have updated the formulas to:

image

stevengj:
Why is that simpler?  The original complex-exponential formula seems simpler to me.

oskooi:
> Why is that simpler? The original complex-exponential formula seems simpler to me.

Note that the original and the updated expressions are *not* equivalent. The updated expression is derived from the fact that the complex conjugate of a product of complex numbers is the *product* of their complex conjugates: $(ab)^* = a^* b^*$ where $a = \vec{E}_{m}(\theta)$ and $b = e^{im\theta}$. 

Previously, we had been using $(ab) + (ab)^* = 2 \Re\left[ab\right]$ which is incorrect.

Issue #2400, opened 2023-02-16
Epigraph slack hyperparameter?


stevengj:
Ardavan found some anecdotal evidence that it is useful to add some "slack" into the initial choice of the dummy variable for an epigraph formulation, to make the dummy-variable constraints inactive for the initial iterations.

Would be good to investigate this further in some of the tutorial problems, in case it holds up more generally and should be documented.

(In the longer run, it would be good to have a CCSA variant that takes the epigraph dummy variable into account analytically, using the fact that it comes linearly into the equations, as described in the Svalberg paper.)

mochen4:
I did some preliminary investigation on the issue. I found that, if the dummy variable was initialized so that the constraints were violated, sometimes (not always), the optimizer may keep adding slack to the dummy variable, even after the constraints were already inactive. 

I set up the experiment based on [the metalens epigraph tutorial](https://nbviewer.org/github/NanoComp/meep/blob/master/python/examples/adjoint_optimization/06-Near2Far-Epigraph.ipynb), where I minimized dummy variable $t$ s.t. $-|E(\omega_i)|^2 - t < 0$ at frequency $\omega_i$. I only had one beta, but initialized with different $t$ values.

Here are some plots (where the red line is the evolution of $t$, and the other lines are the values of $-|E(\omega_i)|^2$:

**Inactive constraint:**
![t_30damping3_-1](https://user-images.githubusercontent.com/25192039/227318827-c7da234b-4c9b-43ca-bc45-bf9482a67c25.png)

**Active constraint but working:**
![t_30damping3_-7](https://user-images.githubusercontent.com/25192039/227318649-81db0970-c1f3-41b4-98cc-0b64529e6089.png)

**Active constraint but not working:**
![t_30damping3_-2](https://user-images.githubusercontent.com/25192039/227319159-890c893a-3187-4038-a6f6-22a87c2f3112.png)
![t_30damping3_-8](https://user-images.githubusercontent.com/25192039/227319240-3b96019c-187a-4d9a-81eb-e12e434eb262.png)

mochen4:
I set verbosity to 1 and re-ran the optimizations. I am not sure how to interpret the results: but for the case that failed case with infeasible starting point (e.g. the last plot above), the printed information for the last several iterations was like:

```
MMA inner iteration: rho -> 1e-05
                 MMA rhoc[0] -> 2.29312
MMA dual converged in 279 iterations to g=-5.40876:
[...]

PR #2862, opened 2024-07-08
Tutorial for shape optimization of a multilayer stack

### LINKED ISSUE #2820: topopt tutorial on 1d multilayer film optimization
As discussed with @oskooi, it might be nice to have a tutorial with "1d" topology optimization, i.e. designing a multilayer film.  This can be done even in 2d and 3d just by making the material grid a 1x1xN grid (which can then be "extruded" over a higher-dimensional block object).

1. For example, designing a multilayer film to minimize transmission should result in essentially a quarter-wave stack (possibly with some modification for the first and last layers).   Just minimize transmission with the abovementioned 1d material grid, ramping up β slowly as usual to let it decide on the number of layers.
2. We should also be able to fix the number of layers.   If you just set β=infinity (with the new subpixel-smoothing algorithm) then the optimization will tend not to change the topology (because the derivative is not correct for changes in topology) unless it happens to take a very large step.   You can probably prevent it from doing this by simply counting the number of layers (from the projected material grid) on each step and adding a large penalty to the objective function if the number of layers changes, which will cause the optimization algorithm to backtrack.

(You could also use a level-set representation if you have a differentiable mapping function that maps a set of layer thickness to a density ρ, ala #2235.)

On a separate (but related) note, we should probably update the `fill_factor` calculation in the subpixel smoothing code [here](https://github.com/NanoComp/meep/blob/3c58a5b2ac6da0c8ab067d23d16bb00ac0628aa9/python/adjoint/filters.py#L798-L826) to always compute the fill factor for a 3d spherical smoothing kernel — i.e. take the perspective that even "2d" and "1d" structures are actually 3d structures that are invariant in 1 or 2 directions, respectively.

oskooi:
Closes #2820.

This PR adds a new tutorial for the adjoint solver which demonstrates shape optimization of a multilayer stack in 1D. The design problem involves finding the layer thicknesses (for a *fixed* number of layers) which minimize the transmittance of a planewave at normal incidence in air with $\lambda = 1.0$ μm through a stack of alternating materials of $n_1=1.0$ and $n_2=1.3$. The layers are arranged as $n_2$, $n_1$, $n_2$, $n_1$, ..., $n_2$. The analytic solution is a quarter-wavelength stack with thicknesses $\lambda / (4 n_1) = 0.25$ and $\lambda / (4 n_2) = 0.1923$.

The optimization uses the [BFGS algorithm of NLopt](https://nlopt.readthedocs.io/en/latest/NLopt_Algorithms/#low-storage-bfgs). An initial value of 0.2 μm is used for the layer thicknesses. Results are shown below for five different cases in which the number of layers are 5, 7, 9, 11, and 13. **The quarter-wavelength stack configuration is found by the optimization only for layer numbers 5 and 7.** Layer numbers 9, 11, and 13 seem to get trapped in a local minima. In all cases, the optimization terminates due to the threshold tolerance of the objective function ([return code 3 of NLopt](https://nlopt.readthedocs.io/en/latest/NLopt_Reference/#return-values)). The thickness of all layers is constrained to be in the range of 0.1 μm to 0.5 μm. Since these are 1D simulations, the runtimes are generally less than a couple of minutes using a single core (Intel Xeon i7-7700K CPU @ 4.20GHz).

A plot of the transmittance vs. iteration number for the five cases of the layer numbers shows that the transmittance decreases with the number of layers, as expected. The optimizer seems to move the design in the *wrong* direction (increasing transmittance) during the early iterations before eventually finding an optimal design which minmizes the transmittance.

We may want to consider creating a unit test based on this tutorial since currently there is no unit test for the 1D adjoint solver.
```
1. NUM_LAYERS = 5 (~quarter-wavelength stack!)

objective value: 0.14224969749140354 (objective value)
layer thicknesses: [0.18767195 0.2477044  0.19506915 0.24954383 0.19187953] (layer thicknesses)


2. NUM_LAYERS = 7 (quarter-wavelength stack!)

objective value: 0.09740488016708304
layer thicknesses: [0.19177897 0.25412021 0.19592259 0.24779526 0.18806041 0.24928526 0.19152152]


3. NUM_LAYERS = 9

objective value: 0.06522007836855086
layer thicknesses: [0.20360086 0.21794387 0.21536126 0.22216203 0.21716955 0.22252077 0.21613793 0.21944627 0.20599552]


4. NUM_LAYERS = 11

objective value: 0.0415530392321388
layer thicknesses: [0.20432596 0.21336318 0.21564032 0.21854779 0.21839713 0.21969996 0.2186816  0.21923049 0.21665256 0.21505099 0.20660389]


5. NUM_LAYERS = 13

objective value: 0.025748647829910033
layer thicknesses: [0.20378588 0.21309369 0.21486671 0.21798861 0.21754488 0.21913944 0.21813383 0.21946564 0.21819879 0.2189978  0.21607473 0.21496185 0.20625602]
```

opt_transmittance_vs_numlayers


stevengj:
For lots of layers, you can easily get a non-quarter-wave structure with very low transmission, and once you get to that point the gradients will be almost zero so probably it will converge very slowly towards a QW solution (it's very slow to try to reduce transmission from `1e-3` to `1e-6`, for example).

@oskooi suggested just taking the log of the transmission.  This seems worth trying, but I'm skeptical — it seems like a free lunch?   It wouldn't help if there is noise in the gradient that is limiting us at low transmission levels, for example.

Another alternative is simply to minimize ∫|E|² in the mirrors (or similar), which will push the optimization towards a field that decays as quickly as possible.   Intuitively, this uses more information than the transmission optimization, since it uses E at every point in the mirror rather than just at the end.

oskooi:
I tried the two suggested objective functions: (1) log of the transmission and (2) ∫|E|² including taking its log. Unfortunately, neither approach seemed to produce better results than the original objective function of just the transmission.

Based on these results, I think the choice of objective function may not be as important as simply running lots of trials with random initial designs in order to try to find the global optima of the quarter-wavelength stack. The design space seems to be highly non convex regardless of the objective function. We may want to consider switching to a different problem.

**1. Objective Function: $\log(E^{*} \times H)$** 

```py
    tran_pt = mp.Vector3(0, 0, 0.5 * size_z_um - PML_UM)

    obj_args = [
        mpa.FourierFields(sim, mp.Volume(center=tran_pt), mp.Ex),
        mpa.FourierFields(sim, mp.Volume(center=tran_pt), mp.Hy),
    ]

    def obj_func(dft_ex, dft_hy):
        """Objective function for the optimization.

        Args:
          dft_ex, dft_hy: the discrete Fourier-transformed Ex and Hy fields.

        Returns:
          The transmittance through the stack.
        """

        return npa.log(npa.real(npa.conj(dft_ex) * dft_hy) / norm_flux)
```


**2. Objective Function: ∫|E|²**
```py
    obj_args = [
        mpa.FourierFields(
            sim,
            matgrid_region.volume,
            mp.Ex
        )
    ]

    def obj_func(dft_ex):
        """Objective function for the optimization.

        Args:
          dft_ex: the discrete Fourier-transformed Ex field.

        Returns:
          The field intensity summed over the design region.
        """

        return npa.log(npa.sum(npa.absolute(dft_ex) **2))
```

stevengj:
When you minimize ∫|E|², did you get a more rapid spatial decay of the fields?   I don't think the transmission should get better, but maybe it is more localized.

stevengj:
(Note that the fact that optimization takes some steps that make things worse initially is a common phenomenon in many algorithms — initially, it takes steps that are too large and then has to backtrack.  After a few iterations, the algorithm has more information about a "good" step size, and in the case of L-BFGS has more information about the 2nd derivative.)

oskooi:
> When you minimize ∫|E|², did you get a more rapid spatial decay of the fields?

I did verify that minimizing ∫|E|² rather than the transmittance does produce more rapid spatial decay of the fields, as expected. The plot below shows a comparison of the field intensity within the multilayer stack (the design region) for the two cases of the objective function: (1) ∫|E|² vs. (2) transmittance. The design produced using (1) had a *smaller* transmittance than (2). This is perhaps a good indication that **(1) may be a more suitable objective function than (2) for this problem**. 

(I was probably unlucky in the local optima I chose for (1) in the results I posted in my [previous comment](https://github.com/NanoComp/meep/pull/2862#issuecomment-2237092977) because that design did not have a lower transmittance than the design obtained using (2).)

**Case 1: objective function = ∫|E|²**
```
tran:, 0.365143 (obj. func. = fields in the stack)
```

**Case 2: objective function = transmittance**
```
tran:, 0.564647 (obj. func. = transmittance)
```

field_decay_in_multilayer_stack

oskooi:
> For example, you can try to design a broad-band AR coating with only 3–4 layers. In this case if you can get an optimum of > 90% at all wavelengths you would be pretty happy, and there is no worry about going from 99.9% to 99.999%. (Either worst-case or average-case broadband would be interesting, perhaps both.)

I changed the original optimization problem of minimizing ∫|E|² for a single wavelength to **multiple wavelengths** using minimax (i.e., worst-case optimization). The updated problem involves two wavelengths (0.95 and 1.05 μm) and nine layers in the stack. Because the epigraph formulation involves the use of nonlinear constraints which are not supported by the BFGS algorithm in NLopt, I switched to using the MMA algorithm instead. The changes to the script `multilayer_opt.py` have been made in the latest commit (e82e8e9).

A plot of the history of the objective function vs. iteration number of the optimizer for the two wavelengths is shown below. Also included in this plot is the epigraph variable. The initial layer thicknesses are chosen randomly. The optimizer seems to converge to a local optima within ~20 iterations. In this case, the optimizer ran for the maximum number of iterations (50).

As expected, the optimal layer thicknesses are quite different from the quarter-wave stack of alternating thicknesses of 0.19 and 0.25 μm.
```
optimal_obj_val:, 13.719162487586791
optimal_layer_thickness_um:, [0.1664, 0.2452, 0.1399, 0.2719, 0.1786, 0.3060, 0.4952, 0.1688, 0.2306]
return_code:, 5
```

multilayer_opt_obj_func_history

Additional results showing the field decay in the stack as well as the transmittance for this design will be added soon.

oskooi:
A plot of the $|E_x|^2$ fields in the stack (design region) for the optimal design from the previous comment is shown below along with the script used to generate this plot. The plot shows that while the DFT fields for each of the two wavelengths are decaying through the stack, the transmittance values are quite different: T($\lambda$ = 0.95 μm) = 0.3140 vs.  T($\lambda$ = 1.05 μm) = 0.7353. This result is consistent with the plot of the objective function vs. iteration number which shows that, because of the minimax setup, the stack is being optimized for $\lambda$ = 0.95 μm and *not* $\lambda$ = 1.05 μm. Also, the magnitude of $|E_x|^2$ at the right edge of the stack is clearly larger for $\lambda$ = 1.05 μm than $\lambda$ = 0.95 μm. This is (again) consistent with the transmittance being larger for $\lambda$ = 1.05 than for $\lambda$ = 0.95 μm.

Perhaps we just happened to choose a poor local optima for this demonstration? Or maybe this is a feature of this particular optimization problem?

![field_decay_in_multilayer_stack_broadband](https://github.com/user-attachments/assets/89716168-d223-4351-a663-800025d5eb40)


```py
rom typing import List, Tuple

import meep as mp
import numpy as np


RESOLUTION_UM = 800
AIR_UM = 1.0
PML_UM = 1.0
NUM_LAYERS = 9
MIN_LAYER_UM = 0.1
MAX_LAYER_UM = 0.5
N_LAYER_1 = 1.0
N_LAYER_2 = 1.3
LAYER_PERTURBATION_UM = 1.0 / RESOLUTION_UM
DESIGN_WAVELENGTHS_UM = (0.95, 1.05)
DESIGN_REGION_RESOLUTION_UM = 10 * RESOLUTION_UM
DESIGN_REGION_UM = mp.Vector3(0, 0, NUM_LAYERS * MAX_LAYER_UM)
NZ_DESIGN_GRID = int(DESIGN_REGION_UM.z * DESIGN_REGION_RESOLUTION_UM) + 1
NZ_SIM_GRID = int(DESIGN_REGION_UM.z * RESOLUTION_UM) + 1

num_wavelengths = len(DESIGN_WAVELENGTHS_UM)
frequencies = [1 / wavelength_um for wavelength_um in DESIGN_WAVELENGTHS_UM]

frequency_center = np.mean(frequencies)

# Set source bandwidth to be larger than the range of design wavelengths.                                         
frequency_width = 1.2 * (np.max(frequencies) - np.min(frequencies))


def design_region_to_grid(nz: int) -> np.ndarray:
    """Returns the coordinates of the grid for the design region.                                                 
                                                                                                                  
    Args:                                                                                                         
      nz: number of grid points.                                                                                  
                                                                                                                  
    Returns:                                                                                                      
      The 1D coordinates of the grid points.                                                                      
    """

    z_grid = np.linspace(
        -0.5 * DESIGN_REGION_UM.z,
        +0.5 * DESIGN_REGION_UM.z,
        nz
    )

    return z_grid


def levelset_and_smoothing(layer_thickness_um: np.ndarray) -> np.ndarray:
    """Returns the density weights for a multilayer stack as a levelset.                                          
                                                                                                                  
    Args:                                                                                                         
      layer_thickness_um: thickness of each layer in the stack.                                                   
                                                                                                                  
    Returns:                                                                                                      
      The density weights as a flattened (1D) array.                                                              
    """

    air_padding_um = 0.5 * (DESIGN_REGION_UM.z - np.sum(layer_thickness_um))

    weights = np.zeros(NZ_DESIGN_GRID)

    # Air padding at left edge                                                                                    
    z_start = 0
    z_end = int(air_padding_um * DESIGN_REGION_RESOLUTION_UM)
    weights[z_start:z_end] = 0

    z_start = z_end
    for j in range(NUM_LAYERS):
        z_end = z_start + int(layer_thickness_um[j] * DESIGN_REGION_RESOLUTION_UM)
        weights[z_start:z_end] = 1 if (j % 2 == 0) else 0
        z_start = z_end

    # Air padding at right edge                                                                                   
    z_end = z_start + int(air_padding_um * DESIGN_REGION_RESOLUTION_UM)
    weights[z_start:z_end] = 0

    # Smooth the design weights by downsampling from the design grid                                              
    # to the simulation grid using bilinear interpolation.                                                        
    z_sim_grid = design_region_to_grid(NZ_SIM_GRID)
    z_design_grid = design_region_to_grid(NZ_DESIGN_GRID)
    smoothed_weights = np.interp(z_sim_grid, z_design_grid, weights)

    return smoothed_weights.flatten()

def input_flux() -> np.ndarray:
    """Returns the flux generated by the source in vacuum."""

    pml_layers = [mp.PML(thickness=PML_UM)]

    size_z_um = PML_UM + AIR_UM + DESIGN_REGION_UM.z + AIR_UM + PML_UM
    cell_size = mp.Vector3(0, 0, size_z_um)

    src_cmpt = mp.Ex
    src_pt = mp.Vector3(0, 0, -0.5 * size_z_um + PML_UM)
    sources = [
        mp.Source(
            mp.GaussianSource(frequency_center, fwidth=frequency_width),
            component=src_cmpt,
            center=src_pt,
        )
    ]

    sim = mp.Simulation(
        resolution=RESOLUTION_UM,
        cell_size=cell_size,
        dimensions=1,
        boundary_layers=pml_layers,
        sources=sources
    )

    tran_pt = mp.Vector3(0, 0, 0.5 * size_z_um - PML_UM)

    flux_mon = sim.add_flux(
        frequencies,
        mp.FluxRegion(center=tran_pt)
    )

    sim.run(
            25.0, src_cmpt, tran_pt, 1e-6
        )
    )

    source_flux = mp.get_fluxes(flux_mon)

    return source_flux

def multilayer_stack(
        input_flux: np.ndarray,
        layer_thickness_um: np.ndarray
) -> Tuple[np.ndarray, np.ndarray]:
    """Returns the DFT fields of a multilayer stack."""

    pml_layers = [mp.PML(thickness=PML_UM)]

    size_z_um = PML_UM + AIR_UM + DESIGN_REGION_UM.z + AIR_UM + PML_UM
    cell_size = mp.Vector3(0, 0, size_z_um)

    src_cmpt = mp.Ex
    src_pt = mp.Vector3(0, 0, -0.5 * size_z_um + PML_UM)
    sources = [
        mp.Source(
            mp.GaussianSource(frequency_center, fwidth=frequency_width),
            component=src_cmpt,
            center=src_pt,
        )
    ]

    mat_1 = mp.Medium(index=N_LAYER_1)
    mat_2 = mp.Medium(index=N_LAYER_2)

    weights = levelset_and_smoothing(layer_thickness_um)
    matgrid = mp.MaterialGrid(
        mp.Vector3(0, 0, NZ_SIM_GRID),
        mat_1,
        mat_2,
        weights
    )

    geometry = [
        mp.Block(
            material=matgrid,
            center=mp.Vector3(),
            size=DESIGN_REGION_UM
        )
    ]

    sim = mp.Simulation(
        resolution=RESOLUTION_UM,
        cell_size=cell_size,
        dimensions=1,
        boundary_layers=pml_layers,
        sources=sources,
        geometry=geometry
    )

    dft_field_mon = sim.add_dft_fields(
        [mp.Ex],
        frequencies,
        center=mp.Vector3(),
        size=DESIGN_REGION_UM,
        yee_grid=True
    )

    tran_pt = mp.Vector3(0, 0, 0.5 * size_z_um - PML_UM)

    flux_mon = sim.add_flux(
        frequencies,
        mp.FluxRegion(center=tran_pt)
    )

    sim.run(
            25.0, src_cmpt, tran_pt, 1e-8
        )
    )

    ex_dft = np.zeros(
        (num_wavelengths, NZ_SIM_GRID),
        dtype=np.complex128
    )
    for j in range(num_wavelengths):
        ex_dft[j, :] = sim.get_dft_array(dft_field_mon, mp.Ex, j)

    tran_flux = mp.get_fluxes(flux_mon)
    transmittance = np.asarray(tran_flux) / np.asarray(norm_flux)

    return transmittance, ex_dft


if __name__ == "__main__":
    norm_flux = input_flux()

    layer_thickness_um = [
        0.1664, 0.2452, 0.1399, 0.2719, 0.1786, 0.3060, 0.4952, 0.1688, 0.2306
    ]
    tran, ex_dft = multilayer_stack(norm_flux, layer_thickness_um)

    markers = ['bo-', 'ro-']
    z_grid = design_region_to_grid(NZ_SIM_GRID)
    fig, ax = plt.subplots()
    for j in range(num_wavelengths):
        ax.plot(
            z_grid,
            np.absolute(ex_dft[j, :])**2,
            markers[j],
            label=(
                f"$\lambda$ = {DESIGN_WAVELENGTHS_UM[j]} μm, "
                f"T = {tran[j]:.4f}"
            )
        )
    ax.set_xlabel("z coordinate (μm)")
    ax.set_ylabel("$|E_x|^2$ in multilayer stack")
    ax.legend()
    if mp.am_master():
        fig.savefig(
            "broadband_field_decay_in_multilayer_stack.png",
            dpi=150,
            bbox_inches="tight"
        )
```

stevengj:
For comparison, a simple hand design would be to use a quarter-wave stack at the center frequency given by the harmonic mean of the wavelengths `λ₀ = 1 / mean(1 ./ [0.95, 1.05]) == 0.9975 ≈ 1`, and hope that the band gap is wide enough to cover both wavelengths.   It would be nice to know how well the optimization does compared to this simple design.

e.g. plot the figure of merit  normalized by the q-wave values.

oskooi:
> e.g. plot the figure of merit normalized by the q-wave values.

> I would use the CCSA-quadratic (#2400) algorithm in NLopt rather than MMA. 

I made a number of changes based on the suggestions which together seem to produce decent results:
1. Changed the optimization algorithm from MMA to CCSA.
2. Used tighter bounds for the layer thicknesses based on the values for the quarter-wavelength stack.
3. Ran the optimization problem 10 times and chose the design with the largest field decay (smallest ∫|E|²).

The two existing figures have been updated with the new results. The plot of the field decay in the stack has been updated to normalize the fields by those from the reference design of the quarter-wavelength stack. A new schematic showing the design problem has also been added.

doc/docs/Python_Tutorials/Adjoint_Solver.md
```diff
@@ -1190,6 +1190,456 @@ if __name__ == "__main__":
     )
 ```
 
+Shape Optimization of a Multilayer Stack
+----------------------------------------
+
+We extend the demonstration of the shape derivative from the previous tutorial to perform shape optimization of a multilayer stack over a broad bandwidth. The 1D design problem is shown in the schematic below and involves finding the layer thicknesses for a fixed number of layers (9) which minimize the largest transmittance (or alternatively, maximize the smallest reflectance) at two wavelengths: $\lambda_1$ = 0.95 μm and $\lambda_2$ = 1.05 μm. However, rather than use the transmittance as the objective function, we use the *field decay* via (the logarithm of) the integral of the fields in the stack ($\int |E_x|^2$). This provides more information to the optimizer which should produce better results.
```
stevengj, line 1196: Maybe we should be minimizing transmittance now that it's not too small.

stevengj: I'm worried that we are saying something that is false here.

stevengj, line 1196:
```suggestion
We extend the demonstration of the shape derivative from the previous tutorial to perform shape optimization of a multilayer stack over a broad bandwidth. The 1D design problem is shown in the schematic below and involves finding the layer thicknesses for a fixed number of layers (9) which minimize the integrated field intensity $\int |E_x|^2$, which demonstrates a `FourierFields` objective function.  (This is equivalent to minimizing absorption if $\varepsilon$ had a small imaginary part, and is related but not precisely equivalent to minimizing transmission or maximizing reflection.)  In particular, we minimize the worst case (largest) of the intensities at two wavelengths: $\lambda_1$ = 0.95 μm and $\lambda_2$ = 1.05 μm, to roughly emulate a broadband problem.
```

doc/docs/Python_Tutorials/Adjoint_Solver.md
```diff
@@ -1190,6 +1190,456 @@ if __name__ == "__main__":
...
+
+The stack consists of two materials of alternating refractive index $n_A$ = 1.3 and $n_B$ = 1.0. The layers are arranged as $n_A$, $n_B$, $n_A$, $n_B$, ..., $n_A$. The semi-infinite regions to the left and right of the stack are vacuum.
+
+![](../images/multilayer_stack_shape_opt.png#center)
+
+A reference design to compare our results against is a [quarter-wavelength stack](https://en.wikipedia.org/wiki/Distributed_Bragg_reflector). The mean frequency of $\lambda_1$ and $\lambda_2$ is $\lambda \approx$ 1.0 μm for which the quarter-wavelength layer thicknesses are approximately $\lambda / (4 n_A)$ = 0.19 μm and $\lambda / (4 n_B)$ = 0.25 μm. These values are used to specify upper and lower bounds for the layer thicknesses. This is important given the non-convex nature of this particular design problem.
```
stevengj, line 1202:
```suggestion
A reference design to compare our results against is a [quarter-wavelength stack](https://en.wikipedia.org/wiki/Distributed_Bragg_reflector). The mean wavelength of $\lambda_1$ and $\lambda_2$ is $\lambda = 1.0 μm for which the quarter-wavelength layer thicknesses are $\lambda / (4 n_A)$ = 0.19 μm and $\lambda / (4 n_B)$ = 0.25 μm. These values are used to specify upper and lower bounds for the layer thicknesses. This is important given the non-convex nature of this particular design problem.
```

doc/docs/Python_Tutorials/Adjoint_Solver.md
```diff
@@ -1190,6 +1190,456 @@ if __name__ == "__main__":
...
+
+The stack consists of two materials of alternating refractive index $n_A$ = 1.3 and $n_B$ = 1.0. The layers are arranged as $n_A$, $n_B$, $n_A$, $n_B$, ..., $n_A$. The semi-infinite regions to the left and right of the stack are vacuum.
+
+![](../images/multilayer_stack_shape_opt.png#center)
+
+A reference design to compare our results against is a [quarter-wavelength stack](https://en.wikipedia.org/wiki/Distributed_Bragg_reflector). The mean wavelength of $\lambda_1$ and $\lambda_2$ is $\lambda = 1.0 μm for which the quarter-wavelength layer thicknesses are $\lambda / (4 n_A)$ = 0.19 μm and $\lambda / (4 n_B)$ = 0.25 μm. These values are used to specify upper and lower bounds for the layer thicknesses. This is important given the non-convex nature of this particular design problem.
+
+(For several layers and a single wavelength, the optimizer can easily find a non-quarter-wavelength structure with very low transmission, and once it gets to that point the gradients will be almost zero so probably it will converge very slowly towards a quarter-wavelength solution because the optimizer is generally very slow to try to reduce transmittance from 1e-3 to 1e-6, for example.)
```
stevengj, line 1204:
```suggestion
```

doc/docs/Python_Tutorials/Adjoint_Solver.md
```diff
@@ -1190,6 +1190,455 @@ if __name__ == "__main__":
...
+
+The stack consists of two materials of alternating refractive index $n_A$ = 1.3 and $n_B$ = 1.0. The layers are arranged as $n_A$, $n_B$, $n_A$, $n_B$, ..., $n_A$. The semi-infinite regions to the left and right of the stack are vacuum.
+
+![](../images/multilayer_stack_shape_opt.png#center)
+
+A reference design to compare our results against is a [quarter-wavelength stack](https://en.wikipedia.org/wiki/Distributed_Bragg_reflector). The mean wavelength of $\lambda_1$ and $\lambda_2$ is $\lambda = 1.0 μm for which the quarter-wavelength layer thicknesses are $\lambda / (4 n_A)$ = 0.19 μm and $\lambda / (4 n_B)$ = 0.25 μm. These values are used to specify upper and lower bounds for the layer thicknesses. This is important given the non-convex nature of this particular design problem.
+
```
stevengj, line 1203:
```suggestion
```

doc/docs/Python_Tutorials/Adjoint_Solver.md
```diff
@@ -1190,6 +1190,454 @@ if __name__ == "__main__":
...
+The worst-case optimization is implemented using the [epigraph formulation](https://nlopt.readthedocs.io/en/latest/NLopt_Introduction/#equivalent-formulations-of-optimization-problems) and the Conservative Convex Separable Approximation (CCSA) algorithm. The optimization is run ten times with random initial designs from which the local optima with the smallest objective value is chosen. For the run involving the best design, a plot of the objective function vs. iteration number for the two wavelengths is shown below. Also included in this plot is the epigraph variable. These results demonstrate that the stack is being optimized for $\lambda$ = 0.95 μm and *not* $\lambda$ = 1.05 μm. We observed this trend for various local optima. The optimizer converges to the local optima using 13 iterations. The nine layer thicknesses (μm) of this design are: 0.1822, 0.2369, 0.1822, 0.2369, 0.1822, 0.2369, 0.1910, 0.2504, 0.1931. Since these are 1D simulations, the runtime for each design iteration is generally fast (about ten seconds using one core of an Intel Xeon i7-7700K CPU @ 4.20GHz).
```
stevengj, line 1204:
```suggestion
The worst-case optimization is implemented using the [epigraph formulation](https://nlopt.readthedocs.io/en/latest/NLopt_Introduction/#equivalent-formulations-of-optimization-problems) and the Conservative Convex Separable Approximation (CCSA) algorithm. The optimization is run ten times with random initial designs from which the local optimum with the smallest objective value is chosen. For the run involving the best design, a plot of the objective function vs. iteration number for the two wavelengths is shown below. Also included in this plot is the epigraph variable. These results demonstrate that the performance of the stack is limited by that of the shorter wavelength $\lambda$ = 0.95 μm, so the $\lambda$ = 1.05 μm performance does not directly affect the result. We observed this trend for various local optima. The optimizer converges to the local optimum after 13 iterations. The nine layer thicknesses of this design are (in μm): 0.1822, 0.2369, 0.1822, 0.2369, 0.1822, 0.2369, 0.1910, 0.2504, 0.1931. Since these are 1D simulations, the runtime for each design iteration is generally fast (about ten seconds using one core of an Intel Xeon i7-7700K CPU @ 4.20GHz).
```

stevengj:
It would be good to double-check the quarter-wave transmission — I'm surprised that the transmission is so low, even though the index contrast is low, though I have no reason to doubt Meep.

Might be nice to just try minimizing transmission too.  (A quarter-wave stack might not be completely optimal even at a single frequency because there might be a better termination for the first and last layers?  I don't remember.)

oskooi:
> It would be good to double-check the quarter-wave transmission — I'm surprised that the transmission is so low, even though the index contrast is low, though I have no reason to doubt Meep.

I confirmed that the relatively large transmittance of ~0.3 of the quarter-wavelength stack in the example is due to the index contrast of 1.3. For comparison, the transmittance is 1e-5 for a quarter-wavelength stack with nine layers for an index contrast of 3.5. This information has been added to the tutorial in #2881.

PR #2895, opened 2024-08-27
Changes to _eval_step_func

kombatEldridge:
In my project, I am wanting to use some step functions while Meep runs. Currently, my entire simulation of Meep runs in a Simulation class and I would like to keep my step functions in this class as well. That being said, they must have `self` as an arg. With the current `_eval_step_func`, the arg handler is a bit too strict. There is probably a reason for this, and please disregard this PR if so.

The changes I am requesting would include the possibility that the step functions contain an arg named `self` but should still keep the intended limitations of the previous version.

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -4996,14 +4996,13 @@ def _combine_step_funcs(*step_funcs):
 
 def _eval_step_func(sim, func, todo):
     num_args = get_num_args(func)
+    name_args = func.__code__.co_varnames
+    self_count = int("self" in name_args)
 
-    if num_args != 1 and num_args != 2:
-        raise ValueError(f"Step function '{func.__name__}'' requires 1 or 2 arguments")
-    elif num_args == 1:
-        if todo == "step":
-            func(sim)
-    elif num_args == 2:
-        func(sim, todo)
+    if num_args not in {1 + self_count, 2 + self_count}:
+        raise ValueError(f"Step function '{func.__name__}' requires 1 or 2 arguments")
+
+    func(sim, todo) if num_args == 2 + self_count else func(sim)
 
 
 def _when_true_funcs(cond, *step_funcs):
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -97,7 +97,7 @@ def get_num_args(func):
     if isinstance(func, Harminv) or isinstance(func, PadeDFT):
         return 2
     elif inspect.ismethod(func):
-        return func.__code__.co_argcount - 1 # remove 'self' from count
+        return func.__code__.co_argcount - 1  # remove 'self' from count
     else:
         return func.__code__.co_argcount
 
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -97,7 +97,7 @@ def get_num_args(func):
     return (
         2
         if isinstance(func, Harminv) or isinstance(func, PadeDFT)
-        else func.__code__.co_argcount - 1 # first argument is "self", passed implicitly for bound methods
+        else func.__code__.co_argcount - 1 # first argument is "self"
         if inspect.ismethod(func)
         else func.__code__.co_argcount
     )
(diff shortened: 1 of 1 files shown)


oskooi:
>  Currently, my entire simulation of Meep runs in a Simulation class and I would like to keep my step functions in this class as well. That being said, they must have self as an arg. With the current _eval_step_func, the arg handler is a bit too strict.

If you haven't already done so, it might be useful to first read [The Run Function is not a Loop](https://meep.readthedocs.io/en/master/The_Run_Function_Is_Not_A_Loop/) which provides a description of the step-function object including examples.

python/simulation.py
```diff
@@ -4996,14 +5004,16 @@ def _combine(sim, todo):
 
 def _eval_step_func(sim, func, todo):
     num_args = get_num_args(func)
+    name_args = get_name_args(func)
+    self_count = int("self" in name_args)
```
stevengj, line 5008: Using `inspect.ismethod(func)` seems like a more reliable way to check for this.  However, this seems unreliable in another way: it assumes that any bound method is a method of the `sim` object.    What if it is a method of some other object?

stevengj:
Note that you can already pass as many additional arguments you want to the step function by wrapping it in a [closure](https://www.pythontutorial.net/advanced-python/python-closures/).

For example, if you have a function `my_step(a,b,c,sim)`, you can pass it as a step function with `lambda sim: my_step(a,b,c,sim)`, in which the additional arguments `a,b,c` are captured by the closure.

Or, if you have a function `foo.my_step(sim)` that is a bound method of some object `foo`, you can pass it as a step function with `lambda sim: foo.my_step(sim)`.

python/simulation.py
```diff
@@ -100,6 +100,14 @@ def get_num_args(func):
     )
 
 
+def get_name_args(func):
+    return (
+        ("sim", "todo")
+        if isinstance(func, Harminv) or isinstance(func, PadeDFT)
+        else func.__code__.co_varnames[: func.__code__.co_argcount]
+    )
```
stevengj, line 108: This can be eliminated if we use `inspect.ismethod` as indicated below.

stevengj:
Overall, I'm in favor of something like this to make it easier to pass bound methods for step functions.

However, what I'd like to do is to simply fix `get_num_args` to subtract 1 if the function is a bound method, as determined by `inspect.ismethod(func)` (not from the argument names!).

python/simulation.py
```diff
@@ -96,6 +97,8 @@ def get_num_args(func):
     return (
         2
         if isinstance(func, Harminv) or isinstance(func, PadeDFT)
+        else func.__code__.co_argcount - 1
```
stevengj, line 100:
```suggestion
        else func.__code__.co_argcount - 1 # first argument is "self", passed implicitly for bound methods
```

stevengj (approved):

python/simulation.py
```diff
@@ -96,6 +97,8 @@ def get_num_args(func):
     return (
         2
         if isinstance(func, Harminv) or isinstance(func, PadeDFT)
+        else func.__code__.co_argcount - 1 # first argument is "self", passed implicitly for bound methods
```
stevengj, line 100:
```suggestion
        else func.__code__.co_argcount - 1 # first argument is "self"
```

python/simulation.py
```diff
@@ -96,6 +97,8 @@ def get_num_args(func):
     return (
         2
         if isinstance(func, Harminv) or isinstance(func, PadeDFT)
+        else func.__code__.co_argcount - 1 # first argument is "self"
```
stevengj, line 100:
```suggestion
        else func.__code__.co_argcount - 1  # first argument is "self"
```

python/simulation.py
```diff
@@ -96,6 +97,8 @@ def get_num_args(func):
     return (
         2
         if isinstance(func, Harminv) or isinstance(func, PadeDFT)
+        else func.__code__.co_argcount - 1  # first argument is "self"
+        if inspect.ismethod(func)
```
oskooi, line 101: For improved readability, I would recommend you replace this ternary operator which spans multiple lines with a full `if` statement. 

See: https://google.github.io/styleguide/pyguide.html#2111-definition.

Issue #2916, opened 2024-10-02
Segmentation fault for 3D simulation using 1D cell


oskooi:
I am running into an unexpected segmentation fault when trying to run a 3D simulation using a 1D cell. This only occurs for certain grid resolutions and Bloch-periodic boundary conditions. 

This issue can be reproduced using the script below which involves a point source in vacuum with Bloch-periodic boundaries in $x$ and $y$ (`k_point = (-0.2, -0.2, 0)`). The simulation is fine when `RESOLUTION_UM` is `51` but produces the segfault for values of 52 and 53. Changing the `k_point` also runs fine (no segfault).

The problem therefore seems to be some combination of the grid resolution and the Bloch-periodic boundaries for a 1D cell. I should also mention that I am using Meep compiled using single-precision floating point and OpenMP.

**Case 1: `RESOLUTION_UM = 51`**
```
Initializing structure...
time for choose_chunkdivision = 7.7198e-05 s
Working in 3D dimensions.
Computational cell is 0.0196078 x 0.0196078 x 12 with resolution 51
time for set_epsilon = 0.00449374 s
-----------
Meep: using complex fields.
field decay(t = 50.009803921568626): 0.2511579661675122 / 0.2511579661675122 = 1.0
field decay(t = 100.01960784313725): 0.25115618711643334 / 0.2511579661675122 = 0.9999929166050117
field decay(t = 150.02941176470588): 1.7969355984235213e-10 / 0.2511579661675122 = 7.154603239719811e-10
run 0 finished at t = 150.02941176470588 (15303 timesteps)

Elapsed run time = 2.3554 s
```

**Case 2: `RESOLUTION_UM = 52`**
```
Using MPI version 3.1, 1 processes
-----------
Initializing structure...
time for choose_chunkdivision = 9.0651e-05 s
Working in 3D dimensions.
Computational cell is 0.0192308 x 0.0192308 x 12 with resolution 52
time for set_epsilon = 0.00381082 s
-----------
Meep: using complex fields.
[devbig095:2604209:0:2604209] Caught signal 11 (Segmentation fault: address not mapped to object at address (nil))
==== backtrace (tid:2604209) ====
 0  /lib64/libucs.so.0(ucs_handle_error+0x2e4) [0x7ff45fc81e44]
 1  /lib64/libucs.so.0(+0x2a4cd) [0x7ff45fc834cd]
 2  /lib64/libucs.so.0(+0x2a6aa) [0x7ff45fc836aa]
[...]

PR #2917, opened 2024-10-03
Tutorial example for radiation pattern of a dipole using 1D calculation and Brillouin-zone integration

### LINKED ISSUE #2866: "1d" Green's functions for near2far
For a "1d" system described by a 2d (or 3d) simulation with Bloch-periodic boundary conditions in the x (or xy) directions, the Green's function is simply a planewave.  It would be nice to support this directly in the near2far feature.

(We support a *finite* periodicity by a lattice sum, but in the limit of zero periodicity this is absurdly inefficient … the lattice sum effectively goes to a quadrature rule.  We should just use the planewave formula.)

This is potentially useful for things like computing the Green's function (dipole emission pattern) for a multilayer film, which corresponds to simply integrating this "1d" near2far field over kx and ky (inside the light code).  Of course, there are many specialized methods to compute multilayer Green's functions that will be vastly more efficient than Meep for this purpose (for free/open-source code, see e.g. https://github.com/modelics/strata and https://github.com/odmiller/MultilayerEM), but it shouldn't be too expensive in Meep.

oskooi:
Closes #2866.

This is still a work in progress. This implementation is based on the procedure described in [#2866 (comment)](https://github.com/NanoComp/meep/issues/2866#issuecomment-2362039118) for computing the radiation pattern of a dipole in vacuum using a 1D calculation and Brillouin-zone integration.

The radiation pattern of a dipole should be similar to a ["donut"](https://en.wikipedia.org/wiki/Dipole_antenna#Radiation_pattern_and_gain) with rotational symmetry around the axis defined by the dipole orientation. However, as shown in the image below, the radiation pattern generated by this script is rotationally symmetric around the $z$ axis. This seems to be incorrect because the dipole is oriented in the $x$ direction.

For reference, the script used to generate the radiation pattern from the farfield data produced as the output from the script is provided below.

```py
from typing import Tuple

import numpy as np


def spherical_to_cartesian(
	polar_rad,
	azimuthal_rad
) -> Tuple[float, float, float]:
    """Converts a point in spherical to Cartesian coordinates."""

    rx = np.sin(polar_rad) * np.cos(azimuthal_rad)
    ry = np.sin(polar_rad) * np.sin(azimuthal_rad)
    rz = np.cos(polar_rad)

    return rx, ry, rz

if __name__ == "__main__":
    data = np.load("dipole_farfields.npz")

    farfield_radius_um = data['FARFIELD_RADIUS_UM']
    polar_rad = data['polar_rad']
    azimuthal_rad = data['azimuthal_rad']
    num_polar = data['NUM_POLAR']
    num_azimuthal = data['NUM_AZIMUTHAL']

    flux_z = np.zeros((num_polar, num_azimuthal))
    xs = np.zeros((num_polar, num_azimuthal))
    ys = np.zeros((num_polar, num_azimuthal))
    zs = np.zeros((num_polar, num_azimuthal))

    for i in range(num_polar):
	for j in range(num_azimuthal):

            flux_z[i, j] = np.real(
		np.conj(data['Ex'][i, j]) * data['Hy'][i, j] -
		np.conj(data['Ey'][i, j]) * data['Hx'][i, j]
            )

            xs[i, j], ys[i, j], zs[i, j] = spherical_to_cartesian(
		polar_rad[i], azimuthal_rad[j]
            )

            xs[i, j] *= flux_z[i, j]
            ys[i, j] *= flux_z[i, j]
            zs[i, j] *= flux_z[i, j]

    fig, ax = plt.subplots(subplot_kw={"projection": "3d"}, figsize=(6, 6))
    ax.plot_surface(xs, ys, zs, cmap="inferno")
    ax.set_title("radiation pattern in 3d")
    ax.set_box_aspect((np.amax(xs), np.amax(ys), np.amax(zs)))
    ax.set_zlabel("z flux (a.u.)", labelpad=30)
    ax.tick_params(axis="z", which="major", pad=15)
    ax.set(xticklabels=[], yticklabels=[])

    fig.savefig(
	"radiation_pattern_3d.png",
	dpi=150,
	bbox_inches="tight",
    )

```

radiation_pattern_3d


stevengj:
You need to take into account $k_x, k_y$ in the near-to-far calculation, since the current sheets are oscillating in $(x,y)$.  (Note that this will also give you nonzero $z$ components in the far field.)

stevengj:
(Note that you might run into an issue with the 1d PML, because when you are right on the light line the fields won't decay.)

oskooi:
> You need to take into account $k_x, k_y$ in the near-to-far calculation, since the current sheets are oscillating in $(x,y)$.  (Note that this will also give you nonzero $z$ components in the far field.)

Here are the equations for the S- and P-polarized farfields given a current sheet oscillating in $(x,y)$.

sheet_current_farfields_Spol

sheet_current_farfields_Ppol

oskooi:
After updating the script to include the corrected formulas for obtaining the farfields given a current sheet, the farfields given an $x$-directed dipole is not rotationally symmetric but is still not quite "donut" like.

radiation_pattern_3d

oskooi:
```py
from typing import Tuple

import numpy as np


def spherical_to_cartesian(
        polar_rad,
        azimuthal_rad
) -> Tuple[float, float, float]:
    """Converts a point in spherical to Cartesian coordinates."""

    rx = np.sin(polar_rad) * np.cos(azimuthal_rad)
    ry = np.sin(polar_rad) * np.sin(azimuthal_rad)
    rz = np.cos(polar_rad)

    return rx, ry, rz


if __name__ == "__main__":
    data = np.load("dipole_farfields.npz")

    farfield_radius_um = data['FARFIELD_RADIUS_UM']
    polar_rad = data['polar_rad']
    azimuthal_rad = data['azimuthal_rad']
    num_polar = data['NUM_POLAR']
    num_azimuthal = data['NUM_AZIMUTHAL']

    flux_x = np.zeros((num_polar, num_azimuthal))
    flux_y = np.zeros((num_polar, num_azimuthal))
    flux_z = np.zeros((num_polar, num_azimuthal))
    xs = np.zeros((num_polar, num_azimuthal))
    ys = np.zeros((num_polar, num_azimuthal))
    zs = np.zeros((num_polar, num_azimuthal))

    for i in range(num_polar):
        for j in range(num_azimuthal):

            flux_x[i, j] = np.real(
                np.conj(data['Ey'][i, j]) * data['Hz'][i, j] -
                np.conj(data['Ez'][i, j]) * data['Hy'][i, j]
            )

            flux_y[i, j] = np.real(
                np.conj(data['Ez'][i, j]) * data['Hx'][i, j] -
                np.conj(data['Ex'][i, j]) * data['Hz'][i, j]
            )

            flux_z[i, j] = np.real(
                np.conj(data['Ex'][i, j]) * data['Hy'][i, j] -
                np.conj(data['Ey'][i, j]) * data['Hx'][i, j]
            )

            flux_r = np.sqrt(flux_x[i, j]**2 + flux_y[i, j]**2 + flux_z[i, j]**2)

            rx, ry, rz = spherical_to_cartesian(polar_rad[i], azimuthal_rad[j])

            xs[i, j] = rx * flux_x[i, j]
            ys[i, j] = ry * flux_y[i, j]
            zs[i, j] = rz * flux_z[i, j]

    fig, ax = plt.subplots(subplot_kw={"projection": "3d"}, figsize=(6, 6))
    ax.plot_surface(xs, ys, zs, cmap="inferno")
    ax.set_title("radiation pattern in 3d")
    ax.set_box_aspect((np.amax(xs), np.amax(ys), np.amax(zs)))
    ax.set_zlabel("z flux (a.u.)", labelpad=30)
    ax.tick_params(axis="z", which="major", pad=15)
    ax.set(xticklabels=[], yticklabels=[])

    fig.savefig(
        "radiation_pattern_3d.png",
        dpi=150,
        bbox_inches="tight",
    )
```

stevengj:
I think you want use
```py
xs[i, j] = rx
ys[i, j] = ry
zs[i, j] = rz
flux_r[i,j] = rx * flux_x[i, j] + ry * flux_y[i, j] + rz * flux_z[i, j]
```
with the surface given by `(xs,ys,zs)` but the color given by `flux_r`.  Here is an [example in matplotlib](https://stackoverflow.com/questions/32413371/color-of-a-3d-surface-plot-in-python) — you have to use the `facecolors` argument to `plot_surface`.  Something like:
```py
c = matplotlib.cm.magma(flux_r)
plot_surface(xs,ys,zs, facecolors=c)
```

oskooi:
I fixed a couple of bugs in the calculation of the far fields. However, even with these fixes the radiation pattern for $\phi = 0$ (i.e., a slice of $\phi$ in the range $[0, 2\pi]$) is quite different than the expected $\cos^2(\theta)$. (A fixed $\phi$ is intended to simplify the analysis from a 3D contour which can be difficult to visualize correctly to a 2D polar plot.) The script used to generate the 2D radiation pattern (polar plot of the radial flux) using the simulation data is provided below. For reference, the radiation pattern for a dipole in 3D resembling a "donut" is shown as a separate figure below.

At this point, it might be worth investigating whether setting up the 3D simulation using a 1D grid itself is bug free (e.g., see #2916).

radiation_pattern_jx

radiation_pattern_of_a_dipole

```py
import math
from typing import Tuple

import matplotlib
import numpy as np


if __name__ == "__main__":
    data = np.load("dipole_farfields.npz")

    polar_rad = data['polar_rad']
    num_polar = data['NUM_POLAR']
    flux_x = np.zeros(num_polar)
    flux_z = np.zeros(num_polar)

    for i in range(num_polar):
        flux_x[i] = np.real(
            np.conj(data['Ey'][i, 0]) * data['Hz'][i, 0] -
            np.conj(data['Ez'][i, 0]) * data['Hy'][i, 0]
	)

	flux_z[i] = np.real(
            np.conj(data['Ex'][i, 0]) * data['Hy'][i, 0] -
            np.conj(data['Ey'][i, 0]) * data['Hx'][i, 0]
        )

    flux_r = np.sqrt(np.square(flux_x) + np.square(flux_z))

    np.savez(
        "dipole_radiation_pattern.npz",
        flux_r=flux_r,
        polar_rad=polar_rad,
    )

    fig, ax = plt.subplots(subplot_kw={"projection": "polar"}, figsize=(6, 6))
    ax.plot(polar_rad, flux_r / np.max(flux_r), 'b-', label="Meep")
    ax.plot(polar_rad, np.square(np.cos(polar_rad)), 'r--', label="cos$^2$(θ)")
    ax.set_theta_direction(-1)
    ax.set_theta_offset(0.5 * math.pi)
    ax.set_thetalim(0, 0.5 * math.pi)
    ax.set_rmax(1)
    ax.set_rticks([0,0.5,1])
    ax.grid(True)
    ax.set_rlabel_position(22)
    ax.legend()

    fig.savefig(
        "radiation_pattern_jx.png",
        dpi=150,
        bbox_inches="tight",
    )
```

python/examples/dipole_in_vacuum_1D.py
```diff
@@ -0,0 +1,353 @@
...
+
+            if DEBUG_OUTPUT:
+                print(f"ex_dft:, {ex_dft}")
+                print(f"ey_dft:, {ey_dft}")
+                print(f"hx_dft:, {hx_dft}")
+                print(f"hy_dft:, {hy_dft}")
+
+            # Rotation angle around z axis to force ky=0. 0 radians is +x.                                   
+            if kx:
+                rotation_rad = -math.atan(ky / kx)
```
oskooi, line 217: Replace `math.atan` with `math.atan2`.

python/examples/dipole_in_vacuum_1D.py
```diff
@@ -0,0 +1,353 @@
...
+                farfields[component] = np.zeros(
+                    (NUM_POLAR, NUM_AZIMUTHAL),
+                    dtype=np.complex128
+                )
+
+            for i, polar in enumerate(polar_rad):
+                for j, azimuthal in enumerate(azimuthal_rad):
+                    rx, ry, rz = spherical_to_cartesian(polar, azimuthal)
+                    r_rotated = rotation_matrix @ np.transpose(np.array([rx, ry]))
+                    phase = np.exp(1j * (np.dot(k_rotated, r_rotated) + kz * rz))
```
oskooi, line 275: Missing `2 * n.pi` in `k_rotated`.

python/examples/dipole_in_vacuum_1D.py
oskooi: Fixed.

python/examples/dipole_in_vacuum_1D.py
oskooi: Fixed.

python/examples/dipole_in_vacuum_1D.py
```diff
@@ -0,0 +1,354 @@
...
+        rx, rz: x,z coordinate of a point on the surface of a hemicircle.                                    
+        current_amplitude: amplitude of the sheet current.                                                   
+        current_component: component of the sheet current.                                                   
+                                                                                                             
+    Returns:                                                                                                 
+        A 3-tuple of the per-polarization electric and magnetic far fields.                                  
+    """
+    if current_component.name == "Jx":
+        # Jx --> (Ex, Hy, Ez) [P pol.]                                                                       
+        ex0 = frequency * current_amplitude / (2 * kz)
```
stevengj, line 134:
```suggestion
        ex0 = kz * current_amplitude / (2 * frequency)
```

python/examples/dipole_in_vacuum_1D.py
```diff
@@ -0,0 +1,354 @@
...
+                        ):
+                            # S polarization                                                                 
+                            farfields["Hx"][i, j] += farfield_pol[0]
+                            farfields["Ey"][i, j] += farfield_pol[1]
+                            farfields["Hz"][i, j] += farfield_pol[2]
+
+            for i in range(NUM_POLAR):
+                for j in range(NUM_AZIMUTHAL):
+                    # Map the rotated point to the closest azimuthal angle.                                  
+                    azimuthal_index = (j + azimuthal_index_shift) % NUM_AZIMUTHAL
```
stevengj, line 315: shifting the index seems wrong.   the point index hasn't changed in this process, you only rotated the fields and r.

oskooi:
The results for the radiation pattern of a dipole in vacuum are improved after making these two modifications to the script: 

1. Correcting the analytic formulas for the far fields from a sheet current. These errors were discovered rederiving the formulas.
2. Specifying a radius of 1 rather than 1e6 for the spherical surface for the far fields. It's not obvious why the radius of the far field surface should affet the results but it does so in a significant way.

The radiation pattern of the dipole computed using the Brillouin-zone integration is now similar to the analytic result. However, the results are not yet identical. The discrepancy should go away when using a finer and finer grid for the Brillouin-zone integration. This is something I am currently investigating.

radiation_pattern_jx_25

stevengj:
1. Probably for large R the far fields E(kx,ky) become very oscillatory functions of (kx,ky), so that your integral is not being computed accurately.  Would be nice to plot E(kx,ky) for a few different R to confirm this.
2. For small R, the angle is not being computed correctly — the angle should be from the location of your dipole source to the far-field point, not from the near-field monitor to the far-field point.

oskooi:
> 1. Probably for large R the far fields E(kx,ky) become very oscillatory functions of (kx,ky), so that your integral is not being computed accurately.  Would be nice to plot E(kx,ky) for a few different R to confirm this.

Increasing $R$ does indeed lead to increasing oscillations in the integrand. The figure below shows the real part of the far field $E_x$ at a fixed position on the hemispherical surface over the ($k_x$, $k_y$) grid for two different $R$: 1.0 vs. 1e6. 

This explains why $R$ had to be made small (< $\lambda$ = 1 μm) in order to be able generate a smooth radiation profile.

farfield_ex_vs_kx_ky_compared_farfield_radius_um

stevengj:
There are numerical methods [(Levin, 1994)](https://www.sciencedirect.com/science/article/pii/0377042794001189) to integrate fast-oscillating integrals like this accurately and efficiently, though I can't find any implementation in Python.  However, in this case we really only care about the $r \to \infty$ limit, which suggests that we really should be able to get an exact asymptotic express (e.g. by doing a saddle-point integration and taking the limit).

I *suspect* that the result in the limit will be proportional to just integrating the he powers from each $k$ separately, though I don't know what the proportionality constant will be.

stevengj:
Note that for the small-R calculation you could just take the far-field point (orange), subtract the source point (red) in Cartesian coordinates, then convert to spherical to get $\theta_2$.

mochen4:
Denote the angle of the orange dot $$\theta_3$$, then you have $$\theta_1 = \theta_2+ \theta_3 $$, and by law of sines you have $$\frac{sin \theta_2 }{R} = \frac{sin \theta_3}{L}$$. Then you have a formula relating $$\theta_1$$ and $$\theta_2$$.

mochen4:
> To generate a range of [0, π/2] for θ 2 (for comparing the Meep data to the expected result of a dipole) requires making R larger. 

For $\theta_2$ to attain π/2, we must have R > L. Alternatively, for R < L, instead of comparing the full range of data, maybe we can just compare the data between [0, $\theta_2^{max}$], where $\theta_2^{max} = arcsin(\frac{R}{L})$.

oskooi:
There seems to be an inconsistency when calculating the radiation profile of an $E_r$ dipole in vacuum in cylindrical coordinates with the expected result of $\cos^2(\theta)$ used to debug this PR which is based on a 1D calculation. The radiation profile calculated in cylindrical coordinates for an $E_r$ (or $E_\phi$) dipole at $r = 0$ and $m=-1$ is equivalent to $(1 + \cos^2(\theta))/2$ which is shown in the plot below. This expression is what we used in [Tutorial/Extraction Efficiency of a Collection of Dipoles in a Disc](https://meep.readthedocs.io/en/latest/Python_Tutorials/Near_to_Far_Field_Spectra/#extraction-efficiency-of-a-collection-of-dipoles-in-a-disc) to empirically determine the weighting factor $s(r_n)$ for a series of $N$ dipoles at position $r_n$ along a line in the $r$ direction.

Why is the radiation profile of a dipole different when computing the result in cylindrical vs. Cartesian coordinates? They should be the same.

dipole_radiation_pattern

Simulation layout of a a dipole at $r = 0$ (red dot) in vacuum in cylindrical coordinates. The blue line denotes the near-field monitor used to enclose the emitter.

dipole_in_vacuum_cyl_layout_PML_UM_2 0

stevengj:
> Why is the radiation profile of a dipole different when computing the result in cylindrical vs. Cartesian coordinates?

Because you don't have a linearly polarized dipole source, you have a circularly polarized dipole source.

The simplest thing in cylindrical coordinates should be to put an Ez source at r=0, with m=0.

If you want an Ex source, it should be equivalent to adding the fields from an Er source with m=+1 and m=-1.   (Alternatively, you can get the fields of m=-1 by taking the fields of m=+1 at a negative frequency -ω and complex conjugating.)

stevengj:
In order to complete this PR request, we really just need to go through the asymptotics carefully to figure out how to semi-analytically evaluate the fast oscillating integral at large r.  (I suspect that the final result will be a simple formula, even if the derivation is tricky.)

oskooi:
> The simplest thing in cylindrical coordinates should be to put an Ez source at r=0, with m=0.

An $E_z$ dipole at $r = 0$ with $m = 0$ produces the expected radiation profile of $\sin^2\theta$. 

I think this demonstration would make for a useful unit test and tutorial.

dipole_radiation_pattern

stevengj:
> The key feature of computing $P = \sum_k P_k$

Note that this cannot work: if you compute the power $P_k$ for each $k$ separately, it will be translation invariant (the same at all points in space) because it is the power of planewave produced by an infinite current sheet.

For highly oscillatory integrals, you could try the routines in QUADPACK for integrating against cos(kx) and sin(kx), which are [available in scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.quad.html) by passing `cos` or `sin` as weight functions to the `scipy.integrate.quad`.

stevengj:
@mfschubert pointed out that physical intuition here tells you the answer: *infinitely* far away, at each direction in the far field you should only see a planewave going radially outwards, which means that for each direction you only need *one* k point.   Given that k point (for both polarizations) you compute the amplitude of the outgoing planewave, get its Poynting flux, and you are done.

Mathematically, this is probably some kind of saddle-point approximation for $\int E_k e^{ik \cdot r} dk$ that makes the integral dominated by $k$ parallel to $r$, which becomes exact in the $\Vert r \Vert \to \infty$ limit.  It would be nice to work this out explicitly algebraically, though I don't have time for this myself right now.

stevengj:
You can check for a bug in your code in the same way that we tested the near2far code in Meep:

1. Pick some random $k_x, k_y$ and calculate the near fields and equivalent currents, and use them to predict the fields at some random small height $z = 4.567\lambda$
2. Compare these predicted "far" fields to the explicit Meep calculation of the same point

oskooi:
Here is the [script](https://gist.github.com/oskooi/8accf884e49dc1032f014e36f12c9fc1) for the planewave exctiation in 1D.

oskooi:
Following the bug fix in #3010 which affects this PR, the output of running the script for a $J_x$ current source is a radiation pattern for $\phi = 0$ (i.e. the $xz$ plane) which matches $\cos^4(\theta)$ rather than the expected $\cos^2(\theta)$. There is likely a bug in the analytic formulas used to compute the surface fields from the surface currents?

The radiation pattern (shown the plot below) is computed for polar angles $\theta$ in $[0, 57^\circ]$. Angles larger than about $57^\circ$ produce oblique planewaves that are poorly absorbed by the PML in the $z$ direction. This causes the fields to not sufficiently decay away thereby preventing the DFT fields from [converging](https://meep.readthedocs.io/en/latest/FAQ/#checking-convergence).

radiation_pattern_1D

oskooi:
This is now working after one important fix and a major simplification of the overall calculation. 

1. The fix involves specifying the source amplitude as $\cos(\theta)$=`kz / frequency`:

```py
sources = [ 
    mp.Source(
        src=mp.GaussianSource(frequency, fwidth=0.1 * frequency),
        component=src_cmpt,
        center=src_pt,
        size=mp.Vector3(),
        amplitude=kz/frequency
     )
]
```

2. The calculation involving the Brillouin-zone integration is straightforward: for each planewave wavevector $(k_x, k_y, k_z)$, compute the flux *directly* from the fields. Each wavevector corresponds to a single direction ($\theta$, $\phi$) of the radiation pattern: $(k_x, k_y, k_z) = \omega(\sin(\theta)\cos(\phi), \sin(\theta)\sin(\phi), \cos(\theta))$. There is no need to use the (original) roundabout process of (1) converting the DFT (near) fields to equivalent currents and (2) obtaining the fields of the planewave from the sheet current using an analytic transformation.

With these changes, the radiation pattern for a $\mathcal{J}_x$ and $\mathcal{J}_y$ dipole (shown below) matches the analytic result for $\phi = 0°$  of $\cos^2(\theta)$ and $1.0$ (constant) for $\theta$ in [0, 90°] where $\theta = 0°$ is the $+z$ direction and $\theta = 90°$ is $+x$. 

One challenge in the 1D simulation is that as $\theta \rightarrow 90°$ (i.e. the light cone), the $z$-PML becomes less absorbing. This means that in practice the Brillouin-zone integration must be restricted to the set of wavevectors slightly within the light cone. This can be seen in the plots of the radiation pattern.

radiation_pattern_1D

radiation_pattern_1D_jy_res50

oskooi:
I found through empirical testing that it is possible to obtain the correct results for the dipole emission *without* needing to specify the source `amplitude` as $\cos(\theta)$. The necessary modification involved the **flux calculation in the radial direction** as $\cos(\theta) * P_z$  whereas previously we were using $\hat{r} \cdot \vec{P}$ with $\hat{r} = (\cos(\phi)\sin(\theta), \sin\(\phi)\sin(\theta), \cos(\theta)$ and $\vec{P}$ is $(P_x, P_y, P_z)$. This makes sense because by Poynting's theorem the *total* flux in the 1D simulation is $P_z$ (i.e. $P_x$ and $P_y$ are ignored).

oskooi:
@stevengj, I think this is finally ready for review.

stevengj:
I'm not convinced that you should just be looking at Pz because it is a 1d computational cell.  If you do a 2d cell with a finite supercell in the $x$ direction and a line source $\sim \exp(ik_x x)$, and then compute the field and Poynting flux at the same point, you should get the same Poynting vector (possibly modulo an overall scale factor because of the way Meep scales point vs. line sources).

doc/docs/Python_Tutorials/Custom_Source.md
```diff
@@ -601,3 +601,50 @@ There are four items to note in this script. (1) Since the exact equivalence bet
...
+
+A point source in a 1D simulation (with discretization along $z$ and Bloch-periodic boundaries in $x$ and $y$) generates a planewave. A dipole can be modeled by combining the response from a series of planewaves using Brillouin-zone integration via the Fourier transform of the Dirac delta function:
+
+$$\delta(x, y) = \frac{1}{2\pi}\int\int e^{i(k_xx + k_yy)}dk_xdk_y$$.
+
+The Fourier coeffcients of a Dirac delta function are a constant (1.0). This integral is calculated over a finite number of planewaves with in-plane wavevector $k_\parallel = (k_x, k_y)$ which are *inside* the light cone: $|k_\parallel| < n_s\omega$ where $n_s$ is the refractive index of the source medium and $\omega$ is the frequency. (Note that $c = 1$ per convention in Meep.) The wavevector of the radiating planewave is $\vec{k} = n_e\omega(k_x, k_y, k_z)$ where $k_z = \sqrt{(n_e\omega)^2 - k_x^2 - k_y^2}$ and $n_e$ is the refractive index of the emission region.
```
stevengj:
```suggestion
The Fourier coeffcients of a Dirac delta function are a constant (1.0). This integral is calculated in principle over the entire $(k_x, k_y)$ plane, but in practice only terms with in-plane wavevector $k_\parallel = (k_x, k_y)$ that are *inside* the light cone: $|k_\parallel| < n_s\omega$ contribute to the far field, where $n_s$ is the refractive index of the source medium and $\omega$ is the frequency.  This finite-domain integral can then be discretized into a sum over a finite set of $k_x, k_y$ by a numerical-integration rule. (Note that $c = 1$ per convention in Meep.) The wavevector of the radiating planewave from each $(k_x,k_y)$ Fourier component of the source is $\vec{k} = n_e\omega(k_x, k_y, k_z)$ where $k_z = \sqrt{(n_e\omega)^2 - k_x^2 - k_y^2}$ and $n_e$ is the refractive index of the emission region.
```

doc/docs/Python_Tutorials/Custom_Source.md
```diff
@@ -601,3 +601,50 @@ There are four items to note in this script. (1) Since the exact equivalence bet
...
+
+In this example, we validate this 1D approach by computing the radiation pattern of a dipole (wavelength of 1 μm) in vacuum ($n_s = n_e = 1$) and comparing with the analytic result from antenna theory. The calculation involves computing the radial flux along the surface of a hemisphere defined in spherical coordinates by $(\theta, \phi)$ where $\theta \in [0, \pi/2]$ and $\phi \in [0, 2\pi]$. (Note: because of symmetry, the radiation pattern can be obtained using a reduced subset of $\phi \in [0, \pi/2]$.) Each grid point on the hemisphere involves a 1D simulation with Bloch-periodic boundaries: $(k_x, k_y, k_z) = n_e\omega(\sin(\theta)\cos(\phi), \sin(\theta)\sin(\phi), \cos(\theta))$.
```
stevengj:
```suggestion
In this example, we validate this 1D approach by computing the radiation pattern of a dipole (wavelength of 1 μm) in vacuum ($n_s = n_e = 1$) and comparing with the analytic result from antenna theory. The calculation involves computing the radial flux along the surface of a hemisphere defined in spherical coordinates by $(\theta, \phi)$ where $\theta \in [0, \pi/2]$ and $\phi \in [0, 2\pi]$. (Note: because of symmetry, the radiation pattern can be obtained using a reduced subset of $\phi \in [0, \pi/2]$.) Each grid point on the hemisphere involves a 1D simulation with Bloch-periodic boundaries: $(k_x, k_y) = n_e\omega(\sin(\theta)\cos(\phi), \sin(\theta)\sin(\phi))$ in $xy$ ($k_z$ is irrelevant due to the PML).
```

doc/docs/Python_Tutorials/Custom_Source.md
```diff
@@ -601,3 +601,50 @@ There are four items to note in this script. (1) Since the exact equivalence bet
...
+
+The 1D cell is truncated with PML in the $z$ direction. One limitation of this approach is that the set of wavevectors used in the Brillouin-zone integration must be restricted to lie within the *interior* of the light cone (i.e., at some distance away from the light line). This is because planewaves with wavevectors near the light line ($|k_\parallel| \approx n_e\omega$) propagate at glancing angles to the PML (i.e. along the $x$ or $y$ directions) and are therefore [not absorbed](https://meep.readthedocs.io/en/latest/FAQ/#why-are-the-fields-not-being-absorbed-by-the-pml). In this example, the set of wavevectors is restricted to those within 95% of the light line. This is not a major limitation in practice because the radiation in the direction to the parallel to the surface of the LED cannot be easily measured using a goniometer.
```
stevengj:
```suggestion
The 1D cell is truncated with PML in the $z$ direction. One limitation of this approach is that the set of wavevectors used in the Brillouin-zone integration must be restricted to lie within the *interior* of the light cone (i.e., at some distance away from the light line). This is because planewaves with wavevectors near the light line ($|k_\parallel| \approx n_e\omega$) propagate at glancing angles to the PML (i.e. along the $x$ or $y$ directions) and are therefore [not absorbed](https://meep.readthedocs.io/en/latest/FAQ/#why-are-the-fields-not-being-absorbed-by-the-pml). In this example, the set of wavevectors is restricted to those within 95% of the light line. This is not a major limitation in practice because the radiation in the direction to the parallel to the surface of the LED cannot be easily measured using a [goniophotometer](https://en.wikipedia.org/wiki/Goniophotometer).
```

doc/docs/Python_Tutorials/Custom_Source.md
```diff
@@ -601,3 +601,50 @@ There are four items to note in this script. (1) Since the exact equivalence bet
...
+
+![](../images/dipole_radiation_pattern_3D_ex.png#center)
+
+![](../images/dipole_radiation_pattern_3D_ey.png#center)
+
+![](../images/dipole_radiation_pattern_phi0_ex.png#center)
+
+![](../images/dipole_radiation_pattern_phi0_ey.png#center)
+
+As a final note, computing the [extraction efficiency of an LED](Local_Density_of_States.md#extraction-efficiency-of-a-light-emitting-diode-led) (not covered in this tutorial) requires the calculation of the total power emitted by the dipole. This can be done in the 1D approach outlined in this tutorial using the local density of states (LDoS) by summing the power from each point source (i.e. planewave) in the Brillouin-zone integration obtained using:
```
stevengj:
```suggestion
As a final note, computing the [extraction efficiency of an LED](Local_Density_of_States.md#extraction-efficiency-of-a-light-emitting-diode-led) (not covered in this tutorial) requires the calculation of the total power emitted by the dipole. This can be done in the 1D approach outlined in this tutorial using the local density of states (LDOS) by summing the power from each point source (i.e. planewave) in the Brillouin-zone integration obtained using:
```

oskooi:
> If you do a 2d cell with a finite supercell...you should get the same Poynting vector 

I verified that the Poynting flux in $x$, $y$, $z$ computed at a single point in vacuum is *identical* in 1D and 2D (as expected). 

The scripts are [here](https://gist.github.com/oskooi/a0fd96696a140ae1e5a3a72e283198f7). A schematic of the setup in 2D involving an $E_z$ line source with spatially varying amplitude $e^{ik_yy}$ and Bloch-periodic boundary conditions $k_y$ is shown below. The 1D simulation involves an $E_y$ point source with planewave wavevector rotation around the $y$ axis of angle $\theta$ in the $xz$ plane. $\theta = 0^\circ$ is the $z$ axis in 1D and $x$ axis in 2D. 

I tested six planewaves with rotation $\theta$ of: 0°, 10.65°, 20.19°, 35.88°, 47.49°, and 55.53°. The flux values are the same in 1D and 2D for all cases.


planewave_in_vacuum_2D_labelled

oskooi:
@mfschubert, in your [FMMAX calculation of the radiation pattern of a dipole](https://mfschubert.github.io/fmmax/notebooks/isolated_dipole_farfield.html#isolated-dipole-farfield), the Poynting flux in the direction $\hat{r}$ of the planewave wavevector $\vec{k}$ (used to obtain one point of the radiation pattern) seems to be computed from the Poynting flux in the $z$ direction (`backward_flux`):

Screenshot 2025-06-17 at 20 24 35

`backward_flux` is then used to compute the radiation pattern:

Screenshot 2025-06-17 at 20 25 25


As a reference, the flux calculation in the $\pm z$ directions is implemented in the function [`directional_poynting_vector`](https://github.com/mfschubert/fmmax/blob/0b760cfaf57fe868b85ba0ed4dcaf368b78b7d54/src/fmmax/_orig/fields.py#L221-L233).

The FMMAX approach seems to be consistent with the approach implemented in this PR as described in this paragraph from the tutorial:

image

Is this correct?

mfschubert:
@oskooi Yes, this seems correct. Is there an issue or are you just validating?

oskooi:
Thanks for confirming.

The issue, as I explained, is the calculation of the flux for each planewave (in the Brillouin-zone integration) in the direction of its wavevector ($P_r$). It seems the correct approach is to use $P_r = \cos(\theta) P_z$. This seems to be the approach implemented in FMMAX. This approach produces a radiation pattern in Meep which (similar to FMMAX) matches the expected result from antenna theory. 

A different approach involving $P_r = \sin(\theta)\cos(\phi)P_x + \sin(\theta)\sin(\phi)P_y + \cos(\theta)Pz$ (discussed with @stevengj) does *not* produce the expected result.

stevengj:
@mfschubert's division of power by $\cos \theta$ to get power per unit solid angle makes a lot of sense: it is basically a Jacobian factor to convert the integration with respect to $k$ to integration with respect to solid angle.

However, I'm still confused as to why you would apply this Jacobian factor to $P_z$, rather than the total Poynting flux of the planewave at that $k$.

mfschubert:
A bit of clarification about fmmax. The quantity computed by `directional_poynting_flux` is the total flux per order per unit cell area. This means e.g. that a diffraction grating excited with a plane wave would have an incident flux vector of `[1, 0, 0, ...]` and a transmitted flux vector of `[0, 1, 0, ...]` if it diffracts light in the 0th order to the 1st order with 100% efficiency.

I haven't followed the remainder of the calculation all the way through, but I expect this distinction is a relevant one.

doc/docs/Python_Tutorials/Custom_Source.md
```diff
@@ -601,3 +601,50 @@ There are four items to note in this script. (1) Since the exact equivalence bet
...
+
+Another feature of the calculation to note is the radial Poynting flux which is computed as $P_r = \hat{r} \cdot \vec{P}$ where $\hat{r} = (\sin(\theta)\cos(\phi), \sin(\theta)\sin(\phi), \cos(\theta))$ and $\vec{P} = (P_x, P_y, P_z)$. By Poynting's theorem, the total flux in the 1D simulation is $P_z$ (i.e. $P_x$ and $P_y$ are ignored) which means $P_r = \cos(\theta) P_z$. The 1D simulation therefore involves a single flux detector for $P_z$ in the emission region.
```
stevengj, line 621:
```suggestion
Getting the correct scaling is a bit tricky.  The total flux per unit area is $P_z$ by Poynting's theorem.  Recall that this is being integrated over $(k_x,k_y)$ in the expansion of the delta function, so $P_z$ is the total flux per $d^2 k = k_{\rho} dk d\phi $ in cylindrical coordinates, but what we *want* is the flux per solid angle, and so there is a $k_{\rho} = k \cos \theta$ Jacobian factor that we need to divide by in order to obtatin the flux per solid angle.  Thus, our final answer is given by $P_z / \cos \theta$.

Another feature of the calculation to note is the radial Poynting flux which is computed as $P_r = \hat{r} \cdot \vec{P}$ where $\hat{r} = (\sin(\theta)\cos(\phi), \sin(\theta)\sin(\phi), \cos(\theta))$ and $\vec{P} = (P_x, P_y, P_z)$. By Poynting's theorem, the total flux in the 1D simulation is $P_z$ (i.e. $P_x$ and $P_y$ are ignored) which means $P_r = \cos(\theta) P_z$. The 1D simulation therefore involves a single flux detector for $P_z$ in the emission region.
```

stevengj:
As a postscript, I had a little bit of time in the airport, and I worked this out in a 2d setting.  I wanted to derive the fact that only a single $k$ contributes, and there is a $\cos \theta$ factor, directly from the near-to-far transform.  That is, at any finite distance to get the field you need to integrate over all $k$, but infinitely far away the contributions only come from a single $k$.

It turns out that it seems to work out straightforwardly: asymptotically for larger $r$, you can do a steepest-descent/saddle-point/stationary-phase approximation to the integral, and you get contributions only from one $k$, and a $\cos \theta$ factor falls out.   Hopefully I'll have time to write up a technical note sometime.

oskooi:
We should therefore rewrite the following paragraph in [Tutorial/Dipole Emission of a Light Emitting Diode as a Multilayer Stack](https://meep.readthedocs.io/en/latest/Python_Tutorials/Custom_Source/#dipole-emission-of-a-light-emitting-diode-as-a-multilayer-stack):

Screenshot 2025-10-24 204705

PR #2951, opened 2024-12-30
Clarify that the radiation pattern of a nonaxisymmetric dipole integrated in the azimuthal direction is rotationally invariant

oskooi:
Adds a note to [Tutorial/Cylindrical Coordinates/Nonaxisymmetric Dipole Sources](https://meep.readthedocs.io/en/latest/Python_Tutorials/Cylindrical_Coordinates/#nonaxisymmetric-dipole-sources) to menion that integrating $\phi$ over the range [0, 2π] for a nonaxisymmetric dipole is essentially equivalent to placing a series of linearly polarized dipoles along the circumference of a circle with fixed radius.


doc/docs/Python_Tutorials/Cylindrical_Coordinates.md
```diff
@@ -867,7 +867,7 @@ A point-dipole source at $r_0 > 0$ can be represented as a Dirac delta function
...
+Physically, the *total* field $E(x,y,z)$ is a sum of $E_m(r,z)e^{im\phi}$ terms, one for the solution at each $m$ (similarly for $H$). Computing the total Poynting flux, however, involves integrating $\Re [E \times H^*]$ over a surface that includes an integral over $\phi$ in the range $[0,2\pi]$. (This is equivalent to arranging a *series* of dipoles with linear polarization and infinitesimal spacing along the circumference of a circle with fixed radius $r_0$. The radiation pattern from such a dipole arrangement is in consequence rotationally invariant.) The key point is that the cross terms $E_mH^*_ne^{i(m-n)\phi}$ integrate to zero due to Fourier orthogonality. **The total Poynting flux is therefore a sum of the Poynting fluxes calculated separately for each $m$.**
```
stevengj, line 870: This is not quite right.  We are averaging/summing the *powers*, which corresponds to *incoherent* dipoles.  If you put a ring of dipoles into a single simulation, then you have *coherent* dipoles.

stevengj:
```suggestion
Physically, the *total* field $E(x,y,z)$ is a sum of $E_m(r,z)e^{im\phi}$ terms, one for the solution at each $m$ (similarly for $H$). Computing the total Poynting flux, however, involves integrating $\Re [E \times H^*]$ over a surface that includes an integral over $\phi$ in the range $[0,2\pi]$. (Summing the powers of a single dipole over all $\phi$ is equivalent to summing the powers of an incoherent ensemble of dipoles distributed uniformly in $\phi$: for an incoherent ensemble, you add powers since interference terms cancel, and each dipole location produces the same field up to a rotation.) The key point is that the cross terms $E_mH^*_ne^{i(m-n)\phi}$ integrate to zero due to Fourier orthogonality. **The total Poynting flux is therefore a sum of the Poynting fluxes calculated separately for each $m$.**
```

villadsegede:
Just out of curiosity: What is your experience with solver efficiency when you introduce complex light sources? In the use cases I've tested, the advantage of rotational design symmetry was lost in practice, and I could imagine it is the same here? (i.e. one might as well have run embarassingly parallel 3D simulations using ODA, for example)

...  And then there can still be other reasons for or against (i.e. once you have the radial basis solutions, you might as well recombine them to represent several sources if so desired)

stevengj:
I'm not sure what you mean by "complex light sources".   Certainly for off-axis planewaves we find it's still more efficient to use cylindrical coordinates, unless the angle is close to 90°.

PR #2970, opened 2025-01-22
Update smoothed projection function

smartalecH:
Here we update the smoothed projection function (`meep.adjoint.smoothed_projection()`) to be smooth as the interface leaves the pixel. We update the variable nomenclature to match our paper too.

I reran all the waveguide crossing and the diffraction grating experiments. All the results (e.g. optimization evolutions, final design topology) look _exactly_ the same (and I confirmed the new projection function is indeed being used here). So the CCSA algorithm is indeed robust to that kink we accidentally introduced in the function.

@hammy4815 can you try this new projection function on your experiment with gridap?


stevengj:
Some MPI tests failing?

smartalecH:
Looks like those MPI failures are also on `master`?

PR #1333, opened 2020-08-28
Singleton DFT fix

smartalecH:
Fixes #1269 

There were actually two bugs here. First, a collapsed array size counter (`reduced_dims`) was uninitialized. If the array to be collapsed only consisted of a single element, it wouldn't update to the correct value. Consequently, the routine would try to create an array the size of whatever garbage was in memory (which would occasionally cause a segfault).

Next, the actual singleton DFT calculation was never copied to the garbage array (assuming no segfault occurred).

Note there are still probably some other bugs in the `collapse_array()` routine (`array_slice.cpp`). If we see more odd behavior with DFT fields (e.g. when symmetries are applied, weird field components, etc.) we should first look here. Then we should look in `_get_dft_array()` inside of `meep.i`.

PR #3010, opened 2025-04-27
Fix bug in `collapse_array` for point objects of DFT fields in 1D cell and add unit test

### LINKED ISSUE #3009: Bug in `Simulation.get_dft_array` for `Simulation.add_dft_fields` for a single point in a 1D simulation
There seems to be a bug in the routine `Simulation.get_dft_array` when used with `Simulation.add_dft_fields` for a **single** point in a 1D simulation. `Simulation.get_dft_array` works fine when used with a finite region.

As a demonstration, the fields from a planewave at two locations in an empty cell are obtained using two different methods:

1. As a slice at the two locations of an array of the fields for the *entire* cell volume.
2. The fields only at the two locations.

The script is [here](https://gist.github.com/oskooi/dced79635c78a5c0a3e8199c418f6117). The results from (1) and (2) should be identical. However, they are not which is an indication of a bug. 

As a check, the results from (1) does show the expected planewave (shown below). This means the bug is related to (2).

Image

oskooi:
Fixes #3009.

This PR fixes a bug in the `collapse_array` routine used by `fields::get_dft_array` shown below whereby specifying a *point* object in a 1D cell (via either a 1D or 3D simulation) resulted in the fields of only a single restricted point to be returned rather than the sum of all the restricted points (including their weights).

A unit test is also included which verifies two properties of the DFT fields of a planewave in vacuum of a 1D cell obtained using a 1D and 3D simulation: (1) the relative phase of two different points in the 1D cell agree with the expected analytic result and (2) the DFT fields at the two points are the same whether they are obtained from either a single point or an array slice of the DFT fields.

https://github.com/NanoComp/meep/blob/bc5ac50542d7227874382020db09099f7375b724/src/dft.cpp#L1268-L1276

### DIFF
--- a/python/Makefile.am
+++ b/python/Makefile.am
@@ -85,6 +85,7 @@ TESTS =                                        \
     $(TEST_DIR)/test_oblique_source.py         \
     $(TEST_DIR)/test_pml_cyl.py                \
     $(TEST_DIR)/test_physical.py               \
+    $(TEST_DIR)/test_planewave_1D.py           \
     $(TEST_DIR)/test_prism.py                  \
     $(TEST_DIR)/test_pw_source.py              \
     $(TEST_DIR)/test_refl_angular.py           \
--- a/python/tests/test_planewave_1D.py
+++ b/python/tests/test_planewave_1D.py
@@ -1,0 +1,139 @@
+# Verifies properties of the fields of a planewave in 1D simulated in 1D or 3D.
+
+
+from typing import Tuple
+import unittest
+
+import meep as mp
+import numpy as np
+
+
+def complex_str(complex_val: complex) -> str:
+    return f"{complex_val.real:+.6f}{complex_val.imag:+.6f}*1j"
+
+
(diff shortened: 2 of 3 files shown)


stevengj:
Looks like this was introduced in #1333 by @smartalecH, any comments?

Seems like this should affect *all* cases where you have a DFT monitor that is a single point, not just 1d.

stevengj:
It would be good to double-check a 2d case with a single-point DFT monitor: presumably that was also wrong before this PR, and is now fixed.

oskooi:
> It would be good to double-check a 2d case with a single-point DFT monitor: presumably that was also wrong before this PR, and is now fixed.

I have verified that this bug was also present in 2D. The test is a 2D version of the 1D test in this PR. This bug fix produces the correct results in 2D as well. 

The test script and results are available in this [gist](https://gist.github.com/oskooi/7ca1402a32175dbfb91403a8822537c6).

PR #3015, opened 2025-05-05
Prepare for 1.30.1 release

oskooi:
The next tarball release (a minor version update mainly because of an important bug fix) is tentatively scheduled for Thursday May 8, 2025.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 ## Meep 1.30.1
 
-5/8/2025
+5/15/2025
 
 * Bug fix in DFT fields at a single point (zero-dimensional volume), where they were only first-order accurate and had the wrong weight ([#3010]).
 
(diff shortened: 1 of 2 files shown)


NEWS.md
```diff
@@ -1,5 +1,11 @@
 # Meep Release Notes
 
+## Meep 1.30.1
+
+5/8/2025
+
+* Bug fix in DFT fields of a single point ([#3010]).
```
stevengj, line 7:
```suggestion
* Bug fix in DFT fields at a single point (zero-dimensional volume), where they were only first-order accurate and had the wrong weight ([#3010]).
```

NEWS.md
```diff
@@ -1,5 +1,11 @@
 # Meep Release Notes
 
+## Meep 1.30.1
+
+5/8/2025
```
stevengj, line 5:
```suggestion
5/15/2025
```

PR #3028, opened 2025-06-10
Fix compatibility with python 3.12

hmaarrfk:
xref: https://github.com/conda-forge/pymeep-feedstock/pull/98

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -576,7 +576,7 @@ else
     if test $have_python = yes; then
 
       AC_MSG_CHECKING([for Python include directory])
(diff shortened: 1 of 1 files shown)


oskooi (approved):
LGTM.

oskooi:
Some additional context for this change: `distutils` was discontinued in Python 3.12 ([PEP 632](https://peps.python.org/pep-0632)). The [migration advice](https://peps.python.org/pep-0632/#migration-advice) suggests replacing `distutils.sysconfig` with `sysconfig` which is implemented in this PR.

PR #3038, opened 2025-06-27
Add VCSEL waveguide simulation at 850 nm

praveenndj:
This PR adds a new example input file for simulating a vertical-cavity surface-emitting laser


oskooi:
Note that there is already a [waveguide tutorial example](https://meep.readthedocs.io/en/master/Python_Tutorials/Basics/#a-straight-waveguide).

stevengj:
I would also expect a VCSEL simulation to have a cavity, and compute the resonant mode or something, not just a waveguide.

Issue #3045, opened 2025-07-03
faster φ quadrature for greencyl near2far


stevengj:
@joshualchen reported that near2far transformations in cylindrical coordinates are extremely slow for large $m$.  Presumably this is just because of the highly oscillatory $e^{i m\phi}$ term in the integrand.

An easy way to fix this would be to pull out the $e^{im \phi}$ term into the quadrature scheme.  For example, this is implemented by QUADPACK and also [by GSL](https://www.gnu.org/software/gsl/doc/html/integration.html#qawo-adaptive-integration-for-oscillatory-functions).  Since this term is the same for all spatial points and all frequencies, we only need to compute the quadrature scheme once.

stevengj:
I don't think it's practical to use the QUADPACK or GSL implementations directly, unfortunately — we'll need our own implementation.  The reason is that we need to integrate 6 complex fields, for both sin and cosine, but QUADPACK and GSL only implement real-valued function.   That means, if we use the GSL functions directly, we'd need to call it 12 times, multiplying the number of Green's-function evaluations by 12.

Losing a factor of 12 is likely to negate any benefit of calling these routines.  So, the only path forward is to implement our own.

I think QAWO is just Clenshaw–Curtis integration with a `cos` or `sin` weight function precomputed, which isn't too hart to implement(??).   Probably better to implement it in a high-level language like Julia first, to do some simple experiments and see how much one can gain.

stevengj:
For example, here is a little implementation in Julia that computes Clenshaw–Curtis quadrature points and weights for $\int_{-1}^{+1} f(x) \text{weight}(x) dx$ for an arbitrary given weight function:
```jl
import FFTW
function clenshaw_curtis(N)
    iseven(N) && N > 0 || throw(ArgumentError("this implementation requires even N > 0"))
    x = cos.(range(0, pi, length=N+1)) # Chebyshev points
    k = 0:N÷2
    d = @. 2 / (1-4k^2) # cosine integrals
    w = FFTW.r2r(d, FFTW.REDFT00) / N # type-I DCT
    w[1] /= 2
    return (; x = x, w = [w; reverse(w[1:end-1])])
end
function clenshaw_curtis(weight::Function, N, Ninner=1024)
    N > 0 || throw(ArgumentError("requires N > 0"))
    x = cos.(range(0, pi, length=N+1)) # Chebyshev points
    clencurt = clenshaw_curtis(Ninner)
    function fourierweight(k)
[...]

PR #3047, opened 2025-07-03
greencyl tol should be independent of m

stevengj:
This PR makes the convergence check in the `greencyl` function independent of $m$, whereas it previously required a more stringent tolerance with increasing $m$ (becoming slower and slower).

The `greencyl` function computes an angular integral $\int d\phi F(\phi)$ of some fields $F(\phi)$ (from the Green's function at a given point in space, integrated over angles).  This is computed by adaptively refining a Riemann sum (exponentially convergent for periodic functions), repeatedly doubling the number of $\phi$ points until some tolerance is achieved.

Currently, the convergence test was $\Vert \Delta \int d\phi F(\phi) \Vert_1 \le \text{tol} \times \Vert \int d\phi F(\phi) \Vert_1$, where $\int$ denotes the estimated integral and $\Delta$ denotes the change from the previous estimate.  `tol` defaults to `1e-3`.  The problem is that $\Vert \int d\phi F(\phi) \Vert_1$ gets smaller and smaller with increasing $m$ (since $F \sim e^{im\phi}$ gets more oscillatory), so effectively we are requiring a more and more stringent tolerance with increasing $m$, and will eventually hit the limits of floating-point precision, so the runtime can explode.

This PR changes the convergence test to $\Vert \Delta \int d\phi F(\phi) \Vert_1 \le \text{tol} \times  \int d\phi \Vert F(\phi) \Vert_1$, which moves the $\Vert \cdots \Vert_1$ to the *integrand* rather than the *integral* on the right-hand side.

This cancels the $e^{im\phi}$ factor: the right hand side is now an upper bound on the integrals for *all* $m$.  This should be appropriate because cylindrical simulations are usually a sum (either explicit or conceptual) over all $m$, and you only need a small error relative to the largest term in the sum.

This PR could be in *addition* to using a more specialized quadrature scheme as described in #3045.

### DIFF
--- a/src/near2far.cpp
+++ b/src/near2far.cpp
@@ -290,12 +290,14 @@ void greencyl(std::complex<double> *EH, const vec &x, double freq, double eps, d
      repeatedly double N until convergence to tol is achieved, re-using previous points. */
   const int N0 = 16 + int(4 * abs(m));
   double dphi = 2.0 / N0; // factor of 2*pi*r is already included in add_dft weight
+  double sumabs = 0;      // integral of L1-norm of integrand
   for (int N = N0; N <= 65536; N *= 2) {
     std::complex<double> EH_sum[6];
     dphi *= 0.5; // delta phi is halved because N doubles
     double dphi2pi = dphi * 2 * pi;
     for (int j = 0; j < 6; ++j)
       EH_sum[j] = EH[j] * 0.5; // re-use previous quadrature points (with halved dphi)
+    sumabs *= 0.5;             // re-use previous quadrature (with halved dphi)
     /* N-point quadrature points i = 0..N-1.  After the first iteration (N==N0), we
        only need to sum over odd i, since the even i were summed for the previous N. */
     for (int i = (N > N0); i < N; i += 1 + (N > N0)) {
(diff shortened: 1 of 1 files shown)


oskooi:
This seems to be working. The [test](https://gist.github.com/oskooi/2796df8d7b89b620462d17028fd841b8) involves computing the emission from a dipole in a disc in cylindrical coordinates (based on a [tutorial example](https://meep.readthedocs.io/en/latest/Python_Tutorials/Near_to_Far_Field_Spectra/#extraction-efficiency-of-a-collection-of-dipoles-in-a-disc)). This calculation involves a Fourier-series expansion of the fields involving the field angular momentum $m$ from $e^{im\phi}$. For each $m$-run in the expansion, I measured the time to run `get_farfield` for a single farfield point. 

A plot of the elapsed time for calling `get_farfield` vs. each $m$ in the expansion is shown below. The time for calling `get_farfields` using the current master branch (5033217) is generally increasing with $m$ whereas it is a constant using this PR.

elapsed_time_get_farfield_vs_m

disc_benchmark_plot2D

oskooi (approved):
LGTM.

stevengj:
And the emission pattern is still accurate, I assume?

(we should probably make the tolerance here a user-adjustable parameter)

oskooi:
> And the emission pattern is still accurate, I assume?

I verified that the far field $E_r$ computed using `get_farfield` in the [test](https://gist.github.com/oskooi/2796df8d7b89b620462d17028fd841b8) using master and this PR are *identical* (to machine precision).

Luochenghuang:
This PR is a fantastic improvement. Thank you for implementing it!
Are there plans for a tarball release in the near future incorporating this PR?

stevengj:
Hopefully it should appear in a release sometime in the next month.

Issue #3063, opened 2025-07-24
Improving the convergence of the Fourier-series expansion of ring currents for simulating nonaxisymmetric dipole sources in cylindrical coordinates


oskooi:
The simulation of a [nonaxisymmetric dipole source](https://meep.readthedocs.io/en/latest/Python_Tutorials/Cylindrical_Coordinates/#nonaxisymmetric-dipole-sources) in cylindrical coordinates involves a Fourier-series expansion of "ring" current sources with azimuthal dependence $e^{im\phi}$ for $m = 0, 1, 2, ..., M$ for some cutoff $M$ which is proportional to the dipole position $r$. When the dipole is positioned far from $r = 0$, $M$ can become large (i.e., 50+) which can significantly increase the total runtime.

It would be useful to somehow speed up the convergence of the Fourier-series expansion.

stevengj:
What you're looking for is some kind of [series extrapolation method](https://en.wikipedia.org/wiki/Category:Series_acceleration_methods).  Not sure which one works best for this particular problem, but you could try a few and experiment.

Basically what you need is a test problem where you've calculated it for lot's of $m$'s, and then compare it to plugging in the first few into a series-extrapolation code, and see which method converges the fastest.

oskooi:
To experiment with the series extrapolation, here is the Fourier-series data for the source and radiated power of a ring current vs. $m$ for a nonaxisymmetric dipole (at $r$ = 6.3 μm) in a dielectric slab ($n$ = 2.4, thickness [in vacuum] of 0.7 um) above a perfect-reflector ground plane. This setup is based on the [tutorial example](https://meep.readthedocs.io/en/latest/Python_Tutorials/Cylindrical_Coordinates/#nonaxisymmetric-dipole-sources). The script used to generate these results is [here](https://gist.github.com/oskooi/e15f375180728eaac9fad0723919c978). 

In this example, the cutoff $M$ is determined by when the *radiated* power (in air) drops to below **1%** its maximum value over the interval of all preceding $m$ values. The value for $M$ found numerically is consistent with the analytic value of roughly $\omega r$ (~42).

The series data is provided at resolutions 50 and 100:

[resolution_50.csv](https://github.com/user-attachments/files/21460953/resolution_50.csv)

[resolution_100.csv](https://github.com/user-attachments/files/21461651/resolution_100.csv)

[...]

PR #3069, opened 2025-07-31
add dft_flux::complexflux

stevengj:
Adds a new `::complexflux()` method, analogous to flux, for `dft_flux` objects that returns the *complex* Poynting flux (not taking the real part).  The imaginary part can be interpreted as the "reactive" flux and some people find it useful.

It may also be useful for #3063.

### DIFF
--- a/python/meep.i
+++ b/python/meep.i
@@ -943,6 +943,14 @@ void _get_gradient(PyObject *grad, double scalegrad,
     delete[] $1;
 }
 
+%typemap(out) std::vector<std::complex<double>> complexflux {
+    size_t size = $1.size();
+    $result = PyList_New(size);
+    for(size_t i = 0; i < size; i++) {
+        PyList_SetItem($result, i, PyComplex_FromDoubles(real($1[i]), imag($1[i])));
+    }
+}
+
 // Typemap suite for dft_force
 
 %typemap(out) double* force {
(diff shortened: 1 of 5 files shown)


oskooi:
Unfortunately, this feature does *not* seem to be working.

Test case adapted from this [script](https://gist.github.com/oskooi/e15f375180728eaac9fad0723919c978):
```py
flux_mon = sim.add_flux(...)

sim.run(...)

radiated_flux = flux_mon.complexflux()
```
produces an error:
```
AttributeError: 'DftFlux' object has no attribute 'complexflux'
```

I have verified that this branch builds on my local machine.

oskooi:
This is working with the fix provided in 46e3010. 

For consistency with the rest of the API, it may be useful to add the wrapper `meep.get_complexfluxes` to `simulation.py`. This is analogous to the existing `meep.get_fluxes`.

```py
def get_complexfluxes(f):
    """
    Given a flux object, returns a list of the current flux spectrum that it has
    accumulated.
    """
    return f.complexflux()
```

stevengj:
I would rather deprecate `get_flux`, which seems not very pythonic. 

Can you add a check to one of the tests that the real part of complexflux matches flux?

oskooi (approved):
A unit test has been added.

LGTM.

PR #3064, opened 2025-07-24
make greencyl_tol a parameter

stevengj:
The tolerance for computing the cylindrical Green's function (`greencyl`) is now a parameter `greencyl_tol` (currently undocumented), rather than being fixed at `1e-3`.   This is especially important with #3047, in case you want to calculate the (small) high-$`m`$ contributions with greater relative accuracy.

PR #3081, opened 2025-08-17
Add missing `greencyl_tol` parameter to functions `save_farfields` and `flux` of class `dft_near2far`

oskooi:
Adds the `greencyl_tol` parameter to the functions `save_farfields` and `flux` of the class `dft_near2far` which were missing from #3064.


oskooi:
Actually, the `flux` function of the `dft_near2far` class does *not* support cylindrical coordinates:

https://github.com/NanoComp/meep/blob/7e7b985e68a7a8111a058923daba42509e20ec59/src/near2far.cpp#L517-L519

PR #3087, opened 2025-09-02
Update SSP with erosions/dilations

smartalecH:
Here we update the SSP algorithm to account for the following features/fixes:
- [ ] (TODO) Generalize the routine to work in 1D and 3D (not just 2D)
- [ ] (in-progress) Add `resolution_simulation` parameter to account for differences in meep and design grid resolutions
- [ ] (in-progress) Augment the algorithm to allow for precise erosions and dilations.

(cc @oskooi, @stevengj )

### DIFF
--- a/python/adjoint/filters.py
+++ b/python/adjoint/filters.py
@@ -5,7 +5,7 @@ transforms.
 """
 
 import sys
-from typing import List, Tuple, Union
+from typing import List, Optional, Tuple, Union
 
 import numpy as np
 from autograd import numpy as npa
@@ -712,6 +712,9 @@ def smoothed_projection(
     beta: float,
     eta: float,
     resolution: float,
+    resolution_simulation: Optional[float] = None,
+    erosion_dilation: float = 0.0,
+
 ):
     """Project using subpixel smoothing, which allows for β→∞.
 
@@ -747,6 +750,10 @@ def smoothed_projection(
         eta: The threshold point in the range [0, 1].
         resolution: resolution of the design grid (not the Meep grid
             resolution).
+        resolution_simulation: resolution of the meep grid.
+        erosion_dilation: erosion (negative) or dilation (positive) transform to
+            apply to the final projected design in "meep units"; The resolution
+            scaling is applied automatically.
     Returns:
         The projected and smoothed output.
 
@@ -763,6 +770,9 @@ def smoothed_projection(
         >>> rho_filtered = conic_filter(rho, filter_radius, Lx, Ly, resolution)
         >>> rho_projected = smoothed_projection(rho_filtered, beta, eta_i, resolution)
     """
+    if resolution_simulation is None:
+        resolution_simulation = resolution
+    
     # Note that currently, the underlying assumption is that the smoothing
     # kernel is a circle, which means dx = dy.
     dx = dy = 1 / resolution
@@ -795,10 +805,13 @@ def smoothed_projection(
     # The distance for the center of the pixel to the nearest interface
     d = (eta - rho_filtered) / rho_filtered_grad_norm_eff
 
+    # Account for an erosion or dilation
+    d = npa.maximum(0, d)
+
     # Only need smoothing if an interface lies within the voxel. Since d is
     # actually an "effective" d by this point, we need to ignore values that may
     # have been sanitized earlier on.
-    needs_smoothing = nonzero_norm & (npa.abs(d) < R_smoothing)
+    needs_smoothing = nonzero_norm & (npa.abs(d + erosion_dilation) < R_smoothing)
 
     # The fill factor is used to perform simple, first-order subpixel smoothing.
     # We use the (2D) analytic expression that comes when assuming the smoothing


smartalecH:
@stevengj  I have the following test for our new erosion and dilation feature:

```
Lx = 2; Ly = 2
resolution = 50
eta_i = 0.5; eta_e = 0.75
lengthscale = 0.1
filter_radius = mpa.get_conic_radius_from_eta_e(lengthscale, eta_e)
Nx = onp.round(Lx * resolution) + 1
Ny = onp.round(Ly * resolution) + 1
rho = onp.random.rand(Nx, Ny)
beta = npa.inf
rho_filtered = mpa.conic_filter(rho, filter_radius, Lx, Ly, resolution)
rho_projected = mpa.smoothed_projection(rho_filtered, beta, eta_i, resolution)
erosion_dilation = 0.5
rho_projected_eroded = mpa.smoothed_projection(rho_filtered, beta, eta_i, resolution, erosion_dilation=-erosion_dilation)
rho_projected_dilated = mpa.smoothed_projection(rho_filtered, beta, eta_i, resolution, erosion_dilation=+erosion_dilation)

plt.figure(figsize=(12,4))
plt.subplot(1,3,1)
plt.imshow(rho_projected_eroded)
plt.subplot(1,3,2)
plt.imshow(rho_projected)
plt.subplot(1,3,3)
plt.imshow(rho_projected_dilated)
```
image

It's obviously not quite there yet.

I'm not sure how to tackle the following challenge: we only use our level-set approximation to identify the boundaries themselves. Anything not on a boundary, we fall back to our brute-force tanh_projection output (hence the `where` statement). 

It seems that in order to do erosions/dilations, we really have to use the level-set across the whole domain, right?

python/adjoint/filters.py
```diff
@@ -795,10 +805,13 @@ def smoothed_projection(
     # The distance for the center of the pixel to the nearest interface
     d = (eta - rho_filtered) / rho_filtered_grad_norm_eff
 
+    # Account for an erosion or dilation
+    d = npa.maximum(0, d)
```
stevengj: I don't understand this line.

stevengj: It seems like you should set `d = d + erosion_dilation`

stevengj:
Your code doesn't seem right?

Let's consider the $\beta = \infty$ case first, where the tanh projection is just the step function and SSP boils down to $F(d)$.  We should just be doing $`F(d + \text{erosion\_dilation})`$, no?

python/adjoint/filters.py
```diff
@@ -795,6 +805,9 @@ def smoothed_projection(
     # The distance for the center of the pixel to the nearest interface
     d = (eta - rho_filtered) / rho_filtered_grad_norm_eff
 
+    # Account for an erosion or dilation
+    d = d + erosion_dilation
```
smartalecH: @stevengj something as simple as this, correct?

unfortunately, we still have the same problem I mention above. At this point, the routine is just falling back to the naive projection everywhere:

image

Note that $\beta=\infty$ for this test.

smartalecH: I think the problem is with `needs_smoothing`.

If I just look at $F$ (and ignore the where clause) I get the expected perturbation. (Note that I have to clip the min and max because things get ugly when the spatial gradient breaks down in uniform regions.)

image

stevengj: What happens if you use `needs_smoothing`, but just do (for $\beta = \infty$):
```jl
return npa.where(
        needs_smoothing,
        F,
        rho_projected,
    )
```
(I just want to isolate the logic for $F(d)$ here — the `npa.where` statement should still be needed since you need to define $F$ piecewise, and the `F` variable is only the polynomial part.)

stevengj:
I think this equation in the SSP paper needs to be modified too:
image
because it is relative to an interface at the wrong (old) position.

However, if you just use $F(d)$ directly for $\beta = \infty$, as I suggested above, you don't need to worry about this for now.  That's why I think it will be easier to debug that case first.

My guess is that the corrected general formula should be something like:

$`\hat{\rho}_{\pm} = P_{\beta,\eta}\left(\tilde{\rho} \pm (\hat{R} F(\mp d) \mp \Delta d) \, \Vert \nabla \tilde{\rho} \Vert  \right)`$

where $\Delta d$ is what you are calling `erosion_dilation`, i.e. you have replaced $d$ everywhere with $d_{\text{new}} = d_{\text{old}} + \Delta d$.   … I have to check the continuity conditions on $\hat{\tilde{\rho}}$, though.

python/adjoint/filters.py
smartalecH: image

Not quite right. Looks like it's now moving the subpixel-smoothed _boundary_ but it's just "superimposing" it on top of the naive projection function (`rho_projected`). Which is indeed what that `where` statement is telling it to do.

stevengj:
Let me make an alternative suggestion.

1. Move the computation of `rho_filtered_grad_norm_eff` **before** the computation of `rho_projected`.
2. Add the following line **before** the computation of `rho_projected`: `rho_filtered = rho_filtered - erosion_dilation * npa.where(nonzero_norm, rho_filtered_grad_norm, 0)`
3. **Remove all of your other changes** to the code (e.g. now `d` will automatically include `erosion_dilation`)

smartalecH:
@stevengj ok so that indeed produces an erosion or dilation:

image

We should do some experiments to show that the resulting transformation really reflects what the user prescribed, and doesn't depend e.g. on the "steepness" of the filtered field like $\eta$ does. The above transformation prescribed a 1% transformation (i.e. 1 pixel)

stevengj:
As a test, you could compare it to non-differentiable morphological open/close operations (maybe even setting `R_smoothing = 0` to get rid of the subpixel smoothing).

To avoid being too sensitive to discretization effects during testing, you'll probably want to make `erosion_dilation` ($\Delta d$) larger than a pixel.  Note that we still want it to be smaller than the filtering radius, i.e. you want $\Delta x \ll \Delta d \ll \tilde{R}$.

Ideally the difference with morphological operations should go to zero as the resolution increases ($\Delta x \to 0$) for fixed $\Delta d$ and $\tilde{R}$.

stevengj:
Note that the same method (simply replacing $\tilde\rho$ with $\tilde{\rho} - \Delta d \Vert \nabla \tilde{\rho}\Vert$ should allow us to more easily implement sidewall angles than in the past, just by making Δd a function of z.  Unlike previous methods where we used a $\Delta \eta$, it should require less calibration to get the desired angle, since the angle no longer depends on $\nabla \tilde \rho$ — you can directly specify whatever height profile you want.

stevengj:
It would be nice to merge an update at some point so that it doesn't assume that Meep and the design grid have the same resolution.

PR #3088, opened 2025-09-09
Fix `Box`-related issues in `plot3D`

### LINKED ISSUE #2913: sim.plot3D fails to plot the 3D layout
I adjusted the bent-waveguide.ipynb example to have a Z component: [bent-waveguide-3D-fails-to-draw-in-3D.ipynb](https://freebsd.org/~yuri/bent-waveguide-3D-fails-to-draw-in-3D.ipynb).

sim.plot3D silently outputs an empty plot:
image

All dependencies are installed.

Version: 1.29.0
Python 3.11
FreeBSD 14.1

### LINKED ISSUE #3000: plot3D incorrect source orientation
plot3D seems to draw the incorrect source orientation. See below a MWE: the source plane should be perpendicular to the geometry.

```
import meep as mp

cell_size = mp.Vector3(1, 1, 10)

sources = [mp.Source(
    mp.GaussianSource(1, fwidth=0.1),
    component=mp.Ex,
    center=mp.Vector3(),
    size=mp.Vector3(1, 1),
)]

sim = mp.Simulation(
    cell_size=cell_size,
    resolution=20,
    geometry=[mp.Cylinder(0.2, height=5, material=mp.Medium(index=2))],
    sources=sources,
)

sim.plot3D()
```

Image

This is similar to the issue in https://github.com/NanoComp/meep/discussions/2429, but I don't think an actual issue has been created.

There also seems to be a plot3D issue with PMLs - even if you define them along only one direction, it draws them in every direction - but that could be a separate issue.

Thanks.

thchr:
This fixes #3000 (and #2429) following the suggestion in https://github.com/NanoComp/meep/issues/3000#issuecomment-2803982467, i.e. wrong orientation of sources and monitors in 3D.

While fixing this, I also noticed that the boundary layer visualization was quite broken, so this fixes it also (by factoring out `get_boundary_volumes` and reusing that). Previously, the implementation assumed that all boundary layers were identical, and also had the orientation wrong on the x- and y-boundaries. This also fixes #2913.

I also took the opportunity to try to make the visualization a bit prettier by (1) using some nicer colors that weren't 100% R, G, or B values, (2) dropping the wireframe on the boundary regions, and (3) adding a faint outline for the simulation cell.

---

Example of change below (photonic crystal with weirdly arranged PMLs, boundary details in code below examples):

| Before | After |
|---|---|
| image | image |

```py
pml_layerZ1 = mp.PML(thickness = .5,   direction=mp.Z, side=mp.Low)
pml_layerZ2 = mp.PML(thickness = 5,    direction=mp.Z, side=mp.High)
pml_layerX  = mp.PML(thickness = .25,  direction=mp.X, side=mp.Low)
pml_layerY  = mp.PML(thickness = .5/3, direction=mp.Y, side=mp.High)
boundary_layers = [pml_layerZ1, pml_layerZ2, pml_layerX, pml_layerY]
```

Here's another before/after:
| Before | After |
|---|---|
| image | image |

```py
pml_layerZ1 = mp.PML(thickness = 3,     direction=mp.Z, side=mp.Low)
pml_layerZ2 = mp.PML(thickness = 5,     direction=mp.Z, side=mp.High)
pml_layerX  = mp.PML(thickness = .125,  direction=mp.X, side=mp.Low)
pml_layerY  = mp.PML(thickness = .25/3, direction=mp.Y, side=mp.High)
boundary_layers = [pml_layerZ1, pml_layerZ2, pml_layerX, pml_layerY]
```

### DIFF
--- a/python/visualization.py
+++ b/python/visualization.py
@@ -662,6 +662,63 @@ def plot_eps(
         return ax
 
 
+def get_boundary_volumes(
+    sim: Simulation,
+    thickness: float,
+    direction: float,
+    side,
+) -> Volume:
+    xmin, xmax, ymin, ymax, zmin, zmax = box_vertices(
+        sim.geometry_center, sim.cell_size, sim.is_cylindrical
+    )
+
+    if direction == mp.X and side == mp.Low:
+        return Volume(
+            center=Vector3(
+                xmin + 0.5 * thickness, sim.geometry_center.y, sim.geometry_center.z
+            ),
+            size=Vector3(thickness, sim.cell_size.y, sim.cell_size.z),
+        )
+    elif (direction == mp.X and side == mp.High) or direction == mp.R:
+        return Volume(
+            center=Vector3(
+                xmax - 0.5 * thickness, sim.geometry_center.y, sim.geometry_center.z
+            ),
+            size=Vector3(thickness, sim.cell_size.y, sim.cell_size.z),
+        )
+    elif direction == mp.Y and side == mp.Low:
+        return Volume(
+            center=Vector3(
+                sim.geometry_center.x, ymin + 0.5 * thickness, sim.geometry_center.z
+            ),
+            size=Vector3(sim.cell_size.x, thickness, sim.cell_size.z),
+        )
+    elif direction == mp.Y and side == mp.High:
+        return Volume(
+            center=Vector3(
+                sim.geometry_center.x, ymax - 0.5 * thickness, sim.geometry_center.z
+            ),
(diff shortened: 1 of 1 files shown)


thchr:
Pre-commit failure seems unrelated afaict (after having been not-unrelated a number of times).

Relative to the before/afters above, the follow-up commit increased the transparency of the boundary layers and removed the coloring on the cell, leaving only its boundary.

oskooi (approved):
LGTM.

(The pre-commit failure involving the missing `pbr` module is indeed unrelated to this PR and can be ignored.)

Issue #2819, opened 2024-04-17
Does not build python module for python 3.12


pbsds:
[PEP-632](https://peps.python.org/pep-0632/) is upon us, and distutils is no longer a part of the default python distribution as of version 3.12, which causes the configure script to fail to locate `Python.h`.

[Our current workaround](https://github.com/NixOS/nixpkgs/pull/304922) consists of replacing `distutils` with `setuptools._distutils` in:

* configure.ac 
* doc/docs/setup.py 
* python/visualization.py

gummyequation:
This is fixed now with Meep 1.31.0.

stevengj:
Closed by #3028.

stevengj:
Actually, that PR only fixed configure.ac — looks like the other two files still need to be patched. See #3093 and #3094.

PR #3093, opened 2025-09-22
fix visualization.py for python 3.12

stevengj:
For #2819 — `distutils.version.LooseVersion` no longer exists, but I think we can delete this check (it's only used to issue a warning for people using versions of matplotlib prior to 2019).


stevengj:
Rebased.

PR #3110, opened 2025-11-06
Ensure the boundary condition on both sides of a direction are the same for `set_boundary`

### LINKED ISSUE #3106: Bug in `Simulation.set_boundary` for 1D simulation
There seems to be a bug in the function [`Simulation.set_boundary`](https://meep.readthedocs.io/en/latest/Python_User_Interface/#field-computations) for the case of a 1D simulation and a metallic boundary condition applied to the boundary at $-z$ (i.e. `side` of `meep.Low`). A demonstration is provided in this [script](https://gist.github.com/oskooi/f36b7cd8a4b8f861669dbbf6eb38d90e). The example involves computing the LDOS for an $E_y$ point source in air with PML in the $+z$ boundary and either a (1) Bloch-periodic or (2) metallic boundary condition in the $-z$ boundary. The LDOS is the *same* for (1) and (2). This result is not expected and an indication of a potential bug.

As an additional test, when the PML is placed at the $-z$ boundary and the Bloch-periodic or metallic boundary condition is applied to the $+z$ boundary, the LDOS is *different* which is expected.

This bug could explain why the LDOS calculation of a cavity in 1D (#3062) produces unexpected results.

oskooi:
Fixes #3106.

### DIFF
--- a/src/boundaries.cpp
+++ b/src/boundaries.cpp
@@ -79,8 +79,8 @@ comms_sequence optimize_comms_operations(const std::vector<comms_operation> &ope
 void fields::set_boundary(boundary_side b, direction d, boundary_condition cond) {
   if (boundaries[b][d] != cond) {
     boundaries[b][d] = cond;
-    // The boundary condition on the opposite side must be the same.
-    if (boundaries[1 - b][d] != cond) boundaries[1 - b][d] = cond;
+    // If it's periodic, the boundary condition on the opposite side must be the same.
+    if (cond == Periodic || boundaries[1 - b][d] == Periodic) boundaries[1 - b][d] = cond;
     // we don't need to call sync_chunk_connections() since set_boundary()
     // should always be called on every process
     chunk_connections_valid = false;


src/boundaries.cpp
```diff
@@ -79,6 +79,8 @@ comms_sequence optimize_comms_operations(const std::vector<comms_operation> &ope
 void fields::set_boundary(boundary_side b, direction d, boundary_condition cond) {
   if (boundaries[b][d] != cond) {
     boundaries[b][d] = cond;
+    // The boundary condition on the opposite side must be the same.
+    if (boundaries[1 - b][d] != cond) boundaries[1 - b][d] = cond;
```
stevengj: This should only be true if it's periodic direction `d`.

stevengj:
```suggestion
    // If it's periodic, the boundary condition on the opposite side must be the same.
    if (cond == Periodic || boundaries[1 - b][d] == Periodic) boundaries[1 - b][d] = cond;
```

PR #3123, opened 2025-12-01
Update docstrings to state usage of border padding

lxvm:
The filter routines stated that they used zero padding of the design variables in convolution filters where they actually used border padding. I updated docstrings to reflect this.


stevengj:
It would be good to expand the docs to precisely define "border padding".

lxvm:
Thanks, I updated the docstrings to explain the implementation and purpose of border padding

PR #3144, opened 2026-02-07
bugfix in mu for pympb

### LINKED ISSUE #3142: MPB: setting `force_mu` causes wrong behavior with `default_material`
Example below (awkwardly, in Julia, calling into Python; sorry):

```jl
using PythonCall

mp = pyimport("meep")
mpb = pyimport("meep.mpb")

ms = mpb.ModeSolver(
    num_bands        = 6,
    resolution       = 12,
    geometry_lattice = mp.Lattice(size=[1,1]),
    default_material = pyfunc(r->mp.Medium(epsilon=2)),
    )
k = mp.Vector3(0,0)

ms.force_mu = false
ms.init_params(p = mp.TM, reset_fields=true)
ms.solve_kpoint(k)
freqs_force_mu_false = ms.get_freqs()

ms.force_mu = true
ms.init_params(p = mp.TM, reset_fields=true)
ms.solve_kpoint(k)
freqs_force_mu_true = ms.get_freqs()

println("with `force_mu = false`: ", freqs_force_mu_false)
println("with `force_mu = true`:  ", freqs_force_mu_true)
```

I would expect this to print the same in both cases, since the `Medium` call above actually only sets an `epsilon` and not any `mu_*` properties, but instead, it prints:

```
with `force_mu = false`: [0.0, 0.707106781292359, 0.7071067813132079, 0.7071067813696875, 0.7071067817245882, 1.0000000002737972]
with `force_mu = true`:  [0.0, 0.5000000006474712, 0.5000000006860166, 0.5000000009506491, 0.5000000041583756, 0.7071067818203879]
```

I am not quite sure what is going wrong, but it seems that when `force_mu` is toggled, then actual mu values might be getting populated with epsilon values? As far as I could see, this issue only happens using `default_material` and not when using plain `geometry` objects.

This came up because I was wanting to set `mu_offdiag` (which shouldn't work unless `force_mu = true` when using `default_material`).

I'm filing this here in the meep repo rather than the mpb repo, because I think the issue probably is in the libpympb files. E.g., in [`dielectric_function`](https://github.com/NanoComp/meep/blob/master/libpympb/pympb.cpp#L98) or in [`mode_solver::material_epsmu`](https://github.com/NanoComp/meep/blob/master/libpympb/pympb.cpp#L543).
I recall having gotten this to work previously in the Scheme API as well.

stevengj:
Fixes #3142, I think.  Also a bit of code cleanup by getting rid of the global-like `eps` member.

@thchr, can you confirm?

### DIFF
--- a/libpympb/pympb.cpp
+++ b/libpympb/pympb.cpp
@@ -129,7 +129,15 @@ static int mean_epsilon_func(symmetric_matrix *meps, symmetric_matrix *meps_inv,
                              const mpb_real r[3], void *edata) {
 
   mode_solver *ms = static_cast<mode_solver *>(edata);
-  return ms->mean_epsilon(meps, meps_inv, n, d1, d2, d3, tol, r);
+  return ms->mean_epsmu(meps, meps_inv, n, d1, d2, d3, tol, r, true);
+}
+
+static int mean_mu_func(symmetric_matrix *meps, symmetric_matrix *meps_inv, mpb_real n[3],
+                        mpb_real d1, mpb_real d2, mpb_real d3, mpb_real tol, const mpb_real r[3],
+                        void *edata) {
+
+  mode_solver *ms = static_cast<mode_solver *>(edata);
+  return ms->mean_epsmu(meps, meps_inv, n, d1, d2, d3, tol, r, false);
 }
 
 /****** utils ******/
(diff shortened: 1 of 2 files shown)


thchr:
Hi Steven; thanks so much for having a look at this and finding a fix..!

I've compiled this branch locally and checked again: everything seems to work now, including anisotropic mu etc.

Many thanks 🙏 !

thchr:
(my `make check` reported a fail (but no details) in `aniso_disp`, but I imagine that's unrelated though)

PR #3150, opened 2026-02-16
Prepare for 1.32 release

oskooi:
The next tarball release is tentatively scheduled for Friday February 20, 2026.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -7,7 +7,7 @@ AC_CONFIG_SRCDIR(src/step.cpp)
 # not the same as the "public" version number.  (Don't worry about this
 # except for public releases.) Note that any change to a C++ class
 # definition (in the .hpp file) generally breaks binary compatibility.
-SHARED_VERSION_INFO="36:0:0" # CURRENT:REVISION:AGE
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -1,13 +1,13 @@
...
-AC_INIT([meep],[m4_esyscmd(./version.sh 1.32.0-beta)])
+AC_INIT([meep],[m4_esyscmd(./version.sh 1.32.0)])
 AC_CONFIG_SRCDIR(src/step.cpp)
 
 # Shared-library version number; indicates api compatibility, and is
 # not the same as the "public" version number.  (Don't worry about this
 # except for public releases.) Note that any change to a C++ class
 # definition (in the .hpp file) generally breaks binary compatibility.
-SHARED_VERSION_INFO="35:0:0" # CURRENT:REVISION:AGE
+SHARED_VERSION_INFO="36:0:0" # CURRENT:REVISION:AGE
```
stevengj:
```suggestion
SHARED_VERSION_INFO="36:0:1" # CURRENT:REVISION:AGE
```
because we added a new interface in #3081 but it is still backwards-compatible with version 35 (because we didn't change the memory layout/fields or any virtual methods).

PR #3155, opened 2026-02-26
Fix bug when combining 2+ mirror symmetries and Bloch-periodic boundary conditions

### LINKED ISSUE #132: bug when combining 3 mirror symmetries with periodic boundary conditions in 3d

oskooi:
Fixes #132.

**Root Cause**

The Bloch phase factor `eikna[d]` is computed using the **symmetry-reduced** grid volume (`gv`) size, while the periodic lattice translation vector (`ilattice_vector`) is computed using the **full** cell size (`user_volume`). This creates a mismatch between the phase applied per periodic translation and the actual translation distance.

**Why it only manifests with 2+ mirror symmetries + periodioc boundaries**

-  **single mirror** (e.g., `Mirror(Y)`) + **periodic in X**: only Y is halved, so `gv.num_direction(X)` equals the original full `user_volume.nx()`. The phase in X is correct. The Y direction typically has k_y = 0 with mirror symmetry, so `eikna[Y] = 1` regardless.
- **2+ mirrors** (e.g., `Mirror(X) + Mirror(Y)`) + **periodic**: Both X and Y are halved. If k_x ≠ 0, then `gv.num_direction(X)` ≈ `user_volume.nx() / 2` and `eikna[X]` is computed for the wrong period. The periodic translation moves by the full cell but applies the phase for half the cell. This produces incorrect field values at boundary conditions leading to expontentially growing errors (instability).
- **k = 0** (pure periodic, no Bloch phase): `eikna = exp(0) = 1` regardless of cell size, so the bug is latent and harmless.

Two new unit tests are added to compare the symmetry-exploiting simulation  against a reference simulation without symmetry, checking field values at multiple points and energies at regular intervals.

### DIFF
--- a/src/boundaries.cpp
+++ b/src/boundaries.cpp
@@ -91,11 +91,15 @@ void fields::use_bloch(direction d, complex<double> kk) {
   k[d] = kk;
   for (int b = 0; b < 2; b++)
     set_boundary(boundary_side(b), d, Periodic);
-  if (real(kk) * gv.num_direction(d) == 0.5 * a) // check b.z. edge exactly
-    eikna[d] = -exp(-imag(kk) * ((2 * pi / a) * gv.num_direction(d)));
+  // Use user_volume (full cell) size, not gv (symmetry-reduced cell) size,
+  // because ilattice_vector and locate_point_in_user_volume translate by
+  // the full cell period. Using gv here would give the wrong Bloch phase
+  // when the cell is halved by mirror symmetries. (See issue #132.)
+  if (real(kk) * user_volume.num_direction(d) == 0.5 * a) // check b.z. edge exactly
+    eikna[d] = -exp(-imag(kk) * ((2 * pi / a) * user_volume.num_direction(d)));
   else {
     const complex<double> I = complex<double>(0.0, 1.0);
-    eikna[d] = exp(I * kk * ((2 * pi / a) * gv.num_direction(d)));
+    eikna[d] = exp(I * kk * ((2 * pi / a) * user_volume.num_direction(d)));
   }
   coskna[d] = real(eikna[d]);
   sinkna[d] = imag(eikna[d]);
--- a/tests/symmetry.cpp
+++ b/tests/symmetry.cpp
@@ -910,6 +910,115 @@ double nonlinear_ex(const grid_volume &gv, double eps(const vec &)) {
   return 1;
 }
 
+// Test mirror symmetry with Bloch-periodic boundaries at the zone edge.
+// This exercises the bug fix from issue #132: eikna[d] must be computed
+// using user_volume (full cell) size, not gv (symmetry-reduced cell) size.
+// Mirror(Y) halves gv in Y, and a non-zero k_y at the Brillouin zone edge
+// (k_y = a/(2*num_y)) means eikna[Y] should be -1. Without the fix,
+// eikna[Y] is computed using the halved cell and gives the wrong phase.
+int test_periodic_mirror_zone_edge(double eps(const vec &)) {
+  double a = 8.0;
+  double ttot = 5.0;
+
+  const grid_volume gv = voltwo(1.0, 1.0, a);
+  the_center = gv.center();
+  const symmetry S = mirror(Y, gv);
+  structure s(gv, eps, no_pml(), S);
+  structure s1(gv, eps);
+  master_printf("Testing mirror symmetry with Bloch-periodic zone-edge k...\n");
+
+  fields f(&s);
+  fields f1(&s1);
+  // k_y at Brillouin zone edge = a/(2*num_y) = 8/(2*8) = 0.5.
+  // This is compatible with Mirror(Y) (fields are anti-periodic in Y).
+  // Mirror(Y) halves gv in Y, so gv.num_direction(Y) != user_volume.num_direction(Y),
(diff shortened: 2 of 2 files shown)


src/boundaries.cpp
```diff
@@ -91,11 +91,15 @@ void fields::use_bloch(direction d, complex<double> kk) {
...
+  // Use user_volume (full cell) size, not gv (symmetry-reduced cell) size,
+  // because ilattice_vector and locate_point_in_user_volume translate by
+  // the full cell period. Using gv here would give the wrong Bloch phase
+  // when the cell is halved by mirror symmetries. (See issue #132.)
+  if (real(kk) * user_volume.num_direction(d) == 0.5 * a) // check b.z. edge exactly
+    eikna[d] = -exp(-imag(kk) * ((2 * pi / a) * user_volume.num_direction(d)));
   else {
     const complex<double> I = complex<double>(0.0, 1.0);
-    eikna[d] = exp(I * kk * ((2 * pi / a) * gv.num_direction(d)));
+    eikna[d] = exp(I * kk * ((2 * pi / a) * user_volume.num_direction(d)));
```
smartalecH, line 102: Wow such a simple fix, for such a hairy bug all these years. Very cool!

stevengj: The odd thing is that mirror symmetries (the halved dimensions) would only ever be in directions $i$ where $k_i = 0$, so it's not clear why this would have an effect?

stevengj: However, it would affect you at $k_i = \pi / a$, which does preserve the mirror symmetry.

stevengj: But then why would this help #132?

stevengj:
Can you verify that this fixes #132?  It's not clear why it would help in a case where $k \ne 0$.

(But even if it fixed an unrelated bug that is nice too.)

oskooi:
It seems the original test case in #132 with `k_point = (0, 0, 0)` actually runs fine (i.e. the fields do *not* blow up) even without this fix. The only test case in which the fields do blow up is `k_point = (0.5, 0.5, 0.5)` (i.e. at the edge of the Brillouin zone, also mentioned in #132) which is fixed by this PR.

PR #3156, opened 2026-02-26
Fix bug in amplitude weight of `Er` source at `r = 0` for `m = ±1` in cylindrical coordinates

### LINKED ISSUE #2704: Incorrect restriction of `Er` source at `r = 0` and `m = ±1` in cylindrical coordinates
In [cylindrical coordinates](https://meep.readthedocs.io/en/latest/Exploiting_Symmetry/#cylindrical-symmetry), $E_r$ is defined at the midpoint of the Yee-grid voxels ($r = 0.5\Delta r$). An $E_r$ point source at $r = 0$ and $m = \pm 1$ requires restriction to $E_r$ grid points at $r = \pm 0.5\Delta r$. There is a [bug in the restriction operation](https://github.com/NanoComp/meep/pull/2538#issuecomment-1777741057) for this case which prevents computing quantities such as the [extraction efficiency](https://meep.readthedocs.io/en/latest/Python_Tutorials/Local_Density_of_States/#extraction-efficiency-of-a-light-emitting-diode-led) involving the ratio of the Poynting flux (`add_flux`) to the LDoS (`dft_ldos`).

Based on the results in [#2538 (comment)](https://github.com/NanoComp/meep/pull/2538#issuecomment-1784250617), a simple workaround is to place the $E_r$ point source at $r = 1.5\Delta r$. This ensures restriction does not involve $E_r$ grid points within the $r = 0$ voxels. A source placed anywhere at $r < 1.5\Delta r$ triggers the bug, however. This slight adjustment of the source position incurs a small error which decreases with increasing grid resolution (smaller $\Delta r$).

oskooi:
Fixes #2704.

**Root Cause**

In cylindrical coordinates, `Er` (and `Dr`) grid points live at half-integer `r` positions due to Yee grid staggering. The first `Er` grid point is at `r = Δr/2`, not `r = 0`. When a user places an `Er` point source at `r = 0`, the `loop_in_chunks` interpolation code computes a bounding box on the `Er` grid that spans `R = -1` (at `r = -Δr/2`) to `R = 1` (at `r = +Δr/2`), and assigns linear interpolation weights of 0.5 to each.

However, `R = -1` (`r = -Δr/2`) is *outside* the physical domain in cylindrical coordinates — negative r doesn't exist. The `owns()` check in `src_vol_chunkloop` skips this non-existent point (`src/sources.cpp:267`), which causes half the source amplitude to be silently lost.

Tracing through the code step by step:

1. `add_volume_source` (`src/sources.cpp:462`) applies delta-function corrections (`data.amp *= gv.a` per collapsed direction) and computes `data.center` from the source position.
2. `loop_in_chunks` (`src/loop_in_chunks.cpp:339`) calls `compute_boundary_weights` which, for a point source at `r = 0` on the Er grid, yields `is.R = -1`, `ie.R = 1` and weights `s0.R = e0.R = 0.5` — a 50/50 split between the two bounding grid points.
3. Chunk clipping (`src/loop_in_chunks.cpp:440-446`) clips `iscS.R` to `max(-1, 1) = 1` (because `little_owned_corner(Er).R = 1`), leaving only the point at `R = 1` with weight 0.5.
4. `src_vol_chunkloop` (`src/sources.cpp:265-305`) loops over grid points, skips `R = -1` via `!fc->gv.owns(iloc)`, and assigns the surviving `R = 1` point an amplitude multiplied by the interpolation weight 0.5.

**Result:** The source amplitude is halved. The same issue affects any component with `iyee.shift(c).R == 1` (i.e., `Er`, `Dr`, `Hz`, `Bz`, `Hp`, `Bp`) when the source is at `r < Δr/2`.

**Why this fix is safe**

1. The compensation only applies to the first grid point in R (`iloc.R == 1`) when the source is very close to the origin (`r_src < Δr/2`). No other grid points or source positions are affected.
2. For integration operations (energy, flux, etc.) that also use `loop_in_chunks`, this code path is never reached because those operations use a different chunkloop.
3. The fix correctly handles all values of `m`, not just `m = ±1`, because the compensation is independent of the angular mode number.

**What this fix enables**

After this fix, the workaround of offsetting the `Er` source to `r = 1.5 / resolution` would no longer be necessary. User could places `Er` sources directly at `r = 0` with correct normalization.

A unit test is added which checks that an `Er` point source at `r = 0` and a source at `r = Δr/2` produce identical fields.

### DIFF
--- a/python/tests/test_pml_cyl.py
+++ b/python/tests/test_pml_cyl.py
@@ -103,7 +103,7 @@ class TestPMLCylindrical(unittest.TestCase):
             ),
         )
 
-        sim.run(until_after_sources=50.94)
+        sim.run(until_after_sources=100.94)
 
         prev_flux = [
             mp.get_fluxes(flux_plus_z)[0],


src/sources.cpp
```diff
@@ -272,6 +272,22 @@ static void src_vol_chunkloop(fields_chunk *fc, int ichunk, component c, ivec is
     vec rel_loc = loc - data->center;
     amps_array[idx_vol] = IVEC_LOOP_WEIGHT(s0, s1, e0, e1, 1) * amp * data->A(rel_loc);
 
+    // In cylindrical coordinates with the grid starting at r=0, for field
+    // components whose first grid point is at r=Δr/2 (not r=0) due to Yee
+    // grid staggering, a point source at r < Δr/2 loses amplitude to a
+    // non-existent grid point at r < 0 (which is skipped by the owns()
+    // check above). Compensate for this lost amplitude by restoring the
+    // interpolation weight to 1.0. (See issue #2704.)
```
stevengj, line 280: This seems a little weird to me.  If you have a source at $\Delta r/2$, shouldn't the $r=0$ boundary conditions imply another source at $-\Delta r/2$?

stevengj:
Closing this PR, as I don't think it's the right fix as commented above.  One of the tricky things here is defining what the normalization of the discretized delta function source even means in cylindrical coordinates.

PR #3157, opened 2026-02-27
Fix bug for 3D simulation using 1D cell and single-precision floating point

### LINKED ISSUE #2916: Segmentation fault for 3D simulation using 1D cell

oskooi:
Fixes #2916.

**Root cause analysis**

The segfault occurs in `fields::step_boundaries()` (`src/step.cpp:263`) when dereferencing a pointer from the `outgoing_connection` vector at an out-of-bounds index.

The bug is in `connect_the_chunks()` (`src/boundaries.cpp`), which has **two passess** over not-owned boundary voxels:

1. **First pass** (line 480): Counts `comm_sizes` for each (`field_type`, `connect_phase`, `chunk_pair`) key. It determines `connect_phase` by comparing the Bloch phase as `complex<double>`: `const connect_phase ip = thephase == 1.0 ? CONNECT_COPY : ...`
2. **Second pass** (line 497): Sets up actual connection pointers. It determines `connect_phase` by comparing the Bloch phase as `complex<realnum>` (= `complx<float>` in single precision): `const connect_phase ip = thephase == static_cast<realnum>(1.0) ? CONNECT_COPY : ...`

When `eikna[X] == eikna[Y]` (because `kx == ky`), a boundary point requiring translation in both X and Y accumulates a combined phase of `conj(eikna[X]) * eikna[Y] = |eikna[X]|²`. In exact arithmetic this equals 1.0, but `cos²(θ) + sin²(θ)` computed in double-precision floating point has a tiny rounding error (~10⁻¹⁶). This error is:
- **Detectable in double precision**: `thephase != 1.0` → classified as `CONNECT_PHASE`
- **Below float precision** (~10⁻⁷): `float(thephase) == 1.0f` → classified as `CONNECT_COPY`

This mismatch means the first pass allocates space for 12 `CONNECT_PHASE` realnums, but the second pass only pushes 8 connection pointers (the other 4 go to `CONNECT_COPY` instead). When `step_boundaries` tries to read all 12, it accesses index 8 of an 8-element vector → segfault.

**Why resolution=51 works**: Different resolution produces different grid dimensions and chunk splits. The angle `θ = k * 2π/a * n` differs, so `cos²(θ) + sin²(θ)` may happen to equal exactly 1.0 in double (or the chunks may not have corner boundary points requiring both X and Y translations).

**Fix**

Change the second pass (line 497-500) to use `thephase_double` (the original `complex<double>` value) instead of `thephase` (the truncated `complex<float>` value) for the connect_phase classification. This ensures both passes use identical comparison.

**Test**

Added `test_1d_cell_bloch()` to `tests/three_d.cpp` that creates a D3 grid with nx=1, ny=1 (1D cell in 3D), resolution=52, `k_point=(-0.2, -0.2, 0)`, PML, and runs for several timesteps. Without the fix, this segfaults on the first step in single precision.

### DIFF
--- a/src/boundaries.cpp
+++ b/src/boundaries.cpp
@@ -479,7 +479,6 @@ void fields::connect_the_chunks() {
           // We're looking at a border element...
           std::complex<double> thephase;
           if (locate_component_point(&c, &here, &thephase) && !on_metal_boundary(here)) {
-            std::complex<realnum> thephase(thephase.real(), thephase.imag());
             for (int j = 0; j < num_chunks; j++) {
               const std::pair<int, int> pair_j_to_i{j, i};
               const bool i_is_mine = chunks[i]->is_mine();


stevengj:
Seems like we should check `if (eikna[X] == eikna[Y]) thephase = 1.0;` and hence use `CONNECT_COPY`, rather than relying on roundoff of `conj(eikna[X]) * eikna[Y]`

src/boundaries.cpp
```diff
@@ -346,6 +346,21 @@ bool fields_chunk::needs_W_notowned(component c) {
...
 }
 
+/* approximate comparisons of phases to 1 or -1.   exp(ix) * exp(-ix) is not exactly 1
+   due to roundoff errors, but we want to treat it equal for the purpose of determining
+   the boundary conditions, because we have optimized b.c.'s for phase +1 and -1, and
+   roundoff errors on the order of 1e-13 are not significant relative to the truncation
+   errors in meep anyway.  (the alternative would be to accumulate symmetry phases more
+   accurately by adding exponents rather than multiplying). */
+static bool phase_isclose(std::complex<double> thephase, double realphase) {
+  return fabs(thephase.imag()) < 1e-13 && fabs(thephase.imag() - realphase) < 1e-13;
```
stevengj:
```suggestion
  return fabs(thephase.imag()) < 1e-13 && fabs(thephase.real() - realphase) < 1e-13;
```

PR #3161, opened 2026-03-01
Add support for `dft_force` with mirror symmetry

### LINKED ISSUE #129: bug when combining dft forces with symmetry objects
There is a bug in the dft force feature when used with symmetry objects. This is demonstrated in the following simple example of 2d vacuum with PML boundaries involving a point source at the origin enclosed by a square box of `force-region`s. Given the symmetry of this arrangement, the total force obtained by summing up the four line regions should be 0. This is the result when no mirror-symmetry objects are used. However, the presence of any mirror-symmetry objects produces a non-zero value.

```scm
(set-param! resolution 10)

(set! geometry-lattice (make lattice (size 5 5 no-size)))

(set! pml-layers (list (make pml (thickness 1))))

(set! sources (list (make source (src (make gaussian-src (frequency 1.0) (fwidth 1.0)))
                          (center 0 0 0) (component Ez))))

                       (make mirror-sym (direction Y))))

(define force-box (add-force 1 0 1
                             (make force-region (direction X) (center +1 0 0) (size 0 2 0) (weight +1.0))
                             (make force-region (direction X) (center -1 0 0) (size 0 2 0) (weight -1.0))
                             (make force-region (direction Y) (center 0 +1 0) (size 2 0 0) (weight +1.0))
                             (make force-region (direction Y) (center 0 -1 0) (size 2 0 0) (weight -1.0))))

(run-sources+ (stop-when-fields-decayed 50 Ez (vector3 0 0 0) 1e-6))

(display-forces force-box)
```
The above script produces the following incorrect non-zero output:
```
force1:, 1.0, 0.38368976835348184
```
When the symmetry objects are removed, the force has the correct value of 0:
```
force1:, 1.0, 1.7105999776390046e-16
```

oskooi:
Fixes #129.

**Context**

#129 reports field instability when combining DFT force monitors with symmetry objects. The root cause is that `fields::add_dft_force()` in `src/stress.cpp` unconditionally calls `S.reduce(where_)` to apply symmetry reduction to the force region volumes. In contrast, `fields::add_dft_flux()` in `src/dft.cpp` has a `use_symmetry` parameter (defaulting to `true`) that allows bypassing `S.reduce()`.

The `S.reduce()` operation is problematic for force calculations because the stress tensor involves pairing different field components (E in force direction vs. E in normal direction) as off-diagonal terms. When symmetry transforms the volume and modifies the component weights, the paired DFT chunk lists (`offdiag1`, `offdiag2`) can end up with mismatched structures, producing incorrect/unstable force values.

**Fix**

**Step 1: Add `use_symmetry` parameter to `add_dft_force` in C++ (3 files)**

`src/meep.hpp` (lines 2165-2174): Add `bool use_symmetry = true` parameter to all three `add_dft_force` overloads, matching the pattern of `add_dft_flux` (lines 2045-2057).

`src/stress.cpp` (lines 153-157): Accept the new `use_symmetry` parameter and conditionally apply `S.reduce()`: `volume_list *where = use_symmetry ? S.reduce(where_): new volume_list(where_)`; This mirrors dft.cpp line 607.

**Step 2: Pass `use_symmetry=False` from Python when symmetries are present (1 file)**

`python/simulation.py` (lines 3399-3404): Modify `_add_force` to pass `use_symmetry=False` to `add_dft_force` when the simulation has symmetry objects. This is done by passing an extra argument through `_add_fluxish_stuff`, which already supports `*args` forwarding (line 3818: `add_dft_stuff(vol_list, freq, decimation_factor, *args)`).

**Step 3: Add unit test (1 file)**

1. Runs a force calculation on a geometry with mirror symmetry using `symmetries=[mp.Mirror(mp.Y)]`.
2. Runs the same geometry without symmetry.
3. Asserts both produce the same force values (within tolerance).

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -3399,8 +3399,9 @@ class Simulation:
     def _add_force(self, freq, forces, decimation_factor):
         if self.fields is None:
             self.init_sim()
+        use_symmetry = not self.symmetries
         return self._add_fluxish_stuff(
-            self.fields.add_dft_force, freq, forces, decimation_factor
+            self.fields.add_dft_force, freq, forces, decimation_factor, use_symmetry
         )
 
     def display_forces(self, *forces):
--- a/python/tests/test_force.py
+++ b/python/tests/test_force.py
@@ -5,50 +5,93 @@ import meep as mp
 
 class TestForce(unittest.TestCase):
     def setUp(self):
+        self.resolution = 20
+        self.cell = mp.Vector3(10, 10)
+        self.pml_layers = mp.PML(1.0)
 
-        resolution = 20
-        cell = mp.Vector3(10, 10)
-        pml_layers = mp.PML(1.0)
-        fcen = 1.0
-        df = 1.0
-        sources = mp.Source(
-            src=mp.GaussianSource(fcen, fwidth=df), center=mp.Vector3(), component=mp.Ez
+        self.fcen = 1.0
+        self.df = 1.0
+        self.sources = mp.Source(
+            src=mp.GaussianSource(frequency=self.fcen, fwidth=self.df),
+            center=mp.Vector3(),
+            component=mp.Ez,
         )
 
-        self.sim = mp.Simulation(
-            resolution=resolution,
-            cell_size=cell,
-            boundary_layers=[pml_layers],
-            sources=[sources],
-        )
-
-        fr = mp.ForceRegion(mp.Vector3(y=1.27), direction=mp.Y, size=mp.Vector3(4.38))
-        self.myforce = self.sim.add_force(fcen, 0, 1, fr, decimation_factor=1)
-        self.myforce_decimated = self.sim.add_force(
-            fcen, 0, 1, fr, decimation_factor=10
+        self.fr = mp.ForceRegion(
+            center=mp.Vector3(0, 1.27), direction=mp.Y, size=mp.Vector3(4.38, 0)
         )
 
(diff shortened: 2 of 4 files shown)


stevengj:
This "fix" seems wrong.  The `volume_list *where = S.reduce(where_);` line in `stress.cpp` *knows* about the force direction that is being computed (via `where->c`), and hence takes this into account in the symmetry reduction, exactly as we do for flux.  Basically there is a `where->weight` that is update as necessary, and is passed as the `extra_weight` parameter in `add_dft`, so that it get included (once) in each stress term.

It could of course be that there is some other bug in how this procedure is being carried out, but the solution is not to remove it entirely, since the symmetry reduction is an important optimization.

stevengj:
It is true, however, that the diagonal and off-diagonal stress-tensor terms should be transformed separately.  And the problem is essentially that the `where->c` transformation is only looking at one of the components, and not both.

(Diagonal terms basically don't pick up any transformation, whereas off-diagonal terms pick up *two* transformations.)

The right fix here is maybe to add a second component to the `volume_list`, defaulting to `Dielectric` (or something that transforms as a scalar), which we can then set appropriately for these transformations.

(For the diagonal terms, we can just set `where->c` to `Dielectric` as well.)

stevengj:
An even easier fix is to optionally let `S.reduce` look at the normal direction in addition to `where->c` in order to decide the phase.

PR #3163, opened 2026-03-06
Fix bug in interpolation of epsilon grid from Meep to MPB

### LINKED ISSUE #1237: 2d eigenmode source and mode decomposition not working in 2d simulation
The following example tries to launch an eigenmode source in a [holey waveguide](https://meep.readthedocs.io/en/latest/Python_Tutorials/Resonant_Modes_and_Transmission_in_a_Waveguide_Cavity/) in a 2d simulation but aborts with an error due to MPB. The same error occurs when trying to use mode decomposition for a 2d eigenmode. Neither `EigenModeSource` or `get_eigenmode_coefficients` has previously been tested for this type of use case. All the examples in [Tutorial/Mode Decomposition](https://meep.readthedocs.io/en/latest/Python_Tutorials/Mode_Decomposition/) and [Tutorial/Eigenmode Source](https://meep.readthedocs.io/en/latest/Python_Tutorials/Eigenmode_Source/) are in 1d.

The schematic below (obtained using [`plot2D`](https://meep.readthedocs.io/en/latest/Python_User_Interface/#data-visualization)) shows the 2d source (red hatches) and mode monitor (blue hatches).

*additional info:* removing the `symmetries` from the `Simulation` constructor does not solve the problem. However, changing the `size` of the source/monitor from a 2d rectangle to a line in *y* (i.e. `size=mp.Vector3(0,sy)`) works fine.

holey-wvg-epsilon

```py
import meep as mp
import numpy as np

resolution = 50 # pixels/μm                                                                                                                                          

dpml = 1

eps = 13          # dielectric constant of waveguide                                                                                                                 
w = 1.2           # width of waveguide                                                                                                                               
r = 0.36          # radius of holes                                                                                                                                  
N = 11            # number of holes in waveguide                                                                                                                     

sx = N
sy = 10
cell_size = mp.Vector3(sx,sy)

                     size=mp.Vector3(mp.inf,w,mp.inf),
                     material=mp.Medium(epsilon=eps))]

for n in range(N):
    geometry.append(mp.Cylinder(radius=r,
                                center=mp.Vector3(-0.5*sx+0.5+n),
                                height=mp.inf,
                                material=mp.air))

f_bandedge = 0.20025  # from MPB                                                                                                                                     
fsrc = 0.9*f_bandedge # frequency of eigenmode source                                                                                                                
kx = 0.4              # initial guess for wavevector in x-direction of eigenmode                                                                                     
bnum = 1              # band number of eigenmode                                                                                                                     

df = 0.1              # frequency width                                                                                                                              

symmetries = [mp.Mirror(mp.Y,phase=-1)]

sources = [mp.EigenModeSource(src=mp.GaussianSource(fsrc,fwidth=df),
                              center=mp.Vector3(0,0),
                              size=mp.Vector3(1,sy),
                              direction=mp.NO_DIRECTION,
                              eig_parity=mp.ODD_Y+mp.EVEN_Z,
                              eig_kpoint=mp.Vector3(kx),
                              eig_band=bnum,
                              eig_match_freq=True)]

                    resolution=resolution,
                    boundary_layers=pml_layers,
                    sources=sources,
                    geometry=geometry,
                    symmetries=symmetries)

```

**output**
```
-----------
Initializing structure...
time for choose_chunkdivision = 0.000737682 s
Working in 2D dimensions.
Computational cell is 11 x 10 x 0 with resolution 50
     block, center = (0,0,0)
          size (1e+20,1.2,1e+20)
          dielectric constant epsilon diagonal = (13,13,13)
     cylinder, center = (-5,0,0)
          radius 0.36, height 1e+20, axis (0, 0, 1)
     cylinder, center = (-4,0,0)
          radius 0.36, height 1e+20, axis (0, 0, 1)
     cylinder, center = (-3,0,0)
          radius 0.36, height 1e+20, axis (0, 0, 1)
     cylinder, center = (-2,0,0)
          radius 0.36, height 1e+20, axis (0, 0, 1)
     cylinder, center = (-1,0,0)
          radius 0.36, height 1e+20, axis (0, 0, 1)
     cylinder, center = (0,0,0)
          radius 0.36, height 1e+20, axis (0, 0, 1)
     cylinder, center = (1,0,0)
          radius 0.36, height 1e+20, axis (0, 0, 1)
     cylinder, center = (2,0,0)
          radius 0.36, height 1e+20, axis (0, 0, 1)
     cylinder, center = (3,0,0)
          radius 0.36, height 1e+20, axis (0, 0, 1)
     ...(+ 2 objects not shown)...
time for set_epsilon = 0.499155 s
-----------
  File "holey-wvg-eigsource-pulse-bug.py", line 54, in <module>
  File "/usr/local/lib/python3.5/site-packages/meep/simulation.py", line 2286, in run
    self.init_sim()
  File "/usr/local/lib/python3.5/site-packages/meep/simulation.py", line 1338, in init_sim
    self.add_source(s)
  File "/usr/local/lib/python3.5/site-packages/meep/simulation.py", line 1624, in add_source
    add_eig_src()
  File "/usr/local/lib/python3.5/site-packages/meep/__init__.py", line 3868, in add_eigenmode_source
    return _meep.fields_add_eigenmode_source(self, *args)
RuntimeError: meep: invalid dielectric function for MPB
```

oskooi:
Fixes #1237.

**Bug Analysis**

The error "invalid dielectric function for MPB" is triggered by MPB's `check_maxwell_dielectric()` (in `maxwell_eps.c:266`), which validates that every grid point's inverse-epsilon tensor (`eps_inv`) is positive-definite.

**Root cause (in `mpb.cpp`, the primary fix)**

The `meep_mpb_eps` callback assembles a 3×3 `eps_inv` tensor at each MPB grid point by querying Meep's `get_chi1inv` for 6 independent components. The **diagonal** components each come from their natural Yee grid positions (Ex grid for `chi1inv_xx`, Ey grid for `chi1inv_yy`, Ez grid for `chi1inv_zz`), which are spatially offset by half a pixel from each other. The **off-diagonal** components were queried from only one grid — e.g., `m01 = chi1inv(Ex, Y, p)` used only the Ex grid.

Near a curved material boundary (like a Cylinder), the Ex grid point may sit right at the interface where subpixel averaging produces a large off-diagonal `chi1inv_xy`, while the Ey grid point (half a pixel away) is inside the high-epsilon material with a small `chi1inv_yy`. The assembled tensor then has `m01² > m00 * m11`, failing positive-definiteness.

Fix: Average each off-diagonal component from both relevant Yee grids:
- `m01 = 0.5 * (chi1inv(Ex,Y,p) + chi1inv(Ey,X,p))`
- `m02 = 0.5 * (chi1inv(Ex,Z,p) + chi1inv(Ez,X,p))`
- `m12 = 0.5 * (chi1inv(Ey,Z,p) + chi1inv(Ez,Y,p))`

This produces off-diagonal values consistent with both corresponding diagonal values.

**Secondary bug (in `monitor.cpp`, defensive six)**

`structure_chunk::get_chi1inv_at_pt` has a frequency-dependent code path (entered whenever `frequency != 0`) that builds a full 3×3 chi1inv tensor using `chi1inv[comp_list[com_it]][dir_int][idx]` for all three components — but `idx` was computed via `gv.index(c, iloc)` for a single component `c`. On the Yee grid, the same flat index refers to a different spatial location for each component, so this reads wrong data. For non-dispersive materials (no susceptibilities or conductivity), this tensor path is a no-op that only introduces errors.

**Fix:** Skip the tensor assembly path when no frequency-dependent material properties exist, returning the direct `chi1inv[c][d][idx]` value instead.

### DIFF
--- a/src/monitor.cpp
+++ b/src/monitor.cpp
@@ -266,8 +266,10 @@ complex<double> structure_chunk::get_chi1inv_at_pt(component c, direction d, int
   if (is_mine()) {
     if (frequency == 0)
       return chi1inv[c][d] ? chi1inv[c][d][idx] : (d == component_direction(c) ? 1.0 : 0);
+    // ----------------------------------------------------------------- //
+    // ---- Step 1: Get instantaneous chi1 tensor ----------------------
+    // ----------------------------------------------------------------- //
 
-    // Determine field type and component list for tensor assembly
     int my_stuff = E_stuff;
     component comp_list[3];
     if (is_electric(c)) {
@@ -299,20 +301,8 @@ complex<double> structure_chunk::get_chi1inv_at_pt(component c, direction d, int
     // (computed for component `c`) to access chi1inv for all three components.
     // On the Yee grid, each component is stored at a different spatial position,
     // so using `c`'s index for other components accesses the wrong grid point.
-    // This is only needed for frequency-dependent materials (susceptibilities or
-    // conductivity). For non-dispersive materials, skip the tensor path and
-    // return the value directly — the result is exact and avoids the index error.
-    bool needs_tensor_path = (chiP[my_stuff] != NULL);
-    for (int com_it = 0; com_it < 3 && !needs_tensor_path; com_it++)
-      for (int dir_int = 0; dir_int < 3 && !needs_tensor_path; dir_int++)
-        if (conductivity[comp_list[com_it]][dir_int]) needs_tensor_path = true;
-
-    if (!needs_tensor_path)
-      return chi1inv[c][d] ? chi1inv[c][d][idx] : (d == component_direction(c) ? 1.0 : 0);
-
-    // ----------------------------------------------------------------- //
-    // ---- Step 1: Get instantaneous chi1 tensor ----------------------
-    // ----------------------------------------------------------------- //
+    // For frequency-dependent materials (susceptibilities or conductivity),
+    // this introduces a first-order error.
 
     std::complex<double> chi1_inv_tensor[9] = {
         std::complex<double>(1, 0), std::complex<double>(0, 0), std::complex<double>(0, 0),


src/monitor.cpp
```diff
@@ -297,6 +295,25 @@ complex<double> structure_chunk::get_chi1inv_at_pt(component c, direction d, int
...
+    // This is only needed for frequency-dependent materials (susceptibilities or
+    // conductivity). For non-dispersive materials, skip the tensor path and
+    // return the value directly — the result is exact and avoids the index error.
+    bool needs_tensor_path = (chiP[my_stuff] != NULL);
+    for (int com_it = 0; com_it < 3 && !needs_tensor_path; com_it++)
+      for (int dir_int = 0; dir_int < 3 && !needs_tensor_path; dir_int++)
+        if (conductivity[comp_list[com_it]][dir_int]) needs_tensor_path = true;
+
+    if (!needs_tensor_path)
+      return chi1inv[c][d] ? chi1inv[c][d][idx] : (d == component_direction(c) ? 1.0 : 0);
```
stevengj: This is not fixing the error (which is a first-order error in the frequency-dependent part), it is just avoiding it if you ask for a frequency-dependent ε but the actual material is frequency-independent.

I would rather fix the underlying problem, but in any case it seems like it belongs in a different issue.

src/monitor.cpp
```diff
@@ -297,6 +295,25 @@ complex<double> structure_chunk::get_chi1inv_at_pt(component c, direction d, int
       my_stuff = B_stuff;
     }
 
+    // The full tensor assembly path below uses the same flat array index `idx`
+    // (computed for component `c`) to access chi1inv for all three components.
+    // On the Yee grid, each component is stored at a different spatial position,
+    // so using `c`'s index for other components accesses the wrong grid point.
+    // This is only needed for frequency-dependent materials (susceptibilities or
+    // conductivity). For non-dispersive materials, skip the tensor path and
```
stevengj: Might be nice to keep a comment here that we are making a first-order error in this case, and leave a TODO

src/monitor.cpp
oskooi: Changes have been reverted.

src/monitor.cpp
oskooi: Done.

PR #3165, opened 2026-03-07
fix symmetry reduction for force

### LINKED ISSUE #129: bug when combining dft forces with symmetry objects

stevengj:
Fixes #129, following discussion in #3161.

The fix is to add an optional second "conjugate" component `cc` to a `volume_list` whose phase (conjugated) is included in symmetry reductions via `reduce`.   For the force calculations, `cc` is set according to the `normal_direction`.

cc @oskooi — this PR is just a rough draft right now, and could use a test (e.g. copied from #3161).

### DIFF
--- a/python/tests/test_force.py
+++ b/python/tests/test_force.py
@@ -35,13 +35,13 @@ class TestForce(unittest.TestCase):
                 weight=-1.0,
             ),
             mp.ForceRegion(
-                direction=mp.Y,
+                direction=mp.X,
                 center=mp.Vector3(0, 1.0, 0),
                 size=mp.Vector3(2.0, 0, 0),
                 weight=1.0,
             ),
             mp.ForceRegion(
-                direction=mp.Y,
+                direction=mp.X,
                 center=mp.Vector3(0, -1.0, 0),
                 size=mp.Vector3(2.0, 0, 0),
                 weight=-1.0,


stevengj:
I copied over the test from the other PR.  Would be good to double-check that it fixes the original test case in #129.

oskooi:
The fix proposed in this PR (ceff208) does *not* produce the expected result of **zero** force for the [test involving an Ez dipole in 2d vacuum](https://gist.github.com/oskooi/4aff61f52a121cbecac934562b7334c7). (The test is based on the one provided in #129 and transcribed from Scheme to Python. The simulation layout is shown below.) Running the test script with the `--with-symmetry` flag (mirror symmetry in $x$ *and* $y$) produces an incorrect force of `0.07380557`. The result is `-1.0482879755219221e-17` without symmetries (which is correct). 

Additionally, using only a mirror symmetry in $x$ produces the correct result . However, a $y$ mirror symmetry produces `0.07380557` which is incorrect. This suggests there is a bug in the $y$ mirror-symmetry case.

force_layout

stevengj:
Okay, so the test proposed in #3161 is insufficient and should be replaced.   Can you update the test in this PR?

Does the Y mirror symmetry alone fail, or only the combination of X and Y?

oskooi:
Here is a summary of the different test cases:

image

python/tests/test_force.py
```diff
@@ -5,50 +5,93 @@
...
         )
+        f_nosym = mp.get_forces(force_nosym)
+
+        # Run with mirror symmetry.
+        sim_sym = mp.Simulation(
+            resolution=self.resolution,
+            cell_size=self.cell,
+            boundary_layers=[self.pml_layers],
+            sources=[self.sources],
+            symmetries=[mp.Mirror(mp.Y)],
```
oskooi:
```suggestion
            symmetries=[mp.Mirror(mp.X), mp.Mirror(mp.Y)],
```

python/tests/test_force.py
```diff
@@ -5,50 +5,93 @@
...
         )
 
-        self.sim = mp.Simulation(
-            resolution=resolution,
-            cell_size=cell,
-            boundary_layers=[pml_layers],
-            sources=[sources],
+        self.fr = mp.ForceRegion(
+            center=mp.Vector3(0, 1.27), direction=mp.Y, size=mp.Vector3(4.38, 0)
         )
```
oskooi, line 23:
```suggestion

        self.fr_box = [
            mp.ForceRegion(
                direction=mp.X,
                center=mp.Vector3(1.0, 0, 0),
	        size=mp.Vector3(0, 2.0, 0),
                weight=1.0
            ),
            mp.ForceRegion(
                direction=mp.X,
	        center=mp.Vector3(-1.0, 0, 0),
	        size=mp.Vector3(0, 2.0, 0),
                weight=-1.0
            ),
            mp.ForceRegion(
                direction=mp.Y,
                center=mp.Vector3(0, 1.0, 0),
                size=mp.Vector3(2.0, 0, 0),
                weight=1.0
            ),
            mp.ForceRegion(
		direction=mp.Y,
                center=mp.Vector3(0, -1.0, 0),
                size=mp.Vector3(2.0, 0, 0),
                weight=-1.0
            ),
        ]
```

python/tests/test_force.py
```diff
@@ -5,50 +5,93 @@
...
+        # Run with mirror symmetry.
+        sim_sym = mp.Simulation(
+            resolution=self.resolution,
+            cell_size=self.cell,
+            boundary_layers=[self.pml_layers],
+            sources=[self.sources],
+            symmetries=[mp.Mirror(mp.Y)],
+        )
+
+        force_sym = sim_sym.add_force(self.fcen, 0, 1, self.fr)
```
oskooi:
```suggestion
        force_sym = sim_sym.add_force(self.fcen, 0, 1, *self.fr_box)
```

python/tests/test_force.py
```diff
@@ -5,50 +5,93 @@
...
+
+        # Run without mirror symmetry.
+        sim_nosym = mp.Simulation(
+            resolution=self.resolution,
+            cell_size=self.cell,
+            boundary_layers=[self.pml_layers],
+            sources=[self.sources],
+        )
+
+        force_nosym = sim_nosym.add_force(self.fcen, 0, 1, self.fr)
```
oskooi:
```suggestion
        force_nosym = sim_nosym.add_force(self.fcen, 0, 1, *self.fr_box)
```

python/tests/test_force.py
```diff
@@ -5,51 +5,121 @@
...
+                weight=-1.0,
+            ),
+            mp.ForceRegion(
+                direction=mp.Y,
+                center=mp.Vector3(0, 1.0, 0),
+                size=mp.Vector3(2.0, 0, 0),
+                weight=1.0,
+            ),
+            mp.ForceRegion(
+                direction=mp.Y,
```
stevengj: This test seems wrong.  For a given force component (the `direction`), you need to integrate the stress tensor over a closed surface.   Here, you have four `ForceRegion`s that together define a closed box, so all of them should have the *same* direction (e.g. `direction=mp.X`).

```suggestion
                direction=mp.X,
                center=mp.Vector3(0, 1.0, 0),
                size=mp.Vector3(2.0, 0, 0),
                weight=1.0,
            ),
            mp.ForceRegion(
                direction=mp.X,
```

stevengj:
I fixed a typo in my PR, should pass now.

PR #3166, opened 2026-03-10
OpenMP parallelize grid initialization (set_epsilon)

Luochenghuang:
The grid initialization loop in `set_chi1inv` iterates over all grid points to compute the inverse permittivity tensor. At high resolutions, this takes awhile on a single core.

Parallelize the loop with OpenMP (`PLOOP_OVER_IVECS_C`) when the material function is thread-safe (standard C++ geometry objects, not Python callbacks). The `trivial[]` flags use OpenMP reduction.

A new virtual method `material_function::is_thread_safe()` returns false by default (safe for Python callbacks). The `geom_epsilon` subclass overrides it to return true unless `MATERIAL_USER` materials are present.

Measured speedup: 30.7s -> 3.2s with 22M voxels (9.7x speed using 22 cores). Set `OMP_NUM_THREADS` to control parallelism.

### DIFF
--- a/src/anisotropic_averaging.cpp
+++ b/src/anisotropic_averaging.cpp
@@ -252,8 +252,8 @@ breakout:
   if (medium.is_thread_safe()) {
     bool trivial0 = true, trivial1 = true, trivial2 = true;
     PLOOP_OVER_IVECS_C(gv, gv.little_corner() + gv.iyee_shift(c),
-                        gv.big_corner() + gv.iyee_shift(c), i,
-                        "omp parallel for collapse(3) reduction(&&:trivial0,trivial1,trivial2)") {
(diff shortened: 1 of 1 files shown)


oskooi:
Here are benchmarking results based on timing `set_epsilon` for a 3D cell of random spheres (position and size) using this [script](https://gist.github.com/oskooi/5340281f45af9f998e407d61d2bf9f82). The speedup seems to be decent (nearly linear).

set_epsilon_runtime_vs_numthreads

Luochenghuang:
bench_speedup
Thank you Ardavan for the benchmark. I too did a benchmark on a AMD EPYC-Milan processor with 83 cores and here are the results.

PR #3167, opened 2026-03-12
Cache `maxwell_data` across `get_eigenmode_coefficients` calls

### LINKED ISSUE #1533: Caching / reusing the results of MPB eigenmode calculations
It seems that every time a mode overlap calculation is performed via `get_eigenmode_coefficients()`, MPB is called to solve for the eigenmodes of the cross section. I am wondering if there is some way to reuse or cache the results of these MPB calculations. Alternatively, perhaps `get_eigenmode_coefficients()` could return a reference to the MPB results that could be used in subsequent calls to `get_eigenmode_coefficients()`. Thus, the eigenmode calculcation could be avoided and the function would only need to perform the overlap calculation. Perhaps this would end up being just a more general interface for performing overlap integrals?

One example of where this might be useful is when using a step function to monitor the convergence of the mode overlaps for deciding when to terminate a simulation. I imagine this feature could also be useful for your adjoint module to avoid redundant eigenmode calculations on the waveguide cross sections (which are not changing during optimization). When many frequencies or monitors are of interest, the overhead here could be significant.

Luochenghuang:
## Summary

When computing diffraction efficiencies via get_eigenmode_coefficients with DiffractedPlanewave, each call redundantly recreates MPB's maxwell_data (FFTW plans + epsilon inverse tensor at every grid point). For structures with many geometry objects, set_maxwell_dielectric dominates the per-call cost.

This PR caches the maxwell_data on the dft_flux object so it persists across separate Python calls to get_eigenmode_coefficients. The first call creates the cache; subsequent calls on the same monitor reuse it. The cache is automatically invalidated when the monitor is removed or destroyed.

Also skips the phase-fixing pass for DiffractedPlanewave since |α|² is phase-invariant (the mode–mode self-overlap normalization cancels any global phase factor).

No Python API changes — the caching is fully transparent.

Partially fixes #1533.

## Changes

**src/meep.hpp**: Add eigenmode_cache field, destructor, and invalidate_eigenmode_cache() to dft_flux. Add skip_phase_fix parameter to get_eigenmode. Add eigenmode_cache parameter to get_eigenmode_coefficients.
**src/mpb.cpp**: Use caller-provided cache in get_eigenmode_coefficients; persist or destroy maxwell_data based on whether a cache pointer is supplied. Guard the phase-fix block behind skip_phase_fix. Implement invalidate_eigenmode_cache().
**src/dft.cpp**: Initialize eigenmode_cache = NULL in all constructors and copy constructor. Add destructor that invalidates cache. Invalidate cache in remove().
**python/meep.i**: Change SWIG wrappers to take dft_flux* (pointer) so cache modifications persist across calls. Pass &flux->eigenmode_cache to the C++ function.
**python/tests/test_binary_grating.py**: Add cache consistency check verifying repeated calls produce identical results.

## Benchmark
The benchmark script can be found [here](https://gist.github.com/Luochenghuang/3ca93f0c92846f3d081cfe07f9a4d17e).
test_binary_grating structure (resolution 30, 60 DiffractedPlanewave calls across 2 flux monitors):

| | Total (s) | Per-call (ms) |
|---|---|---|
| Without cache | 0.252 | 4.2 |
| With cache | 0.042 | 0.7 |
| **Speedup** | **6.0×** | |

## Test plan

[x] test_binary_grating_special_kz (all 3 variants) — energy conservation and cache consistency
[x] Benchmark confirms 6× speedup on test_binary_grating structure
[ ] CI tests

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -2077,15 +2077,20 @@ class Simulation:
             absorbers,
             self.extra_materials,
             self.split_chunks_evenly,
-            False
-            if self.chunk_layout
-            and not isinstance(self.chunk_layout, mp.BinaryPartition)
-            else True,
+            (
+                False
+                if self.chunk_layout
+                and not isinstance(self.chunk_layout, mp.BinaryPartition)
+                else True
+            ),
             None,
             True if self._output_stats is not None else False,
-            self.chunk_layout
-            if self.chunk_layout and isinstance(self.chunk_layout, mp.BinaryPartition)
-            else None,
+            (
+                self.chunk_layout
+                if self.chunk_layout
+                and isinstance(self.chunk_layout, mp.BinaryPartition)
+                else None
+            ),
         )
         self.geps = mp._set_materials(
             self.structure,
@@ -2897,17 +2902,21 @@ class Simulation:
         pts = [s.center for s in self.sources]
 
         src_freqs_min = min(
-            s.src.frequency - 1 / s.src.width / 2
-            if isinstance(s.src, mp.GaussianSource)
-            else mp.inf
+            (
+                s.src.frequency - 1 / s.src.width / 2
+                if isinstance(s.src, mp.GaussianSource)
+                else mp.inf
+            )
             for s in self.sources
         )
         fmin = max(0, src_freqs_min)
 
         fmax = max(
-            s.src.frequency + 1 / s.src.width / 2
-            if isinstance(s.src, mp.GaussianSource)
-            else 0
+            (
+                s.src.frequency + 1 / s.src.width / 2
+                if isinstance(s.src, mp.GaussianSource)
+                else 0
+            )
             for s in self.sources
         )
 
--- a/src/monitor.cpp
+++ b/src/monitor.cpp
@@ -164,7 +164,7 @@ complex<double> fields_chunk::get_field(component c, const ivec &iloc) const {
 }
 
 complex<double> fields::get_chi1inv(component c, direction d, const ivec &origloc, double frequency,
-                                    bool parallel) const {
+                                    bool parallel, bool *is_frequency_dependent) const {
   ivec iloc = origloc;
   complex<double> aaack = 1.0;
   locate_point_in_user_volume(&iloc, &aaack);
@@ -173,7 +173,8 @@ complex<double> fields::get_chi1inv(component c, direction d, const ivec &origlo
       if (chunks[i]->gv.owns(S.transform(iloc, sn))) {
         signed_direction ds = S.transform(d, sn);
         complex<double> val =
-            chunks[i]->get_chi1inv(S.transform(c, sn), ds.d, S.transform(iloc, sn), frequency) *
+            chunks[i]->get_chi1inv(S.transform(c, sn), ds.d, S.transform(iloc, sn), frequency,
+                                   is_frequency_dependent) *
             complex<double>(ds.flipped ^ S.transform(component_direction(c), sn).flipped ? -1 : 1,
                             0);
         return parallel ? sum_to_all(val) : val;
@@ -184,18 +185,18 @@ complex<double> fields::get_chi1inv(component c, direction d, const ivec &origlo
 }
 
 complex<double> fields_chunk::get_chi1inv(component c, direction d, const ivec &iloc,
-                                          double frequency) const {
-  return s->get_chi1inv(c, d, iloc, frequency);
+                                          double frequency, bool *is_frequency_dependent) const {
+  return s->get_chi1inv(c, d, iloc, frequency, is_frequency_dependent);
 }
 
 complex<double> fields::get_chi1inv(component c, direction d, const vec &loc, double frequency,
-                                    bool parallel) const {
+                                    bool parallel, bool *is_frequency_dependent) const {
   ivec ilocs[8];
   double w[8];
   complex<double> res(0.0, 0.0);
   gv.interpolate(c, loc, ilocs, w);
   for (int argh = 0; argh < 8 && w[argh] != 0; argh++)
-    res += w[argh] * get_chi1inv(c, d, ilocs[argh], frequency, false);
+    res += w[argh] * get_chi1inv(c, d, ilocs[argh], frequency, false, is_frequency_dependent);
   return parallel ? sum_to_all(res) : res;
 }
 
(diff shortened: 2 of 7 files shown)


stevengj:
> Also skips the phase-fixing pass for DiffractedPlanewave since |α|² is phase-invariant

People still might want the phase of α, e.g. in order to look at phase delay.

Luochenghuang:
> > Also skips the phase-fixing pass for DiffractedPlanewave since |α|² is phase-invariant
> 
> People still might want the phase of α, e.g. in order to look at phase delay.

@stevengj Good point. I removed `skip_phase_fix` to always run phase fix.

python/simulation.py
```diff
@@ -2897,17 +2902,21 @@ def run_k_point(self, t: float = None, k: Vector3Type = None):
...
 
         fmax = max(
-            s.src.frequency + 1 / s.src.width / 2
-            if isinstance(s.src, mp.GaussianSource)
-            else 0
+            (
+                s.src.frequency + 1 / s.src.width / 2
+                if isinstance(s.src, mp.GaussianSource)
+                else 0
+            )
```
stevengj, line 2919: Note: the changes to simulation.py are purely cosmetic changes, unrelated to this PR, suggested by the linter.

stevengj:
The problem here is that if the ε is frequency-dependent, the cache is no longer valid whenever the frequency changes and you'll at least need to call `maxwell_set_epsilon` again.  This requires two things:

1. Detect whether the eigenmode monitor involves dispersive materials.   One place to do this might be to let you pass an optional `bool *dispersive = NULL` flag to `get_chi1inv` that (if non-NULL) is set to `true` if the queried material is frequency dependent.  Then in `meep_mpb_eps` you would pass this, and track whether `true` was returned for any of the queried points.
2. Store this flag in the `dft_flux` object, and use it to determine whether to re-use the cache.

Luochenghuang:
Added dispersive material handling for the eigenmode cache. The eigenmode cache (`maxwell_data` reuse across `get_eigenmode_coefficients()` calls) assumed frequency-independent epsilon. For dispersive materials, the cached epsilon grid becomes stale when the eigenmode frequency changes. This commit detects dispersive media during epsilon grid construction by comparing `get_chi1inv(freq)` vs `get_chi1inv(freq=0)` for diagonal components at each grid point, short-circuiting after the first dispersive point is found. When the cache is reused at a different frequency on a dispersive monitor, it re-populates the epsilon grid on the existing `maxwell_data` (avoiding full reallocation). The detection only runs once on the first call (cache MISS) — the result is store in `dft_flux.eigenmode_cache_dispersive` and reused on all subsequent calls.

stevengj:
`get_chi1inv` knows whether it is frequency-dependent here: https://github.com/NanoComp/meep/blob/ceb2405b9c71d6fa70db4ed7350e2c79e98ea4d6/src/monitor.cpp#L330C1-L354C6

suggestion: pass an additional optional `bool *is_frequency_dependent = NULL` parameter through `get_chi1inv`.   If this is non-NULL, set it to `false` by default and to `true` if the frequency-dependent code is hit.

src/mpb.cpp
```diff
@@ -48,6 +48,10 @@ typedef struct {
   size_t ncache;  // allocated size of the cache
   size_t icache;  // current position in the cache
   bool use_cache; // whether we are using the cache
+
+  // Track whether any sampled point has frequency-dependent epsilon.
+  // Detected by comparing get_chi1inv(freq=0) vs get_chi1inv(freq=frequency).
+  bool found_dispersive;
```
stevengj, line 53: Needs to be updated to use `is_frequency_dependent`.

Note that after you've sampled all the points you will probably also need a `or_to_all` call to synchronize.

src/monitor.cpp
```diff
@@ -261,9 +264,11 @@ void matrix_invert(std::complex<double> (&Vinv)[9], std::complex<double> (&V)[9]
 }
 
 complex<double> structure_chunk::get_chi1inv_at_pt(component c, direction d, int idx,
-                                                   double frequency) const {
+                                                   double frequency,
+                                                   bool *is_frequency_dependent) const {
   complex<double> res(0.0, 0.0);
   if (is_mine()) {
+    if (is_frequency_dependent) *is_frequency_dependent = false;
```
stevengj: Maybe remove this line, and require the caller to initialize to `false`?  That way you won't need to call `||` on the result of several calls.

src/mpb.cpp
Luochenghuang: Oops, I missed that part. Fixed.

src/monitor.cpp
Luochenghuang: Great idea. Removed.

Luochenghuang:
Code has sinced been updated.

src/mpb.cpp
```diff
@@ -89,9 +92,14 @@ static int meep_mpb_eps(symmetric_matrix *eps, symmetric_matrix *eps_inv, mpb_re
...
-    cache[6 * i] = real(f->get_chi1inv(Ex, X, p, frequency, false));
-    cache[6 * i + 1] = real(f->get_chi1inv(Ey, Y, p, frequency, false));
-    cache[6 * i + 2] = real(f->get_chi1inv(Ez, Z, p, frequency, false));
+    // Pass &is_freq_dep to detect dispersive materials directly from get_chi1inv,
+    // which knows whether frequency-dependent susceptibilities were evaluated.
+    bool is_freq_dep = false;
+    cache[6 * i] = real(f->get_chi1inv(Ex, X, p, frequency, false, &is_freq_dep));
+    cache[6 * i + 1] = real(f->get_chi1inv(Ey, Y, p, frequency, false, &is_freq_dep));
+    cache[6 * i + 2] = real(f->get_chi1inv(Ez, Z, p, frequency, false, &is_freq_dep));
+    if (is_freq_dep) eps_data->found_dispersive = true;
```
stevengj:
```suggestion
    cache[6 * i] = real(f->get_chi1inv(Ex, X, p, frequency, false, &eps_data->found_dispersive));
    cache[6 * i + 1] = real(f->get_chi1inv(Ey, Y, p, frequency, false, &eps_data->found_dispersive));
    cache[6 * i + 2] = real(f->get_chi1inv(Ez, Z, p, frequency, false, &eps_data->found_dispersive));
```

src/mpb.cpp
```diff
@@ -487,6 +502,44 @@ void *fields::get_eigenmode(double frequency, direction d, const volume where, c
...
+      eps_data.cache = summed_cache;
+      eps_data.found_dispersive = or_to_all(eps_data.found_dispersive);
+
+      eps_data.use_cache = true;
+      eps_data.icache = 0;
+      set_maxwell_dielectric(mdata, mesh_size, R, G, NULL, meep_mpb_eps, &eps_data);
+      assert(eps_data.icache == eps_data.ncache);
+
+      free(eps_data.cache);
+      *cache_frequency = frequency;
```
stevengj, line 527: This code needs to be refactored to be unified with the other place we call `set_maxwell_dielectric`.  Either that, or can we just invalidate the cache here so that it gets automatically recomputed elsewhere?

stevengj (approved):

oskooi:
It looks like these changes are causing the address and undefined behavior sanitizers to [fail on the Github CI](https://github.com/NanoComp/meep/actions/runs/24561078320).

Luochenghuang:
> It looks like these changes are causing the address and undefined behavior sanitizers to [fail on the Github CI](https://github.com/NanoComp/meep/actions/runs/24561078320).

Thank you. Fixed on #3202.

PR #3168, opened 2026-03-12
Skip MPB for planewave eigenmodes in homogeneous media

Luochenghuang:
## Summary

When computing diffraction efficiencies via `DiffractedPlanewave` in `get_eigenmode_coefficients`, this PR auto-detects if the flux monitor is in a spatially homogeneous medium and, if so, bypasses MPB entirely by computing the planewave eigenmode fields analytically.

As discussed in #291, the propagating modes in a homogeneous medium are planewaves with analytically known k-vectors from the dispersion relation $k_\perp = \sqrt{n^2\omega^2 - k_\parallel^2}$. Computing the mode coefficients reduces to a discrete Fourier transform of the DFT-accumulated monitor fields at these known k-vectors, which is what the existing overlap integral machinery already does when `eigenmode_amplitude` returns the analytical $e^{i\mathbf{k}\cdot\mathbf{r}}$ planewave.

This eliminates creating `maxwell_data` (epsilon grid setup via `set_maxwell_dielectric`), calling `maxwell_set_planewave`/`maxwell_compute_h_from_H`/`maxwell_compute_d_from_H`/`maxwell_compute_e_from_d`, FFT-grid interpolation in eigenmode_amplitude, and group velocity computation via maxwell_ucross_op.

The existing overlap integral code (`get_mode_flux_overlap`, `get_mode_mode_overlap`) is reused unchanged. Only the mode field evaluation is replaced. This preserves correct handling of Yee grid staggering, forward/backward mode decomposition, and normalization.

## Implementation

- `is_homogeneous_on_monitor()`: samples $\varepsilon$ and $\mu$ at center + edges of the monitor plane to detect spatial uniformity
- `get_eigenmode_planewave()`: computes $\mathbf{k}$, group velocity, and E/H polarization vectors analytically following MPB's `maxwell_set_planewave` convention 
- `eigenmode_amplitude()`: fast path that returns $\mathbf{E}_0 / \mathbf{H}_0 \cdot e^{i\mathbf{k}\cdot\mathbf{r}}$ directly when `is_analytical` is set
- `get_eigenmode_coefficients()`: checks for homogeneous medium when `dp != NULL`; uses analytical path or falls back to MPB transparently

Handles edge cases: evanescent orders (returns `NULL`), grazing modes ($k_\perp = 0$), axis parallel to $\mathbf{k}$ (fallback perpendicular axis), and 2D simulations with out-of-plane $k$ (skips analytical path).

### Sinc correction for Yee grid consistency

The E-field is computed using a Yee-grid-corrected wavevector: each $k_i$ is replaced by $\frac{\sin(\pi k_i / a)}{\pi / a}$, matching the finite-difference curl operator on the Yee grid. Without this correction, the continuous-curl E-H relationship diverges from the FDTD fields for near-grazing modes where $k_\parallel$ approaches the Nyquist frequency. With it, the analytical planewave agrees with MPB's band-number eigensolver to within ${\sim}10^{-4}$ even for the most grazing propagating orders.

### special_kz handling

The analytical path is skipped for 2D simulations with non-zero out-of-plane $k$ (both `kz_2d="real/imag"` where $\beta \neq 0$ and `kz_2d="complex"` where $k_z \neq 0$), since the hybrid continuous/discrete curl in these modes makes the analytical E-H relationship inaccurate.

### DIFF
--- a/src/mpb.cpp
+++ b/src/mpb.cpp
@@ -276,7 +276,7 @@ complex<double> eigenmode_amplitude(void *vedata, const vec &p, component c) {
 
   /* define a macro to give us data(x,y,z) on the grid,
      in row-major order (the order used by MPB): */
-#define D(x, y, z) (data[(((x) * ny + (y)) * nz + (z)) * 3])
+#define D(x, y, z) (data[(((x)*ny + (y)) * nz + (z)) * 3])
   complex<mpb_real> ret;
   ret = (((D(x, y, z) * (1.0 - dx) + D(x2, y, z) * dx) * (1.0 - dy) +
           (D(x, y2, z) * (1.0 - dx) + D(x2, y2, z) * dx) * dy) *
--- a/src/mpb.cpp
+++ b/src/mpb.cpp
@@ -276,7 +276,7 @@ complex<double> eigenmode_amplitude(void *vedata, const vec &p, component c) {
 
   /* define a macro to give us data(x,y,z) on the grid,
      in row-major order (the order used by MPB): */
-#define D(x, y, z) (data[(((x)*ny + (y)) * nz + (z)) * 3])
+#define D(x, y, z) (data[(((x) * ny + (y)) * nz + (z)) * 3])
   complex<mpb_real> ret;
   ret = (((D(x, y, z) * (1.0 - dx) + D(x2, y, z) * dx) * (1.0 - dy) +
           (D(x, y2, z) * (1.0 - dx) + D(x2, y2, z) * dx) * dy) *
@@ -1089,12 +1089,24 @@ void *fields::get_eigenmode_planewave(double frequency, direction d, const volum
   // H₀ = s * shat + p * phat
   complex<double> s_amp = dp->get_s();
   complex<double> p_amp = dp->get_p();
-  complex<double> H0[3], E0[3];
+  complex<double> H0[3];
   for (int i = 0; i < 3; ++i)
     H0[i] = s_amp * shat[i] + p_amp * phat[i];
 
-  // E₀ = -(k × H₀) / (ω * ε) from Maxwell's equations
-  cross3c(E0, kcart, H0);
+  // E₀ = -(k_eff × H₀) / (ω * ε) from Maxwell's equations, using the
+  // Yee-grid-corrected wavevector k_eff.  On the Yee grid the finite-
+  // difference curl replaces each k_i with  sin(π k_i / a) / (π / a),
+  // i.e. the "sinc" correction.  This makes the analytical mode fields
+  // consistent with the FDTD discretization and is critical for near-
+  // grazing modes where k_tangential approaches the Nyquist frequency.
+  const double pi = 0.5 * TWOPI;
+  double kcart_eff[3];
+  for (int i = 0; i < 3; ++i) {
+    double arg = pi * kcart[i] / a; // π k_i Δ  where Δ = 1/a
+    kcart_eff[i] = (fabs(arg) < 1e-20) ? kcart[i] : sin(arg) * a / pi;
+  }
+  complex<double> E0[3];
+  cross3c(E0, kcart_eff, H0);
   for (int i = 0; i < 3; ++i)
     E0[i] = -E0[i] / (frequency * eps);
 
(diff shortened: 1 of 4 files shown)


stevengj:
I'd like to see some more fine-grained benchmarking to understand when MPB would be a bottleneck, especially with the diffracted-planewave feature where it just calls `maxwell_set_planewave` rather than doing a solve.

stevengj:
Regarding the performance issues: Reportedly the time is mostly in `set_maxwell_dielectric` (since the `eigensolve` is skipped when you have a diffractedplanewave object).  This could be fixed by:
1. PR #3167 will mostly eliminate the `set_maxwell_dielectric` cost in a homogeneous non-dispersive media (when you have lots of modes and/or frequencies)
2. We could add an accelerated API to MPB that lets you `set_maxwell_dielectric_constant` for a homogeneous medium.

That being said, this PR is still useful for improved *accuracy* by including the Yee-grid numerical dispersion in planewave modes.

src/mpb.cpp
```diff
@@ -166,7 +170,27 @@ static int pmod(int n, int modulus) {
 /*******************************************************************/
 complex<double> eigenmode_amplitude(void *vedata, const vec &p, component c) {
   eigenmode_data *edata = (eigenmode_data *)vedata;
-  if (!edata || !(edata->mdata)) meep::abort("%s:%i: internal error", __FILE__, __LINE__);
+  if (!edata) meep::abort("%s:%i: internal error", __FILE__, __LINE__);
+
+  // Fast path for analytical planewave in homogeneous medium: skip MPB interpolation
```
stevengj, line 175: This is only for homogeneous isotropic dielectric media. Be careful that we should still call MPB for the anisotropic case.

src/mpb.cpp
```diff
@@ -913,6 +942,208 @@ void fields::add_eigenmode_source(component c0, const src_time &src, direction d
...
+/* Bypasses MPB entirely by computing E and H fields directly  */
+/* from the dispersion relation and polarization vectors.      */
+/***************************************************************/
+static void cross3(double out[3], const double a[3], const double b[3]) {
+  out[0] = a[1] * b[2] - a[2] * b[1];
+  out[1] = a[2] * b[0] - a[0] * b[2];
+  out[2] = a[0] * b[1] - a[1] * b[0];
+}
+
+static void cross3c(complex<double> out[3], const double a[3], const complex<double> b[3]) {
```
stevengj, line 956:
```suggestion
static void cross3(complex<double> out[3], const double a[3], const complex<double> b[3]) {
```
Use C++ static overloading?

src/mpb.cpp
```diff
@@ -913,6 +942,208 @@ void fields::add_eigenmode_source(component c0, const src_time &src, direction d
...
+  out[0] = a[1] * b[2] - a[2] * b[1];
+  out[1] = a[2] * b[0] - a[0] * b[2];
+  out[2] = a[0] * b[1] - a[1] * b[0];
+}
+
+static double norm3(const double v[3]) { return sqrt(v[0] * v[0] + v[1] * v[1] + v[2] * v[2]); }
+
+/* Check whether the medium is spatially homogeneous on the flux monitor plane
+   by sampling epsilon and mu at the center and edges. Returns true if all
+   samples match within a relative tolerance. */
```
stevengj, line 966: This seems a little risky?  Can't we assume it's homogeneous if the user passed a `diffractedplanewave`?

src/mpb.cpp
```diff
@@ -913,6 +942,208 @@ void fields::add_eigenmode_source(component c0, const src_time &src, direction d
...
+  complex<double> H0[3];
+  for (int i = 0; i < 3; ++i)
+    H0[i] = s_amp * shat[i] + p_amp * phat[i];
+
+  // E₀ = -(k_eff × H₀) / (ω * ε) from Maxwell's equations, using the
+  // Yee-grid-corrected wavevector k_eff.  On the Yee grid the finite-
+  // difference curl replaces each k_i with  sin(π k_i / a) / (π / a),
+  // i.e. the "sinc" correction.  This makes the analytical mode fields
+  // consistent with the FDTD discretization and is critical for near-
+  // grazing modes where k_tangential approaches the Nyquist frequency.
```
stevengj, line 1101: Couldn't we get this benefit by passing `k_eff` to `update_maxwell_data_k` when doing `maxwell_data_set_planewave` (or whatever it's called)?

PR #3169, opened 2026-03-13
Tutorial for computing the emission from a cleaved stack using a 2D simulation

### LINKED ISSUE #3128: Tutorial for computing the edge emission of a multilayer stack using a 2D simulation
[Edge-emitting semiconductor lasers](https://www.optica-opn.org/home/articles/volume_30/february_2019/features/semiconductor_lasers_for_3-d_sensing/) are generally based on a multilayer stack where the wafer has been cleaved to expose an edge facet. It would be useful to provide a tutorial demonstration of the computation of the **extraction efficiency** and **radiation pattern** of such a structure. This would be based on a 2D cell in $xy$ and Brillouin-zone integration in $z$ (i.e. the out-of-plane direction). A schematic of the simulation setup is shown below. The active region forms the cavity and consists of an extended (spatially incoherent) source such as a quantum well.

There are three things to note regarding this calculation:

1. A demonstration of Brillouin-zone integration for the case of a 1D cell is provided in [Tutorial/Dipole Emission of a Light Emitting Diode as a Multilayer Stack](https://meep.readthedocs.io/en/latest/Python_Tutorials/Custom_Source/#dipole-emission-of-a-light-emitting-diode-as-a-multilayer-stack). However, in this example, the radiation pattern is computed using a near-to-far field transformation (rather than directly from the DFT flux).
2. The emission from the active region must be computed by combining the response from a [collection of dipoles](https://meep.readthedocs.io/en/latest/Python_Tutorials/Custom_Source/#stochastic-dipole-emission-in-light-emitting-diodes) for which the emission is non zero. The dipole emission is expected to decrease with increasing distance from the edge facet.
3. The radiation pattern $P(\theta, \phi)$ is computed for $\theta$ in [-π/2, π/2] (where $\theta$ = -π/2 is the $-y$ direction and $\theta$ = π/2 is $+y$) and $\phi$ in [0, π/2] (where $\phi$ = 0 is the $xy$ plane and $\phi$ = π/2 is the $yz$ plane).

Image

oskooi:
Closes #3128.

### DIFF
--- a/doc/docs/Python_Tutorials/Custom_Source.md
+++ b/doc/docs/Python_Tutorials/Custom_Source.md
@@ -662,7 +662,7 @@ The emission from a cleaved facet of a multilayer stack (relevant for [edge-emit
 
 There are three important things to note regarding the setup of this calculation:
 
-- A 2D simulation (in $xy$) with out-of-plane wavevector ($k_z$) requires specifying the parameter [`kz_2d`](../2d_Cell_Special_kz.md) of the `Simulation` constructor to `"real/imag"` and using the default value for `force_complex_fields` of `False`. Alternatively, `kz_2d` can be set to "complex" (default value) or "3d" but this requires specifying `force_complex_fields` to `True` (which doubles the storage required for the fields).
+- A 2D simulation (in $xy$) with out-of-plane wavevector ($k_z$) requires specifying the parameter [`kz_2d`](../2d_Cell_Special_kz.md) of the `Simulation` constructor to `"real/imag"` and using the default value for `force_complex_fields` of `False`. Alternatively, `kz_2d` can be set to "complex" (default value) or "3d" but this requires specifying `force_complex_fields` to `True` (to ensure that $k_z=0$ is handled consistently with $k_z \ne 0$, since the latter necessarily use complex fields).
 - The Fourier-series expansion requires only *non-negative* wavevectors $k_z$ since the flux is the same for $\pm k_z$.
 - The criteria for truncating the Fourier series is determined by when the *total* power (computed using the local density of states) has fallen below a given threshold (which needs to be small enough to ensure the results are converged).
 


doc/docs/Python_Tutorials/Custom_Source.md
```diff
@@ -652,3 +652,26 @@ As a final note, computing the [extraction efficiency of an LED](Local_Density_o
...
+
+There are three important things to note regarding the setup of this calculation:
+
+- A 2D simulation (in $xy$) with out-of-plane wavevector ($k_z$) requires specifying the parameters [`kz_2d`](../2d_Cell_Special_kz.md) and `force_complex_fields` of the `Simulation` constructor to `"real/imag"` and `False`, respectively.
+- The Fourier-series expansion requires only *non-negative* wavevectors $k_z$ since the flux is the same for $\pm k_z$.
+- The criteria for truncating the Fourier series is determined by when the flux radiated by the line current (computed using the local density of states) has fallen below a given threshold (which needs to be small enough to ensure the results are converged).
+
+A plot of the flux radiated by the line current and into air versus $k_z$ is shown below for an $E_x$ dipole positioned 0.32 μm from the edge facet. When $k_z$ is sufficiently large, the flux radiated by the line current is negligible. The extraction efficiency computed from this data is 24.12%. This example involved 76 2D simulations.
+
+![](../images/edge_emitter_flux_vs_kz.png#center)
```
stevengj, line 671: It might be nice to label the axis as $k_z / \omega$ to make it clear that 1.0 corresponds to the edge of the light cone, hence the kink

stevengj: and maybe label the lines as "total power" and "extracted power"

doc/docs/Python_Tutorials/Custom_Source.md
```diff
@@ -652,3 +652,26 @@ As a final note, computing the [extraction efficiency of an LED](Local_Density_o
...
+
+![](../images/edge_emitter_layout_2D.png#center)
+
+There are three important things to note regarding the setup of this calculation:
+
+- A 2D simulation (in $xy$) with out-of-plane wavevector ($k_z$) requires specifying the parameters [`kz_2d`](../2d_Cell_Special_kz.md) and `force_complex_fields` of the `Simulation` constructor to `"real/imag"` and `False`, respectively.
+- The Fourier-series expansion requires only *non-negative* wavevectors $k_z$ since the flux is the same for $\pm k_z$.
+- The criteria for truncating the Fourier series is determined by when the flux radiated by the line current (computed using the local density of states) has fallen below a given threshold (which needs to be small enough to ensure the results are converged).
```
stevengj: You might want to reference my comments
> The real/imag trick is essentially equivalent to real fields, i.e. it is half the source, since we only store the real part of one polarization and only the imaginary part of another polarization.

If you use the real/imag trick then `force_complex_fields` is ignored for $k_z \ne 0$ (since the two are incompatible … maybe we should throw an error, actually?).   But for $k_z=0$ then real/imag is ignored because it is irrelevant there and then suddenly `force_complex_fields` becomes relevant.

doc/docs/Python_Tutorials/Custom_Source.md
oskooi: Done.

doc/docs/Python_Tutorials/Custom_Source.md
oskooi: Additional details regarding use of the `kz_2d` feature have been added to the text.

#3170 will be addressed separately.

doc/docs/Python_Tutorials/Custom_Source.md
```diff
@@ -652,3 +652,26 @@ As a final note, computing the [extraction efficiency of an LED](Local_Density_o
...
+
+The emission from a cleaved facet of a multilayer stack (relevant for [edge-emitting semiconductor lasers](https://www.optica-opn.org/home/articles/volume_30/february_2019/features/semiconductor_lasers_for_3-d_sensing/)) can be modeled using a 2D simulation via a Fourier-series expansion of line-current sources. A schematic of the device geometry and simulation layout is shown below. This example involves computing the extraction efficiency of a dipole (wavelength of 1 μm) within an active region ($n$ = 2.3) surrounded by cladding layers ($n$ = 1.5) and placed on a substrate ($n$ = 3.2).
+
+![](../images/edge_emitter_layout_2D.png#center)
+
+There are three important things to note regarding the setup of this calculation:
+
+- A 2D simulation (in $xy$) with out-of-plane wavevector ($k_z$) requires specifying the parameter [`kz_2d`](../2d_Cell_Special_kz.md) of the `Simulation` constructor to `"real/imag"` and using the default value for `force_complex_fields` of `False`. Alternatively, `kz_2d` can be set to "complex" (default value) or "3d" but this requires specifying `force_complex_fields` to `True` (which doubles the storage required for the fields).
```
stevengj:
```suggestion
- A 2D simulation (in $xy$) with out-of-plane wavevector ($k_z$) requires specifying the parameter [`kz_2d`](../2d_Cell_Special_kz.md) of the `Simulation` constructor to `"real/imag"` and using the default value for `force_complex_fields` of `False`. Alternatively, `kz_2d` can be set to "complex" (default value) or "3d" but this requires specifying `force_complex_fields` to `True` (to ensure that $k_z=0$ is handled consistently with $k_z \ne 0$, since the latter necessarily use complex fields).
```

PR #3181, opened 2026-03-26
Exclude tests involving adjoint solver from `make check`

oskooi:
The four tests involving the adjoint solver (particularly `test_adjoint_solver.py`) take an unusually long time to complete which tends to slow down the CI. This PR modifies `make check` to *exclude* these tests and instead creates a `make check-adjoint` whicn can be used to run *all* tests (including those involving the adjoint solver).

### DIFF
--- a/.github/workflows/build-ci-adjoint.yml
+++ b/.github/workflows/build-ci-adjoint.yml
@@ -1,0 +1,116 @@
+name: run adjoint tests
+
+on:
+  pull_request:
+  push:
+  schedule:
+    - cron: 0 22 * * 0 # run at 10 PM UTC every Sunday
+  workflow_dispatch:
+
+jobs:
+  build:
+    name: "serial double-precision adjoint tests"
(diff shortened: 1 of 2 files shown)


stevengj:
Is there a way to conditionally run this?

And/or run it periodically or upon a trigger?

stevengj:
Maybe add a workflow for a new action to run `make check-adjoint` periodically (e.g. once a week), which we can also trigger manually.

oskooi:
A separate GitHub Actions workflow for `make check-adjoint` (adjoint tests only) has been [added](https://github.com/oskooi/meep/actions/runs/25243672874/job/74023985438) which runs once every week (Sundays at 10 pm). `make check` involves only the non-adjoint tests.

stevengj:
It looks like there is a way to set up the adjoint tests to run on PRs on manual request by triggering by a label?

https://stackoverflow.com/questions/62325286/run-github-actions-when-pull-requests-have-a-specific-label

PR #3188, opened 2026-04-02
Fix bug in `fields::_require_component` when allocating new fields

oskooi:
**Bug:** `fields::_require_component()` in [`src/fields.cpp:553`](https://github.com/NanoComp/meep/blob/d4b3822e862e8cfbd4dcaba01681e148c13fc54b/src/fields.cpp#L553-L574) allocates new field arrays when a source is added (via `alloc_f`), and sets `chunk_connections_valid = false`, but does not set `changed_materials = true`. On the next `step()` call, `step_db()` lazily allocates PML auxiliary arrays (`f_u`) for the newly-created field components, which requires `changed_materials` to be `true` per the [assertion](https://github.com/NanoComp/meep/blob/d4b3822e862e8cfbd4dcaba01681e148c13fc54b/src/step_db.cpp#L40).

**Fix:** (`src/fields.cpp:568-573`): Set `changed_materials = true` when `_require_component` allocates new fields.

### DIFF
--- a/src/fields.cpp
+++ b/src/fields.cpp
@@ -570,6 +570,7 @@ void fields::_require_component(component c, bool aniso2d) {
     // we will eventually call sync_chunk_connections(), in either require_component(c)
     // or require_components(), to synchronize this across processes:
     chunk_connections_valid = false;
+    changed_materials = true;
   }
 }
 
--- a/src/fields.cpp
+++ b/src/fields.cpp
@@ -566,8 +566,7 @@ void fields::_require_component(component c, bool aniso2d) {
 
   if (need_to_reconnect) {
     figure_out_step_plan();
-    // we will eventually call sync_chunk_connections(), in either require_component(c)
-    // or require_components(), to synchronize this across processes:
+    // we will eventually call sync_chunk_connections() to synchronize this across processes:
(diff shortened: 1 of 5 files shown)


stevengj:
This could be triggered if you do some timestepping, and *then* add a source (triggering field allocation).   I think we don't see this bug currently because we virtually always create fields at the beginning and then timestep.

src/fields.cpp
```diff
@@ -570,6 +570,7 @@ void fields::_require_component(component c, bool aniso2d) {
     // we will eventually call sync_chunk_connections(), in either require_component(c)
     // or require_components(), to synchronize this across processes:
     chunk_connections_valid = false;
+    changed_materials = true;
```
stevengj: This is is the wrong place to set this, since `changed_materials` is a "collective" flag (it should be set on *all* processes) whereas this `if` statement is per-process.

stevengj:
To do: test case that actually triggers the bug without this patch (e.g. editing one of the C++ tests).

oskooi:
> To do: test case that actually triggers the bug without this patch (e.g. editing one of the C++ tests).

A unit test has been added which fails on master (2d14125).

PR #3197, opened 2026-04-07
Mesh geometry import

### LINKED ISSUE #138: STL import
I have implemented a method to convert the STL file to Yee gird and want to import the grid file to meep.
The function I need is like read_epsilon_file in the meepgeom.cpp. Is there any instruction。

Thank you indeed。

Luochenghuang:
## Summary

Adds Python support for the new `MESH` geometric object type from [Luochenghuang/libctl#mesh-geometry-import](https://github.com/Luochenghuang/libctl/tree/mesh-geometry-import). This enables importing arbitrary triangulated 3D surfaces (e.g., from STL/OBJ files) directly into meep simulations.

Depends on libctl with MESH support (NanoComp/libctl#74).
Closes #138.

## Changes

**`python/geom.py`** — `Mesh` class and file parsers:
- `Mesh(vertices, triangles, center=None, material=...)` — triangulated surface geometry
- `Mesh.from_stl(filename, ...)` — binary and ASCII STL import with vertex welding
- `Mesh.from_obj(filename, ...)` — Wavefront OBJ import with quad triangulation
- No external dependencies (parsers are pure Python using `struct` and `re`)

**`python/typemap_utils.cpp`** — C++ binding:
- `pymesh_to_mesh` — extracts vertices and triangles from Python, calls `make_mesh_with_center`
- `"Mesh"` case added to `py_gobj_to_gobj` dispatch

## Usage

```python
import meep as mp

# From STL file
teapot = mp.Mesh.from_stl("model.stl", scale=0.1,
                           material=mp.Medium(epsilon=12))

# From vertices and triangles directly
cube = mp.Mesh(
    vertices=[(-0.5,-0.5,-0.5), (0.5,-0.5,-0.5), ...],
    triangles=[(0,2,1), (0,3,2), ...],
    material=mp.Medium(epsilon=12),
)

sim = mp.Simulation(cell_size=cell, geometry=[teapot], resolution=10)
```

Subpixel smoothing, epsilon grid queries, and all other meep features work automatically.

## Test plan

5 integration tests in `python/tests/test_mesh.py`:

| Test | What it verifies |
|---|---|
| `test_stl_import` | Binary STL parse + vertex welding (8 verts, 12 faces) |
| `test_point_in_mesh` | Known points inside/outside a cube mesh |
| `test_volume` | Fill fraction from grid sampling matches analytic |
| `test_eps_averaging` | Smoothing produces intermediate epsilon at interfaces |
| `test_mesh_vs_block` | Cube mesh epsilon grid matches native Block exactly |


python/geom.py
```diff
@@ -1343,6 +1343,171 @@ def __init__(
...
+        text = (header + data).decode("ascii", errors="ignore")
+        import re
+
+        for m in re.finditer(
+            r"vertex\s+([-+eE.\d]+)\s+([-+eE.\d]+)\s+([-+eE.\d]+)", text
+        ):
+            x, y, z = float(m.group(1)), float(m.group(2)), float(m.group(3))
+            raw_verts.append((x * scale, y * scale, z * scale))
+
+    # Vertex welding: merge vertices within tolerance using a spatial hash
```
stevengj: Why is a tolerance needed?   Does STL actually store slightly different per-face vertices?  Shouldn't they be exactly equal?

Luochenghuang: In some edge cases, the CAD program _can_ produce slightly different vertices during tessellation while exporting to STL. But this is of course not the expected behavior of the STL file. I suppose the user can use a different library, such as [pymeshfix](https://github.com/pyvista/pymeshfix), to fix the STL file in that case. I will make the tolerance 0 now.

stevengj:
If we still want to support compilation with older libctl versions, then we would need to:

1. add a `configure.ac` check for `make_mesh_with_center` here: https://github.com/NanoComp/meep/blob/52c6b44c8a0e722a64ae1860d814312767bb7f5c/configure.ac#L482-L483
2. guard the C++ code in the `.i` and `.cpp` code with the resulting `#ifdef`, making the routines throw an error in the `#else`.

python/geom.py
```diff
@@ -1343,6 +1343,162 @@ def __init__(
...
+    def from_obj(cls, filename, material=None, center=None, scale=1.0):
+        """Load mesh from Wavefront OBJ file."""
+        vertices, triangles = _parse_obj(filename, scale)
+        kwargs = {}
+        if material is not None:
+            kwargs["material"] = material
+        return cls(vertices=vertices, triangles=triangles, center=center, **kwargs)
+
+
+def _parse_stl(filename, scale=1.0):
```
stevengj: Should we really be including our own STL parser as opposed to just telling the user to call something like `numpy-stl`?

(Similarly for OBJ files below or VRML or whatever.)

Luochenghuang:
Sure, we can certainly use `trimesh` as the parsing library, which also comes with the added benefit of repairing the mesh, ensuring it's watertight with correct winding and compatible with `libctl`. If we go with it, then we can refactor out quite a bit of python code in this branch.

stevengj:
Can there be documentation with an example?

Luochenghuang:
Sure. Documentation is added.

stevengj (approved):

PR #3198, opened 2026-04-11
Update Autoconf macros

oskooi:
Updates to eight Autoconf [macros](https://www.gnu.org/software/autoconf-archive/The-Macros.html) which were obtained from the [archive](https://github.com/autoconf-archive/autoconf-archive) (Github repository):

- `ax_cc_maxopt.m4` (previously `ax_cxx_maxopt.m4`)
- `ax_check_compile_flag.m4` (previously `ax_check_compiler_flags.m4`)
- `ax_compiler_vendor.m4`
- `ax_gcc_archflag.m4`
- `ax_gcc_x86_cpuid.m4`
- `ax_blas.m4`
- `ax_cxx_compile_stdcxx.m4`
- `ax_openmp.m4`


m4/ax_cc_maxopt.m4
```diff
@@ -0,0 +1,196 @@
+# ===========================================================================
```
stevengj, line 1: Does this actually help performance these days, versus just `-O3`?   A lot of these flag choices are pretty old.

PR #3202, opened 2026-04-17
Fix no-MPB stub for `get_eigenmode` to match updated declaration

Luochenghuang:
PR #3167 added `cache_dispersive` and `cache_frequency` parameters to `fields::get_eigenmode` in `meep.hpp`, but only updated the no-MPB stub for `get_eigenmode_coefficients`. The get_eigenmode stub was missed. This causes a compile error when building without MPB (e.g., the ASAN and UBSAN CI jobs):                      
```                                                                                           
mpb.cpp:1075:15: error: out-of-line definition of 'get_eigenmode' does not match any declaration in 'meep::fields'
 1075 | void *fields::get_eigenmode(double frequency, direction d, const volume where, const volume eig_vol,
      |               ^~~~~~~~~~~~~
```
This PR adds the two missing parameters to the no-MPB stub signature.
This PR was tested against the sanitizers CI [here](https://github.com/Luochenghuang/meep/actions/runs/24584241142/job/71889652000).


Luochenghuang:
Tagging @stevengj.

PR #3205, opened 2026-04-24
Include `eps_parameters` for plot3D

drinwater:
Currently plot3D cannot plot materials with frequency-dependent epsilon. This PR fixes that by including eps_parameters kwarg in plot3D

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -4918,7 +4918,7 @@ class Simulation:
         return vis.plot_fields(self, **kwargs)
 
     def plot3D(
-        self, save_to_image: bool = False, image_name: str = "sim.png", **kwargs
+        self, save_to_image: bool = False, image_name: str = "sim.png", eps_frequency: float = 0., resolution: float | None = None, **kwargs
     ):
         """
         Uses vispy to render a 3D scene of the simulation object. The simulation object must be 3D.
@@ -4932,17 +4932,16 @@ class Simulation:
             scale_factor: float, camera zoom factor
             azimuth: float, azimuthal angle in degrees
             elevation: float, elevation angle in degrees
-            eps_parameters: Parameters to plot epsilon:
-                frequency: for materials with a [frequency-dependent
-                            permittivity](Materials.md#material-dispersion) $\\varepsilon(f)$, specifies the
-                            frequency $f$ (in Meep units) of the real part of the permittivity to use in the
-                            plot. Defaults to 0.
-                resolution: the resolution of the $\\varepsilon$ grid. Defaults to the
-                            `resolution` of the `Simulation` object.
+            eps_frequency: for materials with a [frequency-dependent
+                        permittivity](Materials.md#material-dispersion) $\\varepsilon(f)$, specifies the
+                        frequency $f$ (in Meep units) of the real part of the permittivity to use in the
+                        plot. Defaults to 0.
+            resolution: the resolution used for plotting. Defaults to the
+                        `resolution` of the `Simulation` object.
         """
         import meep.visualization as vis
 
-        return vis.plot3D(self, save_to_image, image_name, **kwargs)
+        return vis.plot3D(self, save_to_image, image_name, eps_frequency, resolution, **kwargs)
 
     def visualize_chunks(self):
         """
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -4932,6 +4932,13 @@ class Simulation:
             scale_factor: float, camera zoom factor
             azimuth: float, azimuthal angle in degrees
             elevation: float, elevation angle in degrees
+            eps_parameters: Parameters to plot epsilon:
+                frequency: for materials with a [frequency-dependent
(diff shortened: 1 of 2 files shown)


python/simulation.py
```diff
@@ -4966,6 +4932,13 @@ def plot3D(
             scale_factor: float, camera zoom factor
             azimuth: float, azimuthal angle in degrees
             elevation: float, elevation angle in degrees
+            eps_parameters: Parameters to plot epsilon:
+                frequency: for materials with a [frequency-dependent
+                            permittivity](Materials.md#material-dispersion) $\\varepsilon(f)$, specifies the
+                            frequency $f$ (in Meep units) of the real part of the permittivity to use in the
+                            plot. Defaults to 0.
+                resolution: the resolution of the $\\varepsilon$ grid. Defaults to the
```
stevengj: Couldn't we just have two top-level arguments `eps_frequency` and `resolution`, since `resolution` could in principle apply to any plot (since Meep can linearly interpolate for you), not just epsilon?

drinwater: I have just updated this, could you please take a look?

stevengj (dismissed):

python/visualization.py
```diff
@@ -1123,12 +1124,13 @@ def plot3D(sim, save_to_image: bool = False, image_name: str = "sim.png", **kwar
     ztics = np.linspace(zmin, zmax, Nz)
 
     # Get eps for geometry
-    eps_data = np.round(np.real(sim.get_epsilon_grid(xtics, ytics, ztics)), 2)
 
-    unique = np.unique(np.abs(eps_data)).tolist()
+    eps_data = np.round(np.real(sim.get_epsilon_grid(xtics, ytics, ztics, eos_frequency)), 2)
```
stevengj:
```suggestion
    eps_data = np.round(np.real(sim.get_epsilon_grid(xtics, ytics, ztics, eps_frequency)), 2)
```

python/visualization.py
```diff
@@ -1123,12 +1124,13 @@ def plot3D(sim, save_to_image: bool = False, image_name: str = "sim.png", **kwar
...
-    eps_data = np.round(np.real(sim.get_epsilon_grid(xtics, ytics, ztics)), 2)
 
-    unique = np.unique(np.abs(eps_data)).tolist()
+    eps_data = np.round(np.real(sim.get_epsilon_grid(xtics, ytics, ztics, eps_frequency)), 2)
+
+    unique = np.unique((eps_data)).tolist()
 
     # Remove background material
-    unique.remove(np.round(np.abs(np.asarray(sim.default_material.epsilon_diag)), 2)[0])
+    unique.remove(np.round((np.asarray(sim.default_material.epsilon_diag)), 2)[0])
```
stevengj:
```suggestion
    unique.remove(np.round(np.asarray(sim.default_material.epsilon_diag), 2)[0])
```

python/visualization.py
```diff
@@ -1123,12 +1124,13 @@ def plot3D(sim, save_to_image: bool = False, image_name: str = "sim.png", **kwar
     ztics = np.linspace(zmin, zmax, Nz)
 
     # Get eps for geometry
-    eps_data = np.round(np.real(sim.get_epsilon_grid(xtics, ytics, ztics)), 2)
 
-    unique = np.unique(np.abs(eps_data)).tolist()
+    eps_data = np.round(np.real(sim.get_epsilon_grid(xtics, ytics, ztics, eps_frequency)), 2)
```
stevengj: Not sure why we are rounding here?

python/visualization.py
```diff
@@ -1123,12 +1124,13 @@ def plot3D(sim, save_to_image: bool = False, image_name: str = "sim.png", **kwar
...
+
+    unique = np.unique((eps_data)).tolist()
```
stevengj:
```suggestion
    unique = np.unique(eps_data).tolist()
```

python/visualization.py
drinwater: Previous versions of vispy used to throw an error when not rounded for some reason back when plot3D was first added. I will remove this now since that issue seems to be fixed.

python/visualization.py
```diff
@@ -1123,12 +1124,13 @@ def plot3D(sim, save_to_image: bool = False, image_name: str = "sim.png", **kwar
...
     # Get eps for geometry
-    eps_data = np.round(np.real(sim.get_epsilon_grid(xtics, ytics, ztics)), 2)
 
-    unique = np.unique(np.abs(eps_data)).tolist()
+    eps_data = np.real(sim.get_epsilon_grid(xtics, ytics, ztics, eps_frequency))
+
+    unique = np.unique(eps_data).tolist()
 
     # Remove background material
-    unique.remove(np.round(np.abs(np.asarray(sim.default_material.epsilon_diag)), 2)[0])
```
stevengj, line 1131: @smartalecH, do you remember why `round` was being used here?

PR #3220, opened 2026-05-22
Build optimization for Python/SWIG interface

oskooi:
The SWIG-generated Python interface is the slowest part of the Meep build. SWIG produces monolithic C++ files (typically 50K-100K+ lines) that are expensive to compile, and the current build configuration uses suboptimal SWIG flags and enforces unnecessary serial ordering across directories. The goal is to reduce wall-clock build time via changes to `configure.ac` and `python/Makefile.am`.

**1. Reduce optimization level for SWIG-generated C++ compilation**

The SWIG glue code is not performance-critical – hot loops live in `libmeep.la`. Compiling 50K+ lines at `-O2`/`-O3` (set by `AX_CXX_MAXOPT`) wastes significant time on optimizations (vectorization, inlining) that don't benefit wrapper code.

**2. Fix SWIG flags in both invocations**

These flags primarily improve runtime call performance (10-30% for call-heavy code) but also modestly reduce generated `.cxx` file size, shaving a few second off compilation.

**3. Reorder `SUBDIRS` to unblock parallel builds**

`Makefile.am` (lines 3-18)

Currently: `src` -> `tests` -> `[libpympb]` -> `python`. The `tests` directory (17 C++ test programs) blocks `libpympb` from starting even though `libpympb` only depends on `src/libmeep.la`.

**4. Add `ccache` support to `configure.ac`**

Add an `--enable-ccache` option that prepends `ccache` to `$CC` and `$CXX`. Must go after `CXX=$MPICXX` assignment so it wraps the final compiler. Default to `auto` (use if found, skip silently otherwise).

For iterative development, this is the highest-impact single change: compiling the 50K+ line `meep-python.cxx` goes from 30-60s to ~1s on cache hit.

**5. Suppress warnings on SWIG-generate code**

SWIG-generated code produces many warnings that cannot be fixed. This is a developer-experience improvement rather than a build-time improvement.

### DIFF
--- a/python/Makefile.am
+++ b/python/Makefile.am
@@ -123,10 +123,13 @@ AM_CPPFLAGS = -I$(top_srcdir)/src               \
 
 LIBMEEP = $(top_builddir)/src/libmeep.la
 
+SWIG_WARN_FLAGS = -Wno-unused-function -Wno-unused-parameter -Wno-missing-field-initializers
+
 _meep_la_SOURCES = meep-python.cxx
 _meep_la_LIBADD = $(LIBMEEP) $(PYTHON_LIBS)
 _meep_la_LDFLAGS = -module -version-info @SHARED_VERSION_INFO@
-_meep_la_CPPFLAGS = $(PYTHON_INCLUDES) $(AM_CPPFLAGS)
+_meep_la_CPPFLAGS = $(PYTHON_INCLUDES) $(AM_CPPFLAGS) $(SWIG_WARN_FLAGS)
+_meep_la_CXXFLAGS = -O1
 
 pkgpyexec_LTLIBRARIES = _meep.la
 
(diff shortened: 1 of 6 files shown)


python/Makefile.am
```diff
@@ -179,12 +183,12 @@ SWIG_MEEP_FLAGS = -threads
 endif # MEEP_SWIG_PYTHON_THREADS
 
 meep-python.cxx: $(MEEP_SWIG_SRC) $(HPPFILES)
-	$(SWIG) -Wextra $(SWIG_MEEP_FLAGS) $(AM_CPPFLAGS) -outdir $(builddir) -c++ -nofastunpack -python -o $@ $(srcdir)/meep.i
+	$(SWIG) -Wextra $(SWIG_MEEP_FLAGS) $(AM_CPPFLAGS) -outdir $(builddir) -c++ -O -fastproxy -python -o $@ $(srcdir)/meep.i
```
stevengj: This change seems to be cause compilation errors?

oskooi: Restored `-nofastunpack`. 

It turns out Meep's custom typemaps are incompatible with SWIG's fast unpacking code generation – the `-nofastunpack` flag was there intentionally, not by oversight.

Issue #2463, opened 2023-04-06
asymptotic near2far API


stevengj:
For computing radiation patterns, it would be nice to simplify the calculation of the fields arbitrarily far away.

In particular, the fields arbitrarily far away are proportional exp(ikr)/r (in 3d), so it would be nice to provide an API that computes the fields × r × exp(-ikr) in the limit r → ∞, given only the direction.  (Similarly in 2d by with √r instead of r.)

Basically, the implementation is the same as our existing far-field calculation, except with the Green's function replaced by the asymptotic form (which is *much* simpler).  Indeed, we could even probably do this analytically in cylindrical coordinates (rather than numerically integrating over φ.)

stevengj:
(It will be easy to test.  Just compare it to the current near2far calculation — as r increases, the two should match, with the differences going as 1/r).

PR #3223, opened 2026-06-01
Add support for far-field approximation of the 3D Green's function

### LINKED ISSUE #2269: Simplified near2far for actual far field
Currently, the `Near2Far` features use the full Green's function to project an equivalent source into a user-specified point in the far field. This approach is very _flexible_, but in some cases overkill. For example, we often wish to view the far-field response "infinitely far away", where the Green's function convolution integral [simplifies to a simple Fourier transform](https://doi.org/10.1063/1.335709).

It would be nice to add a feature that does precisely this. It's significantly cheaper than the "brute-force" approach. (of course, one could do this "by hand" using the python API... but a dedicated routine would be much appreciated).

oskooi:
Closes #2269, #2463.

**What the exact `green3d` computes**

The existing function `green3d` (`near2far.cpp:190-214`) evaluates the full free-space Green's function for a point-dipole source at `x0` observed at `x`. It computes three distance-dependent terms:

- **term 1:** `1 - 1/(ikr) + 1/(ikr)²` — radiating (1/r), intermediate (1/r²), and near-field (1/r³) contributions
- **term 2:** `(-1 + 3/(ikr) - 3/(ikr)²) · (p·r̂)` — radial component with all three ranges
- **term 3:** `1 - 1/(ikr)` — curl term with radiating and intermediate contributions

Each cell requires computing `r = |x - x0|`, `r̂ = (x-x0)/r`, `exp(ikr)`, and the complex divisions for these terms — all of which vary per source-observer pair.

**What the new `green3d_farfield` does differently**

The far-field approximation (`near2far.cpp:136-182)` exploits two simplifications valid when `R = |x| >> |x0|` and `R >> λ`:

**1. Geometric approximation**
- `r ≈ R - x̂·x0` (first-order Taylor expansion of distance)
- `r̂ ≈ x̂` (all source points subtend negligible angle)
- `1/r ≈ 1/R` (amplitude is constant across source points)
- The only per-source-point variation is in the phase: `exp(ikr) ≈ exp(ikR) · exp(ik x̂·x0)`

**2. Dropping near-field terms:** With 1/(ikr) ➔ 0, the terms simplify to `term ➔ 1`, `term2 ➔ -p·x̂`, `term3 ➔ 1`. The resulting fields are purely transverse:
- E is the transverse projection of the source dipole: `t = p - (p·x̂)x̂`
- H is `x̂ ⨯ p / Z` (or vice versa for magnetic sources)

The function signature reflects these precomputations — instead of taking the far-field point `x`, it takes `xhat` (unit vector), `k`, `impedance`, and `expfac_base = k·n/(4πR) · exp(i(kR + π/2))`, all computed once per far-field point and reused across every source point on the near-field surface. Per source point, the work reduces to:
1. One dot product `x̂·x0` (3 multiplies)
2. One `std::polar`call for the phase `exp(ik·(...))`
3. One complex multiply to get `expfac`
4. Simple arithmetic for the transverse projection and cross product

**Speedup: two levels**

**Single-point speedup (modest, ~2-3x):** Whe calling `get_farfield` for individual points, the code in `farfield_lowlevel` (line 530-532) still loops over every source point but calls `green3d_farfield` instead of `green3d`. The savings come from avoiding per-pair norm/division/complex-division computations. This gives a roughly 2-3x speedup per point.

**Grid speedup via FFT (massive, O(N_src / log N_src)):** The big win is in `get_farfields_array` (lines 588-769). When `far_field_approx=True` in 3D without periodic replicas, the code activates an FFT-accelerated path:

1. **Preprocessing (once per near-field face):** The DFT-accumulated source currents on each planar face are zero-padded (4x oversampling) and transformed via 2D FFT. Cost: O(N_src · log N_src) total across all faces.
2. **Per output point:** Instead of summing over all N_src source points, the code looks up each face's FFT at the spatial frequency `f = k·x̂·spacing·N/(2π)` via bilinear interpolation (`fft2d_interp`), then applies the geometric phase correction and far-field formula. Cost: O(N_faces) per ouput point.

**Total complexity comparison for a grid of N_out far-field points:**

| Method | Cost |
| ---------- | ------ |
| Exact (`green3d`) | O(N_out ⨯ N_src) |
| Far-field + FFT | O(N_src · log N_src + N_out ⨯ N_faces) |

For a typical 3D problem (resolution 20, near-field box of side 4 ➔ ~6,400 points per face ⨯ 6 faces ≈ 38,400 source points, computing the far fields on a 100⨯100 grid):
- Exact: 10,000 ⨯ 38,400 ≈ **384 million** Green's function evaluations
- FFT: 6 FFTs of size ~256⨯256 + 10,000 ⨯ 6 interpolations ≈ **~2.4 million** operations

That's a **~160x** speedup for this example, and it scales proportionally with N_src — larger near-field surfaces or higher resolution yield even greater speedups (easily 1000x+ for production-size problems).

**Note:** the `<algorithm>` header added in `near2far.cpp` is needed for two things used by the FFT-accelerated far-field path:
1. `std::swap` at line 423 in `fft1d_inplace` — the bit-reversal permutation step.
2. `std::max` at lines 672-673 — computing the zero-padded FFT sizes: `next_pow2(std::max(oversample * cf.n1, 2))`

There's also a pre-existing use of `std::max` at line 790 in the progress reporting, but that code predates this feature. The FFT code alone requires the include.

### DIFF
--- a/src/near2far.cpp
+++ b/src/near2far.cpp
@@ -178,7 +178,7 @@ static void green3d_farfield(std::complex<double> *EH, double xhat_x, double xha
     EH[5] = hf * tz;
   }
   else
-    meep::abort("unrecognized soruce type in green3d_farfield");
+    meep::abort("unrecognized source type in green3d_farfield");
 }
 
 /* Given the field f0 correponding to current-source component c0 at
@@ -587,9 +587,9 @@ double *dft_near2far::get_farfields_array(const volume &where, int &rank, size_t
   bool use_fft = far_field_approx && where.dim == D3 && periodic_n[0] == 0 && periodic_n[1] == 0;
 
   struct chunk_fft {
-    direction normal_d, d1, d2;
-    double face_pos, origin1, origin2, spacing1, spacing2;
-    int n1, n2, nf1, nf2;
+    direction normal_d, d[2];
+    double face_pos, origin[2], spacing[2];
+    int n[2], nf[2];
     component c0;
     std::vector<std::complex<double> > fft_data;
   };
@@ -597,7 +597,6 @@ double *dft_near2far::get_farfields_array(const volume &where, int &rank, size_t
 
   if (use_fft) {
     const int oversample = 4;
-    direction all_d[3] = {meep::X, meep::Y, meep::Z};
 
     for (dft_chunk *f = F; f; f = f->next_in_dft) {
       chunk_fft cf;
@@ -606,79 +605,53 @@ double *dft_near2far::get_farfields_array(const volume &where, int &rank, size_t
       vec rshift(f->shift * (0.5 * inva));
 
       cf.normal_d = NO_DIRECTION;
-      cf.d1 = cf.d2 = NO_DIRECTION;
-      cf.n1 = cf.n2 = 1;
+      for (int i = 0; i < 2; i++) {
+        cf.d[i] = NO_DIRECTION;
+        cf.n[i] = 1;
+      }
       int trans_idx = 0;
       for (int di = 0; di < 3; di++) {
-        direction d = all_d[di];
+        direction d = direction(X + di);
         int nd = (f->ie.in_direction(d) - f->is.in_direction(d)) / 2 + 1;
         if (nd <= 1)
           cf.normal_d = d;
         else {
-          if (trans_idx == 0) {
-            cf.d1 = d;
-            cf.n1 = nd;
-          }
-          else {
-            cf.d2 = d;
-            cf.n2 = nd;
-          }
+          cf.d[trans_idx] = d;
+          cf.n[trans_idx] = nd;
           trans_idx++;
         }
       }
-      if (cf.normal_d == NO_DIRECTION || cf.d1 == NO_DIRECTION) {
+      if (cf.normal_d == NO_DIRECTION || cf.d[0] == NO_DIRECTION) {
         use_fft = false;
         break;
       }
-      if (cf.d2 == NO_DIRECTION) {
-        cf.d2 = cf.d1;
-        cf.n2 = 1;
+      if (cf.d[1] == NO_DIRECTION) {
+        cf.d[1] = cf.d[0];
+        cf.n[1] = 1;
       }
 
-      vec pos0(D3);
-      for (int di = 0; di < 3; di++)
-        pos0.set_direction(all_d[di], f->is.in_direction(all_d[di]) * 0.5 * inva);
-      pos0 = f->S.transform(pos0, f->sn) + rshift;
-
+      vec pos0 = f->S.transform(f->fc->gv[f->is], f->sn) + rshift;
       cf.face_pos = pos0.in_direction(cf.normal_d);
-      cf.origin1 = pos0.in_direction(cf.d1);
-      cf.origin2 = pos0.in_direction(cf.d2);
-
-      cf.spacing1 = inva;
-      if (cf.n1 > 1) {
-        vec pos1(D3);
-        for (int di = 0; di < 3; di++) {
-          double val = f->is.in_direction(all_d[di]) * 0.5 * inva;
-          if (all_d[di] == cf.d1) val += inva;
-          pos1.set_direction(all_d[di], val);
+      for (int i = 0; i < 2; i++) {
+        cf.origin[i] = pos0.in_direction(cf.d[i]);
+        cf.spacing[i] = inva;
+        if (cf.n[i] > 1) {
+          vec pos1 = f->S.transform(f->fc->gv[f->is + unit_ivec(D3, cf.d[i]) * 2], f->sn) + rshift;
+          cf.spacing[i] = pos1.in_direction(cf.d[i]) - cf.origin[i];
         }
-        pos1 = f->S.transform(pos1, f->sn) + rshift;
-        cf.spacing1 = pos1.in_direction(cf.d1) - cf.origin1;
-      }
-      cf.spacing2 = inva;
-      if (cf.n2 > 1) {
-        vec pos1(D3);
-        for (int di = 0; di < 3; di++) {
-          double val = f->is.in_direction(all_d[di]) * 0.5 * inva;
-          if (all_d[di] == cf.d2) val += inva;
-          pos1.set_direction(all_d[di], val);
-        }
-        pos1 = f->S.transform(pos1, f->sn) + rshift;
-        cf.spacing2 = pos1.in_direction(cf.d2) - cf.origin2;
+        cf.nf[i] = next_pow2(std::max(oversample * cf.n[i], 2));
       }
 
-      cf.nf1 = next_pow2(std::max(oversample * cf.n1, 2));
-      cf.nf2 = next_pow2(std::max(oversample * cf.n2, 2));
-      cf.fft_data.assign((size_t)cf.nf1 * cf.nf2 * Nfreq, 0.0);
+      cf.fft_data.assign((size_t)cf.nf[0] * cf.nf[1] * Nfreq, 0.0);
 
       for (size_t fi = 0; fi < Nfreq; fi++) {
-        std::complex<double> *arr = cf.fft_data.data() + fi * cf.nf1 * cf.nf2;
+        std::complex<double> *arr = cf.fft_data.data() + fi * cf.nf[0] * cf.nf[1];
         for (size_t idx_dft = 0; idx_dft < (size_t)f->N; idx_dft++) {
-          int i1 = idx_dft / cf.n2;
-          int i2 = idx_dft % cf.n2;
-          arr[i1 * cf.nf2 + i2] = f->dft[Nfreq * idx_dft + fi];
+          int i1 = idx_dft / cf.n[1];
+          int i2 = idx_dft % cf.n[1];
+          arr[i1 * cf.nf[1] + i2] = f->dft[Nfreq * idx_dft + fi];
         }
-        fft2d_inplace(arr, cf.nf1, cf.nf2, -1);
+        fft2d_inplace(arr, cf.nf[0], cf.nf[1], -1);
       }
       cffts.push_back(std::move(cf));
     }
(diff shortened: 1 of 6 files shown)


src/near2far.cpp
```diff
@@ -348,15 +405,92 @@ void greencyl(std::complex<double> *EH, const vec &x, double freq, double eps, d
...
+    double ang = sign * 2 * pi / len;
+    std::complex<double> wn(cos(ang), sin(ang));
+    for (int i = 0; i < n; i += len) {
+      std::complex<double> w(1.0);
+      for (int j = 0; j < len / 2; j++) {
+        std::complex<double> u = data[i + j];
+        std::complex<double> v = data[i + j + len / 2] * w;
+        data[i + j] = u + v;
+        data[i + j + len / 2] = u - v;
+        w *= wn;
```
stevengj, line 435: Unstable trigonometric recurrence (O(n) roundoff error).

src/near2far.cpp
```diff
@@ -348,15 +405,92 @@ void greencyl(std::complex<double> *EH, const vec &x, double freq, double eps, d
   }
 }
 
-void dft_near2far::farfield_lowlevel(std::complex<double> *EH, const vec &x, double greencyl_tol) {
+static int next_pow2(int n) {
+  int p = 1;
+  while (p < n)
+    p *= 2;
+  return p;
+}
```
stevengj, line 413: There is a cute way to get the next power of 2 using some bit-twiddling, but this is fine.  We don't care about performance here.

src/near2far.cpp
```diff
@@ -435,6 +584,224 @@ double *dft_near2far::get_farfields_array(const volume &where, int &rank, size_t
...
+
+    for (dft_chunk *f = F; f; f = f->next_in_dft) {
+      chunk_fft cf;
+      cf.c0 = component(f->vc);
+      double inva = f->fc->gv.inva;
+      vec rshift(f->shift * (0.5 * inva));
+
+      cf.normal_d = NO_DIRECTION;
+      cf.d1 = cf.d2 = NO_DIRECTION;
+      cf.n1 = cf.n2 = 1;
```
stevengj: Would be better to make these length-2 arrays so that we can use loops.

src/near2far.cpp
```diff
@@ -435,6 +584,224 @@ double *dft_near2far::get_farfields_array(const volume &where, int &rank, size_t
...
+      if (cf.n2 > 1) {
+        vec pos1(D3);
+        for (int di = 0; di < 3; di++) {
+          double val = f->is.in_direction(all_d[di]) * 0.5 * inva;
+          if (all_d[di] == cf.d2) val += inva;
+          pos1.set_direction(all_d[di], val);
+        }
+        pos1 = f->S.transform(pos1, f->sn) + rshift;
+        cf.spacing2 = pos1.in_direction(cf.d2) - cf.origin2;
+      }
```
stevengj, line 689: copy-paste code … use a loop over `cv.n[i]` instead.

src/near2far.cpp
```diff
@@ -435,6 +584,224 @@ double *dft_near2far::get_farfields_array(const volume &where, int &rank, size_t
...
+    double face_pos, origin1, origin2, spacing1, spacing2;
+    int n1, n2, nf1, nf2;
+    component c0;
+    std::vector<std::complex<double> > fft_data;
+  };
+  std::vector<chunk_fft> cffts;
+
+  if (use_fft) {
+    const int oversample = 4;
+    direction all_d[3] = {meep::X, meep::Y, meep::Z};
```
stevengj: This shouldn't be needed because `all_d[i]` is the same thing as `X + i`.

src/near2far.cpp
```diff
@@ -435,6 +584,224 @@ double *dft_near2far::get_farfields_array(const volume &where, int &rank, size_t
...
+      cf.origin2 = pos0.in_direction(cf.d2);
+
+      cf.spacing1 = inva;
+      if (cf.n1 > 1) {
+        vec pos1(D3);
+        for (int di = 0; di < 3; di++) {
+          double val = f->is.in_direction(all_d[di]) * 0.5 * inva;
+          if (all_d[di] == cf.d1) val += inva;
+          pos1.set_direction(all_d[di], val);
+        }
```
stevengj, line 686: Seems like this loop could be simplified to something like:
```c++
vec pos1 = f->fc->gv[f->is + unit_ivec(D3, cf.d1) * 2];
```

src/near2far.cpp
```diff
@@ -435,6 +584,224 @@ double *dft_near2far::get_farfields_array(const volume &where, int &rank, size_t
...
+        break;
+      }
+      if (cf.d2 == NO_DIRECTION) {
+        cf.d2 = cf.d1;
+        cf.n2 = 1;
+      }
+
+      vec pos0(D3);
+      for (int di = 0; di < 3; di++)
+        pos0.set_direction(all_d[di], f->is.in_direction(all_d[di]) * 0.5 * inva);
```
stevengj: Simplify to something like:
```c++
vec pos0 = f->fc->gv[f->is];
```

src/near2far.cpp
```diff
@@ -435,6 +584,224 @@ double *dft_near2far::get_farfields_array(const volume &where, int &rank, size_t
...
+      cf.nf1 = next_pow2(std::max(oversample * cf.n1, 2));
+      cf.nf2 = next_pow2(std::max(oversample * cf.n2, 2));
+      cf.fft_data.assign((size_t)cf.nf1 * cf.nf2 * Nfreq, 0.0);
+
+      for (size_t fi = 0; fi < Nfreq; fi++) {
+        std::complex<double> *arr = cf.fft_data.data() + fi * cf.nf1 * cf.nf2;
+        for (size_t idx_dft = 0; idx_dft < (size_t)f->N; idx_dft++) {
+          int i1 = idx_dft / cf.n2;
+          int i2 = idx_dft % cf.n2;
+          arr[i1 * cf.nf2 + i2] = f->dft[Nfreq * idx_dft + fi];
```
stevengj: Seems better to loop over `i1` and `i2` and compute `idx_dft` from these than vice versa?

src/near2far.cpp
```diff
@@ -435,6 +584,224 @@ double *dft_near2far::get_farfields_array(const volume &where, int &rank, size_t
...
+          }
+          else {
+            std::complex<double> hf = J_tilde * expfac_base / mu;
+            EH1[0] += -hf * crx * impedance;
+            EH1[1] += -hf * cry * impedance;
+            EH1[2] += -hf * crz * impedance;
+            EH1[3] += hf * tx;
+            EH1[4] += hf * ty;
+            EH1[5] += hf * tz;
+          }
```
stevengj, line 816: This code seems like almost a copy-paste of `green_farfield`?

src/near2far.cpp
```diff
@@ -435,6 +584,224 @@ double *dft_near2far::get_farfields_array(const volume &where, int &rank, size_t
...
+            last_point = this_point;
+          }
+          farfield_lowlevel(EH1, x, greencyl_tol, far_field_approx);
+          if (verbosity > 1) all_wait();
+          ptrdiff_t idx = (i0 * dims[1] + i1) * dims[2] + i2;
+          for (size_t i = 0; i < Nfreq; ++i)
+            for (int k = 0; k < 6; ++k) {
+              EH_[((k * 2 + 0) * N + idx) * Nfreq + i] = real(EH1[i * 6 + k]);
+              EH_[((k * 2 + 1) * N + idx) * Nfreq + i] = imag(EH1[i * 6 + k]);
+            }
```
stevengj, line 856: This whole block seems to be a line-for-line copy-paste of the code on line 808 … needs refactoring.

src/near2far.cpp
```diff
@@ -348,15 +405,92 @@ void greencyl(std::complex<double> *EH, const vec &x, double freq, double eps, d
...
 
-void dft_near2far::farfield_lowlevel(std::complex<double> *EH, const vec &x, double greencyl_tol) {
+static int next_pow2(int n) {
+  int p = 1;
+  while (p < n)
+    p *= 2;
+  return p;
+}
+
+static void fft1d_inplace(std::complex<double> *data, int n, int sign) {
```
stevengj, line 415: Note that we already link FFTW if it is present — might as well just call it rather than rewriting an FFT?

src/near2far.cpp
oskooi: Done.

src/near2far.cpp
oskooi: Done.

src/near2far.cpp
oskooi: Done.

src/near2far.cpp
oskooi: Done.

src/near2far.cpp
oskooi: Done.

src/near2far.cpp
oskooi: Note that the right-hand side is different in the two blocks of code.

oskooi:
Empirically confirmed (via C++ instrumentation) that the FFT-accelerated path in `get_farfields_array` was **never executing**. The normal-detection heuristic assumed the surface normal had `nd ≤ 1` grid points, but near2far DFT chunks are **2 grid points thick along their normal** (the tangential field component is offset half a pixel on the Yee grid). So `use_fft` was always disabled, and every `get_farfields(..., far_field_approx=True)` call silently fell back to direct summation — which is why the old `1e-12` test passed trivially (both sides ran identical code).

**The fix (`src/near2far.cpp`)**

1. **Robust normal detection** — pick the thinnest Cartesian direction as the normal; the other two are transverse.
2. **Per-slice handling** — each normal layer is treated as a separate planar current sheet (its own `face_pos`), so the decomposition is exact for any rectangular chunk, including the **edge slivers** (`78×2×2`) that arise when a component (e.g. `Hy`) is tangential to two perpendicular faces. The scatter now uses actual grid indices (`IVEC_LOOP_ILOC`), independent of iteration order.
3. **Bicubic interpolation** — replaced bilinear `fft2d_interp` with Catmull-Rom bicubic. Measured error scaling at the test geometry:

| oversample | bilinear | bicubic |
|-----------------| ---------- | --------- |
| 4×               | 15%      | 2.1%    |
| 8×               | 3.6%     | **0.23%** |
| 16×             | 1.0%     | 0.015% |

4. Settled on **oversample=8 + bicubic** (~0.1-0.2% error, comparable to the radiation-zone approximation's own O(L/R) ≈ 0.4% error) as the accuracy/memory sweet spot.

**Measured speedup** (FFT vs exact `green3d`)

FFT time is nearly constant (~2.5 s, dominated by the O(N_src log N_src) precompute) while direct scales linearly with output points:

| grid     | speedup      |  accuracy (FFT vs exact) |
|----------|-----------------|-----------------------------------|
| 20×20 | 6.4×            | 0.11% |
| 40×40 | 25×            | 0.11% |
| 80×80 | **101×**            | 0.11% |

A notable finding: the per-point `green3d_farfield` simplification alone gives **~no speedup** (direct-approx ≈ exact in wall-clock); essentially all the gain comes from the FFT restructuring.

**Tests** (`python/tests/test_farfield_approx.py`)
- Updated `test_farfield_approx_get_farfields_array` tolerance from the trivially passing `1e-12` to `5e-3` (it now genuinely compares the FFT path vs direct summation).
- Added `test_farfield_approx_fft_grid_vs_exact` — the missing coverage: a finite far-field patch spanning many fractional FFT bins, validated against both the exact near-to-far transform and per-point direct summation at `1e-2`.

stevengj:
cc @lxvm, who implemented FFT-based near-to-farfield transforms in Julia recently for another project.  Can you take a quick look at this?

stevengj:
If you compare the brute-force/slow convolution with the approximate (farfield) Green's function to the FFT-accelerated version, they should the same up to roundoff errors, no?

Hopefully, the 0.11% error you report above is due only to the farfield approximation (and should decay with distance).

src/near2far.cpp
stevengj: If it's *almost* identical, it can still be refactored, no?

lxvm:
I'm having trouble understanding some of the code since I am not fluent in C++, but a few questions that I have are:
* Why is it necessary to assume the far-field approximation in order to apply the FFT-based near-to-far transformation? 
* How is the grid where the far field is evaluated using the FFT decided?
* Interpolating onto the far-field points in two stages (FFT oversampling, i.e. Fourier interpolation, and bicubic interpolation) is a good idea, but the factor of 8 oversampling in the FFT probably makes the transform 8x larger, no? In my package, I implemented the FFT-based convolution using a FFT to transform into the Fourier domain and a chirp-Z transform (also FFT-accelerated) for the inverse Fourier transform. It just accomplishes the Fourier interpolation and is more flexible for sampling target points on grids of arbitrary resolution, but is somewhat more complicated to implement.

A few other comments related to some of Steven's comments:
* The FFT-based near-to-far transformation should match the brute-force convolution up to roundoff error when the source and target grid are the same resolution. Perhaps the FFT upsampling and bicubic interpolation steps add some roundoff error, but 0.11% error sounds off
* Linking against FFTW should provide speedups for obvious reasons, including the fact that you can zero-pad the convolutions to transform sizes other than powers of two and I think powers of 2, 3, 5, and 7 are performant.

PR #3225, opened 2026-06-05
Add optional `sampling_interval` parameter to `stop_when_fields_decayed` termination criterion

oskooi:
The termination criterion `stop_when_fields_decayed` involves calling `get_field_point` at *every* timestep which can be expensive for parallel simulations involving MPI  (because it involves an `MPI_Allreduce` global barrier) even though the decay decision is made once per `dt` window. This PR adds an optional `sampling_interval` parameter to `stop_when_fields_decayed` which can be used to specify the number of timesteps between successive field samples. The default of `1` samples the field at every timestep (the historical behavior).

### DIFF
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -5273,7 +5273,7 @@ def to_appended(fname, *step_funcs):
 
 
 def stop_when_fields_decayed(
-    dt=None, c=None, pt=None, decay_by=None, sample_interval=1
+    dt=None, c=None, pt=None, decay_by=None, sampling_interval=1
 ):
     """
     Return a `condition` function, suitable for passing to `Simulation.run` as the `until`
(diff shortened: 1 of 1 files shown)


python/simulation.py
```diff
@@ -5282,6 +5284,14 @@ def stop_when_fields_decayed(dt=None, c=None, pt=None, decay_by=None):
     over that time period &mdash; in this way, it won't be fooled just because the field
     happens to go through zero at some instant.
 
+    The optional `sample_interval` (a positive integer, default `1`) is the number of
```
stevengj: For Padé DFT we called it `sampling_interval`, and for DFT fields we called it `decimation_factor`.  Seems like we should pick one…

stevengj (approved):

Issue #3229, opened 2026-06-13
Subpixel averaging for dispersive (anisotropic) susceptibilities


oskooi:
**Status:** proposal / design note
**Area:** `src/anisotropic_averaging.cpp`, `src/susceptibility.cpp`, `src/update_pols.cpp`
**Motivation:** permanently fix the anisotropic-Lorentzian boundary instability that #666 worked around with a runtime guard, and remove a known accuracy limitation.

## Background: the #666 guard

`lorentzian_susceptibility::update_P` (`src/susceptibility.cpp`) guards its anisotropic (2×2 and 3×3) branches with a per-point test that skips the polarization update where the diagonal `sigma` is zero:

https://github.com/NanoComp/meep/blob/ea11e6c076653c789b5dabac1f7b7bb8644006d9/src/susceptibility.cpp#L229-L238

It exists to suppress an instability "near the boundaries of materials."

### Root cause (empirically established)

A controlled 2D TE reproducer (anisotropic Lorentzian medium, Gaussian pulse, no PML, watch total energy) isolates the trigger precisely:

| Off-diagonal `sigma` support vs. diagonal support | No-guard result |
|---|---|
| Off-diagonal **exceeds** diagonal | **diverges** |
| Off-diagonal strictly **inside** diagonal | **stable** (decays) |

With the guard, the diverging case becomes stable, and for the well-posed case the guard is a **no-op** (bit-identical results).

The mechanism:

1. The off-diagonal coupling in `update_P` uses the 4-point `OFFDIAG` averaging stencil, which reads `sigma`/field at `i` **and `i+is`** (one cell along the component's own direction).
2. The dispersive `sigma` is **point-sampled** in `structure_chunk::add_susceptibility` (`src/anisotropic_averaging.cpp`) — it is *not* subpixel-averaged. At a material boundary the off-diagonal `sigma` is therefore nonzero one stencil-step beyond where the diagonal `sigma` has dropped to zero.
3. At such a point `P_c` is driven with **no diagonal `sigma` to anchor it**. The local discrete susceptibility tensor is effectively `[[0, σ], [σ, 0]]`, whose eigenvalues are `±σ` — **not positive-semidefinite, i.e. non-passive** — so it pumps energy and the simulation diverges.

[...]

PR #3230, opened 2026-06-14
Add warning for non-positive semidefinite `sigma` tensor for Lorentzian susceptibility

oskooi:
Related to #3229.

This PR adds a warning if the user-supplied anisotropic Lorentzian susceptibility tensor (`sigma`) is non positive semidefinite. Additionally, the guard introduced in #666 to skip update of the polarization fields at material interfaces (to ensure stability) is reorganized to enable loop vectorization.

### DIFF
--- a/src/structure.cpp
+++ b/src/structure.cpp
@@ -474,6 +474,30 @@ void structure::add_susceptibility(double sigma(const vec &), field_type ft,
   add_susceptibility(sig, ft, sus);
 }
 
+/* Heuristic check for a non-positive-semidefinite (non-passive) anisotropic
+   susceptibility tensor: an off-diagonal sigma_ab with sigma_ab^2 > sigma_aa *
+   sigma_bb makes the local 2x2 tensor have a negative eigenvalue, which is
+   non-physical and drives instabilities.  We only test pixels where both
+   diagonal entries are nonzero, so material-boundary pixels (where one diagonal
+   has tapered to zero — a discretization effect handled by the update_P guard)
+   are skipped to avoid false positives. */
+static bool sigma_not_psd(const susceptibility *sus, const grid_volume &gv, field_type ft) {
+  FOR_FT_COMPONENTS(ft, ca) {
+    const realnum *saa = sus->sigma[ca][component_direction(ca)];
+    if (!saa) continue;
+    FOR_FT_COMPONENTS(ft, cb) {
+      if (cb == ca) continue;
+      const realnum *sbb = sus->sigma[cb][component_direction(cb)];
+      const realnum *sab = sus->sigma[ca][component_direction(cb)];
+      if (!sbb || !sab) continue;
+      LOOP_OVER_VOL(gv, ca, idx) {
+        const realnum a = saa[idx], b = sbb[idx], o = sab[idx];
+        if (a > 0 && b > 0 && o * o > a * b * (1 + 1e-6)) return true;
+      }
+    }
+  }
+}
+
 void structure::add_susceptibility(material_function &sigma, field_type ft,
                                    const susceptibility &sus) {
   changing_chunks();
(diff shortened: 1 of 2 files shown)


src/structure.cpp
```diff
@@ -474,6 +474,31 @@ void structure::add_susceptibility(double sigma(const vec &), field_type ft,
...
+    const realnum *saa = sus->sigma[ca][component_direction(ca)];
+    if (!saa) continue;
+    FOR_FT_COMPONENTS(ft, cb) {
+      if (cb == ca) continue;
+      const realnum *sbb = sus->sigma[cb][component_direction(cb)];
+      const realnum *sab = sus->sigma[ca][component_direction(cb)];
+      if (!sbb || !sab) continue;
+      LOOP_OVER_VOL(gv, ca, idx) {
+        const realnum a = saa[idx], b = sbb[idx], o = sab[idx];
+        if (a > 0 && b > 0 && o * o > a * b * (1 + 1e-6)) return true;
```
stevengj, line 495: Why is it checking "2x2" tensors, and not the full 3x3 matrix?

src/susceptibility.cpp
```diff
@@ -226,26 +226,30 @@ void lorentzian_susceptibility::update_P(realnum *W[NUM_FIELD_COMPONENTS][2],
...
-              p[i] = gamma1inv * (pcur * (2 - omega0dtsqr_denom) - gamma1 * pp[i] +
-                                  omega0dtsqr * (s[i] * w[i] + OFFDIAG(s1, w1, is1, is) +
-                                                 OFFDIAG(s2, w2, is2, is)));
-              pp[i] = pcur;
-            }
+            // The s[i] != 0 mask works around instabilities near material
+            // boundaries (see PR #666). sigma is static in time, so the
+            // predicate is the same every step; apply it branchlessly so the
+            // loop still vectorizes. Where s[i] == 0, p[i] and pp[i] are left
+            // unchanged, exactly as the original guarded update did.
```
stevengj, line 233: It would be nice to know how much of a performance difference this makes in two cases:

1. the chunk is completely filled with σ ≠ 0 (so `mask == 1` everywhere).
2. the chunk is mostly σ = 0 (so that you end up doing a bunch of extra computation that would have otherwise been removed by the `if`).

stevengj:
My other concern about this PR is that it is looking at σ on the Yee grid, for which different components of σ are stored at different Yee grid points.   If σ is not constant in space, then it is problematic to simply treat the different components as part of the same σ tensor … you would need to define some effective σ tensor by some kind of spatial average of σ components.

One alternative is simply to check positive-definiteness at a higher level, e.g. in the list of materials/objects in `meepgeom.cpp`.

src/structure.cpp
oskooi: **Why $2\times2$ minors suffice (and 3x3 isn't used)**

1. A **negative $2\times2$ principal minor is *sufficient* to prove the full tensor is not PSD.** A symmetric matrix is PSD iff *all* its principal minors are $\geq$ 0. So finding any negative $2\times2$ minor is conclusive proof of non-passivity – no $3\times3$ work is needed to fire the warning. The function only needs to return `true` on the *common*, *conclusive* case.
2. **The $3\times 3$ determinant alone would not be a valid PSD test anyway.** A $3\times3$ symmetric matrix can have a *positive* determinant and still not PSD (e.g. two negative eigenvalues). A correct full test would require checking *all* principal minors (the three diagonals, the three $2\times2$ minors, *and* the $3\times3$ determinant), or an eigenvalue decomposition – much more expensive, for a warning. The diagonal + $2\times2$ checks catch the realistic failures cheaply.

**The trade-off**

The check can in principle miss a tensor where every $2\times2$ minor is non-negative but the $3\times3$ determinant is negative (one negative eigenvalue spread across all three off-diagonals). The comment on line 477 acknowledges this by calling it  a **"Heuristic check"** – it's tuned to flag the realistic, dominant non-passive inputs with a clear diagnostic, rather than be a mathematically complete PSD test.

src/susceptibility.cpp
oskooi: **Cross-compiler results (ns/element, filled = the common bulk case, L=128)**

|  | branched `if` | branchless mask | mask vs `if` | both vectorize? |
| --------- | ---------- | -------- | -------- | ---------- |
| **GCC 11.5** $3\times3$ | 3.88 | 4.59 | **0.84 $\times$ (slower)** | neither (strided-load blocked) |
| **GCC 11.5** $2\times2$ | 2.23 | 2.84 | **0.79 $\times$ (slower)** | neither |
| **Clang 15** $3\times3$ | 2.43 | 2.43 | **1.00 $\times$ (tie)** | both (width 8) |
| **Clang 15** $2\times2$ | 1.37 | 1.42 | **0.96 $\times$ ($\approx$ tie)** | both (width 8) |

The mask's sole stated rationale – *"apply it branchlessly so the loop still vectorizes"* – is **empirically false on both compilers tested**:
- On Clang it's unnecessary (the branch vectorizes anyway).
- On GCC 11.5 it's ineffective (vectorization is blocked by the strided load, not the branch) and actively harmful.

The `if` guard is **faster-or-equal everywhere measured**, it's bit-for-bit identical in output, and it's the simpler form. So as a performance change the mask doesn't justify itself.

stevengj:
Closing since no performance benefit; positive-definiteness can be checked elsewhere.

PR #3240, opened 2026-06-28
Remove deprecated `monitor_point` C++ class

### LINKED ISSUE #3237: get rid of deprecated monitor_point code
The C++ `monitor_point` interface has been deprecated (including [in the docs](https://github.com/NanoComp/meep/blob/91fc22a68a30a6101f3028b0a8643e159f977b06/doc/docs/C%2B%2B_Developer_Information.md?plain=1#L83)) for a long time now, and is probably worth getting rid of.

(The `monitor_point::fourier_transform` feature is no longer even tested.)

Several of the C++ tests will need to be rewritten to use `get_field_pt` or whatever instead.

oskooi:
Closes #3237.

The C++ class `monitor_point` (`src/meep.hpp:1100-1128`) is a legacy field-sampling helper that has long been documented as deprecated in `doc/docs/C++_Developer_Information.md:83`. It is a thin caching/bookkeeping wrapper: `fields::get_point()` already populates it by calling `fields::get_field()` internally (`src/monitor.cpp:62`). The documented replacement is to call `fields::get_field()` directly and, where spectral analysis is needed, feed a collected array into `do_harminv()`.

`monitor_point` is C++-only — the Python and Scheme front-ends never use it (Python field sampling already goes through `fields::get_field` via `fields::get_field_point`, and Python harminv uses the standalone `do_harminv` wrapper). It is exposed to SWIG only incidentally via `%include "meep.hpp"`, with no explicit typemaps, so removing it does not affect the bindings.

This PR deletes the class and its supporting `fields` methods, migrates the C++ unit tests that use it to `fields::get_field` (+ `do_harminv` where needed), and updates the docs.

### DIFF
--- a/src/monitor.cpp
+++ b/src/monitor.cpp
@@ -22,45 +22,14 @@
 #include "meep.hpp"
 #include "meep_internals.hpp"
 
-#include "config.h"
-#if defined(HAVE_LIBFFTW3)
-#include <fftw3.h>
-#elif defined(HAVE_LIBDFFTW)
-#include <dfftw.h>
-#elif defined(HAVE_LIBFFTW)
-#include <fftw.h>
-#endif
-#if defined(HAVE_LIBFFTW3) || defined(HAVE_LIBFFTW) || defined(HAVE_LIBDFFTW)
-#define HAVE_SOME_FFTW 1
-#else
-#define HAVE_SOME_FFTW 0
-#endif
-
-/* Below are the monitor point routines. */
+/* Below are the field point routines. */
 
 using namespace std;
 
 namespace meep {
 
-monitor_point::monitor_point() { next = NULL; }
-
-monitor_point::~monitor_point() {
-  if (next) delete next;
-}
-
 inline complex<double> getcm(const realnum *const f[2], size_t i) {
   return complex<double>(f[0][i], f[1][i]);
-}
-
-void fields::get_point(monitor_point *pt, const vec &loc) const {
-  if (pt == NULL) meep::abort("Error:  get_point passed a null pointer!\n");
-  for (int i = 0; i < 10; i++)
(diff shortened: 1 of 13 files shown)


src/monitor.cpp
```diff
@@ -22,47 +22,16 @@
...
-#elif defined(HAVE_LIBDFFTW)
-#include <dfftw.h>
-#elif defined(HAVE_LIBFFTW)
-#include <fftw.h>
-#endif
-#if defined(HAVE_LIBFFTW3) || defined(HAVE_LIBFFTW) || defined(HAVE_LIBDFFTW)
-#define HAVE_SOME_FFTW 1
-#else
-#define HAVE_SOME_FFTW 0
-#endif
```
stevengj, line 37: Also get rid of the corresponding `configure.ac` tests if (for now) we have nothing that uses FFTs.

oskooi: FFTW is **still used** — by `src/casimir.cpp` (the [`make_casimir_g` function](https://github.com/NanoComp/meep/blob/3c3bfac411d95c70482a5d355527ef2523d81e14/src/casimir.cpp#L93-L106) uses `fftw_plan_dft_1d`/`fftw_execute` and the FFTW2 paths). The `configure.ac` comment at [line 146](https://github.com/NanoComp/meep/blob/3c3bfac411d95c70482a5d355527ef2523d81e14/configure.ac#L146) even names both files. 

The suggested cleanup is to update the stale comment in `configure.ac` to drop the `src/monitor.cpp` reference, since that file no longer uses FFTW.

stevengj: Okay, let's leave it.   We can probably eliminate FFTW2, but that can be done elsewhere.

PR #3244, opened 2026-07-03
Remove libGDSII dependency; demonstrate gdstk in the tutorials

### LINKED ISSUE #3206: Removing libGDSII support
Currently, [`libGDSII`](https://github.com/HomerReid/libGDSII/) is a [dependency](https://meep.readthedocs.io/en/latest/Build_From_Source/#libgdsii) used for importing GDS files. However, libGDSII is missing important functionality (i.e., [importing layers as tuples](https://meep.readthedocs.io/en/latest/Python_Tutorials/GDSII_Import/#importing-a-gds-layer-using-a-tuple)) and seems to be no longer maintained. A replacement is [`gdstk`](https://heitzmann.github.io/gdstk/index.html) which is [actively maintained](https://github.com/heitzmann/gdstk). The [tutorials](https://meep.readthedocs.io/en/latest/Python_Tutorials/GDSII_Import/) would also need to be updated as part of the deprecation.

oskooi:
Closes #3206.

**Context**

[libGDSII](https://github.com/HomerReid/libGDSII) is an optional Meep dependency used to import GDSII layout files (polygons → Meep prisms / volumes). It is no longer maintained. This change **removes libGDSII entirely** from the Meep codebase — the build hooks, the C++ implementation, the SWIG/Python wrappers, the GDS tests, and the build documentation that referenced it. The Meep-native GDS functions that depended on libGDSII (`get_GDSII_prisms`, `GDSII_vol`, `GDSII_layers`, `with_libGDSII`, `set_geometry_from_GDSII`, `get_GDSII_prism`, `get_GDSII_volume`, `get_GDSII_layers`) are therefore removed as well.

Rather than wiring a replacement library into Meep's build, the [gdstk](https://github.com/heitzmann/gdstk) Python package is demonstrated **directly in the Python tutorials**: users read GDS polygons with gdstk and build `mp.Prism` / `mp.Volume` objects themselves. gdstk is *not* added to Meep's build, requirements, or CI — it is a user-side tool shown in the docs.

Additionally, the GDS-import paths in the Python tests are **deleted** (analytic-prism coverage remains), and CI only has its libGDSII steps removed (gdstk is not added to CI).

**Scope decisions**
- gdstk is **not** integrated into Meep's C++/build/CI — tutorials only.
- Python test GDS paths are **removed**, not rewritten with gdstk.
- `get_center_and_size` (generaly utility, used by `test_simulation.py`) is **kept**.
- `coupler.gds` example asset is **kept** (the rewritten tutorial reads it via gdstk); `python/tests/data/bend-flux.gds` and `python/tests/data/spiral.gds` are **removed**.


stevengj:
I don't want to break working code.  So:

1. Keep GDSII in the code.   (However, it can print a deprecation notice when it is called)
2. Update the tutorials and docs to use gdstk
3. Keep GDSII in the installation manual but note that it is a deprecated feature that will be removed at some point.

PR #3251, opened 2026-07-22
Multithread the `near2far` field transformation with OpenMP

Luochenghuang:
This parallelizes the surface-point integral, via `PLOOP_OVER_IVECS_C` with an array-section reduction over the six field components (split into real/imag double arrays, since `std::complex<double>` isn't an OpenMP reduction type). `IVEC_LOOP_COUNTER` replaces the hand-maintained `idx_dft`, removing the loop-carried dependency so the surface loop can be threaded. The frequency loop moves inside the surface loop, so each point's location, f->S symmetry transform, and periodic-image setup are computed once and reused across all frequencies, and the parallel region is entered once per chunk rather than once per frequency.

We see a ~14.6× speedup of the farfield step on a 44-core machine using 32 threads.
image

### DIFF
--- a/src/near2far.cpp
+++ b/src/near2far.cpp
@@ -363,26 +363,27 @@ void dft_near2far::farfield_lowlevel(std::complex<double> *EH, const vec &x, dou
     component c0 = component(f->vc); /* equivalent source component */
 
     vec rshift(f->shift * (0.5 * f->fc->gv.inva));
-    for (size_t i = 0; i < Nfreq; ++i) {
-      double EHr[6] = {0, 0, 0, 0, 0, 0};
-      double EHi[6] = {0, 0, 0, 0, 0, 0};
-      PLOOP_OVER_IVECS_C(f->fc->gv, f->is, f->ie, idx,
-                         "omp parallel for collapse(3) reduction(+ : EHr[0:6], EHi[0:6])") {
-        (void)idx;
-        size_t idx_dft = IVEC_LOOP_COUNTER;
-        IVEC_LOOP_LOC(f->fc->gv, x0);
-        x0 = f->S.transform(x0, f->sn) + rshift;
-        vec xs(x0);
-        std::complex<double> EH6[6];
-        for (int i0 = -periodic_n[0]; i0 <= periodic_n[0]; ++i0) {
-          if (periodic_d[0] != NO_DIRECTION)
-            xs.set_direction(periodic_d[0], x0.in_direction(periodic_d[0]) + i0 * period[0]);
-          double phase0 = i0 * periodic_k[0];
-          for (int i1 = -periodic_n[1]; i1 <= periodic_n[1]; ++i1) {
-            if (periodic_d[1] != NO_DIRECTION)
-              xs.set_direction(periodic_d[1], x0.in_direction(periodic_d[1]) + i1 * period[1]);
-            double phase = phase0 + i1 * periodic_k[1];
-            std::complex<double> cphase = std::polar(1.0, phase);
+    std::vector<double> EHr(6 * Nfreq, 0.0), EHi(6 * Nfreq, 0.0);
+    double *EHr_p = EHr.data(), *EHi_p = EHi.data();
+    PLOOP_OVER_IVECS_C(
+        f->fc->gv, f->is, f->ie, idx,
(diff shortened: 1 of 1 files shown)


stevengj:
I guess the biggest benefit here is that you might have enough frequencies to use all the threads, whereas you are likely to have enough near-field points (per chunk).

stevengj:
It would be nice to benchmark a case with 32 frequencies, to see whether the speedup is comparable to the old code.

yorickreum:
This is interesting, as we would also benifit from faster near2far transformations.

I took a closer look now. To my understanding, we are reworking parallelism from frequency-parallel only (the outer loop, already parallel) to surface-point/spatial-parallel, while each surface-point worker computes all frequencies. It is sensible that this will win speedwise if there are many more surface points than frequencies, because that exposes much more parallel work while also reusing the frequency-independent geometry and symmetry calculations for each point.

I now benchmarked this since the 32-frequency question was still open. In short, it is noticeably faster.

In detail, I use a 3D box of six near-field surfaces, 239112 surface points. I time `farfield_lowlevel()` over a set of far-field points, best of 3 runs. With more than one rank, I wait until the slowest rank has finished (as `fields::farfield` combines afterwards).


### One rank, varying threads

Seconds per far-field point. Ratio is PR / master, so below 1.00 means this PR is faster.

| Nfreq | | 1 thread | 2 | 4 | 8 | 16 | 24 |
|---|---|---|---|---|---|---|---|
| 1 | master | 1.62e-2 | 1.66e-2 | 1.70e-2 | 1.71e-2 | 1.70e-2 | 1.88e-2 |
| | PR | 1.54e-2 | 7.88e-3 | 4.06e-3 | 2.28e-3 | 1.29e-3 | 1.08e-3 |
| | **ratio** | **0.95** | **0.48** | **0.24** | **0.13** | **0.08** | **0.06** |
| 32 | master | 5.26e-1 | 2.68e-1 | 1.38e-1 | 6.97e-2 | 3.81e-2 | 4.04e-2 |
| | PR | 3.69e-1 | 1.87e-1 | 9.68e-2 | 4.95e-2 | 2.78e-2 | 2.15e-2 |
| | **ratio** | **0.70** | **0.70** | **0.70** | **0.71** | **0.73** | **0.53** |

grafik

What stands out to me: 

- The code is  about 1.4 x faster on a single thread. That is the flat 0.70 ratio in the Nfreq=32 row. So people running serial or pure-MPI get the speedup too.
- It keeps scaling even with more threads than frequencies, as intended. The current code on master only splits work across frequencies, that's why it stops improving at 16 threads and even gets slower at 24. The PR generates a speed up at all thread counts.

### Ranks against threads

By adding more ranks = making the chunks smaller, the PR was still faster in (almost) every combination I tried.

| ranks × threads | Nfreq=1 | Nfreq=4 | Nfreq=16 | Nfreq=32 | Nfreq=64 |
|---|---|---|---|---|---|
| 1 × 24 | 0.06 | 0.11 | 0.35 | 0.53 | 0.63 |
| 2 × 8 | 0.13 | 0.34 | 0.75 | 0.74 | 0.67 |
| 4 × 4 | 0.62 | 0.60 | 0.77 | 0.74 | 0.69 |
| 8 × 2 | 0.67 | 0.80 | 0.76 | 0.74 | 0.71 |
| 16 × 1 | 0.97 | 0.82 | 0.76 | 0.74 | 0.71 |
| 24 × 1 | 0.99 | 0.75 | 0.71 | 0.72 | 0.67 |

The new approach can be slower in one case I can think of: When a chunk contains too few surface points to keep all threads busy. For very small chunks, the fixed overhead of starting and synchronizing the OpenMP loop can dominate because there are too few surface points to keep the threads busy: at 20-40 points per chunk on 24 threads I measured 2-3 x slower than master, and the PR takes 2.5e-4 s per far-field point at Nfreq=1 against 3.5e-4 at Nfreq=32, so 32× the work for 1.4× the time, so the cost is set by overhead rather than by work. As the number of frequencies increases, the additional cost of reducing the per-thread result arrays also grows roughly with the number of threads times the number of frequencies: each thread zeroes and combines its own 12 * Nfreq-double array, which is 288 doubles per parallel region at Nfreq=1 but 18432 at Nfreq=64 on 24 threads. 

It needs a small surface and a high thread count at the same time, so it is an edge case, but it could be worth guarding? A plain serial fallback would not be ideal however, since the old code could still parallelise over frequencies there.

PR #3253, opened 2026-07-26
Cleanup and refactor `eps_fit_lorentzian.py`

oskooi:
**What changed**

**nlopt →SciPy**
- Replaced `nlopt.opt(...)` + the MMA/L-BFGS local-optimizer setup with [`scipy.optimize.least_squares(..., method="trf", bounds=(0, np.inf))`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.least_squares.html) — the idiomatic tool for bound-constrained nonlinear least squares, and the natural fit for this residual-minimization problem.
- Dropped the nlopt-specific in-place-`grad` idiom. `lorentzerr` is replaced by a clean `_residuals` (real/imag-stacked residual vector) plus an **analytic** `_jacobian` passed to `least_squares` (verified against finite differences to ~1e-9).
- `lorentzfit` now returns the optimal params and the residual sum of squares Σ|Δ|² (= `2 * result.cost`), preserving the original error metric.

**Cleanup / refactor**
- Split the monolithic `__main__` block into focused functions: `fit_medium` (random-restart fitting → builds the `Medium`) and `plot_fit` (comparison plot), plus `main()`/`parse_args()`.
- Added **`argparse`** so the hard-coded fitting parameters are now CLI flags: `csv`, `--eps-inf`, `--num-lorentzians`, `--wl-min`, `--wl-max`, `--num-repeat`, `--seed`, `--output` (defaults preserve the original behavior; `--wl-min`/`--wl-max` default to 400/700).
- Added a `--seed` so restarts are **reproducible** (the original used unseeded `np.random`).
- Kept the public `lorentzfunc`/`lorentzfit` names (nothing in the repo imports them, but they're the recognizable API), added type hints and docstrings, and preserved the Drude special-case (ω = 0).

**Validation**
- Analytic Jacobian correct (≈1e-9 vs. finite difference).
- Running the refactored script on the [Sarkar TiO₂ data](https://github.com/polyanskiy/refractiveindex.info-database/blob/main/database/data/main/TiO2/nk/Sarkar.yml) recovers the **same** fit (to be added to `materials.py` in a separate PR): γ→0 (lossless), and A/ω² gives σ = 2.6134 and 0.5625 — matching results from the unmodified script exactly, max |Δn| = 6.0×10⁻⁵
- Compiles, no `nlopt` references remain, `-h` works.

### DIFF
--- a/python/examples/eps_fit_lorentzian.py
+++ b/python/examples/eps_fit_lorentzian.py
@@ -102,6 +102,8 @@ def fit_medium(
     for m in range(num_repeat):
         # Each of the three parameters per term is seeded in [1, 10).
         # Note: for a lossless material γ optimizes toward zero.
+        # FIXME: the ranges for w_n and g_n should depend on the frequency range,
+        #        and the range for a_n should depend on the scale of the ε data
         p_rand = 10 ** rng.random(3 * num_lorentzians)
         params, err = lorentzfit(p_rand, freqs, eps)
         params_str = "( " + ", ".join(f"{p:.4f}" for p in params) + " )"
@@ -175,9 +177,9 @@ def parse_args() -> argparse.Namespace:
     parser.add_argument(
         "--eps-inf",
         type=float,
-        default=1.1,
-        help="instantaneous (infinite-frequency) permittivity. Must be > 1.0 for "
-        "stability; choose so that min(real(eps - eps_inf)) is ~1.0 (default: 1.1)",
+        default=1.0,
+        help="instantaneous (infinite-frequency) permittivity. Must be ≥ 1.0 for "
+        "stability with default Meep Courant factor (default: 1.0)",
     )
     parser.add_argument(
         "--num-lorentzians",
--- a/python/examples/eps_fit_lorentzian.py
+++ b/python/examples/eps_fit_lorentzian.py
@@ -3,7 +3,7 @@ optionally, Drude) susceptibility terms and build a `Medium` from the result.
 
 The complex permittivity is modeled as
 
-    ε(f) = ε_inf + Σ_n  A_n / (ω_n² −  f² − i f γ_n)
+    ε(f) = ε_inf + Σ_n  A_n / (ω_n² − f² − i f γ_n)
 
 and the parameters (A_n, ω_n, γ_n) are found by nonlinear least-squares fitting
 to tabulated data using SciPy (scipy.optimize.least_squares). The fit is
@@ -82,7 +82,7 @@ def lorentzfit(
         gtol=tol,
         max_nfev=maxeval,
     )
-    # least_squares minimizes 0.5 Σ residual²; repot the full Σ|Δ|².
+    # least_squares minimizes 0.5 Σ residual²; report the full Σ|Δ|².
     return result.x, 2 * result.cost
 
 
--- a/python/examples/eps_fit_lorentzian.py
+++ b/python/examples/eps_fit_lorentzian.py
@@ -1,194 +1,249 @@
+"""Fit a complex refractive index profile to a sum of Lorentzian (and,
+optionally, Drude) susceptibility terms and build a `Medium` from the result.
+
+The complex permittivity is modeled as
+
+    ε(f) = ε_inf + Σ_n  A_n / (ω_n² −  f² − i f γ_n)
+
+and the parameters (A_n, ω_n, γ_n) are found by nonlinear least-squares fitting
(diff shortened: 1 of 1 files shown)


oskooi:
Output of `python3 eps_fit_lorentzian.py --help`:
```
usage: eps_fit_lorentzian.py [-h] [--eps-inf EPS_INF] [--num-lorentzians NUM_LORENTZIANS] [--wl-min WL_MIN]
                             [--wl-max WL_MAX] [--num-repeat NUM_REPEAT] [--seed SEED] [--output OUTPUT]
                             [csv]

Fit a complex refractive index profile to a sum of Lorentzian (and,
optionally, Drude) susceptibility terms and build a `Medium` from the result.

The complex permittivity is modeled as

    ε(f) = ε_inf + Σ_n  A_n / (ω_n² − f² − i f γ_n)

and the parameters (A_n, ω_n, γ_n) are found by nonlinear least-squares fitting
to tabulated data using SciPy (scipy.optimize.least_squares). The fit is
repeated from several random starting points to avoid poor local minima.

Usage:
    python eps_fit_lorentzian.py mymaterial.csv --num-lorentzians 2 \
        --eps-inf 1.0 --wl-min 400 --wl-max 700

The input CSV has three comma-separated columns and no header:
    wavelength (nm), real(n), imag(n)

positional arguments:
  csv                   input CSV: wavelength (nm), real(n), imag(n) (default: mymaterial.csv)

options:
  -h, --help            show this help message and exit
  --eps-inf EPS_INF     instantaneous (infinite-frequency) permittivity. Must be > 1.0 for stability; choose so
                        that min(real(eps - eps_inf)) is ~1.0 (default: 1.1)
  --num-lorentzians NUM_LORENTZIANS
                        number of Lorentzian terms in the fit (default: 2)
  --wl-min WL_MIN       minimum wavelength to fit, in nm (default: 400)
  --wl-max WL_MAX       maximum wavelength to fit, in nm (default: 700)
  --num-repeat NUM_REPEAT
                        number of random restarts of the local optimization (default: 30)
  --seed SEED           seed for the random restarts, for reproducitiblity (default: 0)
  --output OUTPUT       filename for the comparison plot (default: eps_fit_sample.png)

Elapsed run time = 0.0124 s
```

python/examples/eps_fit_lorentzian.py
```diff
@@ -1,194 +1,249 @@
...
+    parser.add_argument(
+        "csv",
+        nargs="?",
+        default="mymaterial.csv",
+        help="input CSV: wavelength (nm), real(n), imag(n) (default: mymaterial.csv)",
+    )
+    parser.add_argument(
+        "--eps-inf",
+        type=float,
+        default=1.1,
```
stevengj: Why not 1.0?  That is the physical choice (all physical materials theoretically should have $\varepsilon(\infty) = 1$.)

stevengj:
```suggestion
        default=1.0,
```

python/examples/eps_fit_lorentzian.py
```diff
@@ -1,194 +1,249 @@
...
+        help="instantaneous (infinite-frequency) permittivity. Must be > 1.0 for "
+        "stability; choose so that min(real(eps - eps_inf)) is ~1.0 (default: 1.1)",
```
stevengj:
```suggestion
        help="instantaneous (infinite-frequency) permittivity. Must be ≥ 1.0 for "
        "stability with default Meep Courant factor (default: 1.0)",
```

python/examples/eps_fit_lorentzian.py
```diff
@@ -1,194 +1,249 @@
...
-    # Find the best performing set of parameters.
-    idx_opt = np.where(np.min(mins) == mins)[0][0]
-    popt_str = "( " + ", ".join(f"{prm:.4f}" for prm in ps[idx_opt]) + " )"
-    print(f"optimal:, {popt_str}, {mins[idx_opt]:.6f}")
-
-    # Define a `Medium` class object using the optimal fitting parameters.
-    E_susceptibilities = []
-
+        # Each of the three parameters per term is seeded in [1, 10).
+        # Note: for a lossless material γ optimizes toward zero.
```
stevengj, line 104:
```suggestion
        # Note: for a lossless material γ optimizes toward zero.
        # FIXME: the ranges for w_n and g_n should depend on the frequency range,
        #        and the range for a_n should depend on the scale of the ε data
```

PR #3262, opened 2026-08-09
Fix Python bindings build with SWIG >= 4.1 (Python 2 C-API removal)

jin-castle:
SWIG 4.1 removed the `PyInt_*`/`PyString_*` backward-compatibility defines from
its generated Python runtime, so regenerating the wrappers with any recent SWIG
fails to compile under Python 3:

```
typemap_utils.cpp:507:18: error: 'PyInt_AsLong' was not declared in this scope
numpy.i (pytype_string): error: 'PyString_Check' was not declared in this scope
```

Two small fixes:

- `python/numpy.i`: guard the `pytype_string()` helper for Python 3 and define
  `PyInt_Check`/`PyInt_AsLong` in terms of `PyLong` (this file predates the
  numpy upstream Python-3 rework; this is the minimal change rather than a
  wholesale refresh of `numpy.i`).
- `python/typemap_utils.cpp`: one stray direct `PyInt_AsLong()` call; use the
  existing `PyInteger_AsLong` compatibility macro defined at the top of the
  file.

No functional change to the generated bindings. Verified by regenerating and
compiling the wrappers with SWIG 4.1.1 and 4.5.0 under Python 3.12 (both
previously failed; both now build and pass the Python test suite subset we ran).
CI with SWIG 4.0 is unaffected.

### DIFF
--- a/python/numpy.i
+++ b/python/numpy.i
@@ -105,6 +105,16 @@
 %fragment("NumPy_Utilities",
           "header")
 {
+%#if PY_VERSION_HEX >= 0x03000000
+  /* Python 3 removed the PyInt/PyString C APIs; map the uses in this file
+     onto their Python 3 equivalents (SWIG >= 4.1 no longer provides these
+     backward-compatibility defines itself). */
+%#ifndef PyInt_Check
+%#define PyInt_Check(x) PyLong_Check(x)
+%#define PyInt_AsLong(x) PyLong_AsLong(x)
+%#endif
+%#endif
+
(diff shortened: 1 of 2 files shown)


oskooi:
Thanks for putting this together. I think it would be better to simply drop support in the SWIG typemaps for Python 2 (rather than provide backward compatibility). This is because Meep currently only supports Python 3. See #3268.

oskooi:
Replaced by #3268.

Issue #2578, opened 2023-07-11
Inconsistent finite-difference and adjoint gradients for a simple 3D example


rafael-fuente:
I'm testing a simple 3D optimization setup, shown in the image below, consisting of a source plane, a design layer, and a monitor plane whose objective function is simply the flux of one polarization. 
The setup is periodic in the X and Y axes, and the Z axis contains two PMLs placed at the top and bottom boundaries. 

![Screenshot 2023-07-12 024227](https://github.com/NanoComp/meep/assets/37953893/e995f883-0dca-4ed6-b596-9b00e5de4da1)

Meep version I'm using is 1.27.0 (parallel), but I get similar results with older versions.

For simplicity, I set up the design region resolution exactly equal to the FDTD resolution, which I randomly initialize and then apply a Gaussian filter to smooth it. 
However, I'm unable to get consistent finite-difference and adjoint gradients no matter what is the simulation resolution or the run time.

I attached the code used below:
``` python
import numpy as np
import meep as mp
import scipy
import matplotlib.pyplot as plt
import meep.adjoint as mpa
from autograd import numpy as npa
from scipy.ndimage import gaussian_filter

resolution  = 20

sx = 2.0
sy = 2.0

d_pml = 0.5
thick_air = 0.5
thick_active = 0.5

mat_active = mp.Medium(index=1.4)
mat_air = mp.Medium(index=1.0)

pml_layers = [mp.PML(thickness=d_pml,direction=mp.Z)]
k_boundary = mp.Vector3(0, 0,  0)

sz = d_pml+thick_air+thick_active+thick_air+d_pml
cell_size = mp.Vector3(sx,sy,sz)


# set up sources
wvl_cen = 0.940
frq_cen = 1/wvl_cen
dfrq = 0.2*frq_cen
source_center = mp.Vector3(0.0,0,-0.5*sz+0.5*thick_air+d_pml)
source_size = mp.Vector3(sx,sy,0)
sources = [mp.Source(mp.GaussianSource(frq_cen, fwidth=dfrq), component=mp.Ex,  amplitude=1.0, center=source_center, size=source_size)]


# set up a reference sim to get the FDTD resolution
sim_ref = mp.Simulation(resolution=resolution,
                    cell_size=cell_size,
                    boundary_layers=pml_layers,
                    k_point=k_boundary,
                    default_material=mat_air,
                    sources=sources)
sim_ref.init_sim()
[...]

PR #3263, opened 2026-08-09
Fix 3D MaterialGrid subpixel smoothing: kernel normalization (#3258) and gradient consistency (#3259)

### LINKED ISSUE #3258: MaterialGrid subpixel smoothing kernel integrates to 4/3 in 3D (integer division in get_uproj_w)
### Summary

The 3D branch of `get_uproj_w()` in `src/meepgeom.cpp` normalizes the `MaterialGrid`
subpixel-smoothing kernel by `4 / 3 * pi * rad^3`. `4 / 3` is integer division and
evaluates to `1`, so the denominator is `pi*rad^3` instead of the sphere volume
`(4/3)*pi*rad^3`, and the kernel integrates to **4/3 instead of 1**.

https://github.com/NanoComp/meep/blob/master/src/meepgeom.cpp#L1214-L1223

```c
  if (mgva->dim == meep::D1)
    w = 1 / (2 * mgva->rad);
  else if (mgva->dim == meep::D2 || mgva->dim == meep::Dcyl)
    w = 2 * sqrt(mgva->rad * mgva->rad - x0 * x0) / (meep::pi * mgva->rad * mgva->rad);
  else if (mgva->dim == meep::D3)
    w = meep::pi * (mgva->rad * mgva->rad - x0 * x0) /
        (4 / 3 * meep::pi * mgva->rad * mgva->rad * mgva->rad);   // <-- 4 / 3 == 1
```

Analytically, with `r = 1`:

| dim | ∫ w dx |
|---|---|
| 1D | 1 |
| 2D | 1 |
| **3D** | **4/3** |
| 3D with `4.0 / 3.0` | 1 |

The 1D and 2D branches are correct, so this only ever affected 3D.

### Effect

Wherever the material grid has a nonzero gradient the smoothed tensor is scaled:
the interface-parallel components by 4/3 and the perpendicular component by 3/4.
`fallback_chi1inv_row()` only skips smoothing when `|grad u| < 1e-8`, so for an
isotropic `MaterialGrid` this covers essentially the whole design region, not just
interface pixels. For a silicon design that is a ~30% error in epsilon.

### Reproducing

Removing the material contrast isolates the normalization. Smoothing builds an
anisotropic tensor — harmonic mean along the level-set normal, arithmetic mean
across it — and with real contrast those differ, which moves the ratio for
legitimate reasons and masks the signal. With `medium1 ≈ medium2` both means
collapse onto the same value, so

```
eps(do_averaging=True) / eps(do_averaging=False)  ==  ∫w
```

reads the normalization off directly. A `beta=0` linear ramp in the weights keeps
`|grad u|` constant and nonzero so the smoothing path actually runs.

<details>
<summary>script</summary>

```python
import numpy as np
import meep as mp

DESIGN, PAD, N, RES = 1.0, 0.6, 20, 20

def eps(dim, do_averaging, n1, n2):
    ramp = np.linspace(0.25, 0.75, N)
    weights = np.repeat(ramp[:, None], N, axis=1)[:, :, None]
    grid = mp.MaterialGrid(mp.Vector3(N, N, 1),
                           mp.Medium(index=n1), mp.Medium(index=n2),
                           weights=weights, do_averaging=do_averaging, beta=0)
    sim = mp.Simulation(
        cell_size=mp.Vector3(DESIGN + 2*PAD, DESIGN + 2*PAD,
                             0 if dim == 2 else DESIGN + 2*PAD),
        resolution=RES, default_material=mp.Medium(index=n1),
        geometry=[mp.Block(center=mp.Vector3(),
                           size=mp.Vector3(DESIGN, DESIGN,
                                           mp.inf if dim == 2 else DESIGN),
                           material=grid)],
        dimensions=dim, eps_averaging=True)
    sim.init_sim()
    e = DESIGN * 0.8
    return np.asarray(sim.get_array(component=mp.Dielectric, center=mp.Vector3(),
                                    size=mp.Vector3(e, e, 0 if dim == 2 else e)))

for label, (n1, n2) in {"flat": (1.44, 1.4414), "silicon": (1.44, 3.48)}.items():
    for dim in (2, 3):
        r = eps(dim, True, n1, n2) / eps(dim, False, n1, n2)
        print(f"{label:>8} dim {dim}: {np.median(r):.8f}")
```

</details>

Measured on v1.31.0:

| contrast | dim | v1.31.0 | with `4.0 / 3.0` |
|---|---|---|---|
| none (n1≈n2) | 2 | 1.000000 | 1.000000 |
| none | 3 | **1.058823** | **1.000000** |
| Si/SiO2 | 2 | 0.760190 | 0.7601899795 |
| Si/SiO2 | 3 | **0.732596** | **0.7601899790** |

`1.058823` is `18/17`, the closed form of `3 / (K + 2/K)` at `K = 4/3` (meep's
`Dielectric` output being `1 / mean(chi1inv_diag)`), matching to six digits.

The last row is an independent check: for a design extruded along z the level-set
normal is purely in-plane, so 3D smoothing must reproduce the 2D epsilon exactly.
After the fix it does, to nine digits. Before, it did not.

### Fix

```diff
--- a/src/meepgeom.cpp
+++ b/src/meepgeom.cpp
   else if (mgva->dim == meep::D3)
     w = meep::pi * (mgva->rad * mgva->rad - x0 * x0) /
-        (4 / 3 * meep::pi * mgva->rad * mgva->rad * mgva->rad);
+        (4.0 / 3.0 * meep::pi * mgva->rad * mgva->rad * mgva->rad);
```

Verified against v1.31.0 (`3e7d486`), also present on master (`b3b8aa9`). The full
upstream test suite passes before and after, and 2D/cylindrical results are
unchanged bit-for-bit — `test_adjoint_solver.py` (13 cases, all 2D),
`test_adjoint_cyl.py`, `test_material_grid.py`, `test_get_epsilon_grid.py`,
`test_user_defined_material.py` all pass on both builds.

I'm happy to open a PR with this plus a regression test if that's useful.

### Note on test coverage

Every adjoint test in the tree is 2D or cylindrical, so nothing exercised the D3
branch of `get_uproj_w`. I have a `test_subpixel_3d.py` that fails on v1.31.0 and
passes with the fix; glad to include it.

### LINKED ISSUE #3259: eps_averaging=False does not reach the adjoint gradient: dA/du differentiates a smoothed operator
### Summary

With `eps_averaging=False`, the forward solve correctly drops subpixel smoothing but
the adjoint gradient still differentiates the *smoothed* operator, so the returned
gradient is not the derivative of the simulation that was run.

`structure_chunk::set_chi1inv()` honors the flag:

https://github.com/NanoComp/meep/blob/master/src/anisotropic_averaging.cpp#L218

```c
  if (!use_anisotropic_averaging) maxeval = 0;
```

but `set_materials_from_geom_epsilon()` stores the raw value for later use by the
gradient:

https://github.com/NanoComp/meep/blob/master/src/meepgeom.cpp#L2015-L2017

```c
  // store for later use in gradient calculations
  geps->tol = tol;
  geps->maxeval = maxeval;          // <-- ignores use_anisotropic_averaging
```

and `get_material_gradient()` finite-differences `eff_chi1inv_row()` with that
`geps->maxeval`:

https://github.com/NanoComp/meep/blob/master/src/meepgeom.cpp#L2682-L2684

So the forward/adjoint solves see an unsmoothed operator while `dA/du` is taken of a
smoothed one.

### Reproducing

3D, `MaterialGrid(Nx, Ny, 1)` extruded through a 220 nm silicon slab, eigenmode
ports, resolution 20, `eps_averaging=False`.

Because `eps_averaging=False` already disables `MaterialGrid` smoothing in the
structure, toggling `do_averaging` must change *nothing*. The objective and the
finite differences confirm the forward solve is indeed identical to the last digit.
Only the adjoint moves:

| `do_averaging` | objective | dir 0 FD | dir 0 adjoint | dir 1 FD | dir 1 adjoint |
|---|---|---|---|---|---|
| `False` | 151.06613556484936 | +1.203046e-01 | +1.196406e-01 | +4.417362e+00 | +4.417737e+00 |
| `True` | 151.06613556484936 | +1.203046e-01 | **−4.300552e+00** | +4.417362e+00 | **+1.387950e+01** |

Relative error 68–103%, with a sign flip in one direction.

There is a second path with the same problem: `python/meep.i::_set_materials()` only
reaches `set_materials_from_geom_epsilon()` when `set_materials` is true, so on the
other branch `geps->tol/maxeval` are never assigned and keep the header defaults
`DEFAULT_SUBPIXEL_TOL` / `DEFAULT_SUBPIXEL_MAXEVAL`.

### Fix

```diff
--- a/src/meepgeom.cpp
+++ b/src/meepgeom.cpp
   // store for later use in gradient calculations
   geps->tol = tol;
-  geps->maxeval = maxeval;
+  geps->maxeval = use_anisotropic_averaging ? maxeval : 0;
```

plus the same assignment on the `set_materials == false` branch of
`_set_materials()` in `python/meep.i`.

With the fix, the `do_averaging=True` row becomes identical to the
`do_averaging=False` row — objective, gradient norm, and the finite-difference
agreement (4.54e-05 along the gradient) all match exactly, which is what
`eps_averaging=False` should mean.

Setting `maxeval = 0` does not zero the gradient: `eff_chi1inv_matrix()` still goes
through `get_material_pt()` → `material_grid_val()` → projection → epsilon
interpolation, so `d(eps)/du` remains well defined.

Verified against v1.31.0 (`3e7d486`), also present on master (`b3b8aa9`). The full
upstream suite passes before and after.

### Possibly related

#2578 reports inconsistent FD/adjoint gradients for a 3D example and the reproducer
there sets `eps_averaging=False`. That issue has been open since 2023 without a
root cause; this may be it, though I haven't rerun their exact setup.

Happy to open a PR.

jin-castle:
Fixes #3258 and fixes #3259, with regression tests. (Issue text contains the
full derivations; summary below.)

**Defect A (#3258) — kernel normalization.** `get_uproj_w()`'s 3D branch
divides by `4 / 3 * pi * rad^3`, where `4 / 3` is integer division. The
smoothing kernel therefore integrates to 4/3 instead of 1, scaling the
smoothed tensor everywhere the material grid has a nonzero gradient
(interface-parallel components by 4/3, perpendicular by 3/4) — for an
isotropic `MaterialGrid` that is essentially the whole design region, not just
interface pixels. One-character fix: `4.0 / 3.0`. The 1D/2D branches were
already correct.

**Defect B (#3259) — eps_averaging not propagated to the gradient.**
`structure_chunk::set_chi1inv()` zeroes `maxeval` when averaging is off, but
`set_materials_from_geom_epsilon()` stored the raw `maxeval` into the
`geom_epsilon` used by `get_material_gradient()`, and the `_set_materials`
Python wrapper left the header defaults in place when it skipped
`set_materials_from_geom_epsilon` entirely. The adjoint therefore
finite-differenced a *smoothed* operator that the forward solve never used.
Measured: directional finite-difference vs adjoint disagreement of a factor
~78 in 3D with `do_averaging=True, eps_averaging=False` — the reason 3D
adjoint users have had to disable smoothing entirely.

**Convenience switch.** `Simulation(allow_3d_subpixel=False)` turns off
`do_averaging` on every reachable `MaterialGrid` at structure initialization
in a 3D run, so smoothing can be disabled in one place instead of per-grid.
Default `True` leaves behavior unchanged.

**Tests** (`python/tests/test_subpixel_3d.py`): analytic kernel-normalization
checks in 1D/2D/3D, and 3D directional finite-difference vs adjoint gradient
checks for smoothing on, off, and the mixed case. Before the fix the mixed
case is off by ~7800% and the smoothing-on case silently solves the
mis-normalized epsilon; after it, all three agree.

**Please merge #3264 before this one.** The gradient checks above measure a
quantity that #3264 fixes independently, and on this branch alone they are
only accurate while the design region sits inside a single chunk. The
`fd/adjoint` ratio for the smoothing-on case, measured on this branch by
forcing the split with `Simulation(num_chunks=...)`:

| chunks | 1 | 2 | 4 |
|---|---|---|---|
| this branch alone | 0.99982 | 0.99845 | **0.88371** |
| with #3264 applied | 0.99982 | 0.99982 | 0.99982 |

(Forcing the split serially reproduces the `mpirun` numbers exactly; that was
checked against `-np` 1, 2, 4 and 8 on the pre-fix code.) CI runs its MPI jobs
at `-np 2`, so the checks pass here, but at four chunks the smoothing-on case
misses its 1% tolerance by an order of magnitude.

That is #2578 rather than a defect in this PR. The underlying gradient vector
is 12-44% off in L2 at every split above one chunk; the directional
finite-difference check largely misses this because it probes along `g/|g|`,
i.e. along the erroneous gradient itself. With #3264 in place the ratios are
identical at every chunk count, and the 1% tolerance in these tests could be
tightened again.

### DIFF
--- a/python/Makefile.am
+++ b/python/Makefile.am
@@ -94,6 +94,7 @@ TESTS =                                        \
     $(TEST_DIR)/test_special_kz.py             \
     $(TEST_DIR)/test_source.py                 \
     $(TEST_DIR)/test_stop_when_flux_decayed.py \
+    $(TEST_DIR)/test_subpixel_3d.py            \
     $(TEST_DIR)/test_timing_measurements.py    \
     $(TEST_DIR)/test_user_defined_material.py  \
     $(TEST_DIR)/test_verbosity_mgr.py          \
--- a/python/meep.i
+++ b/python/meep.i
@@ -1939,6 +1939,14 @@ meep_geom::geom_epsilon* _set_materials(meep::structure * s,
         meep_geom::set_materials_from_geom_epsilon(s, geps, use_anisotropic_averaging, tol,
                                              maxeval,alist);
     }
+    else {
+        /* set_materials_from_geom_epsilon() is what normally records these for the
+           gradient calculation. Skipping it leaves the geom_epsilon on its header
+           defaults (DEFAULT_SUBPIXEL_TOL/MAXEVAL), so the adjoint would smooth no
+           matter what the caller asked for. */
+        geps->tol = tol;
+        geps->maxeval = use_anisotropic_averaging ? maxeval : 0;
+    }
 
     if (meep::verbosity > 1 && !split_chunks_evenly && set_materials) {
       int num_procs = meep::count_processors();
--- a/python/simulation.py
+++ b/python/simulation.py
@@ -1246,6 +1246,7 @@ class Simulation:
         progress_interval: float = 4,
         subpixel_tol: float = 1e-4,
         subpixel_maxeval: int = 100000,
+        allow_3d_subpixel: bool = True,
         loop_tile_base_db: int = 0,
         loop_tile_base_eh: int = 0,
         ensure_periodicity: bool = True,
@@ -1395,6 +1396,13 @@ class Simulation:
           effects and irregular
           convergence](Subpixel_Smoothing.md#what-happens-when-subpixel-smoothing-is-disabled).
 
+        + **`allow_3d_subpixel` [ `boolean` ]** — If `False`, then in a 3d simulation
+          any `MaterialGrid` with `do_averaging=True` has its level-set subpixel
+          smoothing turned off at structure-initialization time. This is a
+          convenience switch so a 3d run can drop MaterialGrid smoothing without
+          editing every grid individually; the default `True` leaves behavior
+          unchanged. Has no effect in 1d/2d/cylindrical.
+
         + **`force_complex_fields` [ `boolean` ]** — By default, Meep runs its simulations
           with purely real fields whenever possible. It uses complex fields which require
           twice the memory and computation if the `k_point` is non-zero or if `m` is
@@ -1502,6 +1510,7 @@ class Simulation:
         self.eps_averaging = eps_averaging
         self.subpixel_tol = subpixel_tol
         self.subpixel_maxeval = subpixel_maxeval
+        self.allow_3d_subpixel = allow_3d_subpixel
         self.loop_tile_base_db = loop_tile_base_db
         self.loop_tile_base_eh = loop_tile_base_eh
         self.ensure_periodicity = ensure_periodicity
@@ -2002,11 +2011,45 @@ class Simulation:
 
         return stats
 
+    def _iter_material_grids(self):
+        """Yield every MaterialGrid reachable from this simulation's materials."""
+        candidates = list(self.geometry or [])
+        for obj in candidates:
+            mat = getattr(obj, "material", None)
+            if isinstance(mat, mp.MaterialGrid):
+                yield mat
+        if isinstance(self.default_material, mp.MaterialGrid):
+            yield self.default_material
+        for mat in self.extra_materials or []:
+            if isinstance(mat, mp.MaterialGrid):
+                yield mat
+
+    def _apply_3d_subpixel_policy(self):
+        """Honor `allow_3d_subpixel=False` by disabling MaterialGrid smoothing.
+
+        A single simulation-level switch, so a 3d run can drop level-set subpixel
+        smoothing without editing every MaterialGrid it happens to contain. The
+        default is True, which leaves behavior exactly as it was.
+        """
+        if self.allow_3d_subpixel or self.dimensions != 3:
+            return
+        disabled = 0
+        for grid in self._iter_material_grids():
+            if grid.do_averaging:
+                grid.do_averaging = False
+                disabled += 1
+        if disabled and verbosity.meep > 0:
+            print(
+                f"allow_3d_subpixel=False: disabled do_averaging on {disabled} "
+                "MaterialGrid(s)"
+            )
+
     def _init_structure(self, k=False):
         if verbosity.meep > 0:
             print("-" * 11)
             print("Initializing structure...")
 
+        self._apply_3d_subpixel_policy()
         gv = self._create_grid_volume(k)
         sym = self._create_symmetries(gv)
         br = _create_boundary_region_from_boundary_layers(self.boundary_layers, gv)
--- a/python/tests/test_subpixel_3d.py
+++ b/python/tests/test_subpixel_3d.py
@@ -22,6 +22,11 @@ import numpy as np
 import meep as mp
 
 
+# Allow finite-difference and MPI chunking noise while remaining far below the
+# 70-100% disagreement caused by the regression covered by these tests.
+FD_REL_TOL = 1e-2
+
+
 def _epsilon(dim, do_averaging, n1, n2, resolution=20, n=20):
     """Dielectric array over a MaterialGrid carrying a linear ramp in u."""
     design, pad = 1.0, 0.6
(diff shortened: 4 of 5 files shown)


stevengj:
@smartalecH and @romanodev, can you take a look at this?  Is it done correctly in the [SSP repository](https://github.com/NanoComp/SSP)?

smartalecH:
@stevengj note that this is the cpp path, which we essentially abandoned when doing the paper (all of that was done with python/jax and this feature was disabled). So both the python path and the SSP repo should be fine. 

That being said, it'd be great to get the cpp path working because it would allow us to eg interpolate to a high resolution design using the meep api (that was the main reason we went down this path). 

This LGTM. We should explore what else we can do to revamp meepgeom (i imagine bugs like this are just the tip of the iceberg)

Issue #1984, opened 2022-03-08
Incorrect adjoint gradients with overlapping design variables


mochen4:
When the material grids of design variables are overlapped, e.g. [the eigenmode tutorial example](https://github.com/NanoComp/meep/blob/master/python/examples/adjoint_optimization/01-Introduction.ipynb), the adjoint gradients don't agree with the finite difference gradients. 

In the aforementioned example, the gradients are correct if only one block of the design region is kept in the`` geometry`` but not both:
```
geometry=[
...
mp.Block(center=design_region.center, size=design_region.size, material=design_variables), # design region
mp.Block(center=design_region.center, size=design_region.size, material=design_variables,
             e1=mp.Vector3(x=-1).rotate(mp.Vector3(z=1), np.pi/2), e2=mp.Vector3(y=1).rotate(mp.Vector3(z=1), np.pi/2))
]
```

mochen4:
> IIRC, the error is a constant error, right? meaning we are just off by a scale factor (i.e. it's not as if the entire gradient is just wrong)?

Yes, I think the gradients are off by a scale factor, but that scale varies a bit with different random structures.

smartalecH:
I'm pretty sure the issue is due to the following snippet introduced in #1780.

https://github.com/NanoComp/meep/blob/11545a1b859222df2c404370673115666b589c22/src/meepgeom.cpp#L2758-L2771

MarkMa1990:
> I'm pretty sure the issue is due to the following snippet introduced in #1780.
> 
> https://github.com/NanoComp/meep/blob/11545a1b859222df2c404370673115666b589c22/src/meepgeom.cpp#L2758-L2771

Yes, exactly. In fact, I just compared different versions, it seems that from version 1.21.0, the material derivative to the design variable u is changed: 

![image](https://user-images.githubusercontent.com/10673026/169806823-297068d1-6022-4da2-b3fc-1d5ad2d5ff1a.png)
![image](https://user-images.githubusercontent.com/10673026/169806997-de14c893-0fa7-435d-93b5-9b7b803c5ba9.png)
![image](https://user-images.githubusercontent.com/10673026/169807228-06cb63e6-6a14-47bc-a970-ce464b151c7c.png)

PR #3264, opened 2026-08-09
Make the adjoint design-region gradient independent of the MPI chunk division

### LINKED ISSUE #2578: Inconsistent finite-difference and adjoint gradients for a simple 3D example

jin-castle:
Fixes #2578 (adjoint gradients inconsistent with finite differences in 3D).

**Symptom.** The adjoint gradient of a `MaterialGrid` design region depends on
the number of MPI processes while the objective value does not. Measured on a
20×20×6 design grid (3D waveguide, both with and without subpixel smoothing):

- `f0` identical to 8+ digits at every process count;
- gradient max-norm difference between `-np 1` and `-np 8`: **57%** of
  `|g|_max` (smoothing off) / 18% (smoothing on), spread over ~80% of the
  design nodes in chunk-sized patches;
- directional finite-difference checks that pass at `-np 1` (ratio 1.0002)
  fail under MPI (ratios 0.86–1.22 at `-np 16`).

**Root causes** — all in `material_grids_addgradient()`, which computed the
result from per-chunk DFT views:

1. chunks of *different components* were paired by list index
   (`forward_dft_chunks[ci_forward][cur_chunk]`); the per-component chunk sets
   can legitimately differ near the monitor edges, after which every
   subsequent pair associates the wrong regions, and cross terms were dropped
   entirely when the counts differed (`cur_chunk >= num_f_chunks → continue`);
2. neighboring-point lookups bounds-checked only the chunk-local *linear*
   index, so an out-of-range point in one dimension could alias to a valid
   index of a different point instead of returning zero;
3. the ±1-pixel padding of `persist` dft chunks is clamped to the owning
   fields chunk and cannot cross process boundaries, so the data those
   lookups need does not exist at rank boundaries;
4. the same-component term indexed the forward chunk with the adjoint loop
   index, relying on the chunk-alignment assumption of (1).

**Fix.** Gather each monitor component onto its full (padded) design-region
grid first — monitor-owned points authoritatively (the `is_old..ie_old`
decomposition is disjoint), the padding ring as a multiplicity-average of the
ghost copies so its coverage does not depend on the chunk division — then run
the same per-point math on this chunk-independent view, with the work strided
across processes and per-dimension bounds checks on neighbor lookups. Memory
cost is the gathered design-region DFTs (6 complex arrays over the monitor
grid per frequency), negligible for realistic design regions.

**Validation.**

| check | before | after |
|---|---:|---:|
| `-np 8` vs `-np 1`, 20×20×6, smoothing off | 57% | 7×10⁻¹² |
| `-np 8/16/32/128` vs `-np 1` (both smoothing cases) | up to 57% | ≤7×10⁻¹² |
| FD-vs-adjoint at `-np 16` and `-np 128` (six grid configurations incl. incommensurate/offset/coarse/fine grids) | ratios 0.86–1.22 | identical to serial (1.000 ± FD noise) |
| serial behavior | — | preserved (`test_subpixel_3d`, adjoint E2E unchanged) |

### DIFF
--- a/src/meepgeom.cpp
+++ b/src/meepgeom.cpp
@@ -2851,8 +2851,7 @@ struct gathered_dft_component {
   std::vector<std::complex<double> > data;
   size_t npts = 0;
 
-  gathered_dft_component()
-      : gis(meep::D3), gie(meep::D3), pis(meep::D3), pie(meep::D3) {}
+  gathered_dft_component() : gis(meep::D3), gie(meep::D3), pis(meep::D3), pie(meep::D3) {}
 };
 
 bool ivec_in_range(const meep::ivec &p, const meep::ivec &lo, const meep::ivec &hi,
@@ -2955,9 +2954,8 @@ void gather_dft_component(const std::vector<meep::dft_chunk *> &chunks, size_t n
   out.empty = false;
 }
 
-std::complex<double> gathered_value(const gathered_dft_component &g,
-                                    const meep::grid_volume &gvp, const meep::ivec &p, size_t nf,
-                                    size_t f_i, meep::ndim dim) {
+std::complex<double> gathered_value(const gathered_dft_component &g, const meep::grid_volume &gvp,
+                                    const meep::ivec &p, size_t nf, size_t f_i, meep::ndim dim) {
   // per-dimension bounds check: outside the gathered region is exactly zero
   // (never index-aliased into another point)
   if (!ivec_in_range(p, g.pis, g.pie, dim)) return 0.0;
@@ -3125,10 +3123,9 @@ void material_grids_addgradient(double *v, size_t ng, size_t nf,
               std::complex<double> fwd_avg = std::complex<double>(0.5, 0) * (fwd1 + fwd2);
               meep::vec eps1 = gv[ieps[node]];
               cyl_scale = (gv.dim == meep::Dcyl) ? eps1.r() : 1;
-              material_grids_addgradient_point(v_local + ng * f_i, vec_to_vector3(eps1),
-                                               scalegrad * cyl_scale, geps, adjoint_c, forward_c,
-                                               fwd_avg, std::complex<double>(0.5, 0) * adj,
-                                               frequencies[f_i], gv, du);
+              material_grids_addgradient_point(
+                  v_local + ng * f_i, vec_to_vector3(eps1), scalegrad * cyl_scale, geps, adjoint_c,
+                  forward_c, fwd_avg, std::complex<double>(0.5, 0) * adj, frequencies[f_i], gv, du);
             }
           }
           /********* compute λᵀbᵤ ***************/
--- a/src/meepgeom.cpp
+++ b/src/meepgeom.cpp
@@ -2832,6 +2832,141 @@ void material_grids_addgradient_point(double *v, vector3 p, double scalegrad, ge
   }
 }
 
+namespace {
+/* DFT data for one field component of the design-region monitors, gathered
+   from every chunk on every process onto the full (padded) monitor grid.
+   The gradient computation below then runs on this chunk-independent view,
+   which makes its result independent of the chunk division and hence of the
+   number of MPI processes (the previous per-chunk implementation paired
+   chunks of different components by list index, indexed neighboring points
+   through chunk-local index arithmetic that could alias or silently return
+   zero at chunk boundaries, and dropped cross terms entirely when the
+   per-component chunk counts differed; see NanoComp/meep#2578).
+   Layout matches dft_chunk: frequency f at grid index idx is data[nf*idx+f]. */
+struct gathered_dft_component {
+  bool empty = true;
+  meep::component c = meep::Ex;
+  meep::ivec gis, gie; // monitor-owned global bounds (union over all chunks)
+  meep::ivec pis, pie; // gathered (padded) bounds: owned +- 1 pixel, clamped
+  std::vector<std::complex<double> > data;
+  size_t npts = 0;
+
+  gathered_dft_component()
+      : gis(meep::D3), gie(meep::D3), pis(meep::D3), pie(meep::D3) {}
+};
+
+bool ivec_in_range(const meep::ivec &p, const meep::ivec &lo, const meep::ivec &hi,
+                   meep::ndim dim) {
+  LOOP_OVER_DIRECTIONS(dim, d) {
+    if (p.in_direction(d) < lo.in_direction(d) || p.in_direction(d) > hi.in_direction(d))
+      return false;
+  }
+  return true;
+}
+
+void gather_dft_component(const std::vector<meep::dft_chunk *> &chunks, size_t nf,
+                          meep::grid_volume &gv, gathered_dft_component &out) {
+  // agree on the component and the global monitor bounds across all processes
+  int cint = -1;
+  const int BIG = 1 << 29;
+  int lo[5], hi[5];
+  for (int d = 0; d < 5; ++d) {
+    lo[d] = BIG;
+    hi[d] = -BIG;
+  }
+  for (size_t i = 0; i < chunks.size(); ++i) {
+    cint = (int)chunks[i]->c;
+    LOOP_OVER_DIRECTIONS(gv.dim, d) {
+      lo[(int)d] = std::min(lo[(int)d], chunks[i]->is_old.in_direction(d));
+      hi[(int)d] = std::max(hi[(int)d], chunks[i]->ie_old.in_direction(d));
+    }
+  }
+  cint = meep::max_to_all(cint);
+  if (cint < 0) { // no process has data for this component
+    out.empty = true;
+    return;
+  }
+  out.c = (meep::component)cint;
+  meep::ivec gis(gv.dim), gie(gv.dim), pis(gv.dim), pie(gv.dim);
+  const meep::ivec cis = gv.little_corner() + gv.iyee_shift(out.c);
+  const meep::ivec cie = gv.big_corner() + gv.iyee_shift(out.c);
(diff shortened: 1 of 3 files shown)


stevengj:
Failing single precision tests?

stevengj:
@lxvm, you are running a lot of Meep MPI adjoint jobs, can you look at this?

lxvm:
I tested this PR on my Meep adjoint jobs and can confirm that it dramatically improves the accuracy of the adjoint gradient. I also noticed that changing the number of MPI ranks changed the value of the objective function I calculated and therefore I cannot comment on whether the PR fixes the design-region gradient's dependence on the chunk division because there could be another bug present.

stevengj:
@smartalecH, can you take a look at the PR?

src/meepgeom.cpp
```diff
@@ -2832,6 +2832,139 @@ void material_grids_addgradient_point(double *v, vector3 p, double scalegrad, ge
   }
 }
 
+namespace {
+/* DFT data for one field component of the design-region monitors, gathered
+   from every chunk on every process onto the full (padded) monitor grid.
+   The gradient computation below then runs on this chunk-independent view,
```
stevengj, line 2838: I'm worried that doing it this way hurts the scaling… it would be nice to keep the DFT data parallelized…

smartalecH:
Thanks @jin-castle for pulling this together. You managed to find a few different bugs here. But I'm not confident this is the right approach to resolving them. Here are some thoughts, @stevengj :

1. We definitely have some indexing bugs, and some loose assumptions (e.g. whether component chunks track each other in their respective linked lists). We should absolutely refactor/clean up the code here (although some of the proposed refactoring is more extensive than it needs to be in a single PR).
2. I _don't_ think we need to materialize all the DFT component data to the same process. We added behavior with the `persist` flag that essentially stores an extra array element for each component (and extra "halo") which can be used to access these complicated boundaries. I think there are some bugs with how we access those (particularly in 3D) but we should be able to resolve those easily.

Let me take a stab at making some minimal changes but using your _same test_. Then I'll report back here and iterate as needed.

lxvm:
I can confirm that there is a discrepancy in the calculation of the objective function for my 3d meep optimization problem caused by changing the number of mpi ranks. I can trace it down to a roughly 10% error in the DFT fields, which I should file as a separate issue (unless it is a duplicate of #1051). This pr and #3277 both increased the accuracy of the gradients for a fixed number of MPI ranks

smartalecH:
Wait @lxvm have you rebuilt on master? #3274 was supposed to solve this without needing to serialize the DFT fields.

lxvm:
@smartalecH I also tested on #3277 (stacked on #3274) and found the same discrepancy. I want to try the latest master again anyway in case I missed any commits

smartalecH:
@lxvm got it. Can you provide a MWE to test against? We should figure out what's going on at the core and avoid serialization as much as possible.

lxvm:
I can confirm that the discrepancy I was observing in the forward calculation is a duplicate of #1051. By removing the `geometry_center` from the simulation, now the objective function is consistent to within 14 digits when changing the number of MPI ranks. This was using Meep v1.31. @smartalecH did #3274 make any changes to `geometry_center`? It sounds like I was hitting a different bug

smartalecH:
@lxvm does this PR resolve the discrepancy? I'm not sure how, since it only touches the gradient recombination step and the behavior you're describing is on the forward solve. 

I'm trying to understand what improvements you are seeing with this PR over `master` (and why).

lxvm:
@smartalecH I think the latest master supercedes this PR and I find that both branches give nearly identical adjoint gradients. The implementation in master is better because it keeps the data distributed. I just wanted to return to this discussion to clarify why I was seeing a problem with the forward solve, which it turns out is because my simulations where running into #1051. I also need #3277 in order for my gradients to be accurate enough because I am also running into #1984. Sorry for the confusion!

smartalecH:
@jin-castle I'm going to close this, since we resolved it without needing to serializing. Thanks for surfacing the issue!

PR #3269, opened 2026-08-09
Cleanups and fixes for the SWIG typemaps in `python/typemap_utils.cpp`

oskooi:
Various cleanups and fixes to [`python/typemap_utils.cpp`](https://github.com/NanoComp/meep/blob/master/python/typemap_utils.cpp) including:

**1. Memory corruption (Python-reachable)**

Both in `pymaterial_to_material` / `pymaterial_grid_to_material_grid`, both fixed with one `PyArray_FROMANY(..., NPY_DOUBLE, ...)` call each.

**2. Memory leaks**

`material_free(md)` on both `pymaterial_to_material` failure paths (plus moving four `Py_XDECREF`s above the early return); `delete[] vertices` in `pyprism_to_prism`; `Py_DECREF(py_trans)` in the MultilevelAtom loop.

**3. Unchecked returns**

`Py_XDECREF` where NULL was possible, a NULL guard before `std::string(bytes)`, and `PyErr_Occurred()` checks in `get_attr_int`/`get_attr_dbl` so a stale exception no longer surfaces at an unrelated later call.

**4. Cosmetic**

35 stray null statements, 4 stray `};`, `[[noreturn]]` on `abort_with_stack_trace`, comment typo, `material_grid_object` rename, `Py_ssize_t` for list lengths.

### DIFF
--- a/python/typemap_utils.cpp
+++ b/python/typemap_utils.cpp
@@ -17,8 +17,6 @@
 
 // Utility functions for pymeep typemaps
 
-#include <climits>
-
 #ifndef SWIG_PYTHON_THREAD_SCOPED_BLOCK
 #define SWIG_PYTHON_THREAD_SCOPED_BLOCK SWIG_PYTHON_THREAD_BEGIN_BLOCK
 #endif
@@ -1151,8 +1149,9 @@ static int py_list_to_gobj_list(PyObject *po, geometric_object_list *l) {
   int length = PyList_Size(po);
 
   l->num_items = length;
-  // Value-initialize: geometric object is a POD, so plain new[] would leave the material
-  // points indeterminate and gobj_list_freearg would free garbage.
+  // The trailing () value-initializes. geometric_object is trivially default-constructible,
+  // so plain new[] would leave each material pointer indeterminate and gobj_list_freearg
+  // would pass garbage to material_free.
   l->items = new geometric_object[length]();
 
   for (int i = 0; i < length; i++) {
--- a/python/typemap_utils.cpp
+++ b/python/typemap_utils.cpp
@@ -17,6 +17,8 @@
 
 // Utility functions for pymeep typemaps
 
+#include <climits>
+
 #ifndef SWIG_PYTHON_THREAD_SCOPED_BLOCK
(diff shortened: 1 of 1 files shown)


python/typemap_utils.cpp
```diff
@@ -1030,11 +1049,12 @@ static int pymesh_to_mesh(PyObject *py_mesh, geometric_object *o) {
 
   PyObject *py_verts = PyObject_GetAttrString(py_mesh, "vertices");
   if (!py_verts || !PyList_Check(py_verts)) {
-    meep::abort("Expected Mesh.vertices to be a list\n");
+    PyErr_SetString(PyExc_TypeError, "Expected Mesh.vertices to be a list");
+    Py_XDECREF(py_verts);
```
stevengj, line 1053: It seems to be replacing `meep::abort` (which throws a C++ `runtime_exception` that is caught by SWIG and turned into a generic Python exception) into a more specific Python exception type.

This is fine … but is a little bit different in MPI jobs, where `meep::abort` calls `MPI_Abort` and kills the whole job.    The slight danger here is that, by only throwing a Python exception, you could have an MPI run that hangs because one process halts with an exception while others keep running.  On the other hand, maybe mpi4py takes care of this and calls `MPI_Abort` for uncaught exceptions?

stevengj: Apparently mpi4py [fixes this if you run with `-m mpi4py`](https://mpi4py.readthedocs.io/en/latest/mpi4py.run.html)?  

> To alleviate this issue, [mpi4py](https://mpi4py.readthedocs.io/en/latest/mpi4py.html#module-mpi4py) offers a simple, alternative command line execution mechanism based on using the [-m](https://docs.python.org/3/using/cmdline.html#using-on-cmdline) flag and implemented with the [runpy](https://docs.python.org/3/library/runpy.html#module-runpy) module. To use this feature, Python code should be run passing -m mpi4py in the command line invoking the Python interpreter. In case of unhandled exceptions, the finalizer hook will call MPI_Abort() on the MPI_COMM_WORLD communicator, thus effectively aborting the MPI execution environment.

Arguably this should be done anyway since you could also have uncaught exceptions in pure-Python code that are not thrown by Meep.

PR #3271, opened 2026-08-14
Add `plot1D` function for visualizing layout of 1D simulation

oskooi:
Adds a `plot1D` function for visualizing the layout of 1D simulation. This mirrors the existing [`plot2D`](https://meep.readthedocs.io/en/latest/Python_User_Interface/#data-visualization).

**PEC detection.** A wall is PEC when the cell boundary is metallic — i.e. no `k_point` was given (Meep's default) or `set_boundary(..., mp.Metallic)` was called. Since `fields.boundaries` is only exposed to Python as an opaque SWIG pointer, `Simulation.set_boundary` now records what it set in `self._boundary_conditions` (cleared by `reset_meep`), which `plot1D` consults. This makes the tutorial setup in `antenna_pec_ground_plane_1D.py` (Bloch `k_point` + explicit metallic walls) plot correctly.

Everything is customizable through the same `*_parameters` dict / `show_*` flag convention as `plot2D` (`index_parameters`, `pml_parameters`, `pec_parameters`, `source_parameters`, `monitor_parameters`, `legend_parameters`), and the dispersive-permittivity frequency defaults to the first source's frequency (that lookup is now the shared `get_default_frequency` helper, also used by `plot_eps`).

An example of the output (involving a dielectric slab above a PEC ground plane) produced by `plot1D` is shown below.

layout_1D_plot1D

**Files changed**
- `python/visualization.py` — `plot1D` and its `plot_1d_*` helpers, 1D default-parameter dicts
- `python/simulation.py` — `Simulation.plot1D` with full docstring; boundary-condition tracking
- `python/meep.i` — export `plot1D` from the package
- `python/tests/test_visualization.py` — `test_plot1D` covering legend contents, artist counts, PEC vs. Bloch-periodic, `Absorber`, `set_boundary` override, customization, and the non-1D `ValueError`
- `doc/docs/Python_User_Interface.md{, .in}` — API docs (inserted exactly the block `generate_py_api.py` emits)

**Verification**
- `python/tests/test_visualization.py`: 3 tests pass, serially and under `mpirun -np 2`
- Checked against variants: PML both/one side, `Absorber`, Bloch-periodic, explicit metallic wall, extended sources, dispersive material (Si reads 3.478 at 1.55 μm), nonzero `geometry_center`, `dimensions=1` and unset-`dimensions` 1D cells

### DIFF
--- a/src/boundaries.cpp
+++ b/src/boundaries.cpp
@@ -87,6 +87,10 @@ void fields::set_boundary(boundary_side b, direction d, boundary_condition cond)
   }
 }
 
+boundary_condition fields::get_boundary(boundary_side b, direction d) const {
+  return boundaries[b][d];
+}
+
 void fields::use_bloch(direction d, complex<double> kk) {
   k[d] = kk;
   for (int b = 0; b < 2; b++)
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -653,6 +653,32 @@ section for valid `side`, `direction`, and `boundary_condition` values.
 
 ---
 
+<a id="Simulation.get_boundary"></a>
+
+<div class="class_members" markdown="1">
+
+```python
+def get_boundary(side, direction):
+```
+
+<div class="method_docstring" markdown="1">
+
+Returns the condition of the boundary on the specified side in the specified
+direction. This is the counterpart of `set_boundary`. See the [Constants
+(Enumerated Types)](#constants-enumerated-types) section for valid `side`,
+`direction`, and `boundary_condition` values.
+
+The default boundary condition is `mp.Metallic` (i.e., a perfect electric
+conductor), unless a `k_point` has been specified in which case it is
+`mp.Periodic` (i.e., Bloch-periodic).
+
+</div>
+
+</div>
+
+
+---
+
 <a id="Simulation.phase_in_material"></a>
 
 <div class="class_members" markdown="1">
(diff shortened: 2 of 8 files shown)


stevengj:
Rather than add a Python mirror of the boundary information, I'd rather just add a method to `fields::` that exposes this information, e.g. a `fields::get_boundary` function (mirroring `fields::set_boundary`) or something like that.

oskooi:
> Rather than add a Python mirror of the boundary information, I'd rather just add a method to `fields::` that exposes this information, e.g. a `fields::get_boundary` function (mirroring `fields::set_boundary`) or something like that.

The Python `Simulation._boundary_condition` object is removed. The boundary condition is now read from C++ using a newly added `fields::get_boundary`.

Two things worth flagging:

1. **Fixed a deadlock.** `get_boundary` can call `init_sim()`, which is collective. The original `plot_1d_pec` had an early `if not mp.am_master(): return ax` before the query, so under `mpirun -np 2` only rank 0 entered `init_sim` and the run hung. The query now happens on all ranks before the master-only drawing, with a comment saying why.
2. **The plot got more accurate.** Querying C++ revealed that the original Python mirror was wrong for the case of `set_boundary(mp.Low, mp.Z, mp.Metallic)` on a Bloch-periodic cell which makes *both* walls metallic, because `fields::set_boundary` forces the opposite side to match when it was `Periodic`. The old code drew PEC only on the low side. The updated schematic below confirms this behavior.                                                                                                                                                                                                                                                                                                                              

<img width="803" height="554" alt="layout_1D_plot1D" src="https://github.com/user-attachments/assets/5a5adf77-5c22-4d9e-bfc4-24b2f7c4fcf3" />

src/boundaries.cpp
```diff
@@ -87,6 +87,10 @@ void fields::set_boundary(boundary_side b, direction d, boundary_condition cond)
   }
 }
 
+boundary_condition fields::get_boundary(boundary_side b, direction d) const {
+  return boundaries[b][d];
```
stevengj, line 91: See also https://github.com/NanoComp/meep/issues/3276 — we should add a check on the Python side.  But we have the same problem in `set_boundary` so it is not really a regression.

PR #3273, opened 2026-08-17
Provide `dt` option for `fourier_transform` member function of `GaussianSource`

### LINKED ISSUE #1956: `fourier_transform` function of `GaussianSource` should apply to discrete rather than analytic current amplitude
As discussed in #1931, the `fourier_transform` member function of the [`GaussianSource`](https://meep.readthedocs.io/en/latest/Python_User_Interface/#gaussiansource) class computes the Fourier transform of the analytic current amplitude G(t) rather than the discrete version as implemented in:

https://github.com/NanoComp/meep/blob/35b10e53c181cfae63abfd761bf983d933d8d2e8/src/meep.hpp#L970-L975

https://github.com/NanoComp/meep/blob/35b10e53c181cfae63abfd761bf983d933d8d2e8/src/sources.cpp#L97-L116

As it is currently set up, `fourier_transform` does not seem to have much practical use. #1931 discussed the possibility of using the `fourier_transform` feature as a replacement for the normalization run when computing the incident fields for scattering calculations. This can be useful for a number of applications including [Tutorial/Diffraction Spectrum of a Binary Grating](https://meep.readthedocs.io/en/latest/Python_Tutorials/Mode_Decomposition/#diffraction-spectrum-of-a-binary-grating) which involves an incident planewave source in a homogeneous medium.

(This is somewhat related to a similar feature request for the `GaussianBeamSource` in #1444.)

oskooi:
Closes #1956.

**The problem**

[`GaussianSource.fourier_transform()`](https://meep.readthedocs.io/en/latest/Python_User_Interface/#gaussiansource) returns the **continuous-time** transform of the idealized envelope:

$$\widetilde G(\omega) = w\\, e^{i\omega t_0} e^{-(\omega - \omega_0)^2 w^2 / 2}, \qquad w=1/\Delta f$$

which is $\tfrac{1}{\sqrt{2\pi}} \int e^{i\omega t}G(t)\\,dt$ for $G(t) \approx e^{-(t-t_0)^2 / 2w^2} e^{-i\omega_0(t-t_0)}$. That is not the current the FDTD loop injects.

**What the simulation actually injects**
- `src_time::current(t, dt) = (p(t+dt) -p(t))/dt` — a finite difference of the dipole envelope $p(t)$ ([`src/meep.hpp:989`](https://github.com/NanoComp/meep/blob/e7d46f2a3711aed91b19a33e69399c35793df9eb/src/meep.hpp#L989)), not an analytic derivative.
- Electric sources: `calc_sources(time()+0.5*dt)` → `step_source(D_stuff)` does `D -= J*dt` ([`src/step.cpp:95`](https://github.com/NanoComp/meep/blob/e7d46f2a3711aed91b19a33e69399c35793df9eb/src/step.cpp#L95),[`100`](https://github.com/NanoComp/meep/blob/e7d46f2a3711aed91b19a33e69399c35793df9eb/src/step.cpp#L100)). Magnetic: `calc_sources(time())` → `step_source(B_stuff)` ([`src/step.cpp:64`](https://github.com/NanoComp/meep/blob/e7d46f2a3711aed91b19a33e69399c35793df9eb/src/step.cpp#L64),[69](https://github.com/NanoComp/meep/blob/e7d46f2a3711aed91b19a33e69399c35793df9eb/src/step.cpp#L69))
- DFT monitors stamp E at `time()` and H at `time()-0.5*dt` ([`src/dft.cpp:250-263`](https://github.com/NanoComp/meep/blob/e7d46f2a3711aed91b19a33e69399c35793df9eb/src/dft.cpp#L250-L263))

**So relative to the field times the DFT uses**, the injected current is the *centered* difference $[p(t+\Delta t/2) - p(t-\Delta t/2)]/\Delta t$ whose transform is $-i\omega\\,\mathrm{sinc}(\omega\Delta t/2)\\,\tilde p(\omega)$. With $\tilde p = \widetilde G/(-i\omega_0)$:

$$\widetilde G_{\Delta t}(\omega) = \frac{\omega}{\omega_0}\\,\mathrm{sinc}\\!\left(\frac{\omega\Delta t}{2}\right)\widetilde G(\omega)\\,\qquad \mathrm{sinc}(x)=\sin x/x$$

Two independent corrections: $\omega/\omega_0$ (exact derivative instead of the $-i\omega_0$ approximation) and $\mathrm{sinc}$ (discrete derivative). **No phase factor** — the centered difference is symmetric about the field time. `is_integrated=True` sources subtract $p(t)$ directly at the exact field time ([`src/update_eh.cpp:128-138`](https://github.com/NanoComp/meep/blob/e7d46f2a3711aed91b19a33e69399c35793df9eb/src/update_eh.cpp#L128-L138)), so they get only the $\omega/\omega_0$ factor.

**Verification**

Closed form vs. brute-force DTFT of the actual `swigobj.current` / `swigobj.dipole` samples: agreement to ~5e-11 at `cutoff=10` for both `is_integrated` values and both `resolution=10` and `40`. (At the default `cutoff=5` the residual is ~2e-2 at the band edges — that is the Gaussian's own truncation, not the formula.)

**Note:** this also explains the comment [`python/adjoint/objective.py:101-106`](https://github.com/NanoComp/meep/blob/e7d46f2a3711aed91b19a33e69399c35793df9eb/python/adjoint/objective.py#L101-L106) ("real parts match, imaginary parts are very different") — that code takes a *forward*-difference DTFT on the $t=n\Delta t$ grid, which carries an extra $e^{-i\pi f\Delta t}$ relative to the field-time convention used here.

image

### DIFF
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -7123,20 +7123,19 @@ Construct a `GaussianSource`.
 + **`fourier_transform(f, dt=None)`** — Returns the Fourier transform of the
   current evaluated at frequency $f$ ($\omega=2\pi f$). If `dt` is not specified,
   the transform of the idealized *continuous-time* current is returned:
-  $\widetilde G(\omega) \equiv \frac{1}{\sqrt{2\pi}} \int e^{i\omega t}G(t)\,dt \equiv \frac{1}{\Delta f} e^{i\omega t_0 -\frac{(\omega-\omega_0)^2}{2\Delta f^2}}$
-  where $G(t)$ is the current (not the dipole moment). In this formula, $\Delta f$
-  is the `fwidth` of the source, $\omega_0$ is $2\pi$ times its `frequency`, and
-  $t_0$ is the peak time discussed above.
+  $\widetilde G(\omega) \equiv \frac{1}{\sqrt{2\pi}} \int e^{i\omega t}G(t)\,dt \equiv \frac{\omega}{\omega_0} \frac{1}{\Delta f} e^{i\omega t_0 -\frac{(\omega-\omega_0)^2}{2\Delta f^2}}$
+  where $G(t) = p'(t)$ is the current (the time derivative of a Gaussian dipole moment $p(t)$).
+  In this formula, $\Delta f$ is the `fwidth` of the source, $\omega_0$ is $2\pi$ times
+  its `frequency`, and $t_0$ is the peak time discussed above. The $\omega/\omega_0$
+  factor is the exact time derivative of the Gaussian dipole moment.
 
       The current actually injected by the simulation is not $G(t)$ but rather its
       *discrete-time* counterpart: a finite difference of the Gaussian dipole moment
       $p(t)$ over one timestep $\Delta t$ (`sim.fields.dt`, which equals
       `Courant/resolution`). Passing `dt` returns the Fourier transform of that
-      discrete current, referenced to the field times at which the DFT field monitors
-      accumulate their sums: $\widetilde G_{\Delta t}(\omega) = \frac{\omega}{\omega_0}\mathrm{sinc}\left(\frac{\omega \Delta t}{2}\right)\widetilde G(\omega)$
-      with $\mathrm{sinc}(x) = \sin(x)/x$. The $\omega/\omega_0$ factor is the exact
-      time derivative of the Gaussian dipole moment in place of the $-i\omega_0$
-      approximation made by $\widetilde G(\omega)$, and the sinc factor is the error
+      finite-difference current, referenced to the field times at which the DFT field monitors
+      accumulate their sums: $\widetilde G_{\Delta t}(\omega) = \mathrm{sinc}\left(\frac{\omega \Delta t}{2}\right)\widetilde G(\omega)$
+      with $\mathrm{sinc}(x) = \sin(x)/x$. The sinc factor is the error
       of the discrete-time derivative. For an `is_integrated=True` source, $p(t)$ is
       added to the fields directly rather than differenced, and the sinc factor is
       therefore omitted.
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -6887,10 +6887,9 @@ def eig_power(freq, dt=None):
 
 Returns the total power of the fields from the eigenmode source at frequency `freq`.
 If the timestep `dt` (= `sim.fields.dt` = `Courant/resolution`) is specified, the
-discrete-time Fourier transform of the source is used, which is the power actually
-delivered at a finite resolution; see
-[`GaussianSource`](#gaussiansource). Otherwise, the continuous-time transform is
-used.
+Fourier transform of the discrete-time current is used, which is the power actually
+delivered at a finite resolution; see [`GaussianSource`](#gaussiansource). Otherwise,
+the continuous-time current is used.
 
 </div>
 
--- a/doc/docs/Python_User_Interface.md
+++ b/doc/docs/Python_User_Interface.md
@@ -6880,12 +6880,17 @@ Construct an `EigenModeSource`.
 <div class="class_members" markdown="1">
 
 ```python
-def eig_power(freq):
+def eig_power(freq, dt=None):
 ```
 
 <div class="method_docstring" markdown="1">
 
 Returns the total power of the fields from the eigenmode source at frequency `freq`.
+If the timestep `dt` (= `sim.fields.dt` = `Courant/resolution`) is specified, the
+discrete-time Fourier transform of the source is used, which is the power actually
+delivered at a finite resolution; see
+[`GaussianSource`](#gaussiansource). Otherwise, the continuous-time transform is
+used.
 
 </div>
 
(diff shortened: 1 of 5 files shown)


stevengj:
This is a nice improvement, but note that it is *not* the same thing as the DTFT.  What you are returning here is the *continuous* Fourier transform of the center-differenced pulse.

The true DTFT corresponds to taking this continuous Fourier transform and adding it to all its aliases (shifted by multiples of 2π/Δt).  Nevertheless, the results here closely match the brute-force DTFT because the Fourier transform is so localized in ω, so that the aliases only contribute tiny corrections via their exponential tails.

This should probably be clarified in the comments and docs of this PR.

python/tests/test_source.py
```diff
@@ -80,6 +80,48 @@ def test_source_frequency(self):
         with self.assertRaises(ValueError):
             ContinuousSource()
 
+    def test_gaussian_fourier_transform(self):
+        # The discrete-time Fourier transform of the current injected by the
+        # time stepping should agree with the closed form returned by
+        # GaussianSource.fourier_transform(f, dt). A large cutoff is used so
+        # that the truncation of the Gaussian is negligible.
```
stevengj:
```suggestion
        # The discrete-time Fourier transform of the current injected by the
        # time stepping should very nearly match the closed-form continuous
        # Fourier transform (differing only by exponentially small alias tails) returned by
        # GaussianSource.fourier_transform(f, dt). A large cutoff is used so
        # that the truncation in time of the Gaussian is negligible.
```

python/tests/test_source.py
```diff
@@ -80,6 +80,48 @@ def test_source_frequency(self):
...
+                        / np.sqrt(2 * np.pi)
+                    )
+                    if is_integrated:
+                        dtft *= -1j * 2 * np.pi * f
+                    self.assertAlmostEqual(
+                        dtft / src.fourier_transform(f, dt), 1.0, places=8
+                    )
+
+    def test_gaussian_fourier_transform_continuous(self):
+        # omitting dt gives the continuous-time transform
```
stevengj:
```suggestion
        # omitting dt gives the continuous-time transform without the finite difference
```

stevengj:
Arguably, if `dt` is not specified, we should still include the $\omega / \omega_0$ correction factor, which is the idealized continuous-time Fourier transform including the exact $d/dt$ derivative (and the $-1/i\omega_0$ amplitude factor in the source).  This is the exact $\Delta t \to 0$ result.

src/sources.cpp
```diff
@@ -108,12 +108,33 @@ complex<double> gaussian_src_time::dipole(double time) const {
...
-  return width * polar(1.0, omega * peak_time) * exp(-0.5 * delta * delta);
+  complex<double> ft = width * polar(1.0, omega * peak_time) * exp(-0.5 * delta * delta);
+  if (dt == 0) return ft;
+
+  // -i\omega instead of the -i\omega_0 assumed by the analytic formula above
+  ft *= f / freq;
+  if (!is_integrated) { // discrete-time (centered-difference) derivative
+    double x = 0.5 * omega * dt;
+    ft *= (x == 0) ? 1.0 : sin(x) / x;
+  }
```
stevengj:
```suggestion
  ft *= f / freq; // exact time derivative and i/omega_0 normalization
  if (dt != 0 && !is_integrated) { // discrete-time (centered-difference) derivative
    double x = 0.5 * omega * dt;
    ft *= (x == 0) ? 1.0 : sin(x) / x;
  }
```

python/tests/test_source.py
```diff
@@ -80,6 +80,48 @@ def test_source_frequency(self):
...
+        src = GaussianSource(1.0, fwidth=0.2)
```
stevengj, line 122:
```suggestion
        src = GaussianSource(1.0, fwidth=0.2)
        # to do: implement a more nontrivial test
```

python/source.py
```diff
@@ -306,18 +306,29 @@ def __init__(
...
+              accumulate their sums: $\\widetilde G_{\\Delta t}(\\omega) = \\frac{\\omega}{\\omega_0}\\mathrm{sinc}\\left(\\frac{\\omega \\Delta t}{2}\\right)\\widetilde G(\\omega)$
+              with $\\mathrm{sinc}(x) = \\sin(x)/x$. The $\\omega/\\omega_0$ factor is the exact
+              time derivative of the Gaussian dipole moment in place of the $-i\\omega_0$
+              approximation made by $\\widetilde G(\\omega)$, and the sinc factor is the error
+              of the discrete-time derivative. For an `is_integrated=True` source, $p(t)$ is
+              added to the fields directly rather than differenced, and the sinc factor is
+              therefore omitted.
+
+              Note that this does not include any `amplitude` or `amp_func` factor that you
+              specified for the source.
```
stevengj:
```suggestion
        + **`fourier_transform(f)`** — Returns the Fourier transform of the
          current evaluated at frequency $f$ ($\\omega=2\\pi f$). If `dt` is not specified,
          the transform of the idealized *continuous-time* current is returned:
          $\\widetilde G(\\omega) \\equiv \\frac{1}{\\sqrt{2\\pi}} \\int e^{i\\omega t}G(t)\\,dt \\equiv \\frac{\\omega}{\\omega_0} \\frac{1}{\\Delta f} e^{i\\omega t_0 -\\frac{(\\omega-\\omega_0)^2}{2\\Delta f^2}}$
         where $G(t) = p'(t)$ is the current (the time derivative of a Gaussian dipole moment $p(t)$). In this formula, $\\Delta f$
          is the `fwidth` of the source, $\\omega_0$ is $2\\pi$ times its `frequency`, and
          $t_0$ is the peak time discussed above.  The $\\omega/\\omega_0$ factor is the exact
              time derivative of the Gaussian dipole moment

              The current actually injected by the simulation is not $G(t)$ but rather its
              *discrete-time* counterpart: a finite difference of the Gaussian dipole moment
              $p(t)$ over one timestep $\\Delta t$ (`sim.fields.dt`, which equals
              `Courant/resolution`). Passing `dt` returns the Fourier transform of that
              finite-difference current, referenced to the field times at which the DFT field monitors
              accumulate their sums: $\\widetilde G_{\\Delta t}(\\omega) = \\mathrm{sinc}\\left(\\frac{\\omega \\Delta t}{2}\\right)\\widetilde G(\\omega)$
              with $\\mathrm{sinc}(x) = \\sin(x)/x$. The sinc factor is the error
              of the discrete-time derivative. For an `is_integrated=True` source, $p(t)$ is
              added to the fields directly rather than differenced, and the sinc factor is
              therefore omitted.

              Note that this does not include any `amplitude` or `amp_func` factor that you
              specified for the source.
```

doc/docs/Python_User_Interface.md
```diff
@@ -6975,18 +6980,29 @@ Construct a `GaussianSource`.
...
++ **`fourier_transform(f, dt=None)`** — Returns the Fourier transform of the
+  current evaluated at frequency $f$ ($\omega=2\pi f$). If `dt` is not specified,
+  the transform of the idealized *continuous-time* current is returned:
+  $\widetilde G(\omega) \equiv \frac{1}{\sqrt{2\pi}} \int e^{i\omega t}G(t)\,dt \equiv \frac{1}{\Delta f} e^{i\omega t_0 -\frac{(\omega-\omega_0)^2}{2\Delta f^2}}$
   where $G(t)$ is the current (not the dipole moment). In this formula, $\Delta f$
-  is the `fwidth` of the source, $\omega_0$ is $2\pi$ times its `frequency,` and
-  $t_0$ is the peak time discussed above. Note that this does not include any
-  `amplitude` or `amp_func` factor that you specified for the source.
+  is the `fwidth` of the source, $\omega_0$ is $2\pi$ times its `frequency`, and
+  $t_0$ is the peak time discussed above.
```
stevengj: To do: update `.md` docs from docstring.

python/source.py
stevengj: For some reason github won't let me apply this suggestion.

stevengj: applied manually

oskooi:
The entry for `GaussianSource.fourier_transform` has been updated in `Python_User_Interface.md` based on its docstrings in `sources.py`.

image

Issue #3217, opened 2026-05-08
scalable Pade extrapolation


stevengj:
It would be nice to have a more scalable version of Padé extrapolation (ala #2440) that (a) parallelizes and (b) requires bounded storage independent of the number of time steps.

(a) seems to require implementation at the C++ level so that the computations and storage can be done per-chunk (like DFT accumulation). 

For (b), one idea is (for each DFT point/component), we continue to accumulate the DFT as now, and additionally store the last $n$ timesteps for some $n$ (e.g. $n=20$).    Then we compute the DFT as (accumulated DFT) + (Padé extrapolation from $n$ points).   This could hopefully be very effective because the "tail" of the time series is typically dominated by a small set of resonances, so only a low-order Padé extrapolation (requiring only a few points) should suffice.  

Furthermore, this would give us error measures that we could use to estimate the DFT convergence: the Padé extrapolation itself is an error measure on the accumulated DFT, and we could also use the difference between Padé extrapolation from $n$ points and from $n/2$ points as an error estimate on the extrapolant.

This could be integrated with the existing `dft_fields` objects, with the default being $n=0$ (no extrapolation, i.e. current behavior), so that all existing DFT functions (flux, stress, eigenmode coefficients, adjoints, …) get the option of an extrapolated correction.

stevengj:
Note that you could also then request any frequency you want and it would do a simple low-order interpolation of the accumulated DFT but add the Padé approximant evaluated at *exactly* that frequency.   

In this way you could get "pure" Padé approximation as a special case by setting $n$ to the number of timesteps and having no DFT frequencies.

PR #3277, opened 2026-08-24
[1/n][Adjoint Module] Fix MaterialGrid design-variable attribution in the adjoint gradient

### LINKED ISSUE #1984: Incorrect adjoint gradients with overlapping design variables

smartalecH:
### Stack

| | PR | |
|---:|---|---|
| 1 | #3277 | Fix MaterialGrid design-variable attribution (#1984) **&larr; you are here** |

Fixes #1984.

Stacked on #3274 — based on `fix/adjoint-chunk-pairing-minimal`, so the diff here is just these two commits. Merge #3274 first, then this retargets to `master` cleanly.

While working on a new 3D grating coupler TO example for meep (2 design regions) I noticed that the gradients were rather inaccurate. Claude helped me trace things down to the unique case where we have two adjacent design regions (e.g. a partial etched layer). There were different issues depending on the sizes of the design regions (e.g. if one design region was bigger than the other, then we got a segfault!)

The core issue was that the logic we had for handling multiple design grids during the recombination step wasn't properly bookkeeping the effects near the boundaries. This was because we weren't keeping track of "which grid was which." In addition, we had some hairy "copies" of grids we were trying to use to keep track of multiple grids (hence the crash). So the resulting gradient near the boundary was completely off. To fix this, we simply add a grid id (automatically populated in python) and propagated to the cpp backend via swig.

Practically speaking, this means that **the unit of differentiation is now the design variable, not the grid copy.**

Turns out this bug also affects symmetries (see #1984) and the fix also resolves it.

### DIFF
--- a/src/meepgeom.cpp
+++ b/src/meepgeom.cpp
@@ -15,6 +15,7 @@
 */
 
 #include <algorithm>
+#include <set>
 #include <vector>
 #include "meepgeom.hpp"
 #include "meep_internals.hpp"
@@ -2640,7 +2641,8 @@ get_material_gradient(const meep::vec &r,              // current point
                       geom_epsilon *geps,              // material
                       meep::grid_volume &gv,           // simulation grid volume
                       double du,                       // step size
-                      double *u,                       // matgrid
+                      const std::vector<double *> &us, // matgrid weights, one per
+                                                       // copy of this design grid
                       int idx                          // matgrid index
 ) {
   /*Compute the Aᵤx product from the -λᵀAᵤx calculation.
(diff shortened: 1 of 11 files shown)


lxvm:
I just tested this pr on a two-layer vertical grating coupler with partial etch layer calculation and it dramatically lowered the adjoint error along a directional derivative to 0.2% compared to finite differences. In #3274 the relative error was around 26% and in the latest release the relative error was around 300%. I am a bit surprised that the error dropped so much given that the discretization error for the forward calculation is on the order of 20% at the resolution I am using. Nonetheless, I am looking forward to trying out this pr in an optimization

jin-castle:
This is really nice work, especially the symmetry cases. I ran into a similar, though distinct, issue with symmetries while validating #3274, and it was very difficult to diagnose. I really appreciate the rotated and mirrored regression coverage here.

smartalecH:
Glad to hear it, thanks @lxvm!

> I am a bit surprised that the error dropped so much given that the discretization error for the forward calculation is on the order of 20% at the resolution I am using.

Which "resolution" is this? The simulation resolution, or the "resolution" of your directional derivative (e.g. your Δx)?

Note that meep's hybrid time-/frequency-domain adjoint should match finite difference gradients _regardless of the discretization_. That's because we are differentiating the discretized operator, not the continuous operator (e.g.  differentiate then optimize vs optimize then differentiate). Of course, the step of your finite difference matters (as does whether it's one sided vs centered). But you should get pretty good agreement whether your resolution is 10 or 1000.

We haven't always, and that's because we have bugs 🙃  but let's see if we can track those down now that we have some additional support from coding tools.

lxvm:
Thanks @smartalecH !

The resolution I was referring to was the simulation resolution (i.e. 30 for my simulation for a coupler based on 500nm x 220 nm waveguide dimensions). Thanks for your comment because I had thought meep was using a differentiate then discretize approach, which is not the case. Glad to hear that I should expect good agreement of the adjoint gradients with finite differences (in my test I was using a finite difference step size of 10^-5).

smartalecH:
@lxvm did you try running an optimization after rebasing on this PR?

lxvm:
Yes, I ran an optimization with this pr and it gave improved results compared to the prs leading up to this one. I've been able to do robust optimization of a single mode coupler with length scale and partial etch constraints that attained 50% efficiency which is the best result I've had so far

stevengj:
This looks reasonable to me.

stevengj:
Failing test?

src/meepgeom.cpp
```diff
@@ -2863,6 +2863,7 @@ void material_grids_addgradient_point(double *v, vector3 p, double scalegrad, ge
     do {
       material_data *mg_cur = (material_data *)tpc->objects[oic].o->material;
       if (owner_grid_id < 0 || mg_cur->grid_id == owner_grid_id) {
+        if (udatas.empty()) sz = mg_cur->grid_size;
```
smartalecH, line 2866: cc @stevengj claude picked this up during the rebase (another corner case). It added a test (above) which had a tolerance slightly too tight for single precision.

src/meepgeom.cpp
```diff
@@ -2716,58 +2731,66 @@ With the addition of subpixel smoothing, however, the vJp became
...
+                          dz);
+    int x_list[2] = {x1, x2}, y_list[2] = {y1, y2}, z_list[2] = {z1, z2};
+    int lx = (x1 == x2) ? 1 : 2;
+    int ly = (y1 == y2) ? 1 : 2;
+    int lz = (z1 == z2) ? 1 : 2;
+
+    if (k == 0 && (ukind == material_data::U_MIN || ukind == material_data::U_PROD)) {
+      /* U_MIN and U_PROD carry adjustments defined in terms of the interpolated
+         value at this point. Overlapping grids of these kinds are rejected by
+         the caller, so there is exactly one copy and this is unchanged. */
```
stevengj, line 2762: I don't understand this comment.  Why are overlapping `U_MIN` and `U_PROD` grids "rejected by the caller"?

smartalecH:
Closing in favour of #3297, which is the same branch and the same content, opened from this repository rather than from a fork.

Nothing about the change itself has been revised, and the description is carried over verbatim. The discussion and reviews on this PR stay readable here; please continue on #3297.

Issue #1965, opened 2022-02-25
edge artifacts of `MaterialGrid`


oskooi:
There seems to be unwanted artifacts along the outer edges of a `MaterialGrid` which could introduce errors into a simulation. This can be demonstrated by using `plot2D` to visualize a `MaterialGrid` in a 2d cell consisting of a linear interpolation along a single direction of `medium1` (index `n1`) and `medium2` (index `n2`). The edge artifacts are independent of the `resolution`. The edge artifacts seem to disappear when the `MaterialGrid` is defined as the `default_medium` (rather than a `Block` object) which suggests there is probably a bug in the interpolation for grid voxels near the edges.

(Possibly related to #1399.)

![matgrid_graded_res50_101_101](https://user-images.githubusercontent.com/7152530/155806990-a85ed214-abc6-4a68-8b62-6a2696a215e8.png)

![matgrid_graded_res67_135_135](https://user-images.githubusercontent.com/7152530/155807204-d8e0fbca-be4c-447b-9f4f-e6c45a8b9ca3.png)

![matgrid_graded_default_material_res100_201_201](https://user-images.githubusercontent.com/7152530/155807039-e149c79f-6a2f-4127-8504-8da2f78c3b3c.png)

```py
import meep as mp
import numpy as np
import matplotlib
matplotlib.use('agg')
import matplotlib.pyplot as plt

resolution = 50

cell_size = mp.Vector3(2,2,0)

Nx = 101
Ny = 101
lin_interp = np.linspace(0,1,Nx)
weights = np.tile(lin_interp,(Ny,1))
mat_interp = mp.MaterialGrid(mp.Vector3(Nx,Ny,0),
                             mp.Medium(index=1.5),
                             mp.Medium(index=3.5),
                             weights=weights)

geometry = [mp.Block(material=mat_interp,
                     center=mp.Vector3(),
                     size=mp.Vector3(1,1,0))]

sim = mp.Simulation(cell_size=cell_size,
                    resolution=resolution,
                    geometry=geometry)

sim.plot2D()
plt.title('resolution = {}, (Nx,Ny) = ({},{})'.format(resolution,Nx,Ny))
plt.savefig('matgrid_graded_res{}_{}_{}.png'.format(resolution,Nx,Ny),
            bbox_inches='tight',
            dpi=150)
```

smartalecH:
[...]

PR #3279, opened 2026-08-25
Smooth a MaterialGrid object's own boundary, not just its level set

jin-castle:
`do_averaging=True` promises subpixel smoothing for a `MaterialGrid`, but a
material grid's own object boundary never gets it. In 3d the top and bottom
faces of a design slab — the interfaces that set the vertical confinement of
every strip-waveguide design — stay staircased at every resolution, no matter
what `do_averaging` is set to.

One commit for the defect, a second that adds symmetry-invariance and
anisotropic-media coverage this path never had.

> Split from the original version of this PR at review request: the
> graded-medium harmonic average and the quadrature skip that depends on it are
> a separate defect in the same file and will follow as their own PR.

## The material-grid fallback is the only smoothing path that cannot see an object boundary

There are two paths. `eff_chi1inv_matrix()` calls `get_front_object()` and, if
that succeeds, averages two materials across the object's boundary using its
normal and the pixel's filling fraction. If it fails, `fallback_chi1inv_row()`
integrates along a level set instead.

`get_front_object()` rejected any pixel touching a variable material:

```cpp
if ((o1 && is_variable(o1->material)) || (o2 && is_variable(o2->material)) || ...)
  return false;
```

So a material-grid object's own boundary never reached the first path. And the
fallback takes its normal from `∇u` alone for a grid while a user material
function gets it from `normal_vector()` and integrates `chi1p1()` — both of
which sample the actual material and therefore already see the object boundary.
**Only the material-grid branch is blind to it.**

For an `(n, n, 1)` grid `∇u` has no z component at all. The slab's z faces are
not in the design level set; they are the object's boundary. Nothing in either
path was smoothing them. Those faces do not exist in 2d, and every adjoint test
upstream is 2d or cylindrical, which is why this survived.

### Why the front material may be a grid and the back material may not

The average needs one material that is constant over the pixel to act as the
background, and one boundary with a normal and a filling fraction. `mat_behind`
must therefore stay constant — a variable material there is still rejected.
`mat_front` need not be: the boundary being averaged is the object's own level
set, and the grid can be evaluated at a point before averaging.

Which point matters. The pixel center can sit on the far side of the boundary,
and that is exactly where a `MaterialGrid` is evaluated on its mirrored
extension. For a planar cut of a pixel of size `h` with inside fraction `fill`,
the centroid of the inside part lies `(1-fill)*h/2` from the center along the
inward normal. libctl does not fix the normal's orientation, so both candidates
are tried and the one `point_in_fixed_objectp()` accepts is kept.

Interior pixels must go back to the fallback. A pixel with `fill = 1` has no
boundary in it; averaging it would return the unsmoothed material and silently
disable level-set smoothing across the whole design region. Those are handed
back.

### Changes

1. **Narrow the rejection in `get_front_object()`** to what actually breaks the
   average, and move it after the front/behind assignment so it can distinguish
   the two. A material grid is admissible as the front object's own material
   (not as `default_material`, which has no boundary to average) when
   `do_averaging` is set; `front_is_matgrid` tells the caller to localize it.
2. **`inside_sample_point()`** — evaluate the grid at the centroid of the inside
   part of the pixel, per the derivation above.
3. **Boundary pixels only.** `fill >= 1 - 1e-6` returns `fallback = true`;
   `fill <= 1e-6` takes the no-average path.
4. **Defer the metal test for a front-side grid** until after localization.
   `is_metal()` reads the grid's `medium`, which before `get_material_pt()`
   still holds the previous pixel's interpolation. `mat_behind` is never a grid,
   so its test stays where it was.

## Tests

`TestObjectBoundarySmoothing` in `python/tests/test_subpixel_3d.py`. Filling
`weights` with a constant removes the in-plane level set entirely, leaving the
block's six faces as the only material interfaces — so whatever smoothing the
grid gets there must be the smoothing an ordinary `Medium` block of the same
epsilon gets.

- z profiles across the slab's faces agree with the equivalent plain block to
  `1e-8`;
- the `do_averaging` on/off difference exceeds 0.5, which guards the first
  assertion against passing because both sides are staircased — before this
  commit that difference was identically zero;
- an interior ramp must still show an on/off difference, which is the guard
  that interior pixels are not swallowed by the `fill = 1` boundary path.

A second commit adds coverage for two regimes this path never had tests for:

- **`symmetries` invariance** — the smoothed epsilon assembly is identical with
  and without `mp.Mirror(mp.Y)` for a mirror-symmetric design (measured
  3.6e-15, at `beta=inf` and `beta=0`). Structure only: the adjoint gradient
  under `symmetries` has its own independent, pre-existing defect that nothing
  here touches.
- **Anisotropic media** — a uniform grid of anisotropic media gets the same
  analytic boundary average as a plain block of the interpolated anisotropic
  epsilon (identical to the last bit — same code path), with an on/off guard
  showing the average engages; and in the interior, where the level-set
  fallback deliberately declines anisotropic pixels (its 1d normal average is
  scalar), `do_averaging` is pinned as a finite no-op rather than a
  half-applied average.

One CI note: the MPI jobs initially tripped on
`test_gradient_matches_fd_with_smoothing` — an existing test from #3263 that
holds fd/adjoint to 1%. The excess is the chunk-dependence of the adjoint
gradient that #3274 fixes. After splitting this boundary-only change from the
graded-harmonic work, the two-rank case differs by 10.29% in both precisions;
it vanishes on this branch merged with #3274's head. Until #3274 lands, the
parallel assertion is bounded at 0.11 — still far below the 70–100%
disagreement the test guards against — and the serial assertion keeps its 1%.

Also checked against the pending #3277, since both touch `meepgeom.cpp`: the
merge is conflict-free and both sides' suites pass on the combined build,
including #3277's `test_adjoint_symmetric_grids` (3/3) and
`test_adjoint_adjacent_grids` (5/5). One composition result worth recording: a
`MaterialGrid` overlapped with a mirrored copy of itself (`U_MEAN`, the #1984
pattern) assembles the same smoothed epsilon as a single grid carrying the
explicitly averaged weights, to 2.7e-15.

## What would falsify this

If the uniform-grid z profile does not match the equivalent block, the
localization point or the filling fraction is wrong. If the interior ramp's
on/off difference collapses, interior pixels are being averaged against
`fill = 1` and level-set smoothing is gone. Both are asserted, so the tests are
the falsifier.

## Not addressed

**#1965** (edge artifacts that vanish when the grid is `default_material`
rather than a `Block`) may have the same root as this commit, but I have not
confirmed it.

cc @smartalecH @stevengj

### DIFF
--- a/src/meepgeom.cpp
+++ b/src/meepgeom.cpp
@@ -919,11 +919,16 @@ double geom_epsilon::chi1p1(meep::field_type ft, const meep::vec &r) {
 /* Find frontmost object in v, along with the constant material behind it.
    Returns false if material behind the object is not constant.
 
+   A material grid may be returned as the *front* material, in which case
+   front_is_matgrid is set and the caller must evaluate it at a point before
+   averaging; see eff_chi1inv_matrix().
+
    Requires moderately horrifying logic to figure things out properly,
    stolen from MPB. */
 static bool get_front_object(const meep::volume &v, geom_box_tree geometry_tree, vector3 &pcenter,
                              const geometric_object **o_front, vector3 &shiftby_front,
-                             material_type &mat_front, material_type &mat_behind) {
+                             material_type &mat_front, material_type &mat_behind,
+                             bool &front_is_matgrid) {
   vector3 p;
   const geometric_object *o1 = 0, *o2 = 0;
   vector3 shiftby1 = {0, 0, 0}, shiftby2 = {0, 0, 0};
@@ -1009,9 +1014,8 @@ static bool get_front_object(const meep::volume &v, geom_box_tree geometry_tree,
     shiftby2 = shiftby1;
   }
 
-  if ((o1 && is_variable(o1->material)) || (o2 && is_variable(o2->material)) ||
-      ((is_variable(default_material) || is_file(default_material)) &&
-       (!o1 || is_file(o1->material) || !o2 || is_file(o2->material))))
+  if ((is_variable(default_material) || is_file(default_material)) &&
+      (!o1 || is_file(o1->material) || !o2 || is_file(o2->material)))
     return false;
 
   if (id1 >= id2) {
@@ -1029,7 +1033,51 @@ static bool get_front_object(const meep::volume &v, geom_box_tree geometry_tree,
     mat_front = mat2;
     mat_behind = mat1;
   }
+
+  /* A variable material is not constant over the pixel, so it cannot serve as
+     the *background* of the two-material average.  As the material of the front
+     object a material grid still can: the object's boundary is an ordinary
+     level set, and the caller localizes the grid before averaging.  Rejecting
+     it here as well -- which is what this routine used to do -- meant the
+     boundary of a material-grid object never reached the analytic smoothing at
+     all, so in 3d the top and bottom faces of a design slab stayed staircased
+     no matter what do_averaging was set to.
+
+     A user material function needs no such help: its fallback derives the
+     normal from normal_vector() and integrates chi1p1(), both of which sample
+     the actual material and therefore already see the object boundary.  Only
+     the material-grid fallback is blind to it, because matgrid_grad() and the
+     1d level-set integral that replace them know about u alone. */
+  if (is_variable(mat_behind)) return false;
+  if (is_variable(mat_front)) {
+    if (!is_material_grid(mat_front)) return false;
+    /* only a grid carried by the front object itself has a boundary to average
+       across; a material-grid default_material does not */
+    if (!*o_front || (material_type)(*o_front)->material != mat_front) return false;
+    if (!mat_front->do_averaging) return false;
+    front_is_matgrid = true;
+  }
   return true;
+}
+
+/* Pick the point at which to evaluate a material grid that fills only part of a
+   pixel.  The pixel center can sit on the far side of the object boundary --
+   where a material grid is evaluated on its mirrored extension -- so step to
+   the centroid of the part that is inside: for a planar cut of a pixel of size
+   h with inside fraction `fill`, that centroid lies (1-fill)*h/2 from the
+   center along the inward normal.  The normal's orientation is not fixed by
+   libctl, so try both and keep the candidate that is actually inside.  Returns
+   a lab-frame point. */
+static meep::vec inside_sample_point(const meep::volume &v, const geometric_object &o,
+                                     vector3 shiftby, vector3 normal, double fill) {
+  vector3 center = vec_to_vector3(v.center());
+  vector3 local = vector3_minus(center, shiftby);
+  double step = 0.5 * (1 - fill) * v.diameter();
+  for (int s = 1; s >= -1; s -= 2) {
+    vector3 q = vector3_plus(local, vector3_scale(s * step, normal));
+    if (point_in_fixed_objectp(q, o)) return vector3_to_vec(vector3_plus(q, shiftby));
+  }
+  return v.center();
 }
 
 void geom_epsilon::eff_chi1inv_row(meep::component c, double chi1inv_row[3], const meep::volume &v,
@@ -1071,6 +1119,7 @@ void geom_epsilon::eff_chi1inv_matrix(meep::component c, symm_matrix *chi1inv_ma
   symm_matrix meps;
   vector3 p, shiftby, normal;
   fallback = false;
+  bool front_is_matgrid = false;
 
   if (maxeval == 0) {
   noavg:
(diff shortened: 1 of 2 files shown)


lxvm:
This would be a nice feature to have since the workaround of defining a 3d material grid to smooth out the normal boundaries of a 2d design region with finite thickness could cause you to run out of memory when doing parallel meep calculations since you would have to load a large 3d material grid onto each process

src/meepgeom.cpp
```diff
@@ -1251,7 +1339,7 @@ static number matgrid_inveps_func(int n, number *x, void *mgva_) {
   double u_proj = 0, w = 0;
   matgrid_volavg *mgva = (matgrid_volavg *)mgva_;
   get_uproj_w(mgva, x[0], u_proj, w);
-  return w * ((1 - u_proj) / mgva->eps1 + u_proj / mgva->eps2);
+  return w / ((1 - u_proj) * mgva->eps1 + u_proj * mgva->eps2);
```
smartalecH: This isn't related to the core PR goal (smooth MG boundaries), right? The native cpp MG smoothing code indeed needs to be revamped, but I think we should do that separately.

src/meepgeom.cpp
```diff
@@ -1108,6 +1162,23 @@ void geom_epsilon::eff_chi1inv_matrix(meep::component c, symm_matrix *chi1inv_ma
 
   double fill = box_overlap_with_object(pixel, *o, tol, maxeval);
 
+  if (front_is_matgrid) {
```
smartalecH, line 1165: I'm not sure this part will properly get reflected in the $\frac{dA}{du}$ part of the adjoint's recombination step (currently uses a finite difference over the smoothing routine... but not sure _this_ part of the smoothing routine). Can we do some additional gradient checks to see?

src/meepgeom.cpp
jin-castle: Right on both counts — separate defect, separate PR. I've split it.

**Stays here** (`124554c9`, `4967bd35`, and the two follow-ups): the
`get_front_object()` / `eff_chi1inv_matrix()` change that lets a grid object's
own boundary reach the two-material average, its `TestObjectBoundarySmoothing`
cases, and the symmetry-invariance and anisotropic-media coverage.

**Moves out** (`ccbb655a`, `28cb1512`): the `matgrid_inveps_func` integrand and
the "skip the quadrature when no interface crosses the voxel" shortcut, which
only becomes safe once that integrand is continuous. I'll open them as a
follow-up against master.

One test crosses the line and I want to flag it rather than have it look like a
silent revert: `test_interior_level_set_is_still_smoothed` goes back to its
original form in this PR, and the rewrite of it — the one that distinguishes a
projected (β=∞) ramp, which keeps a real interface in the interior, from an
unprojected (β=0) ramp, which does not — travels with the follow-up, since it
is asserting the behaviour that only exists there.

The split is exact: the two branches applied in sequence reproduce this PR's
tree byte-for-byte, so nothing is being quietly dropped or rewritten in the
process. On a clean MPI-enabled double-precision build of the reduced branch,
`test_subpixel_3d.py` is 11/11 and `test_material_grid.py` 3/3 — the same
subpixel count as before the split, because the one test whose assertion
depended on the follow-up reverts along with it.

For the follow-up PR's benefit rather than this one's, the short version of why
that one line is not cosmetic: `matgrid_inveps_func` integrates
`(1-u)/eps1 + u/eps2`, the inverse of a *binary mixture*, which is the right
quantity only at `u ∈ {0,1}`. The medium the surrounding code assembles at an
interior node is the graded one, `(1-u)*eps1 + u*eps2`. The gap between them
keeps the anisotropic correction finite as `|∇u| → 0`, so the assembled epsilon
is discontinuous in the design weights at the `|∇u| < 1e-8` gate — which is why
it surfaces as a gradient error rather than a small bias. For weights projected
upstream (SSP at β=∞, where 64 % of nodes sit exactly at 0 or 1) the directional
FD error goes 6.6 % → 0.094 % at `dp=1e-3` and 47.5 % → 0.003 % at `dp=3e-3`;
conic-filtered (smooth) weights go 1.8 % → 0.033 %; the `do_averaging=False`
control is 0.000 % before and after. That belongs in its own thread with its own
tests, and it sounds like it overlaps whatever revamp you have in mind, so I'd
rather it be discussed there than bolted onto this one.

lxvm:
I tested this pr on a 3d meep simulation with a vertical grating coupler geometry defined by 2d material grids and got a noticeable 3x accuracy improvement in the calculation of the coupling coefficient compared to before this pr with `do_averaging=False` for the material grid. However, using this PR and `do_averaging=True` introduces a very large error in the adjoint gradient around the boundaries of the grating coupler that renders it impractical for optimization. I haven't tested to find out if the error is also present with `do_averaging=True` and without this pr.

PR #3280, opened 2026-08-25
[1/n][Adjoint Module] Enable proper JAX support in adjoint solver

smartalecH:
### Stack

| | PR | |
|---:|---|---|
| 4 | #3280 | Enable proper JAX support in the adjoint module **&larr; you are here** |

This PR makes Jax a "first-class" citizen within Meep's adjoint model. While meep technically supported Jax already, the interface was buggy, lacked several important features, and didn't play nice with the existing autograd engine. This resolves that.

**Allow users to specify their objective function using either autograd's `numpy` or jax's `numpy`**. The backend handles the vjps automatically. You can't mix and match though (e.g. use meep's autograd filters with a jax objective).
```python

def objective(mode_coeff, dft_fields):
    return npa.abs(mode_coeff)**2 + npa.sum(npa.abs(dft_fields)**2, axis=1)
```
OR
```python
from jax import numpy as npj

def objective(mode_coeff, dft_fields):
    return npj.abs(mode_coeff)**2 + npj.sum(npj.abs(dft_fields)**2, axis=1)
```
Meep handles both just fine:
```python
    simulation=sim,
    objective_functions=[objective],
    objective_arguments=[mode_monitor, dft_monitor],
    design_regions=[design_region],
    frequencies=frequencies,
)

value, gradient = opt([rho])
```

**Extend `MeepJaxWrapper` to properly support single-frequency objectives and vector-valued (multi-frequency) objectives**. There was a bug in the single-frequency case. The multi-frequency case previously required the user to scalarize the objective (e.g. reduce across the frequency/wavelength dimension). You can now specify a vector-valued objective function (like you can with the standard `OptimizationProblem` API) and jax will return a dense jacobian:

```py
def loss(rho, thickness, tilt):
    (dft,) = wrapped_meep([rho])
    return postprocess(dft, thickness, tilt)   # scalar

value, (d_rho, d_thickness, d_tilt) = jax.value_and_grad(
    loss, argnums=(0, 1, 2))(rho, 1.8, 8.0)
```

**Package optimization parameters as pytrees** (and return the gradient as an organized pytree). The end2end jax workflow allows you to chain operations before the meep simulations (preprocessing) and after (postprocessing). The user may want to collect gradients with parameters that tune these routines too. But bookeeping all of this is a pain. So we adopt the same approach e.g. `fmmax` uses via pytrees. You can pass a single pytree (a Dict/NamedTuple) to your objective function and the gradient routine will _also_ produce a pytree. This also works with vector-valued objectives (e.g. for minimax optimization):

```py
class Stack(NamedTuple):
    thickness: jnp.ndarray     # (3,) oxide, antireflection coating, air
    index: jnp.ndarray         # (3,)

class Params(NamedTuple):
    latent: jnp.ndarray        # (601,) design degrees of freedom
    beta: float                # preprocessing: projection strength
    stack: Stack               # post-processing: the layers above the chip
    tilt: float                # post-processing: fiber tilt, in degrees

def project(x, beta, eta=0.5):
    """The usual tanh projection, written in jax.numpy -- see the note below."""
    return (jnp.tanh(beta * eta) + jnp.tanh(beta * (x - eta))) / (
        jnp.tanh(beta * eta) + jnp.tanh(beta * (1 - eta))
    )

def loss(p):
    rho = project(conic_filter(p.latent), p.beta)   # preprocessing, in jax.numpy
    (ez, hx) = wrapped_meep([rho])
    field = propagate(ez, hx, p.stack)          # a stratified-media propagator, say
    return fiber_overlap(field, p.tilt)         # (nfreq,)

values, jacobian = mpa.value_and_jacobian(loss)(params)
jax.tree_util.tree_map(lambda leaf: leaf.shape, jacobian)
```

**Various other bug fixes and removed antipatterns.**

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -37,6 +37,15 @@
   computing a gradient raised `AxisError: axis 1 is out of bounds for array of
   dimension 1`.
 
+* Adjoint solver: `meep.adjoint.value_and_jacobian` transforms a loss function
+  that returns one value per frequency into one that also returns a gradient for
+  each of them, for worst-case (minimax) optimization over a bandwidth. It is the
+  counterpart of `jax.value_and_grad`, accepts arbitrary pytrees, differentiates
+  parameters that never reach Meep alongside the design weights, and costs a
+  single forward and adjoint simulation regardless of the number of frequencies.
+  `jax.jacrev` cannot do this cheaply, since it evaluates the reverse pass -- and
+  so a full timestepping run -- once per output.
+
 ## Meep 1.34.0
 
 7/9/2026
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,38 @@
 # Meep Release Notes
 
+## Meep 1.35.0 (in progress)
+
+* Adjoint solver: objective functions are now differentiated with a single
+  vector-Jacobian product instead of a full frequency Jacobian per objective
+  argument. For an objective of $M$ arguments at $F$ frequencies this reduces
+  the reverse passes from $M \times F$ to one and the peak memory for the
+  objective gradient from $O(F^2 N)$ to $O(F N)$, where $N$ is the number of
+  values a monitor contributes. Gradients are unchanged apart from summation
+  order.
+
+* Adjoint solver: objective functions may be written with `jax.numpy` instead of
+  `autograd.numpy` and passed to `OptimizationProblem` directly, with no wrapper
+  or annotation; they are recognized by the array type they return and
+  differentiated with `jax.vjp`. An objective function may also supply its own
+  pullback via a `vjp(cotangent, *args)` attribute, for an analytic derivative or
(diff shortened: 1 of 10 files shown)


oskooi (approved):
LGTM.

stevengj:
I remember us having  big problems with Jax and MPI on multi-core nodes when @ianwilliamson first implemented jax support.  In particular, JAX tried to grab all of the CPUs on the node, which is bad if you are doing MPI intra-node parallelism.  Is there a way to disable that these days?

smartalecH:
Closing in favour of #3299, which is the same branch and the same content, opened from this repository rather than from a fork.

Nothing about the change itself has been revised, and the description is carried over verbatim. The discussion and reviews on this PR stay readable here; please continue on #3299.

PR #3281, opened 2026-08-26
[5/13][Adjoint Module] Differentiable angular-spectrum propagation through stratified media

smartalecH:


stevengj:
Could you comment a bit on the algorithms here?  In principle it's just the same near2far machinery but with the Green's function of the stratified medium, and there have been lots of papers on computation of Green's functions for stratified media.

python/adjoint/angular_spectrum.py
```diff
@@ -0,0 +1,1196 @@
...
+
+def _safe_sqrt(argument: jnp.ndarray) -> jnp.ndarray:
+    """Square root with the branch chosen so that the imaginary part is >= 0.
+
+    Evanescent orders must decay away from the monitor rather than grow, which
+    fixes the branch. The caller is expected to have regularized the argument so
+    that it never lands exactly on zero; see DEFAULT_LOSS_REGULARIZATION.
+    """
+    root = jnp.sqrt(jnp.asarray(argument, dtype=jnp.complex128))
+    return jnp.where(jnp.imag(root) < 0, -root, root)
```
stevengj, line 160: Note that the default branch cut of `sqrt` always gives a nonnegative imaginary part for *real* arguments, so presumably you are worried about complex arguments here.

But if the argument is complex, it seems a little weird to arbitrary flip the sign here; I'm worried it might be masking a deeper problem.  Can you give an example of where you would be passing in correct inputs that give a negative imaginary part from `sqrt` here?

stevengj:
Could you comment a bit on the algorithms here?  In principle it's just the same near2far machinery but with the Green's function of the stratified medium, and there have been lots of papers on computation of Green's functions for stratified media.

At first glance, it seems like you're doing it by decomposing the input into planewaves (a continuous Fourier transform, or a Fourier series if you have periodic boundaries), and then using the S-matrix (accumulated by the usual Redheffer-product recurrence) for planewaves?   If it's not periodic boundaries, how do you choose the discretization in k space?

At one point I seem to recall talking about using https://github.com/modelics/strata?

PR #3283, opened 2026-08-27
[6/13][Adjoint Module] Adjoint gradients with respect to sources

smartalecH:


stevengj:
This is definitely nice to have.  (Currently @lxvm has been avoiding this issue by using reciprocal simulations, i.e. effectively swapping the sources with output overlap integrals.)

lxvm:
@stevengj I still think the main reason for doing reciprocal simulations is that I am simulating a structure with fewer output channels than input channels. But yes, having adjoints with respect to sources would enable more end-to-end applications where the source terms depend on a design region outside of the meep simulation.

PR #3290, opened 2026-09-02
[2/n][Adjoint Module] Fix source deposition across chunk boundaries

### LINKED ISSUE #1051: Simulation result depends on # of processes, using geometry_center

smartalecH:
### Stack

| | PR | |
|---:|---|---|
| 2 | #3290 | Fix source deposition across chunk boundaries **&larr; you are here** |

Closes #1051 

When users specify a geometry center, it shifts the locations of the sources relative to the chunk pattern (as expected). One bug that was fixed long ago, was when sources were accidentally placed on both chunks after a shift. Another issue (which is resolved here) is when a source shifts, but not all of it is properly shifted due to the way meep handles "ownership."

More specifically, when a zero-thickness source lies on a chunk boundary, Meep must interpolate it onto grid points on both sides. Previously, each chunk clipped that interpolation to its locally owned grid points. With a shifted simulation center, this could omit part of the source or store it on the wrong chunk, making results depend on the chunk layout.
  
The fix computes the complete source interpolation first, then moves each resulting source sample to the chunk that owns the corresponding field update. This makes source deposition independent of domain decomposition.

### DIFF
--- a/python/tests/test_source.py
+++ b/python/tests/test_source.py
@@ -125,6 +125,114 @@ class TestSourceTime(unittest.TestCase):
             self.assertEqual(src.fourier_transform(f), src.swigobj.fourier_transform(f))
 
 
+class TestSourceChunkBoundary(unittest.TestCase):
+    @staticmethod
+    def _get_source(
+        cell_size,
+        geometry_center,
+        source_center,
+        source_size,
+        chunk_layout,
+        run_until=None,
+    ):
+        sim = mp.Simulation(
+            cell_size=cell_size,
+            geometry_center=geometry_center,
+            resolution=8,
+            chunk_layout=chunk_layout,
+            sources=[
+                mp.Source(
+                    mp.ContinuousSource(1.0),
+                    component=mp.Ez,
+                    center=source_center,
+                    size=source_size,
+                )
+            ],
+        )
+        try:
+            sim.init_sim()
+            source = sim.get_source(mp.Ez)
+            energy = None
+            if run_until is not None:
+                sim.run(until=run_until)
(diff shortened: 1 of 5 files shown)


smartalecH:
@lxvm can you try this out?

lxvm:
Sure! Is there an issue I can reference to understand what this should fix?

smartalecH:
@lxvm this should resolve the issue you were referencing in #3264 (i.e. the value of your objective function depends on the chunk layout)

lxvm:
@smartalecH this pr fixes the consistency in my simulations w.r.t. MPI parallelization when I use `geometry_center`. Many thanks!

smartalecH:
@stevengj note that I stacked this on #3277 (to make it easier for Lorenzo to run the full patch suite) so if we approve, we need to merge that one first and rebase here.

smartalecH:
Closing in favour of #3298, which is the same branch and the same content, opened from this repository rather than from a fork.

Nothing about the change itself has been revised, and the description is carried over verbatim. The discussion and reviews on this PR stay readable here; please continue on #3298.

PR #3292, opened 2026-09-02
[3/n][Adjoint Module] Fix adjoint source weighting on a symmetry plane

### LINKED ISSUE #3291: Adjoint gradient is wrong when the objective monitor lies on a symmetry plane (3/4 on-plane; O(dx) crossing)
### Summary

With a mirror symmetry declared, the adjoint design gradient is correct as long as the objective monitor stays clear of the symmetry plane, and wrong when it sits on or crosses it. On the plane it is exactly **3/4** of the correct value; crossing it, the error is **first order in the grid spacing** (4.0% at resolution 20, 1.8% at 40). Without the symmetry the same setups match a finite difference to 0.02%.

An extended monitor straddling the plane is the usual arrangement for a symmetric device, so this affects most symmetric adjoint runs.

### What is *not* wrong

Worth stating up front, because it looks like a bug and is the first thing you notice: with `Mirror(Y)` the design gradient comes back with essentially all of its magnitude in `y >= 0`, as though the routine forgot to mirror the gradient back over the design region. That behaviour is correct. Meep builds the structure on the reduced grid volume, so weights below the plane are never read, and a weight above it drives both halves of the structure at once. Finite differences confirm both halves:

```
symmetry  node             FD dJ/du        adjoint   adj/FD
False     (8,12)      2.773475e+01   2.773700e+01   1.0001
False     (8,3)       2.773475e+01   2.773700e+01   1.0001
True      (8,12)      5.546950e+01   5.341287e+01   0.9629
True      (8,3)       0.000000e+00   0.000000e+00      nan
```

`FD(sym, upper) / FD(nosym, upper) = 2.0000` exactly, at every resolution tested. So the doubling and the zeroed half are both right.

### The actual defect

Sweeping the position of a point objective monitor relative to the plane, at resolution 20:

| monitor y | nosym adj/FD | sym adj/FD |
|---|---|---|
| 0.00 (on plane) | 1.0002 | **0.7484** |
| 0.05 (one pixel off) | 1.0002 | **0.8725** |
| 0.50 | 1.0002 | 1.0002 |
| 1.00 | 1.0002 | 1.0001 |

The on-plane case is resolution independent — 0.7484 at resolution 20, **0.7500** at resolution 40 — so it is a scaling error rather than discretization.

An extended monitor spanning the plane inherits it diluted over the rows it covers, which gives the first-order behaviour:

| resolution | sym adj/FD | error |
|---|---|---|
| 20 | 0.9602 | 3.98% |
| 30 | 0.9715 | 2.85% |
| 40 | 0.9819 | 1.81% |

That is consistent with the single row on the plane being mishandled out of the `O(1/dy)` rows the monitor covers.

### Where it is not

- The forward solve is unaffected: the objective is identical with and without the symmetry to nine digits (`9.65660209e+01` both ways).
- Plain sources on the plane are handled correctly. Running the same `mp.Source` at `y=0` with and without `Mirror(Y)` gives DFT fields agreeing to `4.6e-16`. (Off the plane the symmetric run is 2x, which is the expected replication.)
- The gradient accumulation away from the plane is exact.

Together these point at how the adjoint source is weighted at the symmetry plane — `dft_fields::fourier_sourcedata` in `src/dft.cpp`, which applies `EH0 /= f->S.multiplicity(ix0)` — rather than at `material_grids_addgradient`. I have not isolated the line that produces the 3/4.

Note that `material_grids_addgradient` in `src/meepgeom.cpp` never references the chunk's `S`/`sn`, unlike `fourier_sourcedata` and `dft_chunk::process`. Transforming the deposit position there is *not* the fix — I tried it, and it makes things worse, because the untransformed accumulation is what implements the "one variable drives both halves" doubling. Recording it here so the next person does not repeat the experiment.

### Reproducer

Attached as a test file in the linked PR: off-plane agreement and the inert mirrored half pass; the on-plane and plane-crossing cases are `expectedFailure` so that a fix flips them. Meep currently has no adjoint test that declares a symmetry at all.

Meep 1.35.0-beta, double precision, single rank.

smartalecH:
### Stack

| | PR | |
|---:|---|---|
| 3 | #3292 | Fix adjoint source weighting on a symmetry plane (#3291) **&larr; you are here** |

Fixes #3291.

`fourier_sourcedata` divides each interpolated adjoint source by the multiplicity of the point it sits at, to undo the replication a reduced symmetric domain applies. In the `yee_grid = false` branch the amplitude is split over the four Yee corners of a voxel, but the divisor was evaluated once at the voxel *centre* and reused for all four.

Those disagree exactly on a symmetry plane. A point there is its own image, so its multiplicity is 1, while the centre half a pixel away has 2. The corner on the plane was therefore halved when it should not have been, leaving the adjoint source short.

| objective vs. the plane | before | after |
|---|---|---|
| on the plane | 0.7484 | 1.0002 |
| one pixel off | 0.8725 | 1.0002 |
| clear of it | 1.0002 | 1.0002 |
| extended, spanning it (res 20 / 30 / 40) | 0.9602 / 0.9715 / 0.9819 | 1.0002 / 0.9975 / 1.0000 |

(finite difference = 1; no symmetry declared is 1.0002 throughout)

The on-plane number is exactly 3/4 and resolution independent — the source's three Yee rows carry weights 1/4, 1/2, 1/4 and the halved one is the middle row, so exactly a quarter of the source goes missing. An extended monitor merely spanning the plane dilutes that single row over the `O(1/dx)` rows it covers, which is why it presented as a first-order error rather than a constant one. The last row is the common arrangement for a symmetric device, so most symmetric adjoint runs were carrying a few percent of gradient error.

The fix takes the multiplicity of the site each quarter is actually written to. Off the plane nothing changes, and with no symmetry declared every multiplicity is 1, so this is a no-op for the vast majority of runs.

### Tests

Meep had no adjoint test that declared a symmetry at all. The first commit adds one, including the part that looks like a bug and is not: with `Mirror(Y)` the design gradient puts all of its magnitude in `y >= 0`, because Meep builds the structure on the reduced grid volume, so weights below the plane are never read and a weight above it drives both halves at once. Finite differences confirm both halves — zero below the plane, and exactly `2.0000x` above it.

Those two commits are separated deliberately: the first records the on-plane and plane-crossing failures as `expectedFailure`, and the second turns them into plain assertions, so the diff shows the fix flipping them.

`test_adjoint_solver.py` (44 gradient checks) and `test_adjoint_cyl.py` both pass unchanged.

### DIFF
--- a/python/Makefile.am
+++ b/python/Makefile.am
@@ -34,6 +34,7 @@ ADJOINT_TESTS =                                \
     $(TEST_DIR)/test_adjoint_solver.py         \
     $(TEST_DIR)/test_adjoint_utils.py          \
     $(TEST_DIR)/test_adjoint_cyl.py            \
+    $(TEST_DIR)/test_adjoint_symmetry.py       \
     $(TEST_DIR)/test_adjoint_jax.py
 
 TESTS =                                        \
--- a/python/tests/test_adjoint_symmetry.py
+++ b/python/tests/test_adjoint_symmetry.py
@@ -8,8 +8,12 @@
 # both halves at once. A finite difference confirms both halves of that: the
 # gradient really is zero down there, and really is doubled up here.
 #
-# What is a bug is the scaling when the *objective* sits on the symmetry plane,
-# which these tests pin down rather than paper over.
+# What was a bug is the scaling when the *objective* sits on or spans the
+# symmetry plane. `fourier_sourcedata` divided each quarter of an interpolated
+# adjoint source by the multiplicity of the voxel *centre* rather than of the
+# Yee site it was written to. Those differ exactly on a symmetry plane, where a
+# point is its own image and so has multiplicity 1 while the centre half a pixel
+# away has 2, leaving the source short there. The last two tests cover it.
 
 import unittest
 
@@ -169,14 +173,13 @@ class TestAdjointMirrorSymmetry(unittest.TestCase):
         without = finite_difference(w, False, *CROSSING, ix, iy)
         self.assertAlmostEqual(with_sym / without, 2.0, places=3)
 
-    @unittest.expectedFailure
     def test_gradient_matches_finite_difference_on_plane(self):
-        """Known wrong: an objective on the symmetry plane gives 3/4 the gradient.
+        """An objective sitting on the symmetry plane.
 
-        Exactly 0.7500 at resolution 40 and 0.7484 at 20, so this is a scaling
-        error and not discretization. Everything else in this file passes, which
-        places it in how the adjoint source is weighted at the plane rather than
-        in the gradient accumulation.
+        This gave exactly 3/4 of the correct gradient while the adjoint source
+        was divided by the voxel centre's multiplicity: of the source's three Yee
+        rows, the one on the plane carries half the weight and was the one being
+        halved, so 1/4 of the total went missing regardless of resolution.
         """
         w = self.weights
         ix, iy = 8, 15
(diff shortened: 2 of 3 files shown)


smartalecH:
Closing in favour of #3296, which is the same branch and the same content, opened from this repository rather than from a fork.

Nothing about the change itself has been revised, and the description is carried over verbatim. The discussion and reviews on this PR stay readable here; please continue on #3296.

PR #3297, opened 2026-09-02
[1/13][Adjoint Module] Fix MaterialGrid design-variable attribution in the adjoint gradient

### LINKED ISSUE #1984: Incorrect adjoint gradients with overlapping design variables

smartalecH:
> Recreated so the whole series can form a single GitHub stack. Native stacked pull requests cannot include PRs opened from a fork, and this branch was previously hosted on one. The content and description are unchanged; earlier review discussion is on the superseded PR linked below, which remains readable.

_Supersedes #3277._

Fixes #1984.

Stacked on #3274 — based on `fix/adjoint-chunk-pairing-minimal`, so the diff here is just these two commits. Merge #3274 first, then this retargets to `master` cleanly.

While working on a new 3D grating coupler TO example for meep (2 design regions) I noticed that the gradients were rather inaccurate. Claude helped me trace things down to the unique case where we have two adjacent design regions (e.g. a partial etched layer). There were different issues depending on the sizes of the design regions (e.g. if one design region was bigger than the other, then we got a segfault!)

The core issue was that the logic we had for handling multiple design grids during the recombination step wasn't properly bookkeeping the effects near the boundaries. This was because we weren't keeping track of "which grid was which." In addition, we had some hairy "copies" of grids we were trying to use to keep track of multiple grids (hence the crash). So the resulting gradient near the boundary was completely off. To fix this, we simply add a grid id (automatically populated in python) and propagated to the cpp backend via swig.

Practically speaking, this means that **the unit of differentiation is now the design variable, not the grid copy.**

Turns out this bug also affects symmetries (see #1984) and the fix also resolves it.

### DIFF
--- a/src/meepgeom.cpp
+++ b/src/meepgeom.cpp
@@ -15,6 +15,7 @@
 */
 
 #include <algorithm>
+#include <set>
 #include <vector>
 #include "meepgeom.hpp"
 #include "meep_internals.hpp"
@@ -2640,7 +2641,8 @@ get_material_gradient(const meep::vec &r,              // current point
                       geom_epsilon *geps,              // material
                       meep::grid_volume &gv,           // simulation grid volume
                       double du,                       // step size
-                      double *u,                       // matgrid
+                      const std::vector<double *> &us, // matgrid weights, one per
+                                                       // copy of this design grid
                       int idx                          // matgrid index
 ) {
   /*Compute the Aᵤx product from the -λᵀAᵤx calculation.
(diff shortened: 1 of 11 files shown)


src/meepgeom.cpp
```diff
@@ -2716,58 +2731,66 @@ With the addition of subpixel smoothing, however, the vJp became
...
+                          dz);
+    int x_list[2] = {x1, x2}, y_list[2] = {y1, y2}, z_list[2] = {z1, z2};
+    int lx = (x1 == x2) ? 1 : 2;
+    int ly = (y1 == y2) ? 1 : 2;
+    int lz = (z1 == z2) ? 1 : 2;
+
+    if (k == 0 && (ukind == material_data::U_MIN || ukind == material_data::U_PROD)) {
+      /* U_MIN and U_PROD carry adjustments defined in terms of the interpolated
+         value at this point. Overlapping grids of these kinds are rejected by
+         the caller, so there is exactly one copy and this is unchanged. */
```
stevengj, line 2762: What does it mean that "overlapping grids of these kinds are rejected by the caller"?   The whole point of `U_MIN` and `U_PROD` is that they apply to overlapping grids.

stevengj:
The overall approach seems fine (adding an id per material grid and using that when differentiating with respect to the variables in a single grid), but I want to look closely at some of the details.  One comment in the code, noted above, was disturbing.

PR #3298, opened 2026-09-02
[2/13][Adjoint Module] Fix source deposition across chunk boundaries

### LINKED ISSUE #1051: Simulation result depends on # of processes, using geometry_center

smartalecH:
> Recreated so the whole series can form a single GitHub stack. Native stacked pull requests cannot include PRs opened from a fork, and this branch was previously hosted on one. The content and description are unchanged; earlier review discussion is on the superseded PR linked below, which remains readable.

_Supersedes #3290._

Closes #1051 

When users specify a geometry center, it shifts the locations of the sources relative to the chunk pattern (as expected). One bug that was fixed long ago, was when sources were accidentally placed on both chunks after a shift. Another issue (which is resolved here) is when a source shifts, but not all of it is properly shifted due to the way meep handles "ownership."

More specifically, when a zero-thickness source lies on a chunk boundary, Meep must interpolate it onto grid points on both sides. Previously, each chunk clipped that interpolation to its locally owned grid points. With a shifted simulation center, this could omit part of the source or store it on the wrong chunk, making results depend on the chunk layout.
  
The fix computes the complete source interpolation first, then moves each resulting source sample to the chunk that owns the corresponding field update. This makes source deposition independent of domain decomposition.

### DIFF
--- a/src/loop_in_chunks.cpp
+++ b/src/loop_in_chunks.cpp
@@ -418,8 +418,11 @@ void fields::loop_in_chunks(field_chunkloop chunkloop, void *chunkloop_data, con
       for (int i = 0; i < num_chunks; ++i) {
         if (!chunks[i]->is_mine()) continue;
         grid_volume gvu(chunks[i]->gv);
-        ivec _iscoS(S.transform(gvu.little_owned_corner(cS), sn));
-        ivec _iecoS(S.transform(gvu.big_owned_corner(cS), sn));
+        ivec _iscoS(S.transform(use_symmetry ? gvu.little_owned_corner(cS)
+                                             : gvu.little_corner() + gvu.iyee_shift(cS),
+                                sn));
+        ivec _iecoS(S.transform(
+            use_symmetry ? gvu.big_owned_corner(cS) : gvu.big_corner() + gvu.iyee_shift(cS), sn));
         ivec iscoS(max(user_volume.little_owned_corner(cgrid), min(_iscoS, _iecoS))),
             iecoS(max(_iscoS, _iecoS)); // fix ordering due to to transform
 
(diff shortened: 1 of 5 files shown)


src/sources.cpp
```diff
@@ -501,8 +512,10 @@ void fields::add_volume_source(component c, const src_time &src, const volume &w
       data.amp *= gv.a; // correct units for J delta-function amplitude
   }
   sources = src.add_to(sources, &data.src);
+  register_src_time(data.src);
```
stevengj, line 515:
```suggestion
  register_src_time(data.src); // needed by fix_boundary_sources
```

src/loop_in_chunks.cpp
```diff
@@ -418,8 +418,11 @@ void fields::loop_in_chunks(field_chunkloop chunkloop, void *chunkloop_data, con
       for (int i = 0; i < num_chunks; ++i) {
         if (!chunks[i]->is_mine()) continue;
         grid_volume gvu(chunks[i]->gv);
-        ivec _iscoS(S.transform(gvu.little_owned_corner(cS), sn));
-        ivec _iecoS(S.transform(gvu.big_owned_corner(cS), sn));
+        ivec _iscoS(S.transform(use_symmetry ? gvu.little_owned_corner(cS)
+                                             : gvu.little_corner() + gvu.iyee_shift(cS),
```
stevengj, line 422: Note: `gvu.little_corner() + gvu.iyee_shift(cS)` is the first point in the chunk for that component, whether it is owned or not.

It would be good to add a comment explaining why this is needed for the `!use_symmetry` case.

src/loop_in_chunks.cpp
```diff
@@ -445,6 +448,18 @@ void fields::loop_in_chunks(field_chunkloop chunkloop, void *chunkloop_data, con
...
+        // A chunk origin may lie between points of the global component grid.
+        // For source loops (use_symmetry=false), keep the loop aligned with the
+        // global grid so that interpolation weights are independent of chunking.
+        if (!use_symmetry) {
+          LOOP_OVER_DIRECTIONS(gv.dim, d) {
+            const int parity = (iscS.in_direction(d) - (is - shifti).in_direction(d)) % 2;
+            if (parity) iscS.set_direction(d, iscS.in_direction(d) + 1);
+            if ((iecS.in_direction(d) - (is - shifti).in_direction(d)) % 2)
+              iecS.set_direction(d, iecS.in_direction(d) - 1);
+          }
```
stevengj, line 460: This is a little bit weird?  Won't shifting by 1 shift it to a different Yee grid?

PR #3303, opened 2026-09-04
Draw perfect metals (ε = -∞) in dark red in `plot2D`

oskooi:
**Context**

`mp.metal` and `mp.perfect_electric_conductor` are defined as `Medium(epsilon=-inf)` ([`python/meep.i:1825-1826`](https://github.com/NanoComp/meep/blob/8f399d07bc4044dfd6e02940d89f076f266ec1e8/python/meep.i#L1825-L1826)), and that `-inf` survives as a large finite `-1e20` all the way through `geom_epsilon::chi1p1` → `Simulation.get_epsilon_grid`.

`plot_eps` — the function `plot2D` calls to draw the geometry ([`python/visualization.py:599-732`](https://github.com/NanoComp/meep/blob/8f399d07bc4044dfd6e02940d89f076f266ec1e8/python/visualization.py#L599-L732)) — passes that array straight to `ax.imshow` and, when `colorbar=True`, straight into `mpl.colors.Normalize(np.amin(eps_data), np.amax(eps_data))`. The consequences of this are:

- **Colorbar:** `vmin = -inf`, so the colorbar is meaningless and every finite dielectric collapses onto one end of the scale.
- **Image:** matplotlib's internal `safe_masked_invalid` masks the `-inf` pixels and draws them with the colormap's "bad" color, which defaults to *fully transparent* — metal renders as blank background rather than as a material (see first image below).
- **Contour path:** `levels=np.unique(eps_data)` includes `-inf` as a contour level.

**Changes**

`python/visualization.py`
- new `pec_color: "darkred"` key in `default_eps_parameters` (dropped by `filter_dict` before reaching `imshow`, like the other non-imshow dicts).
- `plot_eps` now masks metallic pixels (`~np.isfinite(eps) | (eps <= -mp.inf)`) and paints them via `cmap.set_bad(pec_color)` on a `copy.copy` of the colormap, so `imshow` autoscales over the finite permittivities only.
- Colorbar `vmin`/`vmax` now come from the unmasked values, with a guard for an all-metal plane.
- Contour path uses `np.unique(eps_data.compressed())` so `-1e20` is no longer emitted as a contour level.
- Fixed a latent aliasing bug: `eps_parameters = default_eps_parameters` handed out the module-level dict itself, which the new code would have mutated — now `copy.deepcopy`.
- Added `_get_colormap()` (handles str or `Colormap`, matplotlib ≥3.5 or older) and used it in `_add_colorbar`; required because a `Colormap` instance is now passed there, and `mpl.cm.get_cmap` is removed in matplotlib 3.11. This also clears the deprecation warning the old cell emitted.

**Verification**

`test_plot2D_pec` (clim finite and equal to (1, 12), bad color = `darkred`, mask non-empty, defaults unmutated, user override honored) and `test_plot2D_pec_contour`.

Here is an example involving a [cleaved multilayer stack](https://meep.readthedocs.io/en/latest/Python_Tutorials/Custom_Source/#dipole-emission-of-a-cleaved-multilayer-stack) in which the substrate is a perfect metal. Metallic/PEC regions render as a distinct dark red, and the color scale (image and colorbar) is driven only by the finite permittivity values.

edge_emitter_layout

edge_emitter_layout

### DIFF
--- a/python/visualization.py
+++ b/python/visualization.py
@@ -690,7 +690,7 @@ def plot_eps(
             return eps_data
 
         # Metals / perfect electric conductors have eps = -inf (see python/meep.i,
-        # where mp.inf is the sentinal 1e20), which would otherwise saturate the
+        # where mp.inf is the sentinel 1e20), which would otherwise saturate the
         # color scale. Mask them out so that the scale is set by the finite
         # permittivities, and draw them in `pec_color` instead.
         eps_data = np.ma.masked_where(
--- a/python/visualization.py
+++ b/python/visualization.py
@@ -14,6 +14,7 @@ from meep.simulation import Simulation, Volume
 
 ## Typing imports
 from matplotlib.axes import Axes
+from matplotlib.colors import Colormap
 from matplotlib.figure import Figure
 from typing import Callable, Union, Any, Tuple, List, Optional
 
@@ -61,6 +62,7 @@ default_eps_parameters = {
     "frequency": None,
     "resolution": None,
     "colorbar": False,
+    "pec_color": "dimgrey",
 }
 
 default_colorbar_parameters = {
@@ -556,9 +558,21 @@ def plot_volume(
     return ax
 
 
+def _get_colormap(cmap: Union[str, Colormap]) -> Colormap:
+    """Return a Colormap object given a colormap name or Colormap object."""
+    import matplotlib as mpl
+
+    if isinstance(cmap, Colormap):
+        return cmap
+    try:
+        return mpl.colormaps[cmap]  # matplotlib >= 3.5
+    except AttributeError:
+        return mpl.cm.get_cmap(cmap)  # matplotlib < 3.5
+
+
 def _add_colorbar(
     ax: Axes,
-    cmap: str,
+    cmap: Union[str, Colormap],
     vmin: float,
     vmax: float,
     default_label: Optional[str] = None,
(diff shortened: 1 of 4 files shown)


python/visualization.py
```diff
@@ -607,7 +620,7 @@ def plot_eps(
 ) -> Union[Axes, Any]:
     # consolidate plotting parameters
     if eps_parameters is None:
-        eps_parameters = default_eps_parameters
+        eps_parameters = copy.deepcopy(default_eps_parameters)
```
stevengj: Probably `copy.copy` is enough here, though `deepcopy` doesn't hurt?

python/visualization.py
```diff
@@ -676,11 +689,22 @@ def plot_eps(
         if not ax:
             return eps_data
 
+        # Metals / perfect electric conductors have eps = -inf (see python/meep.i,
+        # where mp.inf is the sentinal 1e20), which would otherwise saturate the
```
stevengj:
```suggestion
        # where mp.inf is the sentinel 1e20), which would otherwise saturate the
```

python/visualization.py
stevengj:
```suggestion
        eps_parameters = copy.copy(default_eps_parameters)
```

stevengj (approved):

PR #3307, opened 2026-09-11
Fix mpb degenerate band truncation

### LINKED ISSUE #2298: sporadic failure in `test_output_efield_z` of `test_mpb.py`
The `make check` unit test suite which is run automatically once every day as part of the Github Actions CI occasionally reports a failure in `test_output_efield_z` of `test_mpb.py` (see logs of failing scheduled runs in the master repo [here](https://github.com/NanoComp/meep/actions/workflows/build-ci.yml?query=event%3Aschedule+is%3Afailure)).  This failure occurred most recently yesterday in my fork ([log](https://github.com/oskooi/meep/actions/runs/3366721986)). The failure occurs with and without MPI even though MPB itself is always compiled serially. I have also encountered this error when running `make check` on my local machine from time to time.

The [log file](https://github.com/oskooi/meep/suites/9060725220/artifacts/419025647) from the Github Actions CI artifacts shows:

```
FAIL: test_output_efield_z (__main__.TestModeSolver)
  File "/home/runner/work/meep/meep/build/meep-1.25.0-beta/_build/sub/python/../../../python/tests/test_mpb.py", \
line 570, in test_output_efield_z
    self.compare_h5_files(ref_path, res_path)
  File "/home/runner/work/meep/meep/build/meep-1.25.0-beta/_build/sub/python/../../../python/tests/test_mpb.py", \
line 236, in compare_h5_files
    compare_arrays(self, ref[k][()], res[k][()], tol=tol)
  File "/home/runner/work/meep/meep/build/meep-1.25.0-beta/python/tests/utils.py", line 17, in compare_arrays
AssertionError: 1.3740575041517082 not less than 0.001
```

`test_output_efield_z` is relatively short and involves comparing the $E_z$ from a unit-cell calculation of a 2d photonic crystal to precomputed values in an HDF5 file:

https://github.com/NanoComp/meep/blob/ef5a35a9d250c35dbcfde5afaafb9a70f076cc09/python/tests/test_mpb.py#L561-L570

https://github.com/NanoComp/meep/blob/ef5a35a9d250c35dbcfde5afaafb9a70f076cc09/python/tests/test_mpb.py#L49-L67

yorickreum:
Fixes https://github.com/NanoComp/meep/issues/2298 by not splitting degenerate bands.


oskooi (approved):
LGTM.

python/tests/test_mpb.py
```diff
@@ -230,11 +229,25 @@ def check_fields_against_h5(self, ref_path, field, suffix=""):
...
+                paired = {
+                    k[:-2] for k in keys if k.endswith(".r") and k[:-2] + ".i" in keys
+                }
+                for base in sorted(paired):
+                    x = ref[base + ".r"][()] + 1j * ref[base + ".i"][()]
+                    y = res[base + ".r"][()] + 1j * res[base + ".i"][()]
+                    self.assertClose(x, y, epsilon=tol, msg=f"dataset {base}")
+                for k in keys:
+                    if k[-2:] in (".r", ".i") and k[:-2] in paired:
+                        continue
```
stevengj, line 246: Comparing the `.r` and `.i` components as a single complex array makes sense to me.   I'm not sure what it has to do with the degeneracy problem, though?

python/tests/test_mpb.py
```diff
@@ -560,7 +573,10 @@ def test_compute_group_velocity(self):
...
 
     def test_output_efield_z(self):
-        ms = self.init_solver()
+        # Bands 8 and 9 are degenerate at Gamma, so requesting exactly 8 bands
+        # leaves band 8 defined only up to mixing with band 9, and the field
+        # written below irreproducible.  Ten bands ends the block at a real gap.
+        ms = self.init_solver(num_bands=10)
         ms.run_tm()
         mpb.fix_efield_phase(ms, 8)
         mpb.output_efield_z(ms, 8)
```
stevengj, line 582: Won't you still potentially get test failures if different versions of MPB produce a different linear combination for bands 8 and 9 at Gamma?

python/tests/test_mpb.py
yorickreum: Indeed not directly related to the degeneracy problem. https://github.com/NanoComp/meep/pull/3307/changes/c43b9c028bb82da3a267a55dee29f5bdc4235247 is purely about not getting sporadic failures when comparing the .i halves, which for a phase-fixed field are only roundoff, with a relative tolerance. I hope this also becomes clear from the commit message.

It is still needed to get the test reliable, at least on my Windows system. Without the commit in, the test still failed in 2 out of 20 runs. (With both commits in, 20/20 runs are successful.)

python/tests/test_mpb.py
yorickreum: Potentially yes. But imo that's a pre-existing risk which this PR shrinks rather than introduces. And I don't see it occurring, neither on Linux nor on Windows, yet.

Ripping the bands apart at a doublet (bands 8 and 9 at Gamma) is dangerous, because you don't know which vectors you end up with in the eigenspace. Ripping apart at a singlet (there is a sufficient gap of about 7e-2 between 10 and 11 at Gamma) appears safe to me so far (without any more knowledge about how MPB works internally and its different versions...). The run-to-run variation of the band 8 field drops from 3.2e-02 to 1.8e-08 with it.

Alternatively one could output a band that is a singlet at Gamma (7 or 10) instead of band 8, which would remove the ambiguity by construction. That needs a regenerated reference ... but both options should do the trick imo.

Issue #2244, opened 2022-09-24
Error in final "make check" in test_mpb.py for meep 1.24


udaykdk:
This is regarding the "Installation from source on macOS" in the MEEP installation instructions. I followed all the instructions for installing MEEP 1.24 on my Macbook M1 and everything seemed to have gone well. The final instruction was to run a "make check" to run a self-test. Here is where I got one FAIL: 

FAIL: tests/test_mpb.py

Probing the installation log further, I see:

```
grep FAIL python/test-suite.log 
# XFAIL: 0
# FAIL:  1
FAIL: tests/test_mpb
FAIL: test_diamond (__main__.TestModeSolver)
FAIL: test_epsilon_input_file (__main__.TestModeSolver)
FAIL: test_hole_slab (__main__.TestModeSolver)
FAIL: test_output_efield_z (__main__.TestModeSolver)
FAIL: test_run_te_no_geometry (__main__.TestModeSolver)
FAIL: test_strip (__main__.TestModeSolver)
FAILED (failures=6)
FAIL tests/test_mpb.py (exit status: 1)
```

Can someone help debug this error?

 thanks,
 uday

Yaraslaut:
You can try to look at `python/tests/test_mpb.log`  to see what has happened there exactly.

udaykdk:
Thanks. All the errors have one thing in common, they refer to an
"AssertionError", e.g.

self.assertAlmostEqual(r, e, places=3)
AssertionError: 1.4651880980146579 != 1.469257717113039 within 3 places
(0.004069619098381105 difference)

Any tips on how to fix this?
 thanks,
uday

Yaraslaut:
> Thanks. All the errors have one thing in common, they refer to an "AssertionError", e.g. self.assertAlmostEqual(r, e, places=3) AssertionError: 1.4651880980146579 != 1.469257717113039 within 3 places (0.004069619098381105 difference) Any tips on how to fix this? thanks, uday
> […](#)

I think that it is not easy to fix and personally I would  just ignore this small difference in results.

Issue #1853, opened 2021-12-11
improved Apple silicon support


stevengj:
I now have an Apple M4 laptop, and I've verified that Meep `master` and all of its dependencies build from source.   I've also updated the macOS compilation instructions in the documentation.   Some to-do's:

- [x] C++ tests passing
- [ ] Python tests passing
- [ ] Conda packages — miniforge supports M4 these days
- [ ] enable flushing subnormals: https://github.com/NanoComp/meep/issues/1708#issuecomment-991717592
- [ ] CI support — is there a way to enable this?

The C++ tests pass for me, but I'm currently getting some Python test failures that I'm tracking down.

In general, it's nice to have code running on multiple architectures, because it helps expose bugs.

stevengj:
Most of the Python tests are passing.

Several tests are failing simply because I don't have `h5py` installed — `pip3 install h5py` is failing, but presumably I should be able to get it to work if I can tell pip where to find the homebrew hdf5 library.

The only really concerning case is that `tests/test_source` is segfaulting.

stevengj:
The JAX test is failing:
```py
FAIL: tests/test_adjoint_jax
============================

/Users/stevenj/Library/Python/3.8/lib/python/site-packages/jax/_src/lib/__init__.py:32: UserWarning: JAX on Mac ARM machines is experimental and minimally tested. Please see https://github.com/google/jax/issues/5501 in the event of problems.
  warnings.warn("JAX on Mac ARM machines is experimental and minimally tested. "
Traceback (most recent call last):
  File "/Users/stevenj/Library/Python/3.8/lib/python/site-packages/jax/_src/lib/__init__.py", line 37, in <module>
    import jaxlib as jaxlib
ModuleNotFoundError: No module named 'jaxlib'
```
(I did `pip3 install jax`)

stevengj:
The MPB test is failing, possibly just due to the tolerances being too low:
```py
AIL: test_compute_field_energy (__main__.TestModeSolver)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "./tests/test_mpb.py", line 439, in test_compute_field_energy
    self.assertTrue(expected_fp.close(field_pt))
AssertionError: False is not true

======================================================================
FAIL: test_hole_slab (__main__.TestModeSolver)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "./tests/test_mpb.py", line 755, in test_hole_slab
    self.compare_h5_files(ref_path, res_path)
[...]

PR #3308, opened 2026-09-11
Fix test_run_te_no_geometry convergence at degenerate band

yorickreum:
Bands 1 and 2 of the empty lattice are degenerate at k=(0.5,0,0): |k+G| is 0.5 for both G=(0,0,0) and G=(-1,0,0).  The sweep reaches that k-point carrying the previous point's fields as its initial guess, which is a poor basis for the pair, and at 1e-7 -- the only run in this file not using 1e-12 -- the solver stops with band 2 still at 0.5447.  That exceeds the frequency at the next k-point, so the band minimum relocates there and comes out as |(0.5,0.1,0)| = 0.50990 instead of 0.5.

Solving that k-point on its own converges correctly at any tolerance, so what fails is the stopping criterion during the sweep; whether it is met before band 2 settles depends on details of BLAS, FFTW and libm.  The expected values are accurate to 1e-11, four orders tighter than a 1e-7 criterion can justify, so ask for that much convergence.  The assertions are unchanged.

Same class of defect as #2298: a test asserting more precision than its convergence resolves at a degeneracy.  Reported among the test_mpb failures on Apple silicon in #1853 and in #2244.


yorickreum:
Looking at the CI again, just changing the tolerance won't fix it reliably on all platforms. 
Will think about it again.

yorickreum:
Replaced the approach in this PR. 

The original one (tightening `ms.tolerance` to 1e-11) was wrong, as the CI caught, it failed all three jobs on Linux with exactly the symptom it was supposed to fix (detected on a Windows machine).

The tolerance is not the right lever. MPB seeds each k-point with the previous k-point's fields, so changing the tolerance changes what earlier k-points hand over, and its effect on the degenerate pair has no consistent sign:

| | tol 1e-7 | tol 1e-11 |
|---|---|---|
| Linux (CI) | passes | band 2 = 0.54484 |
| mingw-w64 / Windows | band 2 = 0.5447 | passes |

What actually happens at k=(0.5,0,0), from the CI log:

    tefreqs:, 6, 0.5, 0, 0,  0.5, 0.5, 0.54484, 1.12367, 1.13047, 1.13732, ...
    solve_kpoint (0.5,0,0):    Finished solving for bands 1 to 8 after 2 iterations.
    solve_kpoint (0.5,0.1,0):  Finished solving for bands 1 to 8 after 12 iterations.

The whole k-point is unconverged. Bands 3-6 should be 1.118034 four-fold. And the eigensolver stops after two iterations because the inherited basis is close enough to invariant to satisfy the stopping criterion. The next k-point then needs 12 iterations to recover. 

So the PR now randomizes the fields at each k-point and leaves the tolerance at 1e-7.

With that, all eight bands match the analytic empty-lattice |k+G| to under 1e-08. The one expected value that had to change is band 8's maximum: 1.80278 is an eigenvalue at k=(0.5,0,0) but not the eighth, which is 1.5. The inherited guess misses one of the two 1.5 states and picks up an 1.80278 state instead. The randomized approach finds the correct states imo,
over the k-path band 8 actually peaks at 1.64012 at k=(0.3,0,0), which is what the test now asserts.

python/tests/test_mpb.py
```diff
@@ -309,13 +309,16 @@ def test_run_te_no_geometry(self):
             ),
             (
                 (1.4142135623752818, mp.Vector3(0.0, 0.0, 0.0)),
-                (1.8027756376524435, mp.Vector3(0.5, 0.0, 0.0)),
+                (1.6401219539085534, mp.Vector3(0.3, 0.0, 0.0)),
```
oskooi, line 312: Why not change this line to `(1.5, mp.Vector3(0.5, 0.0, 0.0)),` since this is the correct result for the eighth eigenvalue which randomizing the fields actually uncovers (as you described in your bug report)?

yorickreum: 1.5 is the correct eighth eigenvalue at k=(0.5,0,0), but this entry here is the maximum of band 8 over the whole sweep, not its value at that k-point. The maximum sits at k=(0.3,0,0).

```
k=(0.3,0)   band8 = 1.64012195     <- max over the path
k=(0.5,0)   band7 = 1.5, band8 = 1.5, band9 = 1.80277564

```

So `(1.5, mp.Vector3(0.5, 0.0, 0.0))` would fail both assertions, the frequency by 0.14 and the k-point by 0.2.

So, the old value, 1.8027756, is exactly band 9 at k=(0.5,0,0). So whatever generated the reference likely converged onto a subspace that dropped one of the degenerate 0.5 states. Everything below shifts up a slot and the ninth eigenvalue lands in the band 8 entry...

oskooi: You are correct: [`mpb.ModeSolver.band_range_data`](https://mpb.readthedocs.io/en/latest/Python_User_Interface/#output-variables) returns a list of `(frequency_min, kpoint_min), (frequency_max, kpoint_max)`. The proposed change is therefore fine.

oskooi (approved):
LGTM.

Issue #385, opened 2018-06-18
Is it possible to use CMAKE to handle project ,so source code can be compiled on windows?


cathty:
Is it possible to use CMAKE to handle project ,so source code can be compiled on windows?

stevengj:
It wouldn’t really help, since you’d also need to compile all the dependencies. 

The best way to compile on Windows is currently to install Cygwin.

cenit:
I'd like to be able to build meep on Windows "natively", exploiting [vcpkg](https://github.com/Microsoft/vcpkg) for external libraries (blas, lapack, mpi, hdf5, ...), starting from a portable CMakeLists.txt for this code. If I understand correctly, to have only the C++ interface and barely no special output format, the dependencies are even zero, am I right? So that the work can proceed step-by-step, and starting with a barely functional meep with the minimum effort, then expanding to all dependencies. 
Is there any other one interested, to split the work?

stevengj:
I'd rather support Windows via Conda packages, since that's how we are currently distributing Meep binaries.

cenit:
but you still have to build it via Visual Studio, am I wrong?

stevengj:
No, we can use mingw, which is nice because it should allow us to use the existing unixy build system: https://github.com/conda-forge/conda-forge.github.io/issues/112

cenit:
I've lost track of the `MinGW` project long ago, it was really hacky and unsuitable, at least for what I was doing. I will give it a second try, I just finished reading all the thread and it seems that things improved, if Conda-forge has adopted it for many wheels for Windows. I'd like to see this project to run on windows natively asap, and maybe I will keep an eye on a cmake toolchain in the meantime. I hope it won't be too hard (some dependencies like `guile` and `libctl` look more difficult than `meep` itself)!

cenit:
Ok, not my first priority but this is where I started some work in order to build meep (libctl for now ;) ) with mingw on Windows: https://github.com/Microsoft/vcpkg/pull/4997/files

PR #3309, opened 2026-09-11
Native build on Windows

yorickreum:
Turns out that not much needs to be adjusted to be able to build and run MEEP natively on Windows (with MYSY2 UCRT64)... All required changes are in the first commit, the second commit adds a Windows CI runner.

Addresses https://github.com/NanoComp/meep/issues/385


python/tests/test_mpb.py
```diff
@@ -567,7 +566,7 @@ def test_output_efield_z(self):
 
         ref_fname = "tutorial-e.k16.b08.z.tm.h5"
         ref_path = os.path.join(self.data_dir, ref_fname)
-        res_path = re.sub("tutorial", ms.filename_prefix, ref_fname)
+        res_path = ref_fname.replace("tutorial", ms.filename_prefix)
```
oskooi, line 569: These changes seem unrelated to this PR. Perhaps they can be part of a separate PR?

yorickreum: > These changes seem unrelated to this PR. Perhaps they can be part of a separate PR?

They are required to get the test script to run on Windows.

`filename_prefix` is `os.path.join(self.temp_dir, ...)`, so on Windows it's an absolute path like `C:\Users\...\meep-abc\test_foo-0`. As the replacement argument to `re.sub` that's a template. `\U` raises `re.error: bad escape \U`. 

So all of these tests errored out before even reaching `compare_h5_files`. 

Even when it doesn't raise, a prefix containing `\t` or `\1` would be silently rewritten. `str.replace` takes the string literally and behaves identically on POSIX (where the replacement never contains a backslash...).

Issue #2699, opened 2023-11-01
Update meep on Pypi?


jaredmagnusson:
Is there a way to make the latest meep installable with pip, in order to avoid the headache of building from source or requiring conda? I see a version from 2018 on Pypi but nothing later. This would be for machines that already have python environments set up and where meep would be part of a larger workflow.

oskooi:
I agree that PyPI provides a more natural workflow for working with Python packages than Conda which is more geared towards one-off installations. We did consider packaging the Meep binary using PyPI in 2018 but ended up choosing Conda partly because it had better support (at the time) for Meep's non-Python library dependencies including HDF5, harminv, and MPB.

We would consider switching to PyPI if someone could put together the package itself.

zacoons:
+1 a PyPi package would be great

PR #3312, opened 2026-09-14
Build PyPI wheels in CI

yorickreum:
# Build and publish PyPI wheels

Addresses #2699 by building wheels for Python 3.10 through 3.13 for Linux (manylinux, arm64 and x64) and macOS (Apple Silicon / arm64).

This makes installing `meep` via `pip install` (without Conda) straightforward. The wheels package both the serial and MPI binaries (adding around 2 MB per wheel to ship both), so only one package needs to be published, and users can decide during import which version they want to use.

## Naming

The distribution is `pymeep`, matching the Conda package; the import name stays `meep`.

`meep` on PyPI is an unrelated task runner, last released in 2018. It is abandoned, so the name could be requested under PEP 541, but that is a conversation with PyPI that the Meep maintainers would need to have.

## Verified

All platforms above build, repair, and install into a clean environment and run a smoke test that exercises FDTD, an MPB eigensolve, and a real 2-rank MPI job, asserting that both ranks share one communicator and produce the serial result.

Meep's own `test_ring`, `test_mode_decomposition`, `test_dump_load`, `test_eigfreq`, `test_simulation`, and `test_mpb` pass against an installed wheel, including with h5py loaded alongside (which puts two different HDF5 major versions in one process).

**Not verified:** the `publish` job. It needs a real tag and a configured PyPI trusted publisher, so it cannot run before this merges.

## LLM Disclaimer

I used Claude Code (Opus 5) for assistance here.


oskooi (approved):
LGTM.

Thanks for putting together this (long-awaited) package.

stevengj:
In order for this to work, do we need to register a `pymeep` account on PyPI?

python/Makefile.am
```diff
@@ -120,8 +120,13 @@ EXTRA_DIST = $(BUILT_SOURCES) meep.i meep-python.hpp typemap_utils.cpp materials
 CLEANFILES = .coverage
 MAINTAINERCLEANFILES = $(BUILT_SOURCES)
 
+# -I$(top_builddir)/src finds the generated meep/meep-config.h in a VPATH
+# build. It is listed in src/Makefile.am's pkginclude_HEADERS, so a release
+# tarball ships a copy next to the sources and $(top_srcdir)/src is enough;
+# a Git checkout has no such copy.
```
stevengj: I don't think this level of verbosity is needed
```suggestion
```

python/Makefile.am
```diff
@@ -120,8 +120,13 @@ EXTRA_DIST = $(BUILT_SOURCES) meep.i meep-python.hpp typemap_utils.cpp materials
...
 AM_CPPFLAGS = -I$(top_srcdir)/src               \
-              -I$(top_srcdir)/libpympb		\
+              -I$(top_builddir)/src             \
```
stevengj, line 125:
```suggestion
              -I$(top_builddir)/src     # for src/meep/meep-config.h      \
```

doc/docs/Installation.md
```diff
@@ -116,6 +116,50 @@ This will show something like `1.11.0-1-g415bc8eb` where the first three digits
...
+### Parallel (MPI) simulations
+
+The Linux and macOS wheels carry a second, MPI-enabled copy of the extension modules alongside the serial one, and pick between them when `meep` is imported. The parallel copy is built against the [MPICH ABI](https://www.mpich.org/abi/), so it works with MPICH itself or any ABI-compatible derivative: MVAPICH, Intel MPI, HPE Cray MPICH. `libmpi` is deliberately *not* bundled: it has to be the same one your launcher uses.
+
+```bash
+pip install "pymeep[mpi]"
+mpiexec -np 4 python -m mpi4py foo.py
+```
+
+Meep loads the parallel build when it detects that the job was launched under an MPI launcher with more than one process, or when `MEEP_MPI=1` is set explicitly. Plain `python foo.py` runs the serial build in a single process. `meep.MEEP_PARALLEL` reports which of the two is active.
```
stevengj: I'm not sure what the point of building both MPI and non-MPI versions is?

If you want to run in serial, you can just launch the MPI version with one process (which is what happens when you don't use `mpiexec`), and it is functionally equivalent to the serial version.

Building both versions just seems like it wastes space.  Recommendation: build only the MPI version (which works as a serial code).

stevengj:
cc @haard, who maintains a [`meep` package on PyPi](https://pypi.org/project/meep/) that seems to be defunct?  (Its homepage link is broken.)

@haard, if your `meep` package is defunct would it be possible to take over that name for this package (which has been around since 2006)?   No pressure.   It looks like the existing `meep` package continues to be downloaded, in which case we should stick with `pymeep` for our package, which has the advantage of matching the Conda name.

haard:
I'd be happy to, ~not sure about the protocol though~
I can delete the package from pypi, if you are ready to register a new one? Alternatively, I can add you and transfer ownership, it looks like. One issue is that you'd need to set a 2.x version to not collide with already-published releases.

doc/docs/Installation.md
yorickreum: I honestly just copied the Conda approach here (`pymeep` vs `pymeep-parallel`). I'm open to cleaning our deploy code up and shipping only the MPI build. 

The other consequence is that `import meep` would then always initialize MPI, since Meep imports `mpi4py` when built with MPI. Every serial run becomes an MPI job of size 1 and needs a working `libmpi` resolvable at import. Hence it wouldn't necessarily save space, at least not for serial users, since `mpi4py` and `mpich` would then become hard dependencies of `pip install meep` (currently only dragged in with `pip install pymeep[mpi]`). The `mpich` wheel that provides `libmpi` it is about 10 MB depending on platform.

Happy to go either way ... also leaning towards dropping the serial binary, as I do assume most users will want to parallelize anyways ...

yorickreum:
> I'd be happy to, ~not sure about the protocol though~ I can delete the package from pypi, if you are ready to register a new one? Alternatively, I can add you and transfer ownership, it looks like. One issue is that you'd need to set a 2.x version to not collide with already-published releases.

Cool! `meep` is slightly cleaner than `pymeep` of course ... the current GitHub release v1.34.0 already has a higher minor version number than your v1.0.6 on PyPi. Would this be enough or would it really need to be a new major version number?

python/Makefile.am
yorickreum: Fair, I cleaned up most of the comments now.

python/Makefile.am
yorickreum: Aded a comment above now

doc/docs/Installation.md
yorickreum: I thought about it again and went ahead and dropped the separate serial build. Quite a bit cleaner, and the extra MPI dependency seems like a very reasonable trade-off.

yorickreum:
I've changed the package name to `meep` in the CI. If @haard could add @NanoComp as an owner of the PyPI project, this should work cleanly for v1.35.0.

PR #3313, opened 2026-09-15
Compare transmittances relatively in test_material_grid

yorickreum:
`places=5` requires the two values to agree within about 5e-6 in absolute terms. Since the quantity being compared is around 29.5, this corresponds to a relative tolerance of only about 1.7e-7, which is unnecessarily strict for results produced through two different numerical code paths and can lead to spurious failures from small toolchain-dependent rounding differences.

`assertClose`, which is already provided by the test suite, expresses the intended numerical tolerance more directly.


stevengj:
Thanks!

PR #3315, opened 2026-09-15
Fix ring test tolerances

yorickreum:
This fixes two independent but related sources of architecture-dependent failures in the ring tests:

1. `ring-ll` assumes that Harminv always returns the same four modes. One weak mode near a much stronger resonance is marginal and may be omitted depending on small numerical differences between architectures. The test now treats that mode as optional, matches the remaining modes by frequency rather than index, and uses separate relative tolerances for frequency, decay rate, and amplitude.

2. `test_ring.py` uses `assertAlmostEqual(..., places=...)` for Harminv results. Since this is an absolute decimal-place tolerance, it again corresponds to very different relative accuracies for quantities with different magnitudes. The test now also uses `assertClose` with explicit relative tolerances appropriate to each fitted quantity.

These are separate issues, but both stem from the ring tests assuming more numerical reproducibility from Harminv than is warranted across platforms.


stevengj (approved):

stevengj:
Thanks!

PR #3318, opened 2026-09-17
Bound plot3D isosurfaces for grayscale structures

### LINKED ISSUE #3317: add nlevels argument to plot3d to bound the number of isosurfaces
In `plot3d`, the code to compute the number of isosurfaces seems like it could produce a large number of levels in a grayscale structure:
https://github.com/NanoComp/meep/blob/6ef8896d9652cfe457e33b4b283e5ee3c23c73f9/python/visualization.py#L155

Recommendation: pass an `nlevels` argument, defaulting to `None`, which sets the maximum number of levels, defaulting to `10` if `None` is passed (though it could use fewer if the current algorithm results in fewer levels).

(isosurfaces are not an ideal interface for displaying grayscale structures, but we don't want it to kill your machine by generating hundreds of isosurfaces, and more than about 10 items in a legend is probably too much for most cases.)

CAOShurong:
## Summary

- add a public `nlevels` argument to `Simulation.plot3D` for both the native and HTML backends
- default to at most 10 evenly distributed non-background permittivity levels while preserving all levels when fewer are present
- reject non-positive or non-integral limits and document the bounded behavior
- add regression coverage for the default cap, explicit limits, background exclusion, validation, and HTML-backend forwarding

## Testing

- base reproduction: a 12-level grayscale grid generated 11 non-background isosurfaces
- committed-source marching-cubes probe: default generated 10 surfaces; `nlevels=3` generated levels 2, 7, and 12
- focused `test_plot3D_nlevels` execution and HTML dispatch probe
- pre-commit hooks for all four changed files
- `python -m compileall` for the changed Python sources and test
- `python -m mkdocs build --strict -f mkdocs.yml`
- `git diff --check`

Closes #3317

Disclosure: This contribution was developed with OpenAI Codex assistance. I reviewed the implementation and regression coverage.


oskooi (approved):
LGTM.

PR #3321, opened 2026-09-22
Fix per-frequency phase normalization for multifrequency adjoint sources

Jaehyeon-Park:
Corrects the relative half-time-step phase in the scale passed to `FilteredSource` for multifrequency adjoint objectives. The correction $e^{i\pi(f-f_c)\Delta t}$ is applied in the shared `ObjectiveQuantity._adj_src_scale()` method, where $f_c$ is the source center frequency.


smartalecH (changes requested):
Thanks for looking into this.

From a theoretical perspective, I'm not sure why this would be needed. We compensate for the discrete operator here:
https://github.com/NanoComp/meep/blob/0a4e71f9f71b523c64a7eea3f57bf6a5a87549e8/python/adjoint/objective.py#L77-L79
And we brute force the compensation needed for the user's time profile here:
https://github.com/NanoComp/meep/blob/0a4e71f9f71b523c64a7eea3f57bf6a5a87549e8/python/adjoint/objective.py#L81-L99

So why do we need this additional (frequency-dependent) adjustment?

From an empirical perspective, we'd like to see some experiments showing how this improves e.g. gradient accuracy. We don't necessarily need a new test (perhaps we can tighten the tolerance on some of the existing broadband tests?)

Jaehyeon-Park:
`fwd_dtft` is computed for both paths, but only the single-frequency branch uses it:

[https://github.com/NanoComp/meep/blob/0a4e71f9f71b523c64a7eea3f57bf6a5a87549e8/python/adjoint/objective.py#L124-L129](https://github.com/NanoComp/meep/blob/0a4e71f9f71b523c64a7eea3f57bf6a5a87549e8/python/adjoint/objective.py#L124-L129)

The multifrequency `FilteredSource` calls also leave out `time_src`, so `signal_dtft` is 1:

[https://github.com/NanoComp/meep/blob/0a4e71f9f71b523c64a7eea3f57bf6a5a87549e8/python/adjoint/filter\_source.py#L60-L70](https://github.com/NanoComp/meep/blob/0a4e71f9f71b523c64a7eea3f57bf6a5a87549e8/python/adjoint/filter_source.py#L60-L70)

The existing normalization provides a phase reference at the source center frequency, but the multifrequency branch uses that same reference for every frequency. With Meep’s actual Yee timestamps, the adjoint-current prefactor should have the frequency dependence $i\frac{2\sin(\pi f\Delta t)}{\Delta t}$. The current scale additionally carries $e^{-i\pi(f-f_c)\Delta t}$, and `FilteredSource` reproduces that phase as part of the requested spectrum. The proposed factor removes this remaining relative offset while keeping the sine magnitude and the normalization at $f\_c$ unchanged.

I also ran a few double-precision comparisons using the multifrequency cases in [test\_adjoint\_solver.py](https://github.com/NanoComp/meep/blob/0a4e71f9f71b523c64a7eea3f57bf6a5a87549e8/python/tests/test_adjoint_solver.py), and the tests gave relative errors of:

| Test                  | Without correction | With correction |
| --------------------- | -----------------: | --------------: |
| `test_DFT_fields`     |            1.33e-3 |         4.46e-5 |
| `test_complex_fields` |            3.10e-4 |         2.95e-5 |

For `test_multifreq_monitor`, single/multifrequency agreement got slightly worse (`2.72e-3` → `2.79e-3`). But using a separate adjoint mode profile for each frequency (by splitting the broadband adjoint source into five frequency-selective sources in the same simulation), without changing the frequency-domain source coefficients, the correction reduced the discrepancy from `3.76e-4` to `3.91e-5` (looks like the original phase error was partly cancelling out the spatial-mode approximation error in this test).

smartalecH:
> fwd_dtft is computed for both paths, but only the single-frequency branch uses it:

I can't remember why we added this branch... perhaps something to do with `IndexSource`? We should dive into potential motivation here.

> The multifrequency FilteredSource calls also leave out time_src, so signal_dtft is 1:

But the `FilteredSource` is only used for _adjoint sources_, never the forward source, right? So why would this matter?

> With Meep’s actual Yee timestamps, the adjoint-current prefactor should have the frequency dependence...

So rather than adding an additional term that compensates, why not just use a single, correct factor? Can we (or Claude) derive the _proper_ factor we should have regardless if it's narrowband or broadband? That way we can simplify all the above logic?

> I also ran a few double-precision comparisons using the multifrequency cases

As part of this PR, can we update the tests with tighter tolerances, then, to reflect this improvement?

> looks like the original phase error was partly cancelling out the spatial-mode approximation error in this test

This is an interesting hypothesis -- can we run some experiments to verify this? Ideally we have a stacked PR on top that implements the proper broadband support for e.g. eigenmode sources... but that's a lot of work that requires some design decisions. We (Claude) can just hack the code to do the right thing here and validate this hypothesis.

